PR #4, opened 2019-03-23
hack: add test target

tonistiigi:

### DIFF
--- a/hack/test
+++ b/hack/test
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+. $(dirname $0)/util
+set -eu -o pipefail
+
+: ${CONTINUOUS_INTEGRATION=}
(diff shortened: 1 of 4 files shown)


hack/test
```diff
@@ -0,0 +1,57 @@
...
+  tmpfile=$(mktemp -t docker-iidfile.XXXXXXXXXX)
+  buildctl build $progressFlag --frontend=dockerfile.v0 \
+    --local context=. --local dockerfile=. \
+    --frontend-opt target=integration-tests \
+    --output type=docker,name=$iid,dest=$tmpfile
+  docker load -i $tmpfile
+  rm $tmpfile
+  ;;
+*)
+  case $buildmode in
```
AkihiroSuda: this case doesn't need to be nested?

tonistiigi: ptal

PR #8, opened 2019-03-24
build: basis of build command

tonistiigi:
Base implementation for build command. Lots of missing functionality and requires `buildkitd` already running atm. but a base to iterate on.

testing:

```
> make

buildkitd --debug 2>/buildkitd.log &
docker buildx build simple1
```


build/output.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"encoding/csv"
+	"os"
+	"strings"
+
+	"github.com/moby/buildkit/client"
+	"github.com/pkg/errors"
+)
+
+func ParseOutputs(inp []string) ([]client.ExportEntry, error) {
```
AkihiroSuda: can we import https://github.com/moby/buildkit/tree/master/cmd/buildctl/build for deduplication?

tonistiigi: Generally yes, but can probably be a follow-up when things are more stabilized. Looking at it, I do see that there are slight incompatibilities both ways atm that we should definitely fix.

PR #10, opened 2019-03-26
drivers base implementation

tonistiigi:
Starting to implement docker driver

### DIFF
--- a/driver/docker/driver.go
+++ b/driver/docker/driver.go
@@ -2,24 +2,82 @@ package docker
 
 import (
 	"context"
+	"io"
+	"io/ioutil"
 
+	"github.com/docker/docker/api/types"
 	dockertypes "github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/network"
+	dockerclient "github.com/docker/docker/client"
 	"github.com/moby/buildkit/client"
 	"github.com/pkg/errors"
 	"github.com/tonistiigi/buildx/driver"
 )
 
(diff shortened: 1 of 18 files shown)


tonistiigi:
This is ready now. PTAL @tiborvass @AkihiroSuda

driver/manager.go
```diff
@@ -0,0 +1,70 @@
...
+
+type InitConfig struct {
+	// This object needs updates to be generic for different drivers
+	Name           string
+	DockerAPI      dockerclient.APIClient
+	BuildkitConfig BuildkitConfig
+	Meta           map[string]interface{}
+}
+
+var drivers map[string]Factory
```
tiborvass: `var drivers = map[string]Factory{}`

tonistiigi: there is a nil check on line 34

tiborvass: Hence my suggestion, so you can get rid of that :)

driver/docker/driver.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
+	_, conn, err := d.exec(ctx, []string{"buildctl", "dial-stdio"})
+	if err != nil {
+		return nil, err
+	}
+
+	conn = demuxConn(conn)
+
+	return client.New(ctx, "", client.WithDialer(func(string, time.Duration) (net.Conn, error) {
```
AkihiroSuda: Just curious, why not use `docker://...` address?

Issue #13, opened 2019-03-26
buildfile command


tonistiigi:
Add another command that produces a higher level build operation based on config from a file. An input file can be compose or another format that can stack the cli options into separate targets.

On invoking command all builds are invoked concurrently, looking like a single build to the user.

`buildx buildfile -f compose.yml [target] [target]..`

The file allows stacking all cli options, including things like exporter and remote cache settings (but probably excluding entitlements). When using multiple files you can override some properties with a file that is not committed to git. The cli flags can be used for the highest priority override. The non-compose format can support sharing code so that groups of options can be used as part of multiple targets.

The name needs improvement.

AkihiroSuda:
s/buildfile/compose/ ?

tonistiigi:
I don't think compose is ideal format for this, because of the missing code reuse and missing configuration options. Just compose is somewhat similar and widespread so makes sense to support it as well.

tonistiigi:
#16

PR #16, opened 2019-04-10
bake: initial implementation

tonistiigi:
Initial implementation for #13 (formely buildfile)

Testing:

```
> make
cd compose
docker buildx bake
docker buildx bake -f config.hcl release
docker buildx bake -f config.hcl --set db.args.FOO=bar release
# see config.hcl for full data format
# --print shows the parameters that would be used by individual builds without invoking it
```

The compose part is pretty basic and should work as expected. The full internal datastructure for defining targets probably needs some design work and doesn't look right to me.

Alternative designs:
```
group "default" {
	targets = ["db", "webapp"]
}


target "%-cross" {
	platforms = [
		"linux/amd64",
		"linux/arm64"
	]
}

---
docker buildx bake default-cross
```

```
target "default" {
	targets = ["db", "webapp"]
}

target "default-cross" {
	targets = ["db", "webapp-cross"]
}

target "webapp" {
	context = "./"
	args = {
		buildno = "123"
	}
}
```

```
group "default" {
	targets = ["db", "webapp"]
}

group "default-cross" {
        inherits = ["cross"]
	targets = ["db", "webapp"]
}
target "cross" {
	platforms = [
		"linux/amd64",
		"linux/arm64"
	]
}

---
docker buildx bake default-cross
```

Outputs not supported yet, thinking if this should be per target or global.

@tiborvass @AkihiroSuda @seemethere @dmcgowan


AkihiroSuda:
Looks good, but why HCL?

tonistiigi:
@AkihiroSuda I don't really care about the format that much. Could accept anything that we can parse. You can use JSON as well (and thinking if we should allow json5). Examples use HCL cause I personally found it more readable and better than toml on comparisons. I'm not a big fan of yaml but the main thing is that I'd want to avoid something that looks like compose but isn't compose. Atm. `bake` accepts vanilla compose with no changes and I think we should keep it this way. If you want you can use compose with the override logic as well. If you as a user like yaml and compose already you can just keep using that and never care about the other formats.

tonistiigi:
updated buildkit to bring in moby/buildkit#929

bake/bake.go
```diff
@@ -0,0 +1,344 @@
...
+	fnl := strings.ToLower(fn)
+	if strings.HasSuffix(fnl, ".yml") || strings.HasSuffix(fnl, ".yaml") {
+		return ParseCompose(dt)
+	}
+
+	if strings.HasSuffix(fnl, ".json") || strings.HasSuffix(fnl, ".hcl") {
+		return ParseHCL(dt)
+	}
+
+	cfg, err := ParseCompose(dt)
```
tiborvass: Are there compose files without `.yml` or `.yaml` suffixes? Why not error out?

tonistiigi: I have no idea. Does compose force you to use `.yml` suffix currently, I think not.

bake/bake.go
```diff
@@ -0,0 +1,344 @@
...
+
+	bo := &build.Options{
+		Inputs: build.Inputs{
+			ContextPath:    t.Context,
+			DockerfilePath: t.Dockerfile,
+		},
+		Tags:      t.Tags,
+		BuildArgs: t.Args,
+		Labels:    t.Labels,
+		// CacheFrom: t.CacheFrom,
```
tiborvass: ?

tonistiigi: The key does not exist atm. but should later.

commands/bake.go
```diff
@@ -0,0 +1,104 @@
...
+func runBake(dockerCli command.Cli, targets []string, in bakeOptions) error {
+	ctx := appcontext.Context()
+
+	if len(in.files) == 0 {
+		files, err := defaultFiles()
+		if err != nil {
+			return err
+		}
+		if len(files) == 0 {
+			return errors.Errorf("no compose.yml or dockerbuild.hcl found, speficy build file with -f/--file")
```
tiborvass: I think official is `docker-compose.yml`

tiborvass: typo: `specify`

tiborvass: `bakefile.hcl` ? :P

commands/bake.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return err
+	}
+
+	return buildTargets(ctx, dockerCli, bo, in.progress)
+}
+
+func defaultFiles() ([]string, error) {
+	fns := []string{
+		"compose.yml", // support app
```
tiborvass: we need `docker-compose.yml` too, and `.yaml` variants

commands/bake.go
```diff
@@ -0,0 +1,104 @@
...
+		"dockerbuild.json",
+		"dockerbuild-override.json",
```
tiborvass: `dockerbuild.override.json`

FernandoMiguel:
@tonistiigi what ever happened to `--set db.args.FOO=bar` ?

tonistiigi:
@FernandoMiguel wdym?

FernandoMiguel:
> @FernandoMiguel wdym?

@tonistiigi in the OP example
`docker buildx bake -f config.hcl --set db.args.FOO=bar release`

I'm trying to pass args to my buildx

PR #17, opened 2019-04-11
driver: add docker driver

tonistiigi:
Add driver that can work directly with docker storage, old driver is renamed to `docker-container`.

This temporarily breaks things like multi-platform builds (if you are using 19.03).


build/build.go
```diff
@@ -94,6 +95,28 @@ func Build(ctx context.Context, drivers []driver.Driver, opt map[string]Options,
...
 		}
+
+		for i, e := range opt.Exports {
+			if e.Type == "oci" && !d.Features()[driver.OCIExporter] {
+				return nil, notSupported(d, driver.OCIExporter)
+			}
+			if e.Type == "docker" && e.Output != nil && !d.Features()[driver.DockerExporter] {
+				return nil, notSupported(d, driver.DockerExporter)
+			}
+			if e.Type == "image" {
```
tiborvass: Would be useful to automatically add this Export type if `len(opt.Tags) > 0` at line 88. If `image` was not found, add it.

tiborvass: This is to have backwards compat with `build -t foo .`

tonistiigi: Docker driver automatically exports to image store. Unless when we add `-o /dev/null`

tiborvass:
```
/work/simple1 # docker buildx build -t foo .
[+] Building 2.9s (8/8) FINISHED
 => [internal] load build definition from Dockerfile                                                              0.0s
 => => transferring dockerfile: 98B                                                                               0.0s
 => [internal] load .dockerignore                                                                                 0.0s
 => => transferring context: 2B                                                                                   0.0s
 => [internal] load metadata for docker.io/library/alpine:latest                                                  1.2s
 => [1/4] FROM docker.io/library/alpine@sha256:28ef97b8686a0b5399129e9b763d5b7e5ff03576aa5580d6f4182a49c5fe1913   0.6s
 => => resolve docker.io/library/alpine@sha256:28ef97b8686a0b5399129e9b763d5b7e5ff03576aa5580d6f4182a49c5fe1913   0.0s
 => => sha256:bdf0201b3a056acc4d6062cc88cd8a4ad5979983bfb640f15a145e09ed985f92 2.76MB / 2.76MB                    0.2s
 => => sha256:28ef97b8686a0b5399129e9b763d5b7e5ff03576aa5580d6f4182a49c5fe1913 2.36kB / 2.36kB                    0.0s
 => => sha256:5c40b3c27b9f13c873fefb2139765c56ce97fd50230f1f2d5c91e55dec171907 528B / 528B                        0.0s
 => => sha256:cdf98d1859c1beb33ec70507249d34bacf888d59c24df3204057f9a6c758dddb 1.51kB / 1.51kB                    0.0s
 => => extracting sha256:bdf0201b3a056acc4d6062cc88cd8a4ad5979983bfb640f15a145e09ed985f92                         0.2s
 => [internal] load build context                                                                                 0.0s
 => => transferring context: 98B                                                                                  0.0s
 => [2/4] RUN apk add --no-cache curl                                                                             0.9s
 => [3/4] WORKDIR /src                                                                                            0.0s
 => [4/4] COPY . .                                                                                                0.1s
/work/simple1 # docker images
```

build/build.go
```diff
@@ -75,7 +75,19 @@ func Build(ctx context.Context, drivers []driver.Driver, opt map[string]Options,
...
 
-		if len(opt.Exports) > 1 {
+		switch len(opt.Exports) {
+		case 1:
+			// valid
+		case 0:
+			if _, ok := d.(interface {
+				IsDefaultMobyDriver()
+			}); !ok {
+				return nil, errors.Errorf("loading to docker currently not implemented, specify dest file or -")
```
tonistiigi: This should only be shown if exporter type is docker and there is no dest set

tiborvass: `docker` or `image` ?

tonistiigi: docker

PR #27, opened 2019-04-17
build: add more docker UI (context/dockerfile handling, iidfile, add-host, network)

tiborvass:
This should add same level of support for stdin/remote contexts as the buildkit builder in moby.
Also preliminary support for `--iidfile` (only for docker driver and single-platform queries).
Added support for `--add-host` and `--network`


build/build.go
```diff
@@ -208,6 +249,9 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, pw
 			mu.Lock()
 			resp[k] = rr
 			mu.Unlock()
+			if opt.ImageIDFile != "" {
+				return ioutil.WriteFile(opt.ImageIDFile, []byte(rr.ExporterResponse["containerimage.digest"]), 0644)
```
tonistiigi: Add a switch between manifest digest and this. If manifest digest exists use that.

commands/build.go
```diff
@@ -133,7 +132,18 @@ func buildTargets(ctx context.Context, dockerCli command.Cli, opts map[string]bu
 	defer cancel()
 	pw := progress.NewPrinter(ctx2, os.Stderr, progressMode)
 
-	_, err = build.Build(ctx, dis, opts, pw)
+	resp, err := build.Build(ctx, dis, opts, pw)
+	if err != nil {
+		return err
+	}
+
+	for k, sr := range resp {
```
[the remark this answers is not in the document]
tiborvass: debug will remove

build/build.go
```diff
@@ -208,6 +260,9 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, pw
...
```
tonistiigi: this is also not set if no exporter and !moby

build/build.go
```diff
@@ -219,30 +274,109 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, pw
...
+
+	case isLocalDir(inp.ContextPath):
+		target.LocalDirs["context"] = inp.ContextPath
+		switch inp.DockerfilePath {
+		case "-":
+			dockerfileReader = os.Stdin
+		case "":
+			dockerfileDir = inp.ContextPath
+		default:
+			dockerfileDir = filepath.Dir(inp.DockerfilePath)
```
tonistiigi: you shouldn't do this as it exposes a directory potentially outside of context. Also this does not work with symlinks.

build/build.go
tonistiigi: I guess just check if it was present in the response

tonistiigi:
Please handle the remaining comments in follow-up.

PR #33, opened 2019-04-18
build: add cache-from and cache-to support

tonistiigi:
@tiborvass @AkihiroSuda 

### DIFF
--- a/build/cache.go
+++ b/build/cache.go
@@ -1,0 +1,60 @@
+package build
+
+import (
+	"encoding/csv"
+	"strings"
+
+	"github.com/moby/buildkit/client"
+	"github.com/pkg/errors"
+)
+
+func ParseCacheEntry(in []string) ([]client.CacheOptionsEntry, error) {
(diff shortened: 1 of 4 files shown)


build/cache.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"encoding/csv"
+	"strings"
+
+	"github.com/moby/buildkit/client"
+	"github.com/pkg/errors"
+)
+
+func ParseCacheEntry(in []string) ([]client.CacheOptionsEntry, error) {
```
AkihiroSuda: can we import buildctl pkg rather than duplicating the code?

tonistiigi: It is a little different. No legacy mode and accepts`ref` and `ref,ref` for compatibility with existing docker syntax for `cache-from`.

tonistiigi:
@AkihiroSuda If you find a way to normalize parsing across buildkit, cli and buildx then PRs are welcome. It will probably become much easier after we drop support for deprecated versions in buildctl.

PR #40, opened 2019-04-23
Add makefile targets to build easily with just go

seemethere:
Makes it easier to build inside of containers, including the ability to cross compile easily with usage similar to:

    GOOS=darwin make plugin

Shouldn't change current functionality but it'll make it easier when I'm implementing building this in the Docker CLI pipeline.

Signed-off-by: Eli Uriegas <eli.uriegas@docker.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,28 @@
+PKG=github.com/tonistiigi/buildx
+VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
+REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
+LDFLAGS=-X ${PKG}/version.Version=${VERSION} \
+	-X ${PKG}/version.Revision=${REVISION} \
+	-X ${PKG}/version.Package=${PKG}
+
+GOFILES=$(shell find . -type f -name '*.go')
+
+.PHONY: build
+build: plugin
+
+bin/buildx bin/docker-buildx: $(GOFILES)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,3 +1,28 @@
+PKG=github.com/tonistiigi/buildx
+VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
+REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
```
thaJeztah: Do we need an option to override these?

Makefile
```diff
@@ -1,3 +1,28 @@
...
+LDFLAGS=-X ${PKG}/version.Version=${VERSION} \
```
thaJeztah: Same here; wondering if this should be appended to / overridable

tonistiigi:
We should be using the Dockerfile to build binaries. Eg. no point on copying the LDFLAGS logic (or any future logic) and we should always use the go version in the Dockerfile.

Makefile
seemethere: They're overridable if you want:
```
make VERSION=dev REVISION=dev plugin
```

I can add support to override with an environment variable if wanted

seemethere:
@tonistiigi that's fine for this repository but I need to have the ability to build inside of a container (i.e. without a buildkit daemon or docker daemon) since we build our packages inside of a container.

I can either put this build logic here or inside of the installer script in the packaging repository

tonistiigi:
If you need to build some custom version in some custom environment then scripts are probably better. An alternative would be to just use the releases in this repo without trying to rebuild them.

seemethere:
Not necessary for build anymore

PR #41, opened 2019-04-23
bake: add output key

tonistiigi:


bake/bake.go
```diff
@@ -356,12 +367,15 @@ func merge(t1, t2 Target) Target {
 	if t2.Platforms != nil { // no merge
 		t1.Platforms = t2.Platforms
 	}
-	if len(t2.CacheFrom) > 0 { // no merge
-		t1.CacheFrom = t2.CacheFrom
+	if t2.CacheFrom != nil { // no merge
+		t1.CacheFrom = append(t1.CacheFrom, t2.CacheFrom...)
```
tiborvass: what is this doing?

tiborvass: bugfix?

tonistiigi: yes, this makes more sense for this field

PR #56, opened 2019-04-29
ls: platform column now includes user-specified one with driver fallback

tiborvass:
When creating builders with `create --platform`, the user can set a subset
of the platforms supported by the builder, but `ls` should only show
the user-specified platforms, when set. Otherwise, the default platforms.

Signed-off-by: Tibor Vass <tibor@docker.com>


tonistiigi:
Even if you set manual platforms the automatic ones are still used. They just have a different priority. So remove dupes instead. If you want to signify what is running you can add a `*` to manual ones. Or take the automatic ones that get overridden out.

Issue #78, opened 2019-05-18
No declared license.


jrabbit:
i just wrote up a patch but there's no license anywhere.  Also very unclear on contribution guidelines beyond running the test suite; do I need to use the DCO `-s` stuff?

tonistiigi:
Assigned @thaJeztah for these files as he is good at that but the answers are that you do need dco and license is apache2.

tonistiigi:
#85

PR #79, opened 2019-05-18
Change compose file handling to require valid service specifications

jrabbit:
Added the checks and some tests
One of the tests wasn't valid docker-compose.yml, that's been changed.
Bad config throws an error and has a test

Signed-off-by: Jack Laxson <jackjrabbit@gmail.com>

I'm drafting some better tests still and will probably add them when they seem useful.

Also this is pending #78

### DIFF
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -1,6 +1,8 @@
 package bake
 
 import (
+	"fmt"
+
 	"github.com/docker/cli/cli/compose/loader"
 	composetypes "github.com/docker/cli/cli/compose/types"
 )
@@ -33,7 +35,7 @@ func ParseCompose(dt []byte) (*Config, error) {
 		var g Group
 
 		for _, s := range cfg.Services {
-			g.Targets = append(g.Targets, s.Name)
+
 			var contextPathP *string
 			if s.Build.Context != "" {
 				contextPath := s.Build.Context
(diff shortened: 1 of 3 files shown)


bake/compose.go
```diff
@@ -44,6 +46,15 @@ func ParseCompose(dt []byte) (*Config, error) {
 				dockerfilePath := s.Build.Dockerfile
 				dockerfilePathP = &dockerfilePath
 			}
+			// Check if there's actually a dockerfile mentioned
+			if dockerfilePathP == nil && contextPathP == nil {
+				// if not make sure they're setting an image or it's invalid d-c.yml
+				if s.Image == "" {
+					return nil, fmt.Errorf("Compose file invalid: Service %s has neither an image nor a build context specified. At least one must be provided.", s.Name)
+				}
+				break
```
tonistiigi: `continue` ?

bake/compose.go
```diff
@@ -44,6 +46,15 @@ func ParseCompose(dt []byte) (*Config, error) {
 				dockerfilePath := s.Build.Dockerfile
 				dockerfilePathP = &dockerfilePath
 			}
+			// Check if there's actually a dockerfile mentioned
+			if dockerfilePathP == nil && contextPathP == nil {
+				// if not make sure they're setting an image or it's invalid d-c.yml
+				if s.Image == "" {
+					return nil, fmt.Errorf("Compose file invalid: Service %s has neither an image nor a build context specified. At least one must be provided.", s.Name)
```
tonistiigi: `errors.Errorf` and start with a lowercase letter

tonistiigi: actually, should this be invalid? why not just do default build? I agree that we should skip if image is set and there is no build block. and instead of checking a context I think we should check the `s.Build` against the zero value.

jrabbit: It is invalid if you try and validate a broken service definition (like hack/demo-env/examples/compose/docker-compose.yml) w/ docker-compose it spits out that error.

tonistiigi: Ok, does it error every time there is a build block without `context: .`? Is it in the spec? It obviously seems to be supported by the go parser used by stacks etc. Maybe we should patch the python side?

bake/compose.go
jrabbit: oh yup

tiborvass:
@jrabbit FYI we fixed #78!

tiborvass:
@jrabbit I created another branch based off of a rebased version of your PR: https://github.com/tiborvass/buildx/tree/no-build-field I addressed @tonistiigi's comments.

Feel free to update your PR with it and even squash it.

PR #83, opened 2019-05-24
bake: make dockerfile relative to context

tonistiigi:
@tiborvass @dhiltgen 

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"io/ioutil"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/util/platformutil"
+	"github.com/docker/docker/pkg/urlutil"
 	"github.com/moby/buildkit/session/auth/authprovider"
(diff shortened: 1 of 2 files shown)


bake/bake.go
```diff
@@ -275,6 +276,10 @@ func toBuildOpt(t Target) (*build.Options, error) {
 		dockerfilePath = *t.Dockerfile
 	}
 
+	if !path.IsAbs(dockerfilePath) {
+		dockerfilePath = path.Join(contextPath, dockerfilePath)
```
tiborvass: Does this work for remote contexts?

bake/bake.go
```diff
@@ -394,3 +399,7 @@ func removeDupes(s []string) []string {
 	}
 	return s[:i]
 }
+
+func isRemoteResource(str string) bool {
+	return strings.HasPrefix(str, "http://") || strings.HasPrefix(str, "https://") || strings.HasPrefix(str, "git://") || strings.HasPrefix(str, "git@")
```
thaJeztah: could also just check for `://` in the string I guess

PR #86, opened 2019-05-25
Put driver names to create --help

AkihiroSuda:
$ docker buildx driver ls
  NAME             USAGE
  docker           docker
  docker-container docker-container

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>


tonistiigi:
Do you have more plans for the `driver` command? Maybe in the current form, it makes more sense to add this info to `buildx --help` and `buildx create --help`? We should also tweak the `Priority()` logic in the follow-up to show if a driver is properly configured, what properties it accepts and if it can be used with `create`.

@tiborvass

AkihiroSuda:
updated

PR #87, opened 2019-05-29
[Carry #79] Change compose file handling to require valid service specifications

tiborvass:
Closes #79 which is carried in this PR.

This PR basically handles properly compose services without a `build` field, by skipping them, but errors out when neither a `build` nor an `image` field is present.

Tests updated accordingly.


tonistiigi:
@jrabbit

PR #95, opened 2019-06-18
Read COMPOSE_FILE and COMPOSE_PATH_SEPARATOR

sirlatrom:
This PR makes `docker buildx bake` consult the `COMPOSE_FILE` and `COMPOSE_PATH_SEPARATOR` variables the same as `docker-compose` and `docker stack deploy` do.

I'm sure there's a neater way and place to do this, so this PR is just to suggest the general concept.

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/docker/buildx/bake"
 	"github.com/docker/cli/cli/command"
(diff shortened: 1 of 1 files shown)


tonistiigi:
Maybe we should check if the files set in this env actually exist like for `defaultFiles` and not just directly use them. wdyt?

sirlatrom:
@tonistiigi You are completely right, of course. I'll try and add that to the PR next chance I get. Furthermore, after proper research it turns out `docker stack deploy` does _not_ take the Docker Compose variables into account.

commands/bake.go
```diff
@@ -23,9 +24,20 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions) error {
 	ctx := appcontext.Context()
 
 	if len(in.files) == 0 {
-		files, err := defaultFiles()
-		if err != nil {
-			return err
+		var files []string
+		composeFileVar, ok := os.LookupEnv("COMPOSE_FILE")
+		if ok {
+			composePathSeparator, ok := os.LookupEnv("COMPOSE_PATH_SEPARATOR")
```
thaJeztah: Hm, don't think this one is currently supported by `docker stack deploy` 🤔

sirlatrom: I think that counts for both of them, to be honest. Being able to avoid installing `docker-compose`, but rather use `docker buildx bake` and `docker stack deploy` while being able to read `COMPOSE_FILE` and even perhaps a `.env` file would be golden.

thaJeztah:
I think "COMPOSE_FILE" was implemented in the cli (but I'll have to double check)

We should be careful with implementing all options from docker-compose, and review if they're all desirable (I know .env for example has been problematic for some cases, and not well defined, which was a reason to not yet implement it for docker stack)

PR #103, opened 2019-07-08
driver: set netmode to host for container driver

tonistiigi:
with #102 these flags work better in the context of host network. as the container is already privileged atm there isn't an extra security concern.


tonistiigi:
Having second thoughts about this. I think we should leave it optional atm and add smth. like `--driver-opt` for `buildx create` for enabling it.

tonistiigi:
replaced by #122

PR #104, opened 2019-07-08
build: add allowed entitlements

tonistiigi:

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -24,6 +24,7 @@ import (
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/upload/uploadprovider"
+	"github.com/moby/buildkit/util/entitlements"
 	"github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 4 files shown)


build/build.go
```diff
@@ -454,6 +457,7 @@ func toSolveOpt(d driver.Driver, multiDriver bool, opt Options, dl dockerLoadCal
 	switch opt.NetworkMode {
 	case "host", "none":
 		so.FrontendAttrs["force-network-mode"] = opt.NetworkMode
+		so.AllowedEntitlements = append(so.AllowedEntitlements, entitlements.EntitlementNetworkHost)
```
AkihiroSuda: nit: should check whether so.AllowedEntitlements already contains EntitlementNetworkHost?

tonistiigi: It doesn't matter for the builder, but if you want I can clean it up.

tonistiigi:
I think this should be good to go for testing the security modes https://github.com/moby/buildkit/pull/1081 . The host networking will need some additional logic for the container driver as currently it will put you in to "host network" inside buildkit container but I don't think we can merge #103 before the default network in container is bridge (https://github.com/moby/buildkit/pull/1073)

PR #123, opened 2019-08-01
vendor: update buildkit to v0.6.0

tonistiigi:


build/build.go
```diff
@@ -807,33 +803,40 @@ func notSupported(d driver.Driver, f driver.Feature) error {
...
+				close(started)
+				progress.FromReader(prog, "importing to docker", resp.Body)
+			},
+			started: started,
+		}
+		go func() {
+			select {
+			case <-ctx.Done():
+				pr.Close()
+			case <-done:
```
tiborvass: this channel is never closed, so this is only waiting on ctx.Done, in which case you don't need the select, but maybe you meant to close the done channel.

PR #144, opened 2019-09-11
Add FOSSA checks to Jenkins CI

droopy4096:


Jenkinsfile
```diff
@@ -0,0 +1,56 @@
...
+_
+
+pipeline{
+  agent {
+    node {
+      label 'ubuntu-1804-overlay2'
+    }
+  }
+  triggers {
+    pollSCM('H * * * *')
```
zelahi: Do you want to run this on a cron?

Jenkinsfile
```diff
@@ -0,0 +1,56 @@
+fossaCred = [
+    $class: 'StringBinding',
+    credentialsId: 'fossa-api-key',
+    variable: 'FOSSA_API_KEY'
+]
+
+hubCred = [
```
zelahi: Based on comments here: https://github.com/docker/compose/pull/6875#discussion_r323565904,

we want to avoid $class as its legacy and instead follow Nick's suggestion

Makefile.fossa
```diff
@@ -0,0 +1,32 @@
+REPO_PATH?=docker/buildx
+BUILD_ANALYZER?=dockereng/fossa-analyzer
```
zelahi: oo also heads up, I just made the fossa-analyzer image public so you can change this to: 
`docker/fossa-analyzer`?

Jenkinsfile
```diff
@@ -0,0 +1,56 @@
...
+  options{
+    disableConcurrentBuilds()
+  }
+  stages{
+    stage("FOSSA Analyze"){
+      steps {
+
+        withGithubStatus('FOSSA.scan'){
+
+          sh script: 'docker login -u $REGISTRY_USR -p $REGISTRY_PSW'
```
zelahi: You might not need this anymore since we have a public `fossa-analyzer` docker image

Jenkinsfile
```diff
@@ -0,0 +1,56 @@
...
+    node {
+      label 'ubuntu-1804-overlay2'
+    }
+  }
+  triggers {
+    pollSCM('H * * * *')
+  }
+  environment{
+    FOSSA_API_KEY=credentials('fossa-api-key')
+    REGISTRY=credentials('orcaeng-hub.docker.com')
```
zelahi: I don't think you would need this anymore

Jenkinsfile
droopy4096: I've got the impression that's the only way to do it with declarative?

tiborvass:
Is this makefile a generic makefile you have in other repos? I'm asking because it is not using any Makefile specifics, so it might as well be a simple shell script that could be put in `hack`.

droopy4096:
that is correct - that's the boilerplate Makefile we've used in other places. We're working on improvements - this is our first "fast" iteration to get FOSSA coverage across the board.

Makefile.fossa
```diff
@@ -0,0 +1,32 @@
...
+# for use by fossa-cli for license scanning
+.PHONY: fossa-yarn
+fossa-yarn:
+	docker run --rm -i -v "$$(pwd):$$(pwd)" -w "$$(pwd)/ui" node:8-alpine yarn --production=true
+
+clean-fossa-yarn:
+	$(RM) -r ui/node_modules
+
+compile:
+	$(MAKE) vendor
```
tiborvass: You don't need to run vendor, we already have a check for vendor, if it fails we won't merge the PR. And it is not needed for compiling.

droopy4096: removed entire section

Jenkinsfile
```diff
@@ -0,0 +1,30 @@
+
+@Library('jps')
+_
+
+pipeline{
+  agent {
```
tiborvass: `pipeline{` vs `agent {`

Jenkinsfile
```diff
@@ -0,0 +1,30 @@
...
+    stage("FOSSA Analyze"){
+      steps {
+
+        withCredentials([string(credentialsId: 'fossa-api-key', variable: 'FOSSA_API_KEY')]){
+          withGithubStatus('FOSSA.scan'){
+
+            labelledShell returnStatus: false, returnStdout: true, label: "make fossa-analyze",
+                  script:'make -f Makefile.fossa BRANCH_NAME=${BRANCH_NAME} fossa-analyze'
+            labelledShell returnStatus: false, returnStdout: true, label: "make fossa-test",
+                      script: 'make -f Makefile.fossa BRANCH_NAME=${BRANCH_NAME} fossa-test'
```
tiborvass: different indentation than line 22

PR #164, opened 2019-10-16
bake: allow pattern matching for target names in --set

tiborvass:
Although bake is for running multiple targets, --set required a single
target name for overriding a property. This change allows matching
multiple targets for overrides.

bake: add --load and --push shorthands for --set

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,10 +507,12 @@ Options:
 | Flag | Description |
 | --- | --- |
 |  -f, --file stringArray  | Build definition file
+|      --load              | Shorthand for --set=*.output=type=docker
 |      --no-cache          | Do not use cache when building the image
 |      --print             | Print the options without building
 |      --progress string   | Set type of progress output (auto, plain, tty). Use plain to show container output (default "auto")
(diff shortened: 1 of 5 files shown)


FernandoMiguel:
I've been waiting for this for so long.
Thank you.

tonistiigi:
@tiborvass CI red

tiborvass:
Updated with a dependent bugfix in #259

ChrisBr:
@tonistiigi are there any plans to support both, `--load` and `--push`, at the same time?

tonistiigi:
@ChrisBr Yes, but it needs a buildkit update to support multiple exporters.

ChrisBr:
Awesome @tonistiigi, thanks for your response. So is this already supported in Buildkit or needs to be implemented there as well? Sorry for my naive question, we just stumbled upon this earlier and expected we could export and push at the same time. Also happy to have a look if there is already an issue to be worked on etc.

tonistiigi:
@ChrisBr No, there is an old PR https://github.com/moby/buildkit/pull/1788 that got stuck and needs to be picked up or reimplemented.

PR #165, opened 2019-10-16
bake: allow overriding no-cache and pull per target via --set

tiborvass:
Depends on #164


AkihiroSuda:
What's current status?

PR #167, opened 2019-10-21
new driver: kubernetes

### LINKED ISSUE #22: add kubepod driver
Similarly to `docker-container` driver there could be a support to `kubepod` driver that bootstraps itself inside a k8s cluster. Kube endpoints in `docker context` can be used for providing initial configuration. Switching between drivers implemented in #20 . Would be nice if all supported platforms of k8s workers could be also detected automatically.

@AkihiroSuda

AkihiroSuda:
Tested with [`kind`](https://github.com/kubernetes-sigs/kind) and GKE
TODOs:
- [x] Fix ` No Auth Provider found for name "gcp"` error on GKE
- [x] Support consistent hashing
- [ ] Use watch API instead of polling (maybe follow-up PR)
- [x] Use exec API instead of executing `kubectl` binary
- [x] Support rootless
- [x] Support specifying kubeconfig via the current `docker context`
- [x] Change driver name from `kube` to `kubernetes` for `docker` CLI consistency

Note: "nodes" shown in `docker buildx ls` are unrelated to Kubernetes "nodes".
Probably buildx should come up with an alternative term.

## Usage:
```console
  $ kind create cluster
  $ export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
```

```console
  $ docker buildx create --driver kubernetes --driver-opt replicas=3 --use
  $ docker buildx build -t foo --load .
```

`--load` loads the image into the local Docker.

## Driver options
  - `image=IMAGE` - Sets the container image to be used for running buildkit.
  - `namespace=NS` - Sets the Kubernetes namespace. Defaults to the current namespace.
  - `replicas=N` - Sets the number of `Pod` replicas. Defaults to 1.
  - `rootless=(true|false)` - Run the container as a non-root user without `securityContext.privileged`. [Using Ubuntu host kernel is recommended](https://github.com/moby/buildkit/blob/master/docs/rootless.md). Defaults to false.
  - `loadbalance=(sticky|random)` - Load-balancing strategy. If set to "sticky", the pod is chosen using the hash of the context path. Defaults to "sticky"

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>

Close #22


tonistiigi:
> Note: "nodes" shown in docker buildx ls are unrelated to Kubernetes "nodes".
Probably buildx should come up with an alternative term.

Would it make sense that if you do `--replicas 3` then 3 nodes would automatically be generated under the builder instance?

AkihiroSuda:
That doesn't seem to work well with Deployment lifecycle

tonistiigi:
@AkihiroSuda Why? They would all be running or inactive based on if the deployment is running or not. Can't we even have it so that the consistent hashing (optionally) works if you just add two container nodes manually?

AkihiroSuda:
You need to synchronize buildx state with Kube Deployment when pods are rescaled (by either user or HPA) or recreated.

Also, it seems hard to bind pod lifecycle to buildx node without manually managing pods (i.e. reinventing Deployment) or using StatefulSet which is slow.

AkihiroSuda:
Can we rename buildx "Nodes" to "Instances"?

AkihiroSuda:
Or just add a document to clarify buildx "Node" is unrelated to Kubernetes "Node".
Same applies to Swarm-mode "Node" if we are going to add support for `docker service create` driver as well

driver/bkimage/bkimage.go
```diff
@@ -0,0 +1,5 @@
+package bkimage
+
+const (
+	DefaultImage = "moby/buildkit:buildx-stable-1" // TODO: make this verified
```
AkihiroSuda: @tonistiigi Is this different from `moby/buildkit:v0.6.2`?

AkihiroSuda:
Ready for review

tonistiigi:
> Can we rename buildx "Nodes" to "Instances"?

Instance is what we call one named builder already. Instance can have many nodes.

driver/bkimage/bkimage.go
tonistiigi: It is atm, but it is a different tag in case we need to break compatibility in the future. Eg. `moby/buildkit:v0.7.0` might be equal to `moby/buildkit:buildx-stable-1` or `moby/buildkit:buildx-stable-2` depending if it is compatible.

AkihiroSuda:
Subinstance?

AkihiroSuda:
How can we move this forward?

AkihiroSuda:
added new (hacky) commit: kubernetes: show Kubernetes Pods as buildx "Nodes" in `docker buildx inspect`

```console
$ k get pods
NAME                     READY   STATUS    RESTARTS   AGE
kube0-75897b8444-dnfhg   1/1     Running   0          2m
kube0-75897b8444-dsfht   1/1     Running   0          2m
kube0-75897b8444-ghthr   1/1     Running   0          2m

$ k get pods
NAME                     READY   STATUS    RESTARTS   AGE
kube0-75897b8444-dnfhg   1/1     Running   0          2m
kube0-75897b8444-dsfht   1/1     Running   0          2m
kube0-75897b8444-ghthr   1/1     Running   0          2m
```

go.mod
```diff
@@ -80,10 +88,12 @@ require (
...
-	k8s.io/client-go v2.0.0-alpha.0.0.20180806134042-1f13a808da65+incompatible // indirect
+	k8s.io/api v0.0.0-20180712090710-2d6f90ab1293
+	k8s.io/apimachinery v0.0.0-20180621070125-103fd098999d
+	k8s.io/client-go v2.0.0-alpha.0.0.20180806134042-1f13a808da65+incompatible
 	vbom.ml/util v0.0.0-20180919145318-efcd4e0f9787 // indirect
 )
 
 replace github.com/jaguilar/vt100 => github.com/tonistiigi/vt100 v0.0.0-20190402012908-ad4c4a574305
+
+go 1.13
```
tiborvass: If you used `make vendor` it should not have go1.13.

AkihiroSuda: updated

tiborvass:
I'm merging, we can improve in followups.

PR #170, opened 2019-10-23
Support environment variables in docker-container driver

### LINKED ISSUE #169: Feature request: Environment variables in docker-container driver
The docker-container driver currently provides no way to specify environment variables such as `HTTP_PROXY` and `HTTPS_PROXY`, making it impossible to use it behind a proxy.

sirlatrom:
Fixes #169

Example usage:

```bash
docker buildx create --driver-opt env.http_proxy=$http_proxy --driver-opt env.https_proxy=$https_proxy --driver-opt '"env.no_proxy='$no_proxy'"'
```

---

Original suggestion:

```bash
docker buildx create --driver-env http_proxy=$http_proxy --driver-env https_proxy=$https_proxy --driver-env '"no_proxy='$no_proxy'"'
```

### DIFF
--- a/driver/docker-container/factory.go
+++ b/driver/docker-container/factory.go
@@ -2,6 +2,8 @@ package docker
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/docker/buildx/driver"
 	dockerclient "github.com/docker/docker/client"
@@ -39,14 +41,20 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
 	}
 	d := &Driver{factory: f, InitConfig: cfg}
 	for k, v := range cfg.DriverOpts {
-		switch k {
-		case "network":
+		switch {
+		case k == "network":
 			d.netMode = v
 			if v == "host" {
 				d.InitConfig.BuildkitFlags = append(d.InitConfig.BuildkitFlags, "--allow-insecure-entitlement=network.host")
 			}
-		case "image":
+		case k == "image":
 			d.image = v
+		case strings.HasPrefix(k, "env."):
+			envName := strings.TrimPrefix(k, "env.")
+			if envName == "" {
+				return nil, errors.Errorf("invalid env option %q, expecting env.FOO=bar", k)
+			}
+			d.env = append(d.env, fmt.Sprintf("%s=%s", envName, v))
 		default:
 			return nil, errors.Errorf("invalid driver option %s for docker-container driver", k)
 		}
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -29,6 +29,7 @@ type Driver struct {
 	factory driver.Factory
 	netMode string
 	image   string
+	env     []string
 }
 
 func (d *Driver) Bootstrap(ctx context.Context, l progress.Logger) error {
@@ -57,6 +58,7 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 	if d.image != "" {
 		imageName = d.image
 	}
+	env := d.env
 	if err := l.Wrap("pulling image "+imageName, func() error {
 		rc, err := d.DockerAPI.ImageCreate(ctx, imageName, types.ImageCreateOptions{})
 		if err != nil {
@@ -70,6 +72,7 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 
 	cfg := &container.Config{
 		Image: imageName,
+		Env:   env,
 	}
 	if d.InitConfig.BuildkitFlags != nil {
 		cfg.Cmd = d.InitConfig.BuildkitFlags


tonistiigi:
Maybe we could reuse `DriverOpt` for this as they would only be present in certain drivers.

sirlatrom:
> Maybe we could reuse `DriverOpt` for this as they would only be present in certain drivers.

The function for passing in values for DriverOpt does not allow multiple values for the same key; hence the new field.

commands/create.go
```diff
@@ -215,7 +221,7 @@ func csvToMap(in []string) (map[string]string, error) {
 		}
 		for _, v := range fields {
 			p := strings.SplitN(v, "=", 2)
-			if len(p) != 2 {
+			if len(p) < 2 {
```
sirlatrom: I should revert this line if the separate `DriverEnv` field is accepted.

commands/create.go
```diff
@@ -215,7 +221,7 @@ func csvToMap(in []string) (map[string]string, error) {
...
 				return nil, errors.Errorf("invalid value %q, expecting k=v", v)
 			}
 			m[p[0]] = p[1]
```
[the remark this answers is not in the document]
tonistiigi: We could do it either by changing the internal data type to `map[string][]string` or requesting the input as `env.FOO=bar`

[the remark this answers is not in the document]
sirlatrom: Done in ac566dda9c22f0c1bbffa7d6c85e01eb58ee4f3e

commands/create.go
```diff
@@ -215,10 +215,10 @@ func csvToMap(in []string) (map[string]string, error) {
 		}
 		for _, v := range fields {
 			p := strings.SplitN(v, "=", 2)
-			if len(p) != 2 {
+			if len(p) < 2 {
 				return nil, errors.Errorf("invalid value %q, expecting k=v", v)
 			}
-			m[p[0]] = p[1]
+			m[p[0]] = strings.Join(p[1:], "=")
```
tonistiigi: What is this change for?

driver/docker-container/factory.go
```diff
@@ -39,14 +41,20 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
...
+		case k == "network":
 			d.netMode = v
 			if v == "host" {
 				d.InitConfig.BuildkitFlags = append(d.InitConfig.BuildkitFlags, "--allow-insecure-entitlement=network.host")
 			}
-		case "image":
+		case k == "image":
 			d.image = v
+		case strings.HasPrefix(k, "env."):
+			envName := k[4:]
```
tonistiigi: `strings.TrimPrefix` is cleaner

driver/docker-container/factory.go
```diff
@@ -39,14 +41,20 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
...
+			if len(envName) == 0 {
```
tonistiigi: nit: `envName == ""`

commands/create.go
sirlatrom: Environment variable values may contain `=` (and, frankly, `,`, which leads me to think the current approach will not do).

driver/docker-container/factory.go
sirlatrom: Fixed in 046a3b0939a7c7b2c036697a44380b4d7f55151f.

driver/docker-container/factory.go
sirlatrom: Fixed in 046a3b0939a7c7b2c036697a44380b4d7f55151f.

commands/create.go
tonistiigi: It uses `SplitN` so it will never return more than 2 results.

sirlatrom: Fair point, I'll revert this change :+1:

tonistiigi:
@sirlatrom Please squash as well. And avoid putting username mentions in the commit message. I get pinged every time it is pushed or rebased in a fork.

sirlatrom:
> @sirlatrom Please squash as well. And avoid putting username mentions in the commit message. I get pinged every time it is pushed or rebased in a fork.

Squashed and rebased, thanks for reviewing!

spejic88:
when I use `docker buildx create --driver-opt env.http_proxy=$http_proxy --driver-opt env.https_proxy=$https_proxy --driver-opt '"env.no_proxy='$no_proxy'"'`
to set proxy I get `invalid driver option env.http_proxy for docker-container driver`

tavlima:
Hey, @sirlatrom, nice work here. Do you think this could be "ported" to the `remote` driver as well?

sirlatrom:
> Hey, @sirlatrom, nice work here. Do you think this could be "ported" to the `remote` driver as well?

Feel free to borrow what you need from this PR. I'm not currently in a position where I can make additional contributions to this project.

tavlima:
Yeah, I'm aware of what the `remote` driver is, @tonistiigi, but it is just not clear to me yet how the environments (client env, build args, remote env, etc) "interact".

My buildkit nodes run in an environment where all the egress traffic must go through a proxy and I had a hard time getting it to work.

Perhaps something was wrong with my experiments, but AFAIR just making sure the `buildkit` remote nodes had, themselves, the environment variables set (`{NO,HTTP,HTTPS}_PROXY` and their lowercase counterparts) didn't fix the issues while pulling public images. IIRC, I had to also invoke `buildx` like `docker buildx build --builder my_remote_builder --build-arg HTTPS_PROXY=... --build-arg HTTP_PROXY=...`.

Is that expected?

tonistiigi:
Yes, that is expected. To configure proxy for the internal containers created by `RUN` steps, they need to be passed with build-arg. They do not get inherited from the daemon environment.

tavlima:
So, yeah, that is what I was hoping could be addressed in a similar way as this PR. I'd rather not require all my clients to know about the specifics of the cluster environment.

In my case, my team is expected to maintain a cluster with tens of buildkit instances, and I'd rather not have our whole org worrying about what is the right proxy configuration to fetch image XYZ, especially because our environment requires dozens of CIDRs and wildcard domains to be included in the `NO_PROXY` envvar.

Completely abstracting that away from the clients would be ideal. Doing this setup just once (like when "registering" the remote builder in each client) would be an improvement. But having to include `--build-arg` for all proxy environment variables, every time one calls our `--builder`? That will be a hard sell...

tavlima:
FWIW, I just realized what confused me in my experiments: the daemon environment variables are inherited, at least for directives like `FROM`. But, just like you said, @tonistiigi... 

> To configure proxy for the internal containers created by RUN steps, they need to be passed with build-arg

It is really odd that we need to have the same configuration in both places (daemon env and `--build-arg`) in order to build single image (assuming proxies are required for whatever image/command is used in both `FROM`/`RUN`)

PR #184, opened 2019-10-30
Support reading from env on bake --set <t>.args

cpuguy83:
This works just like the `build` command where if you have `--build-arg
FOO`, it will read the variable from env and only set a value if the
variable is defined.


tonistiigi:
@cpuguy83 ci red

cpuguy83:
Fixed

PR #192, opened 2019-11-18
Upgrade to hcl2 to support variables and functions

### LINKED ISSUE #148: HCL interpolation support
error: 
`Error: invalid tag "docker.pkg.github.com/fernandomiguel/buildkit/php:release-${LLB}-${GITHUB_SHA}-${INVOCATION_ID}": invalid reference format`

```hcl
group "default" {
	targets = ["php73", "nginx-base"]
}

target "cache" {
  cache-to = ["type=inline"]
}

target "php73" {
	inherits = ["cache"],
	tags = [
		"docker.pkg.github.com/fernandomiguel/buildkit/php:release-${LLB}-${GITHUB_SHA}-${INVOCATION_ID}",
		"fernandomiguel/php:/php:release-${LLB}-${GITHUB_SHA}-${INVOCATION_ID}"
	]
}
```

vanstee:
* Upgrades hcl parsing to hcl2 package
* Supports variable blocks which are assigned to matching environment variables or default values
* Makes `go-cty` [stdlib functions](https://github.com/zclconf/go-cty/tree/master/cty/function/stdlib) available for use
* Allows for [user defined functions](https://github.com/hashicorp/hcl/tree/hcl2/ext/userfunc)

An example using all the features mentioned above:

```
$ cat <<'EOF' > docker-bake.hcl
function "truncate" {
	params = [string]
	result = "${substr(string, 0, 6)}"
}

variable "TAG" {
	default = "latest"
}

group "default" {
	targets = ["webapp"]
}

target "webapp" {
	tags = ["docker.io/username/webapp:${truncate(TAG)}"]
}
EOF

$ docker buildx bake --print webapp
{
   "target": {
      "webapp": {
         "context": ".",
         "dockerfile": "Dockerfile",
         "tags": [
            "docker.io/username/webapp:latest"
         ]
      }
   }
}

$ TAG=$(git rev-parse HEAD) docker buildx bake --print webapp
{
   "target": {
      "webapp": {
         "context": ".",
         "dockerfile": "Dockerfile",
         "tags": [
            "docker.io/username/webapp:93d5a0"
         ]
      }
   }
}
```

Fixes #148


tonistiigi:
@vanstee What's the state of this? Is it ready for review?

> Can we also support the variables defined in hcl, and should we support helper functions(or even user functions)?

Is this possible? Currently, I couldn't figure out how to provide a default value for a variable. This seems like a quite critical feature to make this useful.

vanstee:
Let me dig through the hcl2 lib a bit more. At first glance, terraform-like variables aren't a part of the core library so to implement that we'll have to use some lower-level apis. But I agree that defining variables with defaults is a critical feature.

* [x] Add examples to the README
* [x] Omit `Name` field from json and `--print` output
* [ ] Define a `Variable` config struct that allows setting defaults that can be overridden by the current environment
* [ ] Use a simple two-pass parsing method to set all necessary variables from defaults or environment overrides
* [ ] Make `go-cty` [stdlib functions](https://github.com/zclconf/go-cty/tree/master/cty/function/stdlib) available for use

Supporting user-defined functions sounds like a good idea. Do you have any use cases in mind that I could test with? This is something that could probably be split out into a follow up ticket as well if needed.

tonistiigi:
> Supporting user-defined functions sounds like a good idea. Do you have any use cases in mind that I could test with? This is something that could probably be split out into a follow up ticket as well if needed.

No specific use case, just seemed an interesting area when I looked at https://github.com/hashicorp/hcl/tree/hcl2/ext/userfunc . Definitely can be follow-up.

vanstee:
Alright, this is ready for a review. User defined functions were easier to add than I thought, so that's now included here as well.

A few minor questions/notes:

Variables are implemented similar to what you'd see in terraform, but are assumed to be strings and only support the `default` attribute (not `description` or `type`). Is this good enough for now?

For now I opted to only include variables from the environment once they've been defined, but we could just always define all environment variables.

User-defined functions are supported, but variables aren't available in that context (stdlib functions are though).

HCL diagnostic errors are returned for things like missing variables or functions. These seem [descriptive enough for now](https://github.com/docker/buildx/pull/192/files#diff-4371e7c5eff6f94d62a0efcb7468e6a8R191), but could be made to better match other `buildx` errors in the future.

tonistiigi:
@tiborvass @AkihiroSuda @cpuguy83 @FernandoMiguel PTAL design.

Design SGTM. Haven't reviewed code yet.

AkihiroSuda:
How is `function "truncate"` called? Couldn't figure out from the example file.

vanstee:
@AkihiroSuda Whoops. Forgot to actually call it in the example in the PR description. Fixed that.

AkihiroSuda:
Thanks, SGTM

cpuguy83:
Wow that's amazing. SGTM.
Had no idea hcl had this stuff.

tonistiigi:
@vanstee Do you want to address any points in https://github.com/docker/buildx/pull/192#pullrequestreview-332894997 or is this ready to merge from your side?

vanstee:
If this is blocking anything else we can go ahead and merge and I'll address those points in a follow up, but I'd like to fix the unshared variable and function context between files since that feels pretty surprising as a user. Just running low on OSS time lately; will probably have that fixed in a week though.

vanstee:
@tonistiigi Rebased! Let me know if this needs anything else before merging.

tonistiigi:
Thanks @vanstee

vanstee:
@cpuguy83 Heads up, hcl2 support is merged

PR #193, opened 2019-11-21
buildx fix type convert compile error on mips64el

XiaodongLoong:
buildx fix type convert compile error;
the type of the "s.Rdev" is 32bit on mip64el, and it cannot convert 'uint32' to 'uint64' implicitly with programming in GO language;
convert the "s.Rdev" type to 'uint64'  explicitly;


tonistiigi:
Update the `moby/moby` dependency in https://github.com/docker/buildx/blob/8bce430f4d74c9a9ca72a46139804e88502d4b78/go.mod#L24 and run `make vendor` to update the files.

PR #219, opened 2020-01-03
Explain binfmt_misc requirements

ArturKlauser:
Getting the QEMU interpreters to work transparently inside containers seems to be a bit of a sticking point, with major distributions still on kernels and/or binfmt-support package versions that don't support the fix_binary flag yet. Give the reader a little more guidance what to look out for.


ArturKlauser:
I added the sentence you suggested.

Unfortunately I haven't found any good pointer for and installation tutorial that actually tells you what the constraints of OS and userspace are that make the difference between an installation that works and doesn't work for use by docker buildx. A while ago I did try to follow a few descriptions of installations, but when they consistently failed on the systems that I tried them on I actually had to learn the hard way what I tried to explain in this PR, i.e. that you need certain versions of the kernel and binfmt-support packages, otherwise following tutorials doesn't get you there. I'll try to write up some post myself, but that might take some time.

tonistiigi:
`docker run --rm --privileged docker/binfmt:66f9012c56a8316f9244ffd7622d7c21c1f6f28d` is the easiest installation instructions. I'm going to merge this but probably would be a good idea to make a new section "Installing QEMU support" in the readme. Move your current addition there with some extra data and link it from the current section. This seems to be a common problem if you look at the issues..

ArturKlauser:
@tonistiigi I've written up a [post](https://nexus.eddiesinentropy.net/2020/01/12/Building-Multi-architecture-Docker-Images-With-Buildx/) describing the Linux environment constraints that need to be met to get docker buildx to work with QEMU. Not sure though if it's appropriate to link to from here and if that is what you had in mind when you suggested to 'point to a tutorial'.

PR #233, opened 2020-03-03
bump to Go 1.13

silvin-lubecki:
Golang 1.13 comes with new restrictions on pseudo-version, so had to fix some of these.
Also bumped buildkit to last `v0.6.4`.


go.mod
```diff
@@ -17,10 +17,10 @@ require (
 	github.com/containerd/containerd v1.4.0-0.20191014053712-acdcf13d5eaf
 	github.com/containerd/typeurl v0.0.0-20190228175220-2a93cfde8c20 // indirect
 	github.com/denisenkom/go-mssqldb v0.0.0-20190315220205-a8ed825ac853 // indirect
-	github.com/docker/cli v1.14.0-0.20191010210043-06f34ba50786
+	github.com/docker/cli v0.0.0-20191017083524-a8ff7f821017
 	github.com/docker/compose-on-kubernetes v0.4.19-0.20190128150448-356b2919c496 // indirect
 	github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
-	github.com/docker/docker v1.14.0-0.20191210192822-1347481b9eb5
+	github.com/docker/docker v1.14.0-0.20190319215453-e7b5f7dbe98c
```
glours: Why don't use `v0.0.0`? insteadt `v1.14.0`  which isn't a `docker/docker` version at all

go.mod
```diff
@@ -29,65 +29,60 @@ require (
...
+	gotest.tools/v3 v3.0.2 // indirect
+	k8s.io/api v0.16.7
+	k8s.io/apimachinery v0.16.7
+	k8s.io/client-go v0.16.7
 	vbom.ml/util v0.0.0-20180919145318-efcd4e0f9787 // indirect
 )
 
+replace github.com/containerd/containerd => github.com/containerd/containerd v1.3.3
+
+replace github.com/docker/docker => github.com/docker/docker v1.4.2-0.20200227233006-38f52c9fec82
```
glours: maybe not need if we use `v0.0.0` version in the declaration above?

go.mod
```diff
@@ -29,65 +29,60 @@ require (
...
-	k8s.io/apimachinery v0.0.0-20180621070125-103fd098999d
-	k8s.io/client-go v2.0.0-alpha.0.0.20180806134042-1f13a808da65+incompatible
+	gotest.tools/v3 v3.0.2 // indirect
+	k8s.io/api v0.16.7
+	k8s.io/apimachinery v0.16.7
+	k8s.io/client-go v0.16.7
 	vbom.ml/util v0.0.0-20180919145318-efcd4e0f9787 // indirect
 )
 
+replace github.com/containerd/containerd => github.com/containerd/containerd v1.3.3
```
tonistiigi: We should use the same versions that are defined in buildkit

silvin-lubecki:
@tonistiigi PTAL, I put the prefixes back and used the same containerd version as buildkit 👍

thaJeztah:
Looks like Jenkins CI here is running on `ci.docker.com` (which was moved to `ci-next.docker.com`) @StefanScherer

thaJeztah:
Looks like it's only the fossa scan in Jenkins (?), so opened https://github.com/docker/buildx/pull/234 to remove it

tonistiigi:
Afaics this takes Buildkit 0.6.4 but buildkit dependencies from master, so it is an untested configuration. (I'm surprised it even builds but I guess buildx doesn't actually use many packages from it). As v0.6.x was never built on 1.13 I'd suggest taking master buildkit as well.

go.mod
```diff
@@ -2,92 +2,74 @@ module github.com/docker/buildx
...
 	github.com/cenkalti/backoff v2.1.1+incompatible // indirect
 	github.com/cloudflare/cfssl v0.0.0-20181213083726-b94e044bb51e // indirect
-	github.com/containerd/console v0.0.0-20191206165004-02ecf6a7291e
-	github.com/containerd/containerd v1.4.0-0.20191014053712-acdcf13d5eaf
-	github.com/containerd/typeurl v0.0.0-20190228175220-2a93cfde8c20 // indirect
+	github.com/containerd/console v0.0.0-20191219165238-8375c3424e4d
+	github.com/containerd/containerd v1.4.0-0
 	github.com/denisenkom/go-mssqldb v0.0.0-20190315220205-a8ed825ac853 // indirect
-	github.com/docker/cli v1.14.0-0.20191010210043-06f34ba50786
+	github.com/docker/cli v1.14.0-0
```
tiborvass: i've never seen this short format. Maybe we should do the same for the docker/docker line ?

tiborvass: Ok I guess I have seen it, just not with `-0`

PR #246, opened 2020-03-25
Add option to build/bake to override instance

cpuguy83:
This helps prevent race conditions with concurrent build invocations.


tonistiigi:
I think this should be a global flag. Eg. like `docker` has `--context` on every command.  Not sure about the flag name, something now quite right about `--use` but can't think of anything better. @tiborvass

tonistiigi:
And if you have time, I guess it would be good to have env for this as well.

cpuguy83:
Every other command, I think, already accepts an argument for the build instance.

tonistiigi:
Not every command. eg https://github.com/docker/buildx/pull/249 does not. I think ENV would be very useful for this.

cpuguy83:
This is updated with the discussed changes.

PR #259, opened 2020-04-21
bake: fix override bug with inheritance

tiborvass:
Added the bug case in tests:

```hcl
target "dep" {
	args {
		VAR_INHERITED = "dep"
	}
}
target "webapp" {
	inherits = ["dep"]
}
```

`bake --set webapp.args.VAR_INHERITED=override` was still inheriting `dep` value.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -23,13 +23,14 @@ func ReadTargets(ctx context.Context, files, targets, overrides []string) (map[s
 		}
 		c = mergeConfig(c, *cfg)
 	}
-	if err := c.setOverrides(overrides); err != nil {
+	o, err := c.newOverrides(overrides)
+	if err != nil {
 		return nil, err
 	}
 	m := map[string]Target{}
 	for _, n := range targets {
 		for _, n := range c.ResolveGroup(n) {
-			t, err := c.ResolveTarget(n)
+			t, err := c.ResolveTarget(n, o)
 			if err != nil {
 				return nil, err
 			}
@@ -105,23 +106,24 @@ func mergeConfig(c1, c2 Config) Config {
 	return c1
 }
 
-func (c Config) setOverrides(v []string) error {
+func (c Config) newOverrides(v []string) (map[string]Target, error) {
+	m := map[string]Target{}
 	for _, v := range v {
 
 		parts := strings.SplitN(v, "=", 2)
 		keys := strings.SplitN(parts[0], ".", 3)
 		if len(keys) < 2 {
-			return errors.Errorf("invalid override key %s, expected target.name", parts[0])
+			return nil, errors.Errorf("invalid override key %s, expected target.name", parts[0])
 		}
 
 		name := keys[0]
 		if len(parts) != 2 && keys[1] != "args" {
-			return errors.Errorf("invalid override %s, expected target.name=value", v)
+			return nil, errors.Errorf("invalid override %s, expected target.name=value", v)
 		}
 
-		t, ok := c.Target[name]
-		if !ok {
-			return errors.Errorf("unknown target %s", name)
+		t := m[name]
+		if _, ok := c.Target[name]; !ok {
+			return nil, errors.Errorf("unknown target %s", name)
 		}
 
 		switch keys[1] {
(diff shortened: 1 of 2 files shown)


bake/bake.go
```diff
@@ -23,22 +23,21 @@ func ReadTargets(ctx context.Context, files, targets, overrides []string) (map[s
...
-	m := map[string]Target{}
 	for _, n := range targets {
 		for _, n := range c.ResolveGroup(n) {
 			t, err := c.ResolveTarget(n)
 			if err != nil {
 				return nil, err
 			}
 			if t != nil {
-				m[n] = *t
+				c.Target[n] = *t
```
tiborvass, line 33: This is incorrect, fixing.

bake/bake_test.go
```diff
@@ -32,9 +35,11 @@ target "webapp" {
 	t.Run("NoOverrides", func(t *testing.T) {
 		m, err := ReadTargets(ctx, []string{fp}, []string{"webapp"}, nil)
 		require.NoError(t, err)
+		require.Equal(t, 1, len(m))
```
tiborvass, line 38: added a couple of hardenings, here and line 77

tiborvass:
Found another bug:

`bake --set webfoo.args.VAR_INHERITED=override` does not override when building `webapp`.

bake/bake_test.go
```diff
@@ -32,42 +35,60 @@ target "webapp" {
...
-		require.Equal(t, m["webapp"].Args["VAR_FROMENV"+t.Name()], "fromEnv")
+			require.Equal(t, m["webapp"].Args["VAR_FROMENV"+t.Name()], "fromEnv")
+
+			require.Equal(t, m["webapp"].Args["VAR_INHERITED"], "override")
+		})
+
+		// building leaf but overriding parent fields
+		t.Run("parent", func(t *testing.T) {
+			m, err := ReadTargets(ctx, []string{fp}, []string{"webapp"}, []string{
+				"dep.args.VAR_INHERITED=override",
```
tiborvass, line 78: this is a regression test

bake/bake.go
```diff
@@ -233,7 +237,7 @@ func (c Config) target(name string, visited map[string]struct{}) (*Target, error
 		}
 	}
 	t.Inherits = nil
-	tt = merge(merge(defaultTarget(), t), tt)
+	tt = merge(merge(merge(defaultTarget(), t), tt), overrides[name])
```
tiborvass, line 240: Note: nothing is setting `c.Target[name]` (aka `t` here), so I believe we could remove the defaultTarget() merge and do `merge(t, tt)` directly.

bake/bake.go
```diff
@@ -233,7 +235,7 @@ func (c Config) target(name string, visited map[string]struct{}) (*Target, error
 		}
 	}
 	t.Inherits = nil
-	tt = merge(merge(defaultTarget(), t), tt)
+	tt = merge(merge(merge(defaultTarget(), tt), t), overrides[name])
```
tiborvass, line 238: @tonistiigi I found another bug (not a regression from this PR). See test with VAR_BOTH. I believe this is correct: `t` is the child target, and `tt` is the parent, we want to merge the child into the parent, and the overrides on top.

bake/bake_test.go
```diff
@@ -32,35 +39,61 @@ target "webapp" {
...
-		require.Equal(t, m["webapp"].Args["VAR_FROMENV"+t.Name()], "fromEnv")
+			require.Equal(t, m["webapp"].Args["VAR_BOTH"], "webapp")
+			require.Equal(t, m["webapp"].Args["VAR_INHERITED"], "override")
+		})
+
+		// building leaf but overriding parent fields
+		t.Run("parent", func(t *testing.T) {
+			m, err := ReadTargets(ctx, []string{fp}, []string{"webapp"}, []string{
+				"dep.args.VAR_INHERITED=override",
+				"dep.args.VAR_BOTH=override",
```
tiborvass, line 91: this one is not, it's a bug in master.

bake/bake.go
tiborvass: An important invariant I can add a comment in the code for, is that `t` should not have the default overrides (context = "." and dockerfile = "."), which is the case here because target() is called before settings those defaults in ResolveTarget.

Issue #269, opened 2020-05-01
bake: HCL does not support pathexpand() function


Cryowatt:
My use case is that I want to do something like this:
```hcl
target "builder" {
  secret = [
    "id=mysecret,src=~/.secrets/mysecret"
  ]
}
```

Docker doesn't resolve `~` in paths, so I tried `"id=mysecret,src=${pathexpand("~/.secrets/mysecret"}"` in hope that the common HCL [functions from Terraform](https://www.terraform.io/docs/configuration/functions/pathexpand.html) also were available here. It failed to parse.

Next I tried just basic interpolation `"id=mysecret,src=${"/users/myuser"}/.secrets/mysecret"}` and it also failed to resolve the path correctly.

So two questions:

1. Is it supposed to work and if not can it be added?
2. Is there a workaround for this? Best option currently seems to be to skip using `docker-bake.hcl` and just script the build commands explicitly in a shell script.

k911:
In your case you could use `$HOME` environment variable that should be available, e.g.:

```hcl
variable "HOME" {
  # value will be overridden by $HOME env variable
  default = "/home/default_user"
}

target "builder" {
  secret = [
    "id=mysecret,src=${HOME}/.secrets/mysecret"
  ]
}
```

As per [docs](https://github.com/docker/buildx#hcl-variables-and-functions), you can find available functions [here](https://github.com/docker/buildx/blob/master/bake/hcl.go#L19-L65)

tiborvass:
It's unclear to me how important pathexpand is. For your particular case, I would agree with @k911's suggestion. If that solves your issue feel free to close, otherwise I'm keeping this open for now for discussion around pathexpand().

tonistiigi:
You should get the function definitions to go-cty first. Then we can include them.

PR #277, opened 2020-05-05
Update go-cty to pull in more stdlib funcs.

cpuguy83:
I needed "split" specifically so I can do something like:

```hcl
variable PLATFORMS {
  default = "linux/amd64"
}

target foo {
  platforms = split(",", "${PLATFORMS}")
  # other stuff
}
```

Where the existing "csvdecode" does not work for this because it parses
the string into a list of objects instead of a list of strings.

I went ahead and just added all the available new functions.

Signed-off-by: Brian Goff <cpuguy83@gmail.com>


bake/hcl.go
```diff
@@ -22,46 +22,76 @@ var (
 		"and":                    stdlib.AndFunc,
 		"byteslen":               stdlib.BytesLenFunc,
 		"bytesslice":             stdlib.BytesSliceFunc,
+		"chomp":                  stdlib.ChompFunc,
+		"chunklist":              stdlib.ChunklistFunc,
+		"ceiling":                stdlib.CeilFunc,
```
cpuguy83, line 27: Should this just be "ceil"?

tiborvass:
I wonder if pathexpand() is also needed, I genuinely don't know. See #269 for a usecase.

bake/hcl.go
tonistiigi: +1

cpuguy83:
I don't see a `PathExpandFunc` in go-cty.

bake/hcl.go
cpuguy83: Done.

bake/hcl.go
```diff
@@ -22,46 +22,76 @@ var (
...
+		"chunklist":              stdlib.ChunklistFunc,
+		"ceil":                   stdlib.CeilFunc,
 		"csvdecode":              stdlib.CSVDecodeFunc,
 		"coalesce":               stdlib.CoalesceFunc,
+		"coalescelist":           stdlib.CoalesceListFunc,
 		"concat":                 stdlib.ConcatFunc,
+		"contains":               stdlib.ContainsFunc,
+		"distinct":               stdlib.DistinctFunc,
 		"divide":                 stdlib.DivideFunc,
+		"elemment":               stdlib.ElementFunc,
```
tonistiigi, line 35: element

cpuguy83: Fixed.

Issue #278, opened 2020-05-07
bake fails to parse JSON Alpine Linux with HCL looking error


jer-k:
Hi, I ran into an issue where I can't build a JSON version of a bake file on an Alpine linux container, but it builds fine on my OSX laptop.

This is the JSON file
```
{                                                                                                                                                                                  
  "group": {                                                                                                                                                                       
    "default": {                                                                                                                                                                   
      "targets":["result","vote","worker"]
      }
    },                                                                                                                                            
    "target":{
      "result":{
        "context":"./result",
        "dockerfile":"Dockerfile",
        "tags":[]
      },                                                                                                     
    "vote":{
      "context":"./vote",
      "dockerfile":"Dockerfile",
      "tags":[]
    },                                                                     
    "worker":{
      "context":"./worker",
      "dockerfile":"Dockerfile",
      "tags":[]
    }
  }
}
```

The result on the container
```
bash-5.0# docker buildx bake -f example-voting-app.json 
examping-voting-app.json:1,1-2: Argument or block definition required; An argument or block definition is required here.
```

I googled `Argument or block definition required; An argument or block definition is required here.` and lots of the results pointed to Terraform issues, thus my thoughts that this is somehow related to HCL, though I'm not quite sure.

On OSX
```
example-voting-app jeremy$ docker buildx bake -f examping-voting-app.json
[+] Building 1.9s (34/34) FINISHED       
```

On the container if I use the `docker-compose.yml` file
```
bash-5.0# docker buildx bake -f docker-compose.yml 
[+] Building 1.8s (34/34) FINISHED  
```

The container information:

The version of Alpine I'm using is 3.11, which comes from here https://github.com/alpinelinux/docker-alpine/blob/c5510d5b1d2546d133f7b0938690c3c1e2cd9549/x86_64/Dockerfile

and the information in the container
```
bash-5.0# uname -a 
Linux 797975ab1268 4.19.76-linuxkit #1 SMP Thu Oct 17 19:31:58 UTC 2019 x86_64 Linux
```

```
bash-5.0# docker version
Client:
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.13.4
 Git commit:        633a0ea838f10e000b7c6d6eed1623e6e988b5bb
 Built:             Mon Nov 25 16:06:35 2019
 OS/Arch:           linux/amd64
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          19.03.8
  API version:      1.40 (minimum version 1.12)
[...]

PR #280, opened 2020-05-08
Support parsing json config with hcl v2

vanstee:
Providing a `docker-bake.json` to `docker buildx bake` fails to parse because we were always assuming `hcl` formatted config after the hcl v2 upgrade.

```
$ cat <<'EOF' > docker-bake.json
{
  "variable": {
    "TAG": {
      "default": "latest"
    }
  },
  "group": {
    "default": {
      "targets": ["webapp"]
    }
  },
  "target": {
    "webapp": {
      "args": {
        "buildno": "${add(123, 1)}"
      }
    }
  }
}
EOF
$ docker buildx bake -f docker-bake.json --print webapp                                                                   
bake.json:1,1-2: Argument or block definition required; An argument or block definition is required here.  
```

Fixed this by using the json parser based on the file suffix.

```
$ docker buildx bake -f docker-bake.json --print webapp                                                                   
{                                                                                                                  
   "target": {                                                                                                     
      "webapp": {                                                                                                  
         "context": ".",                                                                                           
         "dockerfile": "Dockerfile",                                                                               
         "args": {                                                                                                 
            "buildno": "124"                                                                                       
         }                                                                                                         
      }                                                                                                            
   }                                                                                                               
}
```

Also fixes json keys for groups and targets.

Fixes https://github.com/docker/buildx/issues/278

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -72,8 +72,8 @@ func ParseFile(fn string) (*Config, error) {
 
 type Config struct {
 	Variables []*Variable `json:"-" hcl:"variable,block"`
-	Groups    []*Group    `json:"groups" hcl:"group,block"`
-	Targets   []*Target   `json:"targets" hcl:"target,block"`
+	Groups    []*Group    `json:"group" hcl:"group,block"`
+	Targets   []*Target   `json:"target" hcl:"target,block"`
 	Remain    hcl.Body    `json:"-" hcl:",remain"`
 }
 
--- a/bake/hcl.go
+++ b/bake/hcl.go
@@ -74,18 +74,20 @@ type staticConfig struct {
 }
 
 func ParseHCL(dt []byte, fn string) (*Config, error) {
-	var file *hcl.File
-	var diags hcl.Diagnostics
-
-	// Decode user defined functions.
-	fnl := strings.ToLower(fn)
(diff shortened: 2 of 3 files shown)


tonistiigi:
Maybe just fallback on error instead of checking the suffix?

tonistiigi:
@vanstee Also, could you add a test for this?

vanstee:
That could work, but I'm worried we'll be losing out on good error info depending on the default and fallback ordering. For instance, it would be confusing to get errors for parsing an hcl config when you missed a comma in a json file. Let me know what you think; the fallback would be pretty easy to implement.

tonistiigi:
@vanstee I guess if the file has `json` suffix we could just go to json directly. Otherwise, try both and fail with HCL errors. That would be similar to compose detection then https://github.com/docker/buildx/blob/master/bake/bake.go#L54-L70

vanstee:
Nice I like it. Will do.

aloysbaillet:
Hi!
I was updating our CI system to pickup the latest version of `buildx` but we're still using `.json` and I have this exact problem. I'm in no rush to upgrade, but I'd like to know if you're planning on a 4.0.2 version release with this fix?
Cheers,
Aloys

Issue #285, opened 2020-05-12
missing checksums and or GPG signatures for bin files


MaxPeal:
its Best Practices to have code verification with GPG signatures or checksums
https://github.com/docker-library/official-images#security
https://snyk.io/blog/10-docker-image-security-best-practices/

PR #286, opened 2020-05-12
RFC: releases add shasum create for releases bin files

MaxPeal:
the Binary release files are used in many CIs
so shasums and / or gpg sigs are Best Practices.

possible fix for #285 
feedback about the PR is welcome


MaxPeal:
see new PR

PR #288, opened 2020-05-12
RFC: releases add shasum create for releases bin files

MaxPeal:
the Binary release files are used in many CIs
so shasums and / or gpg sigs are Best Practices.

possible fix for #285 
feedback about the PR is welcome

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ env:
 
 script:
   - make binaries validate-all && TARGETPLATFORM="${CROSS_PLATFORMS}" ./hack/cross
-  - for i in ./out/*; do shasum --binary -a 256 "$i" >> "$i.sha256"; done
+  - for i in ./release-out/*; do if [ -f "${i}" ]; then shasum --binary -a 256 "${i}" | tee -a "${i}.sha256"; fi; done; ls -laR ./release-out; ls -laR ./bin;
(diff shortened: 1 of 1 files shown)


tonistiigi:
@MaxPeal CI is not passing.

One problem with this is that we don't have signing/notary for Darwin integrated in CI atm and that is needed to run binaries in newer macos. The way we have done last releases is that after CI has built it I manually pull and sign the mac binary. So with this in mind please exclude darwin binary from the shasums (I guess signing acts as a verification replacement for it anyway)

MaxPeal:
@tonistiigi can you give me a info, about the directory in witch the binaries really are after the full build process, befor it get deployed?

tonistiigi:
In release-out https://github.com/docker/buildx/blob/master/.travis.yml#L29

MaxPeal:
@tonistiigi 
mmm ok, but its not exist after the run is finischd befor the demploment, if i see it corect:
https://travis-ci.org/github/docker/buildx/jobs/694724212#L1517

MaxPeal:
or should i add the creation of the hashum to
https://github.com/docker/buildx/blob/master/Dockerfile#L59
and not to the .travis.yml?

tonistiigi:
We have moved away from travis. Please reopen if you have updated the PR

Issue #315, opened 2020-06-23
ssh context timed out (workaround: ControlPersist)


AkihiroSuda:
```console
$ docker context create my-arm-instance --docker host=ssh://me@my-arm-instance
$ docker buildx ls
NAME/NODE         DRIVER/ENDPOINT  STATUS PLATFORMS
my-arm-instance   docker                  
  my-arm-instance my-arm-instance  listing workers: failed to list workers: rpc error: code = DeadlineExceeded desc = context deadline exceeded
...
```

A workaround is to set ControlPersist in `~/.ssh/config`:
```
ControlMaster auto
ControlPath ~/.ssh/control-%C
ControlPersist yes
```

This workaroud is not needed for `docker --context=my-arm-instance info|version|run`.

BuildX version: v0.4.1

tonistiigi:
Is it just that making a new connection takes longer than the timeout or something else? One way to try would be to make the timeout configurable.

tonistiigi:
Is this duplicate of #358 ?

AkihiroSuda:
Probably yes

PR #323, opened 2020-07-16
Increase ls and inspect timeouts

saulshanabrook:
I was getting a ` listing workers: failed to list workers: rpc error: code = DeadlineExceeded desc = context deadline exceeded
` error when running `docker buildx ls` and `docker buildx inspect` with the kuberentes driver, some of the time:

```bash
$ docker buildx inspect
Name:   cool_fermi
Driver: kubernetes

Nodes:
Name:     buildx0-6dcb5544c6-457mv
Endpoint:
Error:    listing workers: failed to list workers: rpc error: code = DeadlineExceeded desc = context deadline exceeded

$ docker buildx ls
NAME/NODE                  DRIVER/ENDPOINT STATUS PLATFORMS
cool_fermi *               kubernetes
  buildx0-6dcb5544c6-457mv                 listing workers: failed to list workers: rpc error: code = DeadlineExceeded desc = context deadline exceeded
default                    docker
  default                  default         running linux/amd64, linux/arm64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```

I was able to resolve this by increasing the context timeout of both commands from 5 and 7 seconds to 20 seconds. This might be because I am on a satellite internet connection with higher latency to the Kubernetes cluster in the cloud.


tonistiigi:
@AkihiroSuda is this same as #315 ?

Maybe we should make these configurable. 20 sec might be ok but overall this makes the experience worse for the cases where instances never connect.

Another approach might be to try to see if it is in the middle of connecting already. Only cancel with timeout it there was no response at all.

PR #370, opened 2020-08-28
feat: enhance kubernetes driver

morlay:
fix #342 

now we could use kubernetes driver with `--append`

```bash
docker buildx create --use --name=buildkit --platform=linux/amd64 --node=buildkit-amd64 --driver=kubernetes --driver-opt="nodeselector=kubernetes.io/arch=amd64"
docker buildx create --append --name=buildkit --platform=linux/arm64 --node=buildkit-arm64 --driver=kubernetes --driver-opt="nodeselector=kubernetes.io/arch=arm64"
```

```bash
docker buildx inspect buildkit --bootstrap

Name:   buildkit
Driver: kubernetes

Nodes:
Name:      buildkit-amd64
Endpoint:  kubernetes://buildkit/buildkit-amd64
Status:    running
Platforms: linux/amd64*, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6

Name:      buildkit-arm64
Endpoint:  kubernetes://buildkit/buildkit-arm64
Status:    running
Platforms: linux/arm64*, linux/amd64, linux/riscv64, linux/ppc64le, linux/s390x, linux/arm/v7, linux/arm/v6
```

without `binfmt`
```bash
docker buildx inspect buildkit

Name:   buildkit
Driver: kubernetes

Nodes:
Name:      buildkit-amd64
Endpoint:  kubernetes://buildkit/buildkit-amd64
Status:    running
Platforms: linux/amd64*, linux/386

Name:      buildkit-arm64
Endpoint:  kubernetes://buildkit/buildkit-arm64
Status:    running
Platforms: linux/arm64*
```

without `--append`, kubernetes driver works as previous behavior

PR #371, opened 2020-08-31
travis: update base image

tonistiigi:
This should help with the "connecting to buildkit container" issues seen in CI for #370 fyi @morlay 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: bionic
 sudo: required
 
 install:


AkihiroSuda (approved):

AkihiroSuda:
How will this relate to #370?

morlay:
it's not always reproduce.

this error appear only twice. 
image

then i have to push force to re trigger the ci

tonistiigi:
@AkihiroSuda It doesn't. Just coincidence. This error is outside of buildkit afaics. I remember the same thing appeared in buildkit ci as well and it mostly went away after moving away from trusty.

PR #372, opened 2020-08-31
hack/util: take other arches into account on Darwin

thaJeztah:
Noticed that this script hard-coded amd64 on Darwin.

Haven't been able to test the output of `uname -m` on Apple silicon, so if someone's able to verify 🤗 

Also makes some small other changes caught by my linter (see individual commits for details)

### DIFF
--- a/hack/test
+++ b/hack/test
@@ -35,7 +35,7 @@ case $buildmode in
 esac
 
 cacheVolume="buildx-cache"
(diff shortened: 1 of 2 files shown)


thaJeztah:
> Haven't been able to test the output of uname -m on Apple silicon, so if someone's able to verify

Asked internally, and looks like it prints `arm64`, so this change should work with that 👍

Issue #132, opened 2019-08-13
buildx is not a docker command on linux/amd64 ?


pujag-xlnx:
Hi,
I have docker 19.03 on ubuntu amd64 and I tried the steps in your readme but fail to get buildx command working. It complains command not found. What do I need to get it working? I tried beta release as well and same issue. I am looking to compile images for arm64.

Appreciate the help a lot,
Puja
$ cat ~/.docker/config.json
{
"experimental": "enabled"
}
$ ls -l ~/.docker/cli-plugins/docker-buildx
total 55936
drwxr-xr-x 2 pujag newhiredefaultgrp 4096 Aug 12 14:05 .
drwxr-xr-x 3 pujag newhiredefaultgrp 4096 Aug 12 14:01 ..
-rwxr-xr-x 1 pujag newhiredefaultgrp 57036919 Aug 12 14:04 buildx-v0.2.0.linux-amd64
$ docker version
Client: Docker Engine - Community
Version: 19.03.1
API version: 1.40
Go version: go1.12.5
Git commit: 74b1e89
Built: Thu Jul 25 21:21:05 2019
OS/Arch: linux/amd64
Experimental: true

Server: Docker Engine - Community
Engine:
Version: 19.03.1
API version: 1.40 (minimum version 1.12)
Go version: go1.12.5
Git commit: 74b1e89
Built: Thu Jul 25 21:19:41 2019
OS/Arch: linux/amd64
Experimental: false
containerd:
Version: 1.2.6
GitCommit: 894b81a4b802e4eb2a91d1ce216b8817763c29fb
runc:
Version: 1.0.0-rc8
GitCommit: 425e105d5a03fabd737a126ad93d62a9eeede87f
docker-init:
Version: 0.18.0
GitCommit: fec3683

gupta05:
Hi,
I have docker 19.03 on ubuntu amd64 and I tried the steps in your readme but fail to get buildx command working. It complains command not found. What do I need to get it working? I tried beta release as well and same issue. I am looking to compile images for arm64.

Appreciate the help a lot,
Puja 
$ cat ~/.docker/config.json
{
"experimental": "enabled"
}
$ ls -l ~/.docker/cli-plugins/docker-buildx
total 55936
drwxr-xr-x 2 pujag newhiredefaultgrp     4096 Aug 12 14:05 .
drwxr-xr-x 3 pujag newhiredefaultgrp     4096 Aug 12 14:01 ..
[...]

PR #384, opened 2020-09-08
clarified that buildx is not included in all linux distros

admackin:
As noted in #132 there is a lot of confusion from people expecting it to be present. I see [there is code to install it in docker CE](https://github.com/docker/docker-ce/blob/b63a7966fe7bc67dc15403c29f41f060ba8cb44a/components/packaging/plugins/buildx.installer) but it seems that linux distros (at least Amazon Linux for me) don't always include this even with version 19.03. I think this update reflects the true state of things.


README.md
```diff
@@ -45,7 +45,7 @@ Using `buildx` as a docker CLI plugin requires using Docker 19.03. A limited set
 
 ### Docker CE
 
-`buildx` comes bundled with Docker CE starting with 19.03, but requires experimental mode to be enabled on the Docker CLI.
+`buildx` comes bundled by default with Docker Desktop Docker CE starting with 19.03, but may not be present in all repackaged versions for Linux distributions (in which case follow the 'Binary release' instructions below). When it is included, it requires experimental mode to be enabled on the Docker CLI.
```
AkihiroSuda, line 48: "Docker Desktop Docker CE" -> "Docker Desktop"

admackin, line 48:
```suggestion
`buildx` comes bundled by default with Docker Desktop and Docker CE starting with 19.03, but may not be present in all repackaged versions for Linux distributions (in which case follow the 'Binary release' instructions below). When it is included, it requires experimental mode to be enabled on the Docker CLI.
```

prutschman:
Merging this would have saved me a couple hours :|

crazy-max:
@admackin Needs DCO and rebase

README.md
```diff
@@ -49,7 +49,9 @@ Using `buildx` as a docker CLI plugin requires using Docker 19.03 or newer. A li
 
 ### Docker
 
-`buildx` comes bundled with Docker Desktop and in latest Docker CE packages.
+`buildx` comes bundled by default with Docker Desktop, and with Docker CE packages starting with 19.03, but may not be present in all repackaged versions for Linux distributions (in which case follow the 'Binary release' instructions below).
```
admackin, line 52:
```suggestion
`buildx` comes bundled by default with Docker Desktop (and with Docker CE packages) starting with 19.03, but may not be present in all repackaged versions for Linux distributions (in which case follow the 'Binary release' instructions below). 
```

PR #385, opened 2020-09-08
Allow users to have separate store paths

### LINKED ISSUE #308: allow more control to where builder configs are stored
I am using buildx inside a container in GitHub Actions. Since `$DOCKER_CONFIG/buildx` is no inside a container, I have to mount it in order to persist builder config files between invocations of buildx.
It would be nice if builder configs either didn't have to stored on disk at all, or perhaps there existed some way of explicitly passing just a single file around instead of having to share a directory in an implicit way.

At the moment my solution is to have a symlink inside the builx image (`ln -s -f /github/workspace/.buildx ${DOCKER_CONFIG}/buildx`), and ensure that empty directory exists at the top-level of GitHub Actions workspace.

So, right now effectively this is what happens:

```
mkdir -p /github/workspace/.buildx
cd /github/workspace
docker run -v  $(pwd):/github/workspace $buildx_image docker buildx create > /github/workspace/.buildx_builder
docker run -v  $(pwd):/github/workspace  $buildx_image docker buildx build  --builder $(cat .buildx_builder) <flags...>
docker run -v  $(pwd):/github/workspace  $buildx_image docker buildx build  --builder $(cat .buildx_builder) <flags...>
```

It would be nice if instead, I could use something like:

```
mkdir -p /github/workspace/.buildx
cd /github/workspace
docker run -v  $(pwd):/github/workspace $buildx_image docker buildx create --write-config buildx_builder.json
docker run -v  $(pwd):/github/workspace  $buildx_image docker buildx build  --builder-config .buildx_builder.json <flags...>
```

The main challenge is that passing builder by name implies the state is stored locally, it would be nice to be able to pass the state explicitly.

errordeveloper:
- decouple store path from `$DOCKER_CONFIG`
- improve containerised build setup
- introduce new `$BUILDX_CONFIG` environment variable

Related to #308

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -22,12 +22,25 @@ import (
 
 // getStore returns current builder instance store
 func getStore(dockerCli command.Cli) (*store.Txn, func(), error) {
-	dir := filepath.Dir(dockerCli.ConfigFile().Filename)
-	s, err := store.New(dir)
+	s, err := store.New(getConfigStorePath(dockerCli))
 	if err != nil {
 		return nil, nil, err
 	}
 	return s.Txn()
+}
+
+// getConfigStorePath will look for correct configuration store path;
+// if `$BUILDX_CONFIG` is set - use it, otherwise use parent directory
+// of Docker config file (i.e. `${DOCKER_CONFIG}/buildx`)
+func getConfigStorePath(dockerCli command.Cli) string {
+	if buildxConfig := os.Getenv("BUILDX_CONFIG"); buildxConfig != "" {
+		logrus.Debugf("using config store %q based in \"$BUILDX_CONFIG\" environment variable", buildxConfig)
+		return buildxConfig
+	}
+
+	buildxConfig := filepath.Join(filepath.Dir(dockerCli.ConfigFile().Filename), "buildx")
+	logrus.Debugf("using default config store %q", buildxConfig)
+	return buildxConfig
 }
 
 // getCurrentEndpoint returns the current default endpoint value
--- a/store/store.go
+++ b/store/store.go
@@ -15,7 +15,6 @@ import (
 )
 
 func New(root string) (*Store, error) {
-	root = filepath.Join(root, "buildx")
 	if err := os.MkdirAll(filepath.Join(root, "instances"), 0700); err != nil {
 		return nil, err
 	}


commands/util.go
```diff
@@ -22,14 +22,66 @@ import (
...
+	if path == "" {
+		var err error
+		path, err = os.Getwd()
+		if err != nil {
+			logrus.Debugf("unable to get working directory: %s", err.Error())
+			return ""
+		}
+	}
+
+	isRoot := path == "/"
```
errordeveloper, line 65: TODO: find a way to check this on Windows...

errordeveloper:
Two UX questions:
- should new behaviour be an opt-in to begin with, or since buildx is beta, it's reasonable to enable it right away?
- this change will always log store path, should it not log when default is used?

tonistiigi:
I'm ok with the `BUILDX_CONFIG` var but not sure about finding `.buildx` dir. I consider the config directory structure internal but `.buildx` should be something that can be potentially committed to git.

I think loading a builder config from a file or having project-specific builders is a valid use case but probably these should be handled separately.

errordeveloper:
@tonistiigi thanks for the review, and I think you are right. I will revisit this ping you once it is ready for review.

errordeveloper:
@tonistiigi please have another look now!

commands/util.go
```diff
@@ -22,14 +22,27 @@ import (
...
 	}
 	return s.Txn()
 }
 
+// getConfigStorePath will look for correct configuration store path;
+// if `$BUILDX_CONFIG` is set - use it, otherwise use parent directory
+// of Docker config file (i.e. `${DOCKER_CONFIG}/buildx`)
+func getConfigStorePath(dockerCli command.Cli) string {
+	if buildxConfig := os.Getenv("BUILDX_CONFIG"); buildxConfig != "" {
+		logrus.Infof("using config store %q based in \"$BUILDX_CONFIG\" environment variable", buildxConfig)
```
tonistiigi, line 37: could you change these to `Debugf`

errordeveloper: done 👍

tonistiigi (approved):

errordeveloper:
Thanks!

PR #403, opened 2020-09-28
fixes #401 : Replace error generated by `quiet` option by a warning.

### LINKED ISSUE #401: --quiet should not return an error but a warning
As discussed in [Docker Community Slack](https://dockercommunity.slack.com/archives/C7S7A40MP/p1601030935005400) `--quiet` in `docker buildx build` command return an error (from a code point of view), a message telling that quiet is not implemented that stop the build process.

That implementation cause some script to failed when used in combination of [`docker build` aliasing](https://github.com/jygastaud/buildx#setting-buildx-as-default-builder-in-docker-1903).

## Reproducing

* Install buildx
* Set buildx as default builder
* Run `docker build -q .` and you will get the message `quiet currently not implemented` and no build

## Expecting result

From my opinion, `--quiet` flag absence should not be a blocker.
When running `docker build -q .` we should have a warning, then the build still continue.

jygastaud:
As suggested in #401, replace error generated by `--quiet` option by a warning.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -15,6 +15,7 @@ import (
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/util/appcontext"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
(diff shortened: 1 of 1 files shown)


tonistiigi:
@tiborvass Why is it unimplemented. Is it just something that we forgot to remove. `-q` is just print `iidfile`. Progress is already all in stderr now so maybe don't even need to hide it. Could set a none-printer for progress of course but even the simple fix shouldn't break any existing scripts.

thaJeztah:
ping @tiborvass ptal

tonistiigi (approved):

PR #404, opened 2020-09-29
Remove "version" annotations from du and prune commands

thaJeztah:
These annotations were picked up by the YAML docs generator, and showed up as "minimum API version" (https://github.com/docker/docker.github.io/pull/11447). I couldn't find a reference to these annotations in the PR that added them, so thought it would be ok to remove


thaJeztah:
Waiting for CI to see if it was used anywhere 😅

tonistiigi (approved):

PR #425, opened 2020-10-22
allow builder flag to switch to default instance

tonistiigi:
Currently `docker buildx --builder=default build .` produces error `error: open /root/.docker/buildx/instances/default: no such file or directory`. This makes it work properly and produces appropriate error if context name is used instead.

@cpuguy83 


cpuguy83 (approved):
LGTM

PR #428, opened 2020-10-26
fixes #427: Handle empty strings in elements enabling conditional logic

### LINKED ISSUE #427: How to do optional Tags in bake?
I am trying to only add tags if a variable is none empty, primarily fetching if the git commit has a version tag and if so i would like to push that tag. below is an example bake file demonstrating where i am at.

```
variable "TAG" {default="" }

group "default" {
	targets = [
    "app",
  ]
}

target "app" {
  context="."
  dockerfile="Dockerfile"
  tags = [
    "my-image:latest",
    notequal("",TAG) ? "my-image:${TAG}": null,
  ]
}
```

null type is not allowed in arrays. An empty string will throw an error in the docker build. Not sure how to implement optional entries into an array. I also could not find a way to do appending with the tags assign.

wonder if it would be wiser to have empty strings be ignored for tagging to allow empty string results.

zanven42:
this change aims to resolve #427 by allowing either the compact function to be used and also allowing empty strings to be present in the resulting bake file and stripping them out of the final file before actioning.

example file 1 with compact.
```
variable "TAG" {default="" }

group "default" {
	targets = [
    "app",
  ]
}

target "app" {
  context="."
  dockerfile="Dockerfile"
  tags = "${compact([
      "my-image:latest",
      notequal("",TAG) ? "my-image:${TAG}-1": "",
      notequal("",TAG) ? "my-image:${TAG}-2": null,
      null,
    ])}"
}

```

Example 2 with allowing empty strings in arrays will no longer cause any issues with docker push as they are stripped.
```variable "TAG" {default="" }

group "default" {
	targets = [
    "app",
  ]
}

target "app" {
  context="."
  dockerfile="Dockerfile"
  tags = [
      "my-image:latest",
      notequal("",TAG) ? "my-image:${TAG}-1": "",
      notequal("",TAG) ? "my-image:${TAG}-2": "",
  ]
}
```

the above two config files results in the below print
```
❯ docker buildx bake --print
{
   "target": {
      "app": {
         "context": ".",
         "dockerfile": "Dockerfile",
         "tags": [
            "my-image:latest"
         ]
      }
   }
}

❯ TAG="ok" docker buildx bake --print
{
   "target": {
      "app": {
         "context": ".",
         "dockerfile": "Dockerfile",
         "tags": [
            "my-image:latest",
            "my-image:ok-1",
            "my-image:ok-2"
         ]
      }
   }
}```

### DIFF
--- a/README.md
+++ b/README.md
@@ -741,6 +741,43 @@ $ docker buildx bake --print webapp
 }
 ```
 
+Example of only adding tags if a variable is not empty using an `notequal` function:
+
+```
+$ cat <<'EOF' > docker-bake.hcl
+variable "TAG" {default="" }
+
+group "default" {
+	targets = [
+    "webapp",
+  ]
+}
+
+target "webapp" {
+  context="."
+  dockerfile="Dockerfile"
+  tags = [
+      "my-image:latest",
+      notequal("",TAG) ? "my-image:${TAG}": "",
+  ]
+}
+EOF
+
+$ docker buildx bake --print webapp
+{
+   "target": {
+      "webapp": {
+         "context": ".",
+         "dockerfile": "Dockerfile",
+         "tags": [
+            "my-image:latest"
+         ]
+      }
+   }
+}
+```
+
+
 ### `buildx imagetools create [OPTIONS] [SOURCE] [SOURCE...]`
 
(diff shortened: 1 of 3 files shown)


zanven42:
Hi @tonistiigi i would really like to get this into the next release if possible, if anything needs to be changed or modified please let me know.

tonistiigi (approved):

PR #433, opened 2020-11-04
docker-container: ensure credentials are passed when pulling buildkit

tonistiigi:
Anonymous pulls may get rate limited differently so make sure to send user credentials when pulling buildkit image for creating builder container.

If anyone hits this live workaround is to run `docker pull moby/buildkit:buildx-stable-1` manually before `docker buildx create`.

@AkihiroSuda you probably want to make a similar change for kube driver.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -127,7 +127,7 @@ func allIndexes(l int) []int {
 	return out
 }
 
-func ensureBooted(ctx context.Context, drivers []DriverInfo, idxs []int, pw progress.Writer) ([]*client.Client, error) {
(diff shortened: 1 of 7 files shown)


AkihiroSuda:
> you probably want to make a similar change for kube driver.

We could upload ImagePullSecret from local ~/docker/config.json and associate the ImagePullSecret with our deployment.spec.podSpecTemplate.

However, I assume Kube users are highly likely to associate ServiceAccounts with their pre-uploaded ImagePullSecrets, lest they will hit the limit for other DockerHub images.

So I guess we don't need to duplicate ImagePullSecret.

AkihiroSuda (approved):

arkodg (approved):

PR #436, opened 2020-11-05
Git Ignore More Things

kensipe:
Signed-off-by: Ken Sipe <kensipe@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,31 @@
 bin
 cross-out
+
+### mac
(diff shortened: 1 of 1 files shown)


tonistiigi:
If you're using a specific editor, why not put the extra files that editor uses in the global gitignore file instead of all projects trying to cover all the possible editors.

kensipe:
I'll move that way... thanks

PR #444, opened 2020-11-19
build: handle lowercase Dockerfile name as a fallback

### LINKED ISSUE #415: 'docker build' error: "failed to solve with frontend dockerfile.v0"
- [x] I have tried with the latest version of my channel (Stable or Edge)
  - [x] I have uploaded Diagnostics
  - Diagnostics ID: DAA27533-4D93-4AF4-916C-8492CC6BF66B/20201015121621
 
### Expected behavior
I expect to build a docker image from my Dockerfile for my R Shiny application.

### Actual behavior
I get the resulting error from the docker build command:
Capture

### Information
Docker version is:
Capture

Windows Version: **Windows 10 Home, version 1909**

And `docker run hello-world` does work as expected:
Capture

I'm working with **Docker Desktop for Windows 10 Home** and I've followed this guide for the installation:
[Install Docker Desktop on Windows Home](https://docs.docker.com/docker-for-windows/install-windows-home/)

Since I think I have followed the guide correctly, and passed the `docker run hello-world` test, I'm not sure how to begin debugging this error. 

I would really appreciate any help on clarifying what the error means, and how to potentially address it. Thank you!

tonistiigi:
fixes #415

This was supported by the legacy builder: moby/moby#10858

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -796,11 +796,13 @@ func LoadInputs(inp Inputs, target *client.SolveOpt) (func(), error) {
 	if dockerfileName == "" {
 		dockerfileName = "Dockerfile"
 	}
-	target.FrontendAttrs["filename"] = dockerfileName
 
 	if dockerfileDir != "" {
 		target.LocalDirs["dockerfile"] = dockerfileDir
+		dockerfileName = handleLowercaseDockerfile(dockerfileDir, dockerfileName)
 	}
+
+	target.FrontendAttrs["filename"] = dockerfileName
 
 	release := func() {
 		for _, dir := range toRemove {
@@ -888,4 +890,35 @@ func (w *waitingWriter) Close() error {
 		return w.err
 	}
 	return err
+}
+
+// handle https://github.com/moby/moby/pull/10858
+func handleLowercaseDockerfile(dir, p string) string {
+	if filepath.Base(p) != "Dockerfile" {
+		return p
+	}
+
+	f, err := os.Open(filepath.Dir(filepath.Join(dir, p)))
+	if err != nil {
+		return p
+	}
+
+	names, err := f.Readdirnames(-1)
+	if err != nil {
+		return p
+	}
+
+	foundLowerCase := false
+	for _, n := range names {
+		if n == "Dockerfile" {
+			return p
+		}
+		if n == "dockerfile" {
+			foundLowerCase = true
+		}
+	}
+	if foundLowerCase {
+		return filepath.Join(filepath.Dir(p), "dockerfile")
+	}
+	return p
 }


build/build.go
```diff
@@ -796,12 +796,14 @@ func LoadInputs(inp Inputs, target *client.SolveOpt) (func(), error) {
 	if dockerfileName == "" {
 		dockerfileName = "Dockerfile"
```
haker19904, line 797: 6

build/build.go
```diff
@@ -889,3 +891,34 @@ func (w *waitingWriter) Close() error {
...
+	if filepath.Base(p) != "Dockerfile" {
+		return p
+	}
+
+	f, err := os.Open(filepath.Dir(filepath.Join(dir, p)))
+	if err != nil {
+		return p
+	}
+
+	names, err := f.Readdirnames(-1)
```
thaJeztah, line 906: Was wondering why you were using `f.Readdirnames` here, but then realised`os.Stat()` returns the name you pass to it, not the actual filename on disk (preserving case) 😞 

Wondering, if `Readdirnames()` would be an issue if there's lots of files in the directory.

Would something like this work?

```go
dockerfiles, err := filepath.Glob(p + "/[Dd]ockerfile")
```

I did a quick test on my machine;

```go
package main

import (
	"fmt"
	"path/filepath"
)

func main() {
	dockerfileDir, _ := filepath.Abs(".")
	dockerfileName := "Dockerfile"
	f := handleLowercaseDockerfile(dockerfileDir, dockerfileName)
	fmt.Println("Using:", f)
}

// handle https://github.com/moby/moby/pull/10858
func handleLowercaseDockerfile(dir, p string) string {
	if filepath.Base(p) != "Dockerfile" {
		return p
	}
	dockerfiles, err := filepath.Glob(filepath.Join(dir, "[Dd]ockerfile"))
	if err != nil || len(dockerfiles) != 1 {
		return p
	}
	// either Dockerfile or dockerfile was found; use what's found
	return filepath.Base(dockerfiles[0])
}
```

```bash
rm -f Dockerfile && touch dOckErFILE && go run main.go
rm -f Dockerfile && touch Dockerfile && go run main.go
rm -f Dockerfile && touch dockerfile && go run main.go

Using: Dockerfile
Using: Dockerfile
Using: dockerfile
```

tonistiigi: `Glob` is implemented with `Readdirnames(-1)`

thaJeztah: sigh 😔

it really surprised me that os.Stat doesn't return the actual filename

thaJeztah (approved):
LGTM

PR #368, opened 2020-08-28
feat: use k8s cluster config when buildx used in k8s cluster

morlay:
grant permission to `service account` of pod.

PR #460, opened 2020-12-03
fix: buildx in k8s pod with kube client config in cluster

morlay:
Signed-off-by: Wang Jinglei <morlay.null@gmail.com>

use kube config in cluster not work with last PR https://github.com/docker/buildx/pull/368

sorry for checking so late.

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -203,9 +203,23 @@ func driversForNodeGroup(ctx context.Context, dockerCli command.Cli, ng *store.N
 						logrus.Error(err)
 					}
 				}
+
+				tryToUseKubeConfigInCluster := false
 				if kcc == nil {
-					kcc = driver.KubeClientConfigInCluster{}
+					tryToUseKubeConfigInCluster = true
+				} else {
+					if _, err := kcc.ClientConfig(); err != nil {
+						tryToUseKubeConfigInCluster = true
+					}
 				}
+				if tryToUseKubeConfigInCluster {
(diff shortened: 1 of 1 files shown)


commands/util.go
```diff
@@ -203,9 +203,23 @@ func driversForNodeGroup(ctx context.Context, dockerCli command.Cli, ng *store.N
 						logrus.Error(err)
 					}
 				}
+
+				tryToUseKubeConfigInCluster := false
 				if kcc == nil {
-					kcc = driver.KubeClientConfigInCluster{}
+					tryToUseKubeConfigInCluster = true
+				} else {
+					if _, err := kcc.ClientConfig(); err != nil {
```
AkihiroSuda, line 211: Wouldn't hurt to logrus.Debug(err) here?

AkihiroSuda (approved):

commands/util.go
morlay: i found kcc not be nil in cluster pod.

but not found other way to check local kube config exists.

not sure which logrus.Debug(err) you means? I haven't notice logrus.Debug in Client Config

AkihiroSuda: I thought printing the err with `logrus.Debug(err)` might be useful for debugging potential failures, but either is fine to me

tonistiigi:
@morlay needs rebase (and maybe update)

morlay:
@tonistiigi updated

tonistiigi (approved):

PR #451, opened 2020-11-24
Refactor validate for GitHub Actions

crazy-max:
This PR is the beginning of a best-effort to migrate to GitHub Actions since travis-ci.org [will be officially closed down completely on December 31st, 2020](https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom?e=%5BUNIQID%5D). Migrate to [travis-ci.com will require a limited number of credits](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing):

> ...we will upgrade you to our trial (free) plan with a 10K credit allotment (which allows around 1000 minutes in a Linux environment).

While waiting to switch to GitHub Actions, we should consider contacting Travis about [their OSS credits plan](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing):

> We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment. Please include:
> 
> * Your account name and VCS provider (like travis-ci.com/github/[your account name] )
> * How many credits (build minutes) you’d like to request (should your run out of credits again you can repeat the process to request more or discuss a renewable amount)

Ref.

* moby/buildkit#1736
* moby/buildkit#1821
* moby/buildkit#1824
* moby/buildkit#1829
* moby/buildkit#1831
* moby/buildkit#1833
* moby/buildkit#1835

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #463, opened 2020-12-05
GitHub Actions for test

crazy-max:
Follow-up #451 

* `make test` GHA compatible
* Add local cache for GHA (binaries)
* Add badges in README
* Add godev workflow to "ping" pkg.go.dev for changes when a release is produced (see [Adding a package](https://go.dev/about)).
* Add codecov

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,90 @@
+name: build
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - 'v*'
+  pull_request:
+    branches:
+      - 'master'
+
+env:
+  REPO_SLUG_ORIGIN: "moby/buildkit:master"
+  CACHEKEY_BINARIES: "binaries"
+
+jobs:
+  base:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Cache ${{ env.CACHEKEY_BINARIES }}
+        uses: actions/cache@v2
+        with:
+          path: /tmp/.buildx-cache/${{ env.CACHEKEY_BINARIES }}
+          key: ${{ runner.os }}-buildx-${{ env.CACHEKEY_BINARIES }}-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-${{ env.CACHEKEY_BINARIES }}-
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          driver-opts: image=${{ env.REPO_SLUG_ORIGIN }}
+      -
+        name: Build ${{ env.CACHEKEY_BINARIES }}
+        run: |
+          ./hack/build_ci_first_pass binaries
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+      - 'master'
+    tags:
+      - 'v*'
+  pull_request:
+    branches:
+      - 'master'
+
+env:
+  REPO_SLUG_ORIGIN: "moby/buildkit:master"
+  PLATFORMS: "linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64,linux/s390x,linux/ppc64le"
```
tonistiigi, line 18: looks unused

crazy-max: I'm a little ahead of schedule :sweat_smile:

codecov.yml
```diff
@@ -0,0 +1 @@
+comment: false
```
tonistiigi, line 1: want to make same changes as https://github.com/moby/buildkit/pull/1878 for this?

hack/test
```diff
@@ -3,51 +3,39 @@
...
+: ${TEST_COVERAGE=}
+
+importCacheFlags=""
+if [ -n "$cacheRefFrom" ]; then
+  if [ "$cacheType" = "local" ]; then
+    for ref in $cacheRefFrom; do
+      importCacheFlags="$importCacheFlags--cache-from=type=local,src=$ref "
+    done
+  else
+    importCacheFlags="--cache-from=type=registry,ref=$cacheRefFrom:integration-tests "
```
tonistiigi, line 15: don't think image like `$cacheRefFrom:integration-tests` exists

hack/test
```diff
@@ -3,51 +3,39 @@
...
-docker run --rm -v /tmp --volumes-from=$cacheVolume --privileged $iid go test ${TESTFLAGS:--v} ${TESTPKGS:-./...}
+buildxCmd build $importCacheFlags \
+  --target "integration-tests" \
+  --output "type=docker,name=$iid" \
+  $currentcontext
 
-if [ -n "$BUILDX_NOCACHE" ]; then
-  docker rm -v $cacheVolume
-fi
+cacheVolume=$(docker create -v /root/.cache -v /root/.cache/registry -v /go/pkg/mod alpine)
```
tonistiigi, line 36: what are the registry references for. same for next line

tonistiigi, line 36: previous code has capabilities for reusing the previous cache. without it separate volumes do not add much.

.github/workflows/build.yml
crazy-max: @tonistiigi Done

hack/update-vendor
```diff
@@ -1,45 +1,16 @@
 #!/usr/bin/env bash
 
 . $(dirname $0)/util
-set -eu -o pipefail -x
+set -eu
```
tonistiigi, line 4: could we put non-test things to separate commit.

codecov.yml
crazy-max: I suspect that the issue on buildkit is related to testing on the windows environment (test-os job). I think we can leave it here and see what happens in the future.

hack/test
crazy-max: Good point, none actually

tonistiigi (approved):

PR #467, opened 2020-12-07
Get multi-platform buildkit frontend opt from args

### LINKED ISSUE #158: Option to disable platform output namespace?
I have some some build targets setup for multiple platforms which takes care of namespacing outputs itself e.g. `COPY --from=build /path/to/build/bin/* /bin/${TARGETPLATFORM}/`

However when building with buildx and multiple platforms (and seemingly only multiple platforms?) buildx namespaces further as `<TARGETOS>_<TARGETARCH>[_<TARGEVARIANT>]/<OUTPUT>/` which ends up being redundant and requires an extra merge step.

Hoping to have some flag or some way to signal to not do the top-level namespacing.

cpuguy83:
This allows builders to opt into determnistic output regardless of
multi-platform output or not.

This resolves the real intention behind 158; deterministic output.

Closes #158

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -331,6 +331,12 @@ func toSolveOpt(d driver.Driver, multiDriver bool, opt Options, dl dockerLoadCal
 		AllowedEntitlements: opt.Allow,
 	}
 
+	if v, ok := opt.BuildArgs["BUILDKIT_MULTI_PLATFORM"]; ok {
+		if v, _ := strconv.ParseBool(v); v {
+			so.FrontendAttrs["multi-platform"] = "true"
+		}
+	}
+
 	if multiDriver {
 		// force creation of manifest list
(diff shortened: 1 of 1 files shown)


tonistiigi:
if you want to set the frontend opt you should do it directly. `multiDriver` has a different meaning where build is in multiple nodes and a merge needs to happen from buildx cli.

tonistiigi:
nit while you fix the rest: typo in commit title

tonistiigi (changes requested):
^

tonistiigi (approved):

tonistiigi:
@cpuguy83 I think we should add this into the Dockerfile frontend as well. Otherwise, `buildx` behavior is inconsistent from other tools using BuildKit.

PR #475, opened 2020-12-15
driver: do not insert "platform" as driver-opt

tonistiigi:
Addresses https://github.com/docker/setup-buildx-action/issues/45

Simple repro:
```
$ buildx create --platform linux/amd64 --use
$ buildx build - <<EOF
from scratch
EOF
```

Since https://github.com/docker/buildx/pull/370 a `platform` driver-opt was automatically inserted with the value specified by `--platform` flag on regardless of the type of driver, even though it was only used in the kubernetes driver. However, because the docker-container driver is pedantic about the options being passed, it errored out.

Another side-effect I suspect is that with the kubernetes driver it was now possible to specify the platforms in two different ways: `--driver-opt platform=...` and `--platform`.

This patch reverts completely the `platform` driver-opt and instead ensures the platforms information is passed onto the kubernetes driver via variables.

Signed-off-by: Tibor Vass <tibor@docker.com>

cc @morlay

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/driver"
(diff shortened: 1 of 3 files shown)


tonistiigi:
LGTM

morlay:
This pr works well. with kubernetes
```
docker buildx create --use --name buildkit --platform=linux/amd64 --node=buildkit-amd64 --driver=kubernetes --driver-opt=namespace=gitlab
docker buildx create --append --name buildkit --platform=linux/arm64 --node=buildkit-arm64 --driver=kubernetes --driver-opt=namespace=gitlab
```

tiborvass (approved):
Gotta try approving my own PR :P

crazy-max (approved):
LGTM

PR #489, opened 2020-12-19
Allow to read env from dotenv file

### LINKED ISSUE #282: docker buildx bake should apply env variable from .env
docker buildx should read .env file like compose. Kudos

docker-compose.yml
```
version: '3.6'

services:
    ssr:
        build:
            context: .
            target: ssr
            args:
                COMMIT_SHA: ${COMMIT_SHA}
```
.env
```
COMMIT_SHA=a06baf1a
```

docker-compose config
```
services:
  ssr:
    build:
      args:
        COMMIT_SHA: a06baf1a
      context: 
      target: ssr
version: '3.6'
```

docker buildx bake --print
```
{
   "target": {
      "ssr": {
         "context": ".",
         "dockerfile": "Dockerfile",
         "args": {
            "COMMIT_SHA": ""
         },
         "target": "ssr"
      }
   }
}
```

crazy-max:
Fixes #282

### DIFF
--- a/bake/env.go
+++ b/bake/env.go
@@ -1,0 +1,27 @@
+package bake
+
+import (
+	"os"
+	"strings"
+
+	"github.com/imdario/mergo"
+	"github.com/joho/godotenv"
+)
+
+func readEnv() (envs map[string]string, err error) {
+	envs, _ = godotenv.Read()
+	err = mergo.Merge(&envs, envMap(os.Environ()), mergo.WithOverride)
+	return
+}
+
+func envMap(env []string) map[string]string {
+	result := make(map[string]string, len(env))
+	for _, s := range env {
+		kv := strings.SplitN(s, "=", 2)
+		if len(kv) != 2 {
+			continue
+		}
+		result[kv[0]] = kv[1]
+	}
+	return result
+}
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"reflect"
-	"strings"
 
 	"github.com/docker/cli/cli/compose/loader"
 	composetypes "github.com/docker/cli/cli/compose/types"
@@ -15,26 +14,20 @@ func parseCompose(dt []byte) (*composetypes.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	envs, err := readEnv()
+	if err != nil {
+		return nil, err
+	}
+
 	return loader.Load(composetypes.ConfigDetails{
 		ConfigFiles: []composetypes.ConfigFile{
 			{
 				Config: parsed,
 			},
 		},
-		Environment: envMap(os.Environ()),
+		Environment: envs,
 	})
-}
-
-func envMap(env []string) map[string]string {
-	result := make(map[string]string, len(env))
-	for _, s := range env {
-		kv := strings.SplitN(s, "=", 2)
-		if len(kv) != 2 {
-			continue
-		}
-		result[kv[0]] = kv[1]
-	}
-	return result
 }
 
 func ParseCompose(dt []byte) (*Config, error) {
--- a/bake/hcl.go
+++ b/bake/hcl.go
@@ -1,7 +1,6 @@
 package bake
 
 import (
-	"os"
 	"strings"
 
 	"github.com/hashicorp/go-cty-funcs/cidr"
@@ -189,9 +188,11 @@ func parseHCL(dt []byte, fn string) (_ *Config, err error) {
 	}
 
 	// Override default with values from environment.
-	for _, env := range os.Environ() {
-		parts := strings.SplitN(env, "=", 2)
-		name, value := parts[0], parts[1]
+	envs, err := readEnv()
+	if err != nil {
+		return nil, err
+	}
+	for name, value := range envs {
 		if _, ok := variables[name]; ok {
 			variables[name] = cty.StringVal(value)
 		}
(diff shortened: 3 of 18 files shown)


bake/env.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"os"
+	"strings"
+
+	"github.com/imdario/mergo"
+	"github.com/joho/godotenv"
+)
+
+func readEnv() (envs map[string]string, err error) {
+	envs, _ = godotenv.Read()
```
tonistiigi, line 12: @thaJeztah I know there were some weirdnesses in compose env file. Is this the correct way to read that file? Or at least the way we want to support it.

crazy-max: `.env` file processing has been intriduced in docker-compose since [1.7.0 (2016-04-13)](https://github.com/docker/compose/blob/8633939080095bc81f03b8a6a517d5c83d0dee93/CHANGELOG.md#170-2016-04-13) and looks for `.env` file in the directory where it's run and reads any environment variables defined inside through the `python-dotenv`.

In this PR `godotenv` module has the same behavior as the python lib. The only difference with compose is if the The `--project-directory` option is used in compose, it will resolve the `.env` file only in that path (since [1.23.0](https://github.com/docker/compose/blob/8633939080095bc81f03b8a6a517d5c83d0dee93/CHANGELOG.md#1230-2018-10-30)).

thaJeztah: Sorry for the delay.

Hmm, yes, I have thoughts on `.env`. I do agree we should try to make `bake` be compatible with `docker-compose build`, to allow existing projects to be built, but we should probably have a close look "how" to achieve this.

The `.env` file _in general_ is a bit problematic (IMO);

1. the `.env` file is not specific to docker-compose

    A project could happen to have a `.env` file (for the project itself) that is not intended to be used by Compose. In these situations, users will have to use workarounds to make Compose ignore the file (https://github.com/docker/compose/issues/6741)

2. the format is under-defined (and validation was too relaxed)

    For example, compose at some point accepted environment-variables with white-space in their name, and lines having `export`, mostly because validation was lacking, and when validation was added (https://github.com/docker/compose/pull/6403), this broke existing users (see https://github.com/docker/compose/issues/6511). While supporting (or "ignoring") `export` lines could be "ok", it may also set the expectation that other shell-like constructs are possible, but they are not (https://github.com/docker/compose/pull/7598)

3. the `.env` file is not associated with a specific docker-compose file

    Not sure what the current state is for compose, but (see https://github.com/docker/compose/issues/7928#issuecomment-733986258) docker compose traverses parent directories _both_ for the compose file, and for the `.env` file, which will result in confusing behavior.

4. Naming

   This is more of an implementation detail in Compose; naming is hard. There's some ambiguity between `.env` and `env_file` (equivalent to `docker run --env-file`). Compose has a top-level `--env-file` flag that allows specifying an alternative `.env` file. Often, users confuse the Compose `--env-file` flag with the `docker run` `--env-file`, which has a very different meaning.

thaJeztah: So, back to making `buildx bake` compatible with Docker Compose projects.

- Perhaps we should _NOT_ automatically load `.env`, but require a path to be specified for a file to load env-vars from. Doing so prevents the ambiguity about "where" to load the `.env` from (same dir as compose-file?, project-dir? any "parent" dir that contains a `.env` ?). I think this makes things more flexible (e.g. users can use an `.env` file with any build-definition file, not just when using `docker-compose.yml`), and we don't paint ourself in a corner.
- Make sure to document the file-format and behavior; mostly thinking of:
    - wether or not we ignore `export FOO=bar` lines (do we print a warning? do we silently ignore? do we error out?)
    - wether or not the file support substitution (`FOO=${BAR:-hello}`), and if so, which variants thereof (e.g. "required" variables)
    - wether or not the file supports multi-line environment variables
    - if we do expand variables, what's the default/behavior if the env-var is not set? (`FOO=$NOSUCHVAR`); empty string? `FOO` "not set"?
    - does the format have the same semantics as `--env-file`? (`FOO` means "take value from `$FOO` if it exists, or don't set `FOO` otherwise, and `FOO=` means "set `$FOO` to an empty string)

tonistiigi: @crazy-max What do you think about adding a flag instead (and maybe a warning if file exists but flag not used).

@chris-crone Any thoughts about the issues about the format issues of the file @thaJeztah listed? Also, I guess `compose` itself is not interested in fixing the ambiguity issues?

crazy-max: @thaJeztah 

> Compose has a top-level `--env-file` flag that allows specifying an alternative `.env` file.

Didn't know about that flag, that could be confusing as you say.

> Perhaps we should NOT automatically load .env, but require a path to be specified for a file to load env-vars from.

I'm ok with a new flag to load env vars to avoid any confusion or misleading about "default" paths. I think we should discuss about that and find a common implementation to use with [compose-cli](https://github.com/docker/compose-cli) (docker/compose-cli#1152)?

> Make sure to document the file-format and behavior; mostly thinking of

Agree about documentation. If we use godotenv lib we can use a similar usage doc as [this one](https://github.com/bkeepers/dotenv#usage) to start with.

tonistiigi: I think we should also start to move over to https://github.com/compose-spec/compose-go/ . Had a quick look and seems it has some env file support builtin in parser. Not sure if we still want changes for the default `.env` behavior because of the issues above. I guess https://github.com/docker/cli/tree/master/cli/compose is now deprecated although don't see it mentioned. @ndeloof

ndeloof: cli/compose is not deprecated but is only relevant in the context of `stack deploy` command and should not be used in another context. compose-go has diverged in the meantime to match compose-spec needs, but should be preferred as a "compose library".

crazy-max:
I think we can close this PR since #541 and use compose-go in a follow-up.

PR #497, opened 2020-12-30
feat: store kube config file to make buildx builder switchable

morlay:
issue from slack by David Giffin

> Is there a way to pass a kubeconfig to "docker buildx"? I need to be able to switch between clusters / builders easily. It looks like buildx is using the default context / cluster when creating buildkit deployments, running builds, etc

failed use docker context, the auth way not support only `username` & `password` (https://github.com/docker/cli/blob/master/cli/context/kubernetes/load.go#L73-L74), 
which will break other auth ways like `client-key` & `client-cert`.

just added query parameter config-file  (from `KUBECONFIG` when builder created). and dynamic load kube config when use the kubernetes builder.

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"encoding/csv"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 
@@ -145,7 +146,14 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
 
 		if in.driver == "kubernetes" {
 			// naming endpoint to make --append works
-			ep = fmt.Sprintf("%s://%s?deployment=%s", in.driver, in.name, in.nodeName)
+			ep = (&url.URL{
+				Scheme: in.driver,
+				Path:   "/" + in.name,
+				RawQuery: (&url.Values{
+					"deployment": {in.nodeName},
+					"kubeconfig": {os.Getenv("KUBECONFIG")},
+				}).Encode(),
+			}).String()
 		}
 
 		m, err := csvToMap(in.driverOpts)
--- a/commands/util.go
+++ b/commands/util.go
@@ -2,8 +2,10 @@ package commands
 
 import (
 	"context"
+	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/docker/buildx/build"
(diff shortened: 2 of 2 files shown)


commands/create.go
```diff
@@ -145,7 +146,14 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
 
 		if in.driver == "kubernetes" {
 			// naming endpoint to make --append works
-			ep = fmt.Sprintf("%s://%s?deployment=%s", in.driver, in.name, in.nodeName)
+			ep = (&url.URL{
+				Scheme: in.driver,
+				Path:   "/" + in.name,
+				RawQuery: (&url.Values{
+					"deployment":  {in.nodeName},
+					"config-file": {os.Getenv("KUBECONFIG")},
```
AkihiroSuda, line 154: "kubeconfig" rather than "config-file" might be more clear

commands/util.go
```diff
@@ -12,11 +14,14 @@ import (
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/context/docker"
 	"github.com/docker/cli/cli/context/kubernetes"
+	ctxstore "github.com/docker/cli/cli/context/store"
 	dopts "github.com/docker/cli/opts"
+	apis "github.com/docker/compose-on-kubernetes/api"
```
AkihiroSuda, line 19: Can we eliminate dependency on docker/compose-on-kubernetes?

commands/create.go
morlay: `config-file` from https://docs.docker.com/engine/reference/commandline/context_create/#create-a-context-with-a-docker-and-kubernetes-endpoint 😆 


should follow ? or rename to `kubeconfig` ?

commands/util.go
morlay: it is a deps in `github.com/docker/cli/cli/context/kubernetes`.

I will the use clientcmd instead.

commands/create.go
morlay: updated to `kubeconfig`

AkihiroSuda: Hmm, then config-fine makes sense as well... Either is fine to me 😅

AkihiroSuda (approved):

PR #504, opened 2021-01-11
Allow to set default progress through env var

crazy-max:
In my case I want to always use plain to show container output without specifying `--progress` flag every time. This PR allows to set the default `progress` value through the `BUILDX_PROGRESS_DEFAULT` env var.


tonistiigi (approved):

tonistiigi:
I'm also ok with setting the preferred default for known CI systems.

PR #91, opened 2019-06-06
[WIP] Add YAML-docs generator

thaJeztah:
opening as a WIP PR

PR #506, opened 2021-01-13
Split docs to separate files

thaJeztah:
This moves the reference docs to the docs directory, splitting the readme to dedicated markdown files for each subcommand.

See individual commits for more details; with this change, https://github.com/docker/buildx/pull/91 is able to include the extended descriptions and examples in the generated YAML docs, which are used for docs.docker.com; I opened a draft PR that updates the docs based on this pull request: https://github.com/docker/docker.github.io/pull/12092; preview here; https://deploy-preview-12092--docsdocker.netlify.app/

- build: https://deploy-preview-12092--docsdocker.netlify.app/engine/reference/commandline/buildx_build/#examples
- bake: https://deploy-preview-12092--docsdocker.netlify.app/engine/reference/commandline/buildx_bake/


docs/reference/buildx.md
```diff
@@ -0,0 +1,32 @@
+# buildx
```
thaJeztah, line 1: Note; moving the docs to be in `docs/reference`; this is more in line with other repositories that generate yaml docs (for integration in docs.docker.com); having these in a separate directory keeps the top `docs/` directory available for other (non-reference) docs. If we don't anticipate other docs to be added, we could move to the top-level `docs/`

README.md
```diff
@@ -159,697 +159,6 @@ Currently, the bake command supports building images from compose files, similar
 
 There is also support for custom build rules from HCL/JSON files allowing better code reuse and different target groups. The design of bake is in very early stages and we are looking for feedback from users.
```
thaJeztah, line 160: Probably could move the "introduction" part to be included in the `buildx` top-level command, but can do that in a follow-up.

README.md
```diff
@@ -29,16 +29,16 @@ Key features:
...
-    + [`buildx create [OPTIONS] [CONTEXT|ENDPOINT]`](#buildx-create-options-contextendpoint)
-    + [`buildx use NAME`](#buildx-use-name)
-    + [`buildx inspect [NAME]`](#buildx-inspect-name)
-    + [`buildx ls`](#buildx-ls)
-    + [`buildx stop [NAME]`](#buildx-stop-name)
-    + [`buildx rm [NAME]`](#buildx-rm-name)
-    + [`buildx bake [OPTIONS] [TARGET...]`](#buildx-bake-options-target)
-    + [`buildx imagetools create [OPTIONS] [SOURCE] [SOURCE...]`](#buildx-imagetools-create-options-source-source)
-    + [`buildx imagetools inspect NAME`](#buildx-imagetools-inspect-name)
+    + [`buildx build [OPTIONS] PATH | URL | -`](docs/reference/buildx_build.md)
```
thaJeztah, line 32: Perhaps we should simplify these to be just `buildx build` (not listing the options/usage)

crazy-max:
@thaJeztah That's really cool thanks for that! See also #504 about a recent change with the `--progress` flag for `buildx build` and `buildx bake`. Maybe add a note about that new env var?

thaJeztah:
> See also #504 about a recent change with the --progress flag for buildx build and buildx bake. Maybe add a note about that new env var?

Oh! Thanks for the link. We need to be a bit creative, because that option is not yet in the version of buildx that's currently shipping with Docker Desktop. I can create some follow-up PRs after this, and we can time merging those in the right order (so that I can generate the docs from master until these changes are in a release).

crazy-max:
@thaJeztah Yes exactly not yet released, like you said a follow-up PR is a good move everytime the doc is impacted. Btw, there is also [BUILDX_BUILDER](https://github.com/docker/buildx/blob/d1a46faf8475564fbb27c9228b79a806e59b8f65/commands/root.go#L55) to take into account.

tonistiigi:
I think having the command outputs copy-pasted from the user's terminals is very error-prone. If we are going with this way then please add a script that generates them and validates that they are up-to-date in the CI. Should be easy to run buildx in a container or in a Dockerfile and copy outputs to the scratch stage.

I do think having separate pages affects the readability, especially in github where there is no sidebar. We should add extra links back to the TOC from a Documentation section in the page so if someone reads through the page they realize there is more context in the separate pages. If possible we could also consider adding markers to the main readme instead of separate pages to have the possibility to separate commands .

dieucao:
Was chatting with @tonistiigi and I wondered if it might make sense to have a single page with multiple section links similar to how the Dockerfile reference is laid out at https://docs.docker.com/engine/reference/builder/ with section links along the right side, instead of splitting into multiple pages, as that seems a little closer to how it's laid out on the buildx Readme currently. @usha-mandya thoughts?

usha-mandya:
Dieu, thanks for looping me in. I believe a separate page for each sub command makes sense as a single page with all the commands and their extended descriptions might increasing the page scroll. With a page each for the subcommands, readers will still be able to access other commands using the left nav. This is also inline with all other pages in the CLI reference section.

tonistiigi:
Added generators from code and CI validation. Let's get this merged.

@thaJeztah Didn't make text changes. What's left is TOC visibility in the main readme and fix the titles of the flag definitions. These are not "examples" (that are not organized in any way so the only way to find information is to read through the whole page), but descriptions for the flags. Of course, flag descriptions can have examples and there can be general examples in a special section. If you need any more markers or yaml output, they can be added (probably in follow-up).

docs/reference/buildx_build.md
```diff
@@ -0,0 +1,271 @@
...
+The `buildx build` command starts a build using BuildKit. This command is similar
+to the UI of `docker build` command and takes the same flags and arguments.
+
+For documentation on most of these flags refer to the [`docker build`
+documentation](https://docs.docker.com/engine/reference/commandline/build/). In
+here we’ll document a subset of the new flags.
+
+## Examples
+
+### <a name="platform"></a> Set the target platforms for the build (--platform)
```
thaJeztah, line 54: Let me try updating my docs PR, and see if it likes this, or if it expects the "extended" syntax; https://www.markdownguide.org/extended-syntax/#heading-ids

```suggestion
### Set the target platforms for the build (--platform) {#platform}
```

docs/reference/buildx_rm.md
```diff
@@ -1,15 +1,11 @@
...
 
-Options:
-      --builder string   Override the configured builder instance
-```
-
-## Description
 
-Removes the specified or current builder. It is a no-op attempting to remove the
-default builder.
+<!---MARKER_GEN_END-->
```
thaJeztah: I guess this was not removed on purpose (I can update if it was indeed accidental)


```suggestion
<!---MARKER_GEN_END-->

## Description

 Removes the specified or current builder. It is a no-op attempting to remove the
 default builder.
```

tonistiigi: I think this was accidental indeed. I removed something where there was no actual description and the generated line from the command description was just repeated. But this one does have a little bit more context.

docs/reference/buildx_build.md
tonistiigi: If they are functionally equivalent then we can use the markdown style. Need to update the generator as well as it is looking for HTML style marker atm. to see if link is possible.

thaJeztah: Ok, I tried manually updating to the markdown syntax in https://github.com/docker/docker.github.io/pull/12092 (preview  at https://deploy-preview-12092--docsdocker.netlify.app/engine/reference/commandline/buildx_build/

So both work, but looks like the Markdown syntax works better, and is cleaner; in that case Jekyll accepts the custom ID; skipping its own generated one. So in that case there won't be two IDs (one custom and one generated);

Before:

Screenshot 2021-03-24 at 16 14 43


After (with Markdown syntax):

Screenshot 2021-03-24 at 16 41 23


I haven't tested what it looks like on GitHub though. Let's try if it works in this comment

### Example with custom anchor-id  {#mycustomid}

this is a link [to the custom ID](#mycustomid)


Sigh... doesn't seem to work in GFM 😞 (I hate all the different flavours of markdown); https://github.github.com/gfm/

Let me dig a bit more to see if it's possible, otherwise we use the current approach

docs/reference/buildx_rm.md
thaJeztah: Fixed

docs/reference/buildx_build.md
thaJeztah: Looks like it's not possible with GitHub's markdown flavour, so let's keep the markdown as-is; perhaps I'll change the yamldocs generator to change it, but that's not for this PR

thaJeztah (commented):
LGTM

thaJeztah:
@tonistiigi I think this is good to merge (thanks for writing the generator code!); we can further improve after this

PR #531, opened 2021-02-05
don't require entitlement for network none

### LINKED ISSUE #524: buildx with network=none generates entitlement error
If I try to run with a builder with `--network=none`, I get the following error:

```
error: failed to solve: rpc error: code = Unknown desc = granting entitlement network.host is not allowed by build daemon configuration
```

I think the issue is from [here](https://github.com/docker/buildx/blob/33e3ca524e2c90e4dcd3a6d63eefc9559ea8236e/build/build.go#L467-L474), but unsure:

```go
	switch opt.NetworkMode {
	case "host", "none":
		so.FrontendAttrs["force-network-mode"] = opt.NetworkMode
		so.AllowedEntitlements = append(so.AllowedEntitlements, entitlements.EntitlementNetworkHost)
	case "", "default":
	default:
		return nil, nil, errors.Errorf("network mode %q not supported by buildkit", opt.NetworkMode)
	}
```

I could see why that would be the case if I set `--network=host`, which is _elevated_ privileges, but why do I get the error with `--network=none`, which is _reduced_ privileges?

Is there a workaround I could use?

tonistiigi:
fix #524

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -465,9 +465,11 @@ func toSolveOpt(ctx context.Context, d driver.Driver, multiDriver bool, opt Opti
 
 	// setup networkmode
 	switch opt.NetworkMode {
-	case "host", "none":
+	case "host":
 		so.FrontendAttrs["force-network-mode"] = opt.NetworkMode
 		so.AllowedEntitlements = append(so.AllowedEntitlements, entitlements.EntitlementNetworkHost)
+	case "none":
+		so.FrontendAttrs["force-network-mode"] = opt.NetworkMode
 	case "", "default":
 	default:
 		return nil, nil, errors.Errorf("network mode %q not supported by buildkit", opt.NetworkMode)


deitch:
What's the pipeline and timelines for this to get into docker engine?

AkihiroSuda (approved):

deitch:
Thanks @AkihiroSuda . is there an ETA for it getting into a release? And do I need to wait for that, or can I get ahead of the curve and use it with an existing version of docker?

AkihiroSuda:
I don't know about the release plan, but the code in the git master can be tried with `make binaries install`

crazy-max:
@deitch You can also use the [produced artifacts by the CI](https://github.com/docker/buildx/actions/runs/555391945): https://github.com/docker/buildx/suites/2012268309/artifacts/40286252

deitch:
Thanks

PR #532, opened 2021-02-05
bake: allow variables in user functions

tonistiigi:
Currently, all variables used in functions need to be passed in as parameters. This is quite cumbersome and ofter means variables need to be redefined as parameters. Eg. https://github.com/tonistiigi/binfmt/blob/master/docker-bake.hcl#L47 . This change allows user functions to see global variables by just changing their parsing order.

@vanstee @crazy-max 


crazy-max (approved):
LGTM

vanstee (approved):

PR #538, opened 2021-02-10
bake: allow variables across files

tonistiigi:
depends on #532

Updates HCL parsing so that when multiple files are specified, one file can use variables defined in another file.

@crazy-max @vanstee

Issue #445, opened 2020-11-21
Provide support for resources interpolation


anouvel-ec:
Hi,

In complex bake files, I would like to use advanced interpolation syntax in order reduce duplication:

- Currently

```hcl
target "front" {
  context = "front"
  tags    = ["front"]
}
```

- Expected

```hcl
target "front" {
  # using the name of the target
  context = target.front.name

  # and / or using embedded target settings
  tags    = [target.front.context]
}
```

This should be helpful also for groups.

Thanks !

smangels:
hi, I wonder whether there is a way to reference the name/context of the current target? Or could this be achieved by other methods in HCL files?

```hcl
target "_base" {
    platforms = [
        "linux/arm64",
        "linux/amd64"
    ]
    tags = [
        "myrepo.com/${target.this.name}:latest"
    ]
}

group "default" {
    targets = [ "backend", "ui" ]
}

target "backend" {
    inherits = ["_base"]
    dockerfile = "Dockerfile.backend"
}

target "ui" {
    inherits = ["_base"]
    dockerfile = "Dockerfile.ui"
}
```

Is there a way to make that result in images being tagged `myrepo/backend:latest` as well as `myrepo/ui:latest` using the HCL example above?

crazy-max:
> I was also thinking about a `this` name, either as a sub-property of `target` or at the top-level. I am a bit worried about the idea of having "special" variable names, since if users start using that name, you start getting slightly weird results.
> 
> That said, I think we should probably have a `this` name somehow, it probably wouldn't be hugely difficult to implement - WDYT @crazy-max?

I'm not sure how we could handle inheritance for this use case. Here `target.this.name` would be `_base` imo so maybe in our implementation `this` should be the caller?

jedevc:
Oh nooo. Making that work nicely with inheritance is gonna be practically very difficult. We'd need to move the `inherits` processing into `hclparser` as well, making that much less simple to do.

PR #541, opened 2021-02-11
bake: allow attributes in global scope

tonistiigi:
depends on #539 #538

I propose this as a base for https://github.com/docker/buildx/pull/489#pullrequestreview-561253366 

You can now define global scope attributes in HCL/JSON and use them for code reuse and setting values for variables. This means you can do a "data-only" HCL file with the values you want to set/override and use it in the list of regular output files.

Eg.

```
variable "REPO" {
  default = "tonistiigi/repo"
}

APP_TAG="foo"

target "app" {
  tags = ["${REPO}/${APP_TAG}"]
  args = {
     variant = APP_TAG
  }
}
```

You can use this file directly or create an override configuration file

```
WHOAMI="myuser"
REPO="${WHOAMI}/foo"
APP_TAG="bar-by-${WHOAMI}"
```

and invoke bake together with both of the files.

So instead of loading a `.env` automatically with another custom format, you can just pass any config file. The HCL attributes are not 100% compatible with env file format so I think for compose support we still should have a fallback, and maybe still automatically load it if you parse compose. Possibly we can just use the new compose library that does that automatically on parsing compose. For everything else, you can use HCL/JSON attributes.

In addition to strings HCL file attributes also support custom data types: arrays, booleans etc. You can refer to other attributes or call built-in functions(user functions not supported atm).

The difference between variables and attributes is that only variables are configurable by the user on invocation. I think this distinction is important, for example in the future I would expect that we add a "description" property to variables so we can show the user what variables are accepted and what do they mean.

I think we should also do https://github.com/docker/buildx/issues/445

@crazy-max @tiborvass @thaJeztah @vanstee

### DIFF
--- a/bake/hcl.go
+++ b/bake/hcl.go
@@ -133,6 +133,8 @@ type StaticConfig struct {
 	Variables []*Variable `hcl:"variable,block"`
 	Remain    hcl.Body    `hcl:",remain"`
 
+	attrs hcl.Attributes
+
 	defaults map[string]*hcl.Attribute
 	env      map[string]string
 	values   map[string]cty.Value
@@ -146,6 +148,9 @@ func mergeStaticConfig(scs []*StaticConfig) *StaticConfig {
 	sc := scs[0]
 	for _, s := range scs[1:] {
 		sc.Variables = append(sc.Variables, s.Variables...)
+		for k, v := range s.attrs {
+			sc.attrs[k] = v
+		}
 	}
 	return sc
 }
@@ -168,6 +173,12 @@ func (sc *StaticConfig) Values(withEnv bool) (map[string]cty.Value, error) {
 
 	sc.values = map[string]cty.Value{}
 	sc.progress = map[string]struct{}{}
+
+	for k := range sc.attrs {
+		if _, err := sc.resolveValue(k); err != nil {
+			return nil, err
+		}
+	}
 
 	for k := range sc.defaults {
 		if _, err := sc.resolveValue(k); err != nil {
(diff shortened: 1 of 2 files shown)


crazy-max:
@tonistiigi 

> So instead of loading a .env automatically with another custom format, you can just pass any config file. The HCL attributes are not 100% compatible with env file format so I think for compose support we still should have a fallback, and maybe still automatically load it if you parse compose. Possibly we can just use the new compose library that does that automatically on parsing compose. For everything else, you can use HCL/JSON attributes.

Indeed, I think it is a better approach compared to what I proposed in #489 and also more intuitive for the HCL format.

> In addition to strings HCL file attributes also support custom data types: arrays, booleans etc. You can refer to other attributes or call built-in functions(user functions not supported atm).

About that I think something like I suggested in https://github.com/docker/buildx/pull/491#issue-544732880 for value types accepted for a variable like [Terraform](https://www.terraform.io/docs/configuration/variables.html#declaring-an-input-variable) does would be nice. But that could be done in a further PR.

tonistiigi:
> About that I think something like I suggested in #491 (comment) for value types accepted for a variable like Terraform does would be nice. But that could be done in a further PR.

Yes, validation would be nice (in follow-up). For type, I just detect it from the default value atm. but that gets tricky with list types. Maybe for lists, you would need to define it.

tonistiigi:
Dependant PRs have been merged.

crazy-max (approved):

tiborvass (approved):

PR #558, opened 2021-03-05
Add quiet mode to progress printer

ulyssessouza:
Allows it to write to `os.DevNull` and by this silencing the printer.

Related to https://github.com/docker/compose-cli/pull/1390

### DIFF
--- a/util/progress/printer.go
+++ b/util/progress/printer.go
@@ -2,11 +2,20 @@ package progress
 
 import (
 	"context"
+	"io"
+	"io/ioutil"
 	"os"
 
 	"github.com/containerd/console"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/util/progress/progressui"
+)
+
+const (
+	PrinterModeAuto  = "auto"
+	PrinterModeTty   = "tty"
+	PrinterModePlain = "plain"
+	PrinterModeQuiet = "quiet"
 )
 
 type Printer struct {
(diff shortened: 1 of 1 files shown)


util/progress/printer.go
```diff
@@ -9,6 +9,12 @@ import (
 	"github.com/moby/buildkit/util/progress/progressui"
 )
 
+const (
+	PrinterModeAuto  = "auto"
+	PrinterModeTty   = "tty"
+	PrinterModeQuiet = "quiet"
```
ndeloof, line 15: defining those constant, should also create one for `plain` mode

ndeloof (approved):

util/progress/printer.go
ulyssessouza: Yep! Done

ndeloof (approved):

ulyssessouza:
@tonistiigi Could you PTAL?

util/progress/printer.go
```diff
@@ -34,15 +41,25 @@ func NewPrinter(ctx context.Context, out console.File, mode string) *Printer {
...
 		mode = v
 	}
 
 	go func() {
 		var c console.Console
-		if cons, err := console.ConsoleFromFile(out); err == nil && (mode == "auto" || mode == "tty") {
-			c = cons
+		switch mode {
+		case PrinterModeQuiet:
+			devNull, err := os.Open(os.DevNull)
```
tonistiigi, line 52: just set out to `ioutil.Discard` , no `fallthrough` needed

util/progress/printer.go
```diff
@@ -34,15 +41,25 @@ func NewPrinter(ctx context.Context, out console.File, mode string) *Printer {
...
+			if err == nil {
+				out = devNull
+			}
+			fallthrough
+		case PrinterModeAuto, PrinterModeTty:
+			if cons, err := console.ConsoleFromFile(out); err == nil {
+				c = cons
+			}
 		}
+
```
tonistiigi, line 62: This could be connected to `-q` as well. Otherwise, the ux is confusing unless we want to completely remove `-q`. But there are some things to figure out like outputs can pipe to stdout if image ID is printed there so I guess it could be follow up. Another way to make it faster is to just read the channel here directly until end if the content is disarded anyway. `DisplaySolveStatus` does need to analyze the objects and sort/cache them so has some overhead.

util/progress/printer.go
ulyssessouza: `ioutil.Discard` is a `Writer` and `out` is a `File`.

About the `fallthrough` I'm not sure if I understood, but it is needed in order to execute `console.ConsoleFromFile(out)` and assign to `c`. Otherwise, how could it affect the console? Note that if not, `plain` output is printed.

tonistiigi: The parameter that `DisplaySolveStatus` takes is an `io.Writer`. If you discard the output then you don't have a console. That should answer to the `fallthrough` case as well.

PR #575, opened 2021-03-27
bake: allow user functions in variables and vice-versa

tonistiigi:
~depends on #541~

Removes previous limitations where user-defined functions could not be used in variables and function definitions could not call other functions.

I at least temporarily needed to fork `userfunc` package to make this work. The package is not very big but I already sent a PR to upstream with the same refactor. A more complicated story is parsing function dependencies from JSON that isn't really available with the current API. I added a hacky version with reflect package and opened an issue in upstream to find a better solution.

Some of the parsing logic could use a reorganization but didn't want to do it yet as much work is still needed to get #445 working first. After that, we can think about cleanups.


bake/hcl.go
```diff
@@ -171,79 +185,325 @@ func (sc *StaticConfig) Values(withEnv bool) (map[string]cty.Value, error) {
...
+	return arr, nil
+}
 
-	def, ok := sc.attrs[name]
+func appendJSONFuncCalls(exp hcl.Expression, m map[string]struct{}) error {
+	v := reflect.ValueOf(exp)
+	if v.Kind() != reflect.Ptr || v.IsNil() {
+		return errors.Errorf("invalid json expression kind %T %v", exp, v.Kind())
+	}
+	src := v.Elem().FieldByName("src")
```
tiborvass, line 273: this will panic if v.Elem() is not of kind struct

bake/hcl.go
```diff
@@ -171,79 +185,325 @@ func (sc *StaticConfig) Values(withEnv bool) (map[string]cty.Value, error) {
...
+	arr := make([]string, 0, len(m))
+	for n := range m {
+		arr = append(arr, n)
+	}
+	return arr, nil
+}
 
-	def, ok := sc.attrs[name]
+func appendJSONFuncCalls(exp hcl.Expression, m map[string]struct{}) error {
+	v := reflect.ValueOf(exp)
```
tiborvass, line 269: if you want to make this hackier but more readable, you might as well abuse //go:linkname to import those private structs.

tonistiigi: not that it is any cleaner but I don't think it would even work. I don't just need access to a private function/var in here.

crazy-max:
@tonistiigi Tested on my side and LGTM. Maybe we should wait for hashicorp/hcl#455 before merge?

tonistiigi:
I don't think we can wait for hashicorp/hcl#455 . If there is movement we can replace the code.

crazy-max (approved):

bake/hcl.go
crazy-max: Yeah looks like it panics if v's `Elem` is not `Interface` or `Ptr`.

PR #577, opened 2021-03-30
use default socket when no --ssh option given

alexcb:
Signed-off-by: Alex Couture-Beil <alex@mofo.ca>


alexcb:
https://github.com/docker/buildx/pull/578 is a simpler fix.

PR #578, opened 2021-03-30
set --ssh default when omitted

alexcb:
this fixes https://github.com/moby/buildkit/issues/1035

Signed-off-by: Alex Couture-Beil <alex@mofo.ca>

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -309,7 +309,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 	flags.StringArrayVar(&options.secrets, "secret", []string{}, "Secret file to expose to the build: id=mysecret,src=/local/secret")
 
-	flags.StringArrayVar(&options.ssh, "ssh", []string{}, "SSH agent socket or keys to expose to the build (format: default|<id>[=<socket>|<key>[,<key>]])")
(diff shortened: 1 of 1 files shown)


tonistiigi:
No, `--ssh default` is considered insecure if it is added without any opt-in by the user. But going back to https://github.com/moby/buildkit/issues/1035 I think we can consider the user using `git@` ssh address in context as equivalent opt-in (when user does not specify other ssh config).

tonistiigi:
@alexcb Not `in.buildArgs` but `contextPath`. Otherwise sgtm. There is also bake command, if you can you can look into covering that as well.

alexcb:
@tonistiigi I was hoping to use https://github.com/moby/buildkit/blob/master/client/llb/source.go#L209 to do something like:

	sshSpecs := in.ssh
	if len(sshSpecs) == 0 && llb.getGitProtocol(in.contextPath) == llb.gitProtocolSSH {
		sshSpecs = []string{"default"}
	}
	ssh, err := build.ParseSSHSpecs(sshSpecs)
	...

However those aren't exported; How do you feel about me moving the git protocol detection code into a new buildkit util, under `buildkit/util/gitutil`?

tonistiigi:
@alexcb sgtm

PR #581, opened 2021-04-01
include default ssh socket when given an ssh-based git url

alexcb:
Signed-off-by: Alex Couture-Beil <alex@earthly.dev>


alexcb:
@tonistiigi here's my changes to make use of https://github.com/moby/buildkit/pull/2056

I'm unsure how I should update go.mod; if I run `go get -u github.com/moby/buildkit@df49b648c8bfef657ba99543390516a3f518d22a` it updates 5 of so packages, is that the desired way to pull in the latest buildkit code? I can then run `hack/update-vendor` to pull in the deps. Just wanted to double if there's a different way to do this with more fine-control?

tonistiigi:
@alexcb You can just update to go.mod file to point to the sha and then run `hack/update-vendor` that will fix everything up. It is expected that it also updates records of the dependencies that have been updated.

alexcb:
thanks for guiding me through the process @tonistiigi , I wasn't sure if it was expected for this PR to end up being 312 files changed.

tonistiigi (approved):

PR #582, opened 2021-04-03
Fail fast on multi platform build with load

rumpl:
Avoid doing a build if we cannot load the images at the end.

This doesn't really fix anything but at least users will see the error right away and not after a potentially long build.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -456,6 +456,9 @@ func toSolveOpt(ctx context.Context, d driver.Driver, multiDriver bool, opt Opti
 			return nil, nil, notSupported(d, driver.OCIExporter)
 		}
 		if e.Type == "docker" {
+			if len(opt.Platforms) > 1 {
+				return nil, nil, errors.Errorf("docker exporter does not currently support exporting manifest lists")
+			}
 			if e.Output == nil {
 				if d.IsMobyDriver() {
 					e.Type = "image"


tonistiigi:
This should fail fast already without building. Should block it in moby level if not.

For buildx side, this check should be moved to `build/build.go` so that builds from `bake` also pass validation.

tonistiigi:
hmm 

https://github.com/docker/buildx/blob/84a734dc876a177eca8eb709dc479c3d88220429/build/build.go#L460-L462

rumpl:
Ok so it's building before failing if the driver is `docker-container`, which has `driver.MultiPlatform: true`

tonistiigi:
@rumpl Ah, ok. That is different. Add this extra condition to https://github.com/docker/buildx/blob/master/build/build.go#L415

rumpl:
> ^

👍 I'm moving right now, will get at this next week, sorry

build/build.go
```diff
@@ -520,6 +520,10 @@ func toSolveOpt(ctx context.Context, d driver.Driver, multiDriver bool, opt Opti
 		so.FrontendAttrs["platform"] = strings.Join(pp, ",")
 	}
 
+	if len(opt.Platforms) != 1 && d.Factory().Name() == "docker-container" {
```
tonistiigi, line 523: This does not look correct. Multi-platform builds with `docker-container` driver are definitely supported. If I understood your description correctly then the condition you want is to check if exporting is happening to docker. Also, we should never check for driver names. There is a features list for that but don't think you need that atm., just check for the exporter.

tonistiigi (changes requested):

rumpl:
I should check that the exporter is “docker” right?

tonistiigi:
> I should check that the exporter is “docker” right?

Correct

build/build.go
```diff
@@ -520,6 +520,10 @@ func toSolveOpt(ctx context.Context, d driver.Driver, multiDriver bool, opt Opti
 		so.FrontendAttrs["platform"] = strings.Join(pp, ",")
 	}
 
+	if len(opt.Platforms) > 1 && opt.Exports[0].Type == "docker" {
```
tonistiigi, line 523: Looks correct but move it to line 458 so all the docker exporter handling is in one place.

tonistiigi (changes requested):

crazy-max (approved):
LGTM

tonistiigi (approved):

PR #622, opened 2021-05-29
fixes #617: change delimeter for nodeselector

### LINKED ISSUE #617: nodeselector keyvalues are not parsed properly for kubernetes driver
# Issue description
Based on the [docs](https://github.com/docker/buildx/blob/master/docs/reference/buildx_create.md#-set-additional-driver-specific-options---driver-opt), it should be possible to pass multiple nodeselector keyvalues using 
```
nodeselector="label1=value1,label2=value2"
```
By quoting the nodeselector value, it should be parsed with [this line](https://github.com/docker/buildx/blob/4fec647b9d8f34f8569141124d8462c912858144/driver/kubernetes/factory.go#L95) becoming a list of keyvalues.

It seems that the second keyvalue, `label2` in this case, is treated as another driver-opt rather than a keyvalue for nodeselector. It might be caused by the way driver-opts are parsed (using csvToMap func).

# Reproducing the issue
```
$ docker buildx create --driver kubernetes --driver-opt replicas=1,nodeselector="kubernetes.io/arch=arm64,kubernetes.io/arch=arm" --use
unruffled_elion
$ touch Dockerfile
$ docker buildx build . 
[+] Building 0.0s (0/0)                                                                                                                                                                       
error: no valid drivers found: invalid driver option kubernetes.io/arch for driver kubernetes
```

MichalAugustyn:
As described in #617, there is a conflict between the comma splitting labels and driver opts

### DIFF
--- a/docs/reference/buildx_create.md
+++ b/docs/reference/buildx_create.md
@@ -125,7 +125,7 @@ Passes additional driver-specific options. Details for each driver:
     - `image=IMAGE` - Sets the container image to be used for running buildkit.
     - `namespace=NS` - Sets the Kubernetes namespace. Defaults to the current namespace.
     - `replicas=N` - Sets the number of `Pod` replicas. Defaults to 1.
-    - `nodeselector="label1=value1,label2=value2"` - Sets the kv of `Pod` nodeSelector. No Defaults. Example `nodeselector=kubernetes.io/arch=arm64`
+    - `nodeselector="label1=value1;label2=value2"` - Sets the kv of `Pod` nodeSelector. No Defaults. Use `;` to split labels. Example `nodeselector=kubernetes.io/arch=arm64`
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Can we fix the parser to support "," properly?

MichalAugustyn:
That would be possible if the quotation marks were kept inside the driverOpt parameter value.
As far as I know, buildx is not even aware of the quotation marks as these are removed in bash before buildx runs.

```
bin/buildx create --driver kubernetes --name buildx --driver-opt "replicas=3,nodeselector='label1=value1,label2=value2,namespace=5'"
```
would do the job, but I feel like using `;` is more convenient.

jedevc:
@MichalAugustyn since this PR was opened, we've updated the kubernetes driver docs https://docs.docker.com/build/drivers/kubernetes/#node-assignment:

> Due to quoting rules for shell commands, you must wrap the nodeselector and tolerations parameters in single quotes. You can even wrap all of --driver-opt in single quotes, ...

Does that help resolve the issue? If not, we should probably try and work out what to do with this PR.

PR #639, opened 2021-06-21
Minor fix to links in the document

eitsupi:
Reflect the branch name change of the linked repository.
(It is automatically redirected, so the current link is also working.)

I just happened to open this link and checked the others, but they didn't seem to need similar modifications.


tonistiigi:
@eitsupi missing DCO

tonistiigi:
Looks like still doesn't pass validator https://github.com/docker/buildx/pull/639/checks?check_run_id=2905022559 . I think you need to change the committer name in your git config so it matches your name.

eitsupi:
Thank you for explaining about the operation. I learned a lot.

crazy-max (approved):
LGTM

PR #645, opened 2021-06-25
New bake parser

tonistiigi:
New implementation with `hclparser` package that understands variables, functions and attributes. Makes sure parsing is generic and doesn't know about the expected schema. Also removes weird concepts like `StaticConfig`.

This makes it possible to add linking of values in targets but more work is needed before that. The support for inherits and overrides needs to be moved to the parser level so that if these values are used in variables they get the correct values.


crazy-max (approved):
LGTM

PR #653, opened 2021-06-29
Dockerfile: update xx

tonistiigi:
Buildx was still using v1 version of xx


crazy-max (approved):
LGTM

Issue #21, opened 2019-04-16
add process driver


tonistiigi:
Add driver that can detect buildkitd in the `$PATH` and run it directly. For example, this driver can be used with #12 . Pidfile should be used for keeping track of the daemon and monitoring its state.

errordeveloper:
> Pidfile should be used for keeping track of the daemon and monitoring its state.

Process management can be tricky, perhaps `docker-container` driver does a pretty good job at this!

Perhaps this kind of driver might be more feasible for a setting where the `buildkitd` runs in foreground for a single build and expected to exit right away?

crazy-max:
I thought a bit more about this driver, quick notes:

* Check buildkitd in path
  * If not available we can download latest buildkit release and extract to `~/.docker/buildx/.buildkit/<version>` (use GitHub API to retrieve artifacts)
    * Also be able to set buildkit version through driver-opt
* PID in `~/.docker/buildx/.buildkit/<builder_name>/buildkitd.pid`
* Define default BuildKit Daemon configuration for this driver with defaults (can be default buildkitd-flags too):

```toml
root = "~/.docker/buildx/.buildkit/<builder_name>/state"
insecure-entitlements = [ "network.host", "security.insecure" ] # not sure

[worker.oci]
  enabled = true
  snapshotter = "auto" # overlayfs or native, default value is "auto".
  rootless = true # rootless should be enabled by default (or disabled if "docker buildx create" run as root)
  binary = "~/.docker/buildx/.buildkit/<version>/buildkit-runc"
```

PR #656, opened 2021-06-29
Create buildx image

### LINKED ISSUE #647: Is there an official docker buildx docker image?
I know that there is a dockerhub image for docker (`docker:latest`) and there is one for docker in docker (`docker:dind`) but is there one for running buildx (without having to run n number of commands to do so?)

I build one that works pretty well (for me at least) and I decided to push it to dockerhub reg here: https://hub.docker.com/r/maniator/dind-buildx

Is there anything wrong with that?

What type of attribution do I need to add?

crazy-max:
Fixes #647

Could be used like this:

```Dockerfile
FROM docker
COPY --from=docker/buildx:latest /buildx /usr/libexec/docker/cli-plugins/docker-buildx
```

e.g. https://hub.docker.com/r/crazymax/buildx/tags?page=1&ordering=last_updated

image

cc @chris-crone 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ on:
       - 'master'
 
 env:
+  REPO_SLUG: "docker/buildx-bin"
   REPO_SLUG_ORIGIN: "moby/buildkit:master"
   CACHEKEY_BINARIES: "binaries"
   PLATFORMS: "linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64,linux/s390x,linux/ppc64le,linux/riscv64"
@@ -173,6 +174,38 @@ jobs:
           name: buildx
           path: ./release-out/*
           if-no-files-found: error
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            ${{ env.REPO_SLUG }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+      -
+        name: Login to DockerHub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Build and push image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          target: binaries
+          push: ${{ github.event_name != 'pull_request' }}
+          cache-from: type=local,src=/tmp/.buildx-cache/${{ env.CACHEKEY_BINARIES }}
+          platforms: ${{ env.PLATFORMS }},darwin/amd64,darwin/arm64,windows/amd64,windows/arm64
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
       -
         name: GitHub Release
         if: startsWith(github.ref, 'refs/tags/v')
--- a/README.md
+++ b/README.md
@@ -20,25 +20,28 @@ Key features:
 # Table of Contents
 
 - [Installing](#installing)
+  - [Docker](#docker)
+  - [Binary release](#binary-release)
+  - [From `Dockerfile`](#from-dockerfile)
 - [Building](#building)
-    + [with Docker 18.09+](#with-docker-1809)
-    + [with buildx or Docker 19.03](#with-buildx-or-docker-1903)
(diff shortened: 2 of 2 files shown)


tonistiigi:
As this is not a runnable image it might lead to even more confusion. Maybe if it was under a different name that makes it more clear. But logically it should just be inside the docker image(or image that contains both).

crazy-max:
> As this is not a runnable image it might lead to even more confusion. Maybe if it was under a different name that makes it more clear.

Agree, changed to `docker/buildx-bin` if it makes sense.

> But logically it should just be inside the docker image(or image that contains both).

I think we can start with a dedicated non-runnable buildx image for now and could be included later in the official docker image using `COPY --from`? See also docker-library/docker#243.

README.md
```diff
@@ -60,8 +60,17 @@ Change the permission to execute:
...
+### From `Dockerfile`
+
+Here is how to use buildx inside a Dockerfile through the [`docker/buildx-bin`](https://hub.docker.com/r/docker/buildx-bin) image:
 
+```Dockerfile
+FROM docker
+COPY --from=docker/buildx-bin:latest /buildx /usr/libexec/docker/cli-plugins/docker-buildx
+RUN docker buildx version
+```
+
```
crazy-max, line 72: Added this section in the README to avoid confusion.

tonistiigi:
SGTM

@AkihiroSuda @chris-crone

.github/workflows/build.yml
```diff
@@ -173,6 +174,38 @@ jobs:
...
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+      -
+        name: Login to DockerHub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
crazy-max, line 197: Would need to add these secrets on the repo.

chris-crone:
SGTM as buildx can be used standalone. I agree with @tonistiigi that we should have everything in the `docker` image and that should be most people's default.

maniator:
> As this is not a runnable image it might lead to even more confusion. Maybe if it was under a different name that makes it more clear. But logically it should just be inside the docker image(or image that contains both).

Agreed

maniator:
> SGTM as buildx can be used standalone. I agree with @tonistiigi that we should have everything in the `docker` image and that should be most people's default.

Cant we do something like I did here? https://hub.docker.com/r/maniator/dind-buildx

crazy-max:
@chris-crone has created the secrets. SGTM to merge.

README.md
```diff
@@ -60,6 +63,17 @@ Change the permission to execute:
...
 
+### From `Dockerfile`
+
+Here is how to use buildx inside a Dockerfile through the [`docker/buildx-bin`](https://hub.docker.com/r/docker/buildx-bin) image:
+
+```Dockerfile
+FROM docker
+COPY --from=docker/buildx-bin:latest /buildx /usr/libexec/docker/cli-plugins/docker-buildx
+RUN docker buildx version
+```
```
maniator, line 74: I am not sure how this helps users use `buildx` in dind though?

They would still need to build their own image, no?

crazy-max: @maniator Atm yes but could be done in a follow-up on https://github.com/docker-library/docker repo to include this instruction.

maniator: Shouldnt we also have a PR there in tandem?

crazy-max: Could open an an issue [docker-library/docker](https://github.com/docker-library/docker) when this PR is merged as a follow-up if you want yes. Will see what their pov is on the inclusion of buildx in the official image.

tonistiigi (approved):

tonistiigi:
> As this is not a runnable image it might lead to even more confusion. Maybe if it was under a different name that makes it more clear. But logically it should just be inside the docker image(or image that contains both).

I'm starting to change my mind about this one. It might be kind of useful to have it for working with remote instances or testing out various versions with `docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock docker/buildx` . That image should be on top of alpine and, contain ca-certificates etc. We also need to improve the ux. We can inject environment variable into container so that if image gets run without a correct endpoint it would show messages about the docker flags that need to be exposed (eg. docker socket or if context is empty then volume for that or same for credentials). I guess same could be possible when image runs in k8s pod but because k8s API is authenticated I don't think it is as common as `docker.sock`.

Would be cool if `docker run --privileged docker/buildx` would also work without any docker socket or remote address. I guess with process driver https://github.com/docker/buildx/issues/21 . We do ship daemonless buildctl script with `moby/buildkit` but that is `buildctl` and doesn't have a good ux. Maybe a special tag that contains buildkit binaries already so they don't need to be pulled on each invocation.

That doesn't change that buildx should be included in docker image.

PR #659, opened 2021-06-30
Ignore release-out folder

crazy-max:
Also remove unused `TAG` env var and wrap binaries in shell script.

### DIFF
--- a/hack/release
+++ b/hack/release
@@ -1,21 +1,11 @@
 #!/usr/bin/env bash
 
-TAG=$1
-OUT=$2
+OUT=${1:-release-out}
 
 . $(dirname $0)/util
 set -eu -o pipefail
 
 : ${PLATFORMS=linux/amd64}
-
-usage() {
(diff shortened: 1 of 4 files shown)


hack/release
```diff
@@ -1,21 +1,10 @@
...
-
-usage() {
-  echo "usage: ./hack/release <tag> <out>"
-  exit 1
-}
-
-if [ -z "$TAG" ] || [ -z "$OUT" ]; then
-  usage
-fi
+: ${EXPORT_LOCAL=release-out}
```
tonistiigi, line 7: Not very fond of this env. The name isn't very clear and setting dest path is kind-of expected to be a parameter for the release. Maybe leave it as positional arg and default to `./release-out`. Unless there is some well-known env that is used in these cases that I can't think of atm.

tonistiigi:
see ci. Looks like it is pushing the whole vm.

PR #661, opened 2021-07-01
Keep git dir

crazy-max:


tonistiigi:
But this is not being built from git?

crazy-max:
Yes you're right my bad

tonistiigi:
Maybe it should be though

PR #670, opened 2021-07-12
Example with gha cache

crazy-max:
Follow-up #535 

### DIFF
--- a/docs/reference/buildx_build.md
+++ b/docs/reference/buildx_build.md
@@ -201,10 +201,17 @@ single-platform build result to `docker images`.
 --cache-from=[NAME|type=TYPE[,KEY=VALUE]]
 ```
 
-Use an external cache source for a build. Supported types are `registry` and `local`.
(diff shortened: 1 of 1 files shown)


docs/reference/buildx_build.md
```diff
@@ -217,27 +224,36 @@ $ docker buildx build --cache-from=user/app:cache .
 $ docker buildx build --cache-from=user/app .
 $ docker buildx build --cache-from=type=registry,ref=user/app .
 $ docker buildx build --cache-from=type=local,src=path/to/cache .
+$ docker buildx build --cache-from=type=gha .
 ```
 
+More info about cache exporters and available attributes: https://github.com/moby/buildkit#export-cache
```
thaJeztah, line 230: We should probably include a table somewhere in a Markdown document inside `docs/`  (perhaps_ move_ the information from the readme to that page), otherwise the docs at docs.docker.com will link to the GitHub page (instead of showing the information in the docs)

PR #671, opened 2021-07-13
bake: allow BAKE_CMD_CONTEXT builtin var

### LINKED ISSUE #641: bake: local context not supported with remote git bake definition
Currently when using remote `git` bake definition `buildx bake git://` then context is fixed as a relative directory inside the git location. It should be possible for remote bake to use local context. Possible solutions would be to define the path as `local://` or special rule for path starting with `./`

tonistiigi:
fix #641 

Allows accessing the main context for bake command from bake
file that has been imported remotely.

```bash
# docker buildx bake "git://github.com/docker/cli" --print
{
  "target": {
    "binary": {
      "context": "git:/github.com/docker/cli",
      "dockerfile": "Dockerfile",
      "args": {
        "BASE_VARIANT": "alpine",
        "GO_STRIP": "",
        "VERSION": ""
      },
      "target": "binary",
      "platforms": [
        "local"
      ],
      "output": [
        "build"
      ]
    }
  }
}

/tmp/a # touch foo bar
/tmp/a # docker buildx bake "git://github.com/tonistiigi/buildx#remote-test"
...
 > [4/4] RUN ls -l && stop:
#9 0.207 total 0
#9 0.207 -rw-r--r--    1 root     root             0 Jul 13 03:39 bar
#9 0.207 -rw-r--r--    1 root     root             0 Jul 13 03:39 foo
#9 0.207 /bin/sh: stop: not found


 # docker buildx bake "git://github.com/tonistiigi/buildx#remote-test" "git://github.com/docker/cli"
...
#8 0.232 -rw-r--r--    1 root     root          1893 Jul 13 03:58 poule.yml
#8 0.232 drwxr-xr-x    7 root     root          4096 Jul 13 03:58 scripts
#8 0.232 drwxr-xr-x    3 root     root          4096 Jul 13 03:58 service
#8 0.232 drwxr-xr-x    2 root     root          4096 Jul 13 03:58 templates
#8 0.232 drwxr-xr-x   11 root     root          4096 Jul 13 03:58 vendor
#8 0.232 -rwxr-xr-x    1 root     root         10123 Jul 13 03:58 vendor.conf
#8 0.232 /bin/sh: stop: not found

#   docker buildx bake "git://github.com/tonistiigi/buildx#remote-test" "https://ftp.gnu.org/gnu/binutils/binutils-2.36.tar.gz"
...
 > [4/4] RUN ls -l && stop:
#8 0.250 total 4
#8 0.251 drwxrwxrwx   19 root     root          4096 Jul 13 04:00 binutils-2.36
#8 0.251 /bin/sh: stop: not found
```

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -61,8 +61,8 @@ func ReadLocalFiles(names []string) ([]File, error) {
 	return out, nil
 }
 
-func ReadTargets(ctx context.Context, files []File, targets, overrides []string) (map[string]*Target, error) {
-	c, err := ParseFiles(files)
+func ReadTargets(ctx context.Context, files []File, targets, overrides []string, defaults map[string]string) (map[string]*Target, error) {
+	c, err := ParseFiles(files, defaults)
 	if err != nil {
 		return nil, err
 	}
@@ -86,7 +86,7 @@ func ReadTargets(ctx context.Context, files []File, targets, overrides []string)
 	return m, nil
 }
 
-func ParseFiles(files []File) (_ *Config, err error) {
+func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error) {
 	defer func() {
 		err = formatHCLError(err, files)
 	}()
@@ -120,6 +120,7 @@ func ParseFiles(files []File) (_ *Config, err error) {
 	if len(fs) > 0 {
 		if err := hclparser.Parse(hcl.MergeFiles(fs), hclparser.Opt{
 			LookupVar: os.LookupEnv,
+			Vars:      defaults,
 		}, &c); err.HasErrors() {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


tonistiigi:
This also fixed the issue where remote URLs were used in bake targets. Previously the values always passed `path.Clean` that corrupted the url.

crazy-max (approved):
@tonistiigi LGTM. In a follow-up, we could reserve variables starting with `BAKE_` as we discussed.

bake/bake.go
```diff
@@ -526,6 +527,12 @@ func updateContext(t *build.Inputs, inp *Input) {
 		t.ContextPath = inp.URL
 		return
 	}
+	if strings.HasPrefix(t.ContextPath, "cwd://") {
```
crazy-max, line 530: Document `cwd` protocol

tonistiigi: I'm not sure if it is a good idea. I don't want users to write it in the bake file. If they do then they break the remote URL case. They should use the variable and this should be the internal behavior of the variable. We should be able to change that to `workdir://` for example and no old files should be broken.

We should document the builtin var, and also more docs for vars/attrs in general.

crazy-max: > they break the remote URL

Ah yes right

> We should document the builtin var, and also more docs for vars/attrs in general.

:+1:

PR #672, opened 2021-07-13
Keep BuildKit state in a volume

crazy-max:
Retain BuildKit state inside a named volume if the container is removed for the `docker-container` driver. `--volume` flag has also been added to the `rm` command to be able to force volume removal.

### DIFF
--- a/commands/rm.go
+++ b/commands/rm.go
@@ -11,7 +11,8 @@ import (
 )
 
 type rmOptions struct {
-	builder string
+	builder   string
+	keepState bool
 }
 
 func runRm(dockerCli command.Cli, in rmOptions) error {
@@ -28,7 +29,7 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
 		if err != nil {
 			return err
 		}
-		err1 := stop(ctx, dockerCli, ng, true)
+		err1 := rm(ctx, dockerCli, ng, in.keepState)
 		if err := txn.Remove(ng.Name); err != nil {
 			return err
 		}
@@ -40,7 +41,7 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
 		return err
 	}
 	if ng != nil {
-		err1 := stop(ctx, dockerCli, ng, true)
+		err1 := rm(ctx, dockerCli, ng, in.keepState)
 		if err := txn.Remove(ng.Name); err != nil {
 			return err
 		}
@@ -66,10 +67,13 @@ func rmCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 		},
 	}
 
+	flags := cmd.Flags()
+	flags.BoolVar(&options.keepState, "keep-state", false, "Keep BuildKit state")
+
 	return cmd
 }
 
-func stop(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, rm bool) error {
+func rm(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, keepState bool) error {
 	dis, err := driversForNodeGroup(ctx, dockerCli, ng, "")
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


tonistiigi (commented):
But where is the volume created on `create`?

I wonder if for backward compatibility we should remove volume by default. Or print a warning that volume is still there and data was not released. Maybe there is a good opt-in  flag that is opposite of `--rm`. Most of the time I think users want to remove volume. @thaJeztah 

 I guess we could revisit the deletion path once `upgrade` command is added in the future. I guess then `rm` always deletes volume(or at least by default). And `upgrade` performs `rm+create` without touching the volume.

driver/docker-container/driver.go
```diff
@@ -103,6 +105,13 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 		hc := &container.HostConfig{
 			Privileged: true,
 			UsernsMode: "host",
+			Mounts: []mount.Mount{
+				{
+					Type:   mount.TypeVolume,
+					Source: d.Name,
```
tonistiigi, line 111: Sorry, didn't see this the first time I looked at it. Is this enough to create volume if it does not exist and reuse it if it does? I wonder if we should add a suffix to `d.Name`? Maybe "-state".

crazy-max: > Sorry, didn't see this the first time I looked at it. Is this enough to create volume if it does not exist and reuse it if it does?

Yes it's enough, tested it live to be sure.

> I wonder if we should add a suffix to `d.Name`? Maybe "-state".

Was wondering too about that. I'm ok with a suffix.

crazy-max:
@tonistiigi 

> Most of the time I think users want to remove volume. 

Agree, just wanted to keep the same UX as other commands of the Docker CLI but maybe it's not legitimate in this case.

thaJeztah:
> Retain BuildKit state inside a named volume if the container is removed for the `docker-container` driver. `--volume` flag has also been added to the `rm` command to be able to force volume removal.

Haven't looked closely at the changes, but be aware that `--volume` only removes anonymous (unnamed) volumes, and won't remove "named" volumes; from `docker rm --help`:

```
-v, --volumes   Remove anonymous volumes associated with the container
```

docs/reference/buildx_rm.md
```diff
@@ -7,6 +7,13 @@ docker buildx rm [NAME]
 <!---MARKER_GEN_START-->
 Remove a builder instance
 
+### Options
+
+| Name | Description |
+| --- | --- |
+| `--builder string` | Override the configured builder instance |
+| `--keep-state` | Keep BuildKit state |
```
tonistiigi, line 15: so it can be reused by a new builder with the same name. Only supported by the `docker-container` driver at the moment.

This can also be in a separate block explaining the flag.

driver/docker-container/driver.go
```diff
@@ -17,14 +17,18 @@ import (
...
 	"github.com/docker/docker/api/types/network"
 	dockerclient "github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/moby/buildkit/client"
+	"github.com/moby/buildkit/util/appdefaults"
 	"github.com/moby/buildkit/util/tracing/detect"
 	"github.com/pkg/errors"
 )
 
+const volumeStatePrefix = "state_"
```
tonistiigi, line 30: I think suffix would be better for this. If it is a prefix it should start with `buildkit_`

commands/rm.go
```diff
@@ -66,10 +67,16 @@ func rmCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 		},
 	}
 
+	flags := cmd.Flags()
+
+	flags.BoolVar(&options.rmVolume, "volume", false, "Remove volume associated with the builder")
+
+	_ = flags
```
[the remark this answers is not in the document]
tonistiigi: ^

thaJeztah:
> In our case it's not so yes that's confusing.

Reason for that is that anonymous volumes are commonly used as temporary / ephemeral storage. `docker rm -v <some container>` is somewhat the equivalent of `docker run --rm <some image with a VOLUME>`. When _naming_ a volume, in most cases that's for a reason (being able to reference the volume later), and we don't want to inadvertently remove `my_important_database_volume`.

tonistiigi (approved):

PR #674, opened 2021-07-15
Create checksums for artifacts

### LINKED ISSUE #285: missing checksums and or GPG signatures for bin files

crazy-max:
Fixes #285 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -164,6 +164,7 @@ jobs:
           ./hack/release ${{ env.RELEASE_OUT }}
         env:
           PLATFORMS: ${{ env.PLATFORMS }},darwin/amd64,darwin/arm64,windows/amd64,windows/arm64
+          CHECKSUMS: 1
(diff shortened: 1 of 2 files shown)


tonistiigi:
See https://github.com/docker/buildx/pull/288#issuecomment-630288471 . I think we need to exclude darwin for that reason.

https://github.com/docker/buildx/pull/288/files also created file per binary. I don't really have a preference.

crazy-max:
@tonistiigi 

> See #288 (comment) . I think we need to exclude darwin for that reason.

Yes indeed, I will exclude darwin bins. We could use [gon](https://github.com/mitchellh/gon) to sign/notarize binaries in our pipeline in the future I guess.

> https://github.com/docker/buildx/pull/288/files also created file per binary. I don't really have a preference.

I think it's more convenient to have a single checksum file that handles file trees.

tonistiigi (approved):

Issue #675, opened 2021-07-15
failed to run any build if using a builder instance with multiple native nodes


chestack:
I want to try building multi-platform images using one builder instance with multiple nodes(docker daemon) as that in [ buildx README](https://github.com/docker/buildx#building-multi-platform-images).

steps:
1. one arm64 node with docker daemon remote API enabled
2. one x86 node, create docker context arm-builder, create build instance with two nodes(x86 and arm)
3. build multi-platform images

following is detailed commands result
```
[root@build-cloud-product-clone-4 Dockerfiles]# docker context ls
NAME            DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT   ORCHESTRATOR
arm-builder *   arm build server                          tcp://172.38.0.16:2375
default         Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                         swarm
[root@build-cloud-product-clone-4 Dockerfiles]# docker ps
CONTAINER ID   IMAGE                                  COMMAND                  CREATED          STATUS             PORTS                  NAMES
1be76a989346   moby/buildkit:buildx-stable-1          "buildkitd"              17 minutes ago   Up 17 minutes                             buildx_buildkit_insecure-builder1
cd79793acc5d   hub.easystack.io/arm64v8/registry:v2   "/entrypoint.sh /etc…"   25 hours ago     Up About an hour   0.0.0.0:80->5000/tcp   registry-hub
[root@build-cloud-product-clone-4 Dockerfiles]# docker buildx ls
NAME/NODE           DRIVER/ENDPOINT             STATUS  PLATFORMS
insecure-builder *  docker-container
  insecure-builder0 unix:///var/run/docker.sock running linux/arm64*, linux/amd64*, linux/386
  insecure-builder1 arm-builder                 running linux/arm64
arm-builder         docker
  arm-builder       arm-builder                 running linux/arm64
default             docker
  default           default                     running linux/amd64, linux/386
[root@build-cloud-product-clone-4 Dockerfiles]# docker -D buildx build --push --platform linux/amd64,linux/arm64 -t hub.easystack.cn:80/buildx/multiple-platform .
DEBU[0000] using default config store "/root/.docker/buildx"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xca36c2]

goroutine 15 [running]:
github.com/moby/buildkit/client.(*Client).controlClient(...)
[...]

PR #37, opened 2019-04-22
build: multi-node build support

tonistiigi:
based on #35 

Automatically split a build to chunks based on platform if needed, distribute between nodes and merge back to a manifest list when all nodes have completed.

PR #686, opened 2021-07-21
fix: should ignore nil client

morlay:
maybe fix https://github.com/docker/buildx/issues/675

I have another issue

With multi-node builder, but single platform build `buildx build --platform=linux/amd64`, will cause the nil issue too

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -197,6 +197,10 @@ func resolveDrivers(ctx context.Context, drivers []DriverInfo, auth Auth, opt ma
 
 	eg, ctx := errgroup.WithContext(ctx)
 	for i, c := range clients {
+		if c == nil {
+			continue
+		}
+
 		func(i int, c *client.Client) {
 			eg.Go(func() error {
 				clients[i].Build(ctx, client.SolveOpt{}, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {


crazy-max (approved):
LGTM, ptal @AkihiroSuda

AkihiroSuda (approved):

build/build.go
```diff
@@ -197,6 +197,10 @@ func resolveDrivers(ctx context.Context, drivers []DriverInfo, auth Auth, opt ma
 
 	eg, ctx := errgroup.WithContext(ctx)
 	for i, c := range clients {
+		if c == nil {
```
thaJeztah, line 200: Fix looks ok to address the issue, but perhaps we should also look into why this slice contains nil values in the first place, @crazy-max ?

crazy-max: Multi-node build support and the underlying clients iteration was first introduced while resolving drivers (#37). Client nil check was already there

https://github.com/docker/buildx/blob/7d312eaa0a04e03810f4a3a32d80e3e96b14f6da/build/build.go#L283-L285

but was left out to be included in #535 as well.

> but perhaps we should also look into why this slice contains nil values in the first place,

Yes agree, might be an issue with the context dialer while creating the buildkit client for docker-container driver. Will take a look.

thaJeztah: Was initially thinking, perhaps the slice is initialized with a length/size (not taking into account that filtering might happen). Haven't looked at the code though (was reading from my phone), but though it wouldn't hurt to leave a comment ☺️🤗

crazy-max: Yes you did well :)

tonistiigi:
This does not look correct and iiuc can return cleanly on error cases as skip over the builds. We should first figure out why this happens. Probably the correct solution is to return an error. If skipping over a node is determined to be better it should be  in `resolveDriversBase` and pick a replacement node instead of skipping over a (part of) build.

PR #682, opened 2021-07-21
[kubernetes] Support --config to mount buildkit.toml and -driver-opt=qemu.install=true,qemu.image=tonistiigi/binfmt:latest for qemu installing

morlay:
fix https://github.com/docker/buildx/issues/603

PR #691, opened 2021-07-24
Setup integration testing

morlay:
Signed-off-by: Morlay <morlay.null@gmail.com>

**driver** (`docker-container`, `kubernetes`)  X  **mode** (`single-node`, `multi-nodes`)

could added `--config` mode, but block by https://github.com/docker/buildx/pull/682

To run  cases:
* buildx build single platform 
* buildx build multiple platforms

https://github.com/morlay/buildx/actions/runs/1062152109


testings/builder/create_multiple_nodes_builder.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
AkihiroSuda, line 2: set -eu -o pipefail

tonistiigi:
Maybe as a stopgap we could start with this but generally I would prefer buildkit-style integration testing with dockerfile describing the environment. Not github actions specific scripts that will have scaling issues.

crazy-max (commented):
Like @tonistiigi said it would be better to sandbox as much as possible to be able to repro locally.

PR #692, opened 2021-07-25
Allow booting builder after creation

### LINKED ISSUE #189: add `--bootstrap` to `--create`
From user's point of view, bootstrapping might be considered as a part of `create` rather than `inspect`.

So `docker buildx create` should support `--bootstrap` as well as `docker buildx inspect`.

crazy-max:
Fixes #189

Allow booting builder after creation with `docker buildx create --bootstrap ...`.

~Also adds `buildx start` command.~

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -1,11 +1,13 @@
 package commands
 
 import (
+	"context"
 	"encoding/csv"
 	"fmt"
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/docker/buildx/driver"
 	"github.com/docker/buildx/store"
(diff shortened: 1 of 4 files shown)


tonistiigi:
Not sure if we need a command for this. And as we have `buildx stop` it should probably be `start` then. But I think we can start with just a flag.

crazy-max:
@tonistiigi 

> And as we have `buildx stop` it should probably be `start` then.

:+1:

> But I think we can start with just a flag.

Yeah I thought ~`bootstrap`~ `start` was a legit command to have here cause of the `stop` one. Let me know if you don't think so.

crazy-max:
Remove `start` command as discussed. PTAL @tonistiigi

tonistiigi (approved):

Issue #62, opened 2019-05-06
simplify controlling plain output


tonistiigi:
Some CI systems run through terminal even if they really shouldn't and setting the `--progress=plain` flag in scripts is hard. We should also provide an environment variable that is much easier to configure for the `auto` behavior. Maybe we could just detect `CONTINUOUS_INTEGRATION=1` directly but some users may prefer tty output even in ci.

`BUILDKIT_PROGRESS_PLAIN=1`
`BUILDKIT_PROGRESS=plain`

@tiborvass

tiborvass:
-1 on honoring `CONTINUOUS_INTEGRATION`.
+1 on `BUILDKIT_PROGRESS=plain`, hoping we won't need more of those `BUILDKIT_*` env vars.

velcrine:
Hy where is the documentation for all these ENV options. It seems only developers are aware of these options

mbuotidem:
ENV options docs : https://docs.docker.com/build/building/env-vars/

PR #693, opened 2021-07-25
Duplicated progress env var

crazy-max:
This PR reverts `BUILDX_PROGRESS_DEFAULT` env var that was introduced in #504 because it was already available through #69 with `BUILDKIT_PROGRESS` env var.

Docs have also been updated as stipulated in https://github.com/docker/buildx/issues/62#issuecomment-886162128

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -352,13 +352,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 func commonBuildFlags(options *commonOptions, flags *pflag.FlagSet) {
 	options.noCache = flags.Bool("no-cache", false, "Do not use cache when building the image")
-
-	defaultProgress, ok := os.LookupEnv("BUILDX_PROGRESS_DEFAULT")
-	if !ok {
-		defaultProgress = "auto"
-	}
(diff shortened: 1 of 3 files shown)


thaJeztah (approved):
LGTM, looks like it was not yet in use by anyone 🤞https://grep.app/search?q=BUILDX_PROGRESS_DEFAULT

crazy-max:
@thaJeztah 

> looks like it was not yet in use by anyone

Yes hopefully :)

thaJeztah:
Perhaps we should add a section related to build env-vars in https://github.com/docker/cli/blob/master/docs/reference/commandline/cli.md#environment-variables

(also make sure that `docker build` without buildx supports this)

crazy-max:
@thaJeztah 

> Perhaps we should add a section related to build env-vars

Done https://github.com/docker/cli/pull/3208

I think we should also mutualize the progress package in a follow-up:

* https://github.com/docker/buildx/tree/master/util/progress
* https://github.com/moby/buildkit/tree/master/util/progress

thaJeztah:
> I think we should also mutualize the progress package in a follow-up:

Possibly yes; but I see the buildkit version has more subpackages. Haven't looked closely yet, but if you do so, be sure check that it doesn't bring complexity and/or dependencies that are not desirable here.

PR #700, opened 2021-07-27
Update bake docs

crazy-max:
Follow-up #532 #538 #539 #541 #575 #671

### DIFF
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -34,16 +34,19 @@ Build from a file
 Bake is a high-level build command. Each specified target will run in parallel
 as part of the build.
 
-Read [High-level build options](https://github.com/docker/buildx#high-level-build-options) for introduction.
+Read [High-level build options](https://github.com/docker/buildx#high-level-build-options)
+for introduction.
 
-Please note that `buildx bake` command may receive backwards incompatible features in the future if needed. We are looking for feedback on improving the command and extending the functionality further.
+Please note that `buildx bake` command may receive backwards incompatible
+features in the future if needed. We are looking for feedback on improving the
+command and extending the functionality further.
 
 ## Examples
 
 ### <a name="file"></a> Specify a build definition file (-f, --file)
 
-By default, `buildx bake` looks for build definition files in the current directory,
-the following are parsed:
+By default, `buildx bake` looks for build definition files in the current
+directory, the following are parsed:
 
 - `docker-compose.yml`
 - `docker-compose.yaml`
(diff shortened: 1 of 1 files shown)


docs/reference/buildx_bake.md
```diff
@@ -361,14 +521,127 @@ EOF
...
+
+function "tag" {
+    params = [tag]
+    result = ["${REPO}:${tag}"]
+}
+
+target "webapp" {
+    tags = tag("v1")
+}
+EOF
```
thaJeztah, line 552: We were already doing this, but HEREDOC can be a bit tricky when we print it in the documentation, as it makes it hard to copy. They also don't work on Windows / PowerShell / CMD. e.g. see https://docs.docker.com/engine/reference/commandline/buildx_bake/#hcl-variables-and-functions

Screenshot 2021-07-28 at 11 57 02

We should consider splitting these to

1. show the file content
2. show the `docker builds bake .....` command that uses it.

Doing so also allows us to highlight the `hcl` document (if Jekyll / rouge supports the language)

```suggestion

`docker-bake.hcl`:

```hcl
variable "REPO" {
    default = "user/repo"
}

function "tag" {
    params = [tag]
    result = ["${REPO}:${tag}"]
}

target "webapp" {
    tags = tag("v1")
}
```

docs/reference/buildx_bake.md
```diff
@@ -361,14 +521,127 @@ EOF
...
+      "dockerfile": "Dockerfile",
+      "tags": [
+        "my-image:latest"
+      ]
+    }
+  }
+}
+```
+
+Example of using variables in functions:
```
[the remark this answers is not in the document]
crazy-max: Agree done!

crazy-max:
PTAL @thaJeztah

docs/reference/buildx_bake.md
```diff
@@ -211,18 +320,83 @@ Complete list of valid target fields:
...
 to define variables with values provided by the current environment, or a default
 value when unset.
 
+A [set of generally useful functions](https://github.com/docker/buildx/blob/master/bake/hclparser/stdlib.go)
+provided by [go-cty](https://github.com/zclconf/go-cty/tree/main/cty/function/stdlib)
+are available for use in HCL files. In addition, [user defined functions](https://github.com/hashicorp/hcl/tree/main/ext/userfunc)
+are also supported.
 
-Example of using interpolation to tag an image with the git sha:
+#### Using interpolation to tag an image with the git sha
```
thaJeztah, line 396: Not a show-stopper (would be ok for a follow-up), but it'd be nice to have a short introduction for each example. Admitted, with a descriptive "heading" it's a bit of "boiler plating", but it looks nice if there's a short introduction instead of a code-block directly after a heading, and it can lead in to what the example is illustrating.

Things along the lines of;

> The following example shows a docker-bake HCL file that specifies a `TAG` variable to allow customizing the build:

docs/reference/buildx_bake.md
```diff
@@ -234,45 +408,42 @@ group "default" {
...
+      "tags": [
+        "docker.io/username/webapp:latest"
+      ]
+    }
+  }
 }
+```
 
+```console
 $ TAG=$(git rev-parse --short HEAD) docker buildx bake --print webapp
```
thaJeztah, line 429: And then before this example, it could describe that the default value is overridden using the env-var.

thaJeztah (commented):
overall looks great. I left a suggestion for a follow-up (likely), one minor nit, and a comment about the `git://` scheme (could be a bug, not sure)

docs/reference/buildx_bake.md
```diff
@@ -89,6 +92,111 @@ $ docker buildx bake -f docker-compose.dev.yaml backend database
  ...
 ```
 
+You can also use a remote `git` bake definition:
+
+```console
+$ docker buildx bake "git://github.com/docker/cli#master" --print
```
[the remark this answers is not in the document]
crazy-max: You have to use a plain http request in this case so:

```console
$  docker buildx bake "https://raw.githubusercontent.com/docker/cli/master/docker-bake.hcl" --print
#1 [internal] load remote build context
#1 DONE 0.0s
{
  "target": {
    "binary": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "BASE_VARIANT": "alpine",
        "GO_STRIP": "",
        "VERSION": ""
      },
      "target": "binary",
      "platforms": [
        "local"
      ],
      "output": [
        "build"
      ]
    }
  }
}
```

But yeah we could enhance this behavior and detect the smart server response.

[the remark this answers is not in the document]
thaJeztah: Yeah, that's what I was wondering. I know buildkit already has code for that, so that's why it surprised me it didn't work in this case

docs/reference/buildx_bake.md
crazy-max: Done

thaJeztah (approved):
LGTM! Thank you!

PR #707, opened 2021-08-02
Enhance readme

crazy-max:
Readme enhanced and also wrapped to 80-chars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@
 [![Go Report Card](https://goreportcard.com/badge/github.com/docker/buildx)](https://goreportcard.com/report/github.com/docker/buildx)
 [![codecov](https://codecov.io/gh/docker/buildx/branch/master/graph/badge.svg)](https://codecov.io/gh/docker/buildx)
 
-`buildx` is a Docker CLI plugin for extended build capabilities with [BuildKit](https://github.com/moby/buildkit).
+`buildx` is a Docker CLI plugin for extended build capabilities with
+[BuildKit](https://github.com/moby/buildkit).
 
 Key features:
 
@@ -23,6 +24,7 @@ Key features:
   - [Docker](#docker)
   - [Binary release](#binary-release)
   - [From `Dockerfile`](#from-dockerfile)
+- [Set buildx as the default builder](#set-buildx-as-the-default-builder)
 - [Building](#building)
   - [with Docker 18.09+](#with-docker-1809)
   - [with buildx or Docker 19.03](#with-buildx-or-docker-1903)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -31,62 +33,82 @@ Key features:
...
+    - [`buildx imagetools inspect`](docs/reference/buildx_imagetools_inspect.md)  
+  - [`buildx inspect`](docs/reference/buildx_inspect.md)
+  - [`buildx install`](docs/reference/buildx_install.md)
+  - [`buildx ls`](docs/reference/buildx_ls.md)  
+  - [`buildx prune`](docs/reference/buildx_prune.md)  
+  - [`buildx rm`](docs/reference/buildx_rm.md)
+  - [`buildx stop`](docs/reference/buildx_stop.md)
+  - [`buildx uninstall`](docs/reference/buildx_uninstall.md)
+  - [`buildx use`](docs/reference/buildx_use.md)
+  - [`buildx version`](docs/reference/buildx_version.md)
```
crazy-max, line 52: I guess we could remove that from TOC as it's already available in `docs/reference/buildx.md`

README.md
```diff
@@ -95,65 +117,150 @@ $ make install
...
+including the new features in Docker 19.03 such as outputs configuration, inline
+build caching or specifying target platform. In addition, buildx supports new
+features not yet available for regular `docker build` like building manifest
+lists, distributed caching, exporting build results to OCI image tarballs etc.
+
+Buildx is supposed to be flexible and can be run in different configurations
+that are exposed through a driver concept. Currently, we support a
+[`docker` driver](docs/reference/buildx_create.md#driver) that uses the BuildKit
+library bundled into the Docker daemon binary, and a
+[`docker-container` driver](docs/reference/buildx_create.md#driver) that
```
AkihiroSuda, line 150: And kubernetes driver

crazy-max: Good point thanks for your feedback

crazy-max:
PTAL @thaJeztah

README.md
```diff
@@ -95,65 +117,152 @@ $ make install
...
+distributed caching, and exporting build results to OCI image tarballs.
+
+Buildx is supposed to be flexible and can be run in different configurations
+that are exposed through a driver concept. Currently, we support a
+[`docker` driver](docs/reference/buildx_create.md#docker-driver) that uses
+the BuildKit library bundled into the Docker daemon binary, a
+[`docker-container` driver](docs/reference/buildx_create.md#docker-container-driver)
+that automatically launches BuildKit inside a Docker container and a
+[`kubernetes` driver](docs/reference/buildx_create.md#kubernetes-driver) to
+spin up pods with defined BuildKit container image to build your images. We
```
tonistiigi, line 153: `for example, one that would allow  ...` I think we can remove this. Not sure what it even means.

README.md
```diff
@@ -95,65 +117,152 @@ $ make install
...
-```
+BuildKit is designed to work well for building for multiple platforms and not
+only for the architecture and operating system that the user invoking the build
+happens to run.
+
+When you invoke a build, you can set the `--platform` flag to specify the target
+platform for the build output, (for example, `linux/amd64`, `linux/arm64`, or
+`darwin/amd64`).
+
+When the current builder instance is backed by the `docker-container` driver,
```
tonistiigi, line 210: or kubernetes

README.md
```diff
@@ -95,65 +117,152 @@ $ make install
...
 ```
 
-Finally, depending on your project, the language that you use may have good support for cross-compilation. In that case, multi-stage builds in Dockerfiles can be effectively used to build binaries for the platform specified with `--platform` using the native architecture of the build node. List of build arguments like `BUILDPLATFORM` and `TARGETPLATFORM` are available automatically inside your Dockerfile and can be leveraged by the processes running as part of your build.
+Finally, depending on your project, the language that you use may have good
+support for cross-compilation. In that case, multi-stage builds in Dockerfiles
+can be effectively used to build binaries for the platform specified with
+`--platform` using the native architecture of the build node. A list of build
+arguments like `BUILDPLATFORM` and `TARGETPLATFORM` is available automatically
+inside your Dockerfile and can be leveraged by the processes running as part
+of your build.
```
tonistiigi, line 263: We can probably add a link to https://github.com/tonistiigi/xx for more advanced cases

docs/reference/buildx_create.md
```diff
@@ -92,17 +92,25 @@ See an [example buildkitd configuration file](https://github.com/moby/buildkit/b
...
+the [`--load`](buildx_build.md#--load) flag is implied by default on
+`buildx build`. However, building multi-platform images or exporting cache is
+not currently supported.
+
+#### `docker-container` driver
+
+Uses a BuildKit container that will be spawned via docker. With this driver,
+both building multi-platform images and exporting cache are supported.
+
+However, images built **will not automatically appear in `docker images`**
```
tonistiigi, line 107: Unlike `docker` driver, built images will not automatically appear in `docker images` and `--load` needs to be used to achieve that. (same for k8s)

tonistiigi (commented):
For the wrapping, I'm fine with changes but we shouldn't make it a requirement. It is an unnecessary maintenance burden to wrap changes by hand or set up special tooling. Markdown does not require it.

crazy-max:
@tonistiigi 

> For the wrapping, I'm fine with changes but we shouldn't make it a requirement. It is an unnecessary maintenance burden to wrap changes by hand or set up special tooling. Markdown does not require it.

Yeah it's kind-of controversial, it's an issue of style as well as ease of use. Don't think either we should make it a requirement but 80-100 characters is a good line length imo. The diff might get messy and make it hard to work out what's visually changed but it somehow drives me nuts to scroll horizontally _ad infinitum_ :sweat_smile:. Ofc the editor can take care of soft-wrapping so I'm not against reverting the wrapping.

tonistiigi (approved):

PR #712, opened 2021-08-03
Add HCL `env` func

crazy-max:
Atm we have to define a variable block to be able to retrieve an environment variable in a subsequent function or target. This PR implements an HCL `env` func to be able to retrieve an environment variable without defining a variable block.

Before:

```hcl
variable "GITHUB_REPOSITORY" {
}

variable "GITHUB_REF" {
}

target "build" {
  args = {
    GITHUB_REPOSITORY = GITHUB_REPOSITORY
    GITHUB_REF = GITHUB_REF
  }
  dockerfile = "./Dockerfile"
  target = "build"
}
```

```console
$ GITHUB_REPOSITORY=crazy-max/test docker buildx bake --print build
```

```json
{
  "target": {
    "build": {
      "context": ".",
      "dockerfile": "./Dockerfile",
      "args": {
        "GITHUB_REF": "",
        "GITHUB_REPOSITORY": "crazy-max/test",
      },
      "target": "build"
    }
  }
}
```

After:

```hcl
target "build" {
  args = {
    GITHUB_REPOSITORY = env("GITHUB_REPOSITORY")
    GITHUB_REF = env("GITHUB_REF")
  }
  dockerfile = "./Dockerfile"
  target = "build"
}
```

```console
$ GITHUB_REPOSITORY=crazy-max/test docker buildx bake --print build
```

```json
{
  "target": {
    "build": {
      "context": ".",
      "dockerfile": "./Dockerfile",
      "args": {
        "GITHUB_REF": "",
        "GITHUB_REPOSITORY": "crazy-max/test",
      },
      "target": "build"
    }
  }
}
```

### DIFF
--- a/bake/hclparser/stdlib.go
+++ b/bake/hclparser/stdlib.go
@@ -1,12 +1,15 @@
 package hclparser
 
 import (
+	"os"
+
 	"github.com/hashicorp/go-cty-funcs/cidr"
 	"github.com/hashicorp/go-cty-funcs/crypto"
 	"github.com/hashicorp/go-cty-funcs/encoding"
 	"github.com/hashicorp/go-cty-funcs/uuid"
 	"github.com/hashicorp/hcl/v2/ext/tryfunc"
 	"github.com/hashicorp/hcl/v2/ext/typeexpr"
+	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/function"
 	"github.com/zclconf/go-cty/cty/function/stdlib"
 )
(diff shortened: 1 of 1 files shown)


tonistiigi:
I'm not sure about this one. It's better for the build to define it's input, not where the values are coming from. Eg. when bake is moved to daemon side it does not have access to all env but the config can be still passed. Same for example in CI or then you want to save build config.

crazy-max:
@tonistiigi 

> when bake is moved to daemon side

Yes that makes sense.

Issue #396, opened 2020-09-18
Documentation: multiple tags by simply using multiple --tag parameters


jonstelly:
I had cobbled together a build process using buildx where I ran `buildx build ... --push` twice (with different tags) so I could build and publish for multiple tags.

```
docker buildx build . --platform=linux/arm64,linux/amd64 --tag my/image:0.1 --pull --push --no-cache
docker buildx build . --platform=linux/arm64,linux/amd64 --tag my/image:latest
```

It didn't immediately jump out at me that the `--tag` parameter could be specified multiple times until I happened to see `stringArray` as the type for the `--tag` parameter.  This lets me build and push to multiple tags in one step:

```
docker buildx build . --platform=linux/arm64,linux/amd64 --tag my/image:0.1 --tag my/image:latest --pull --push --no-cache
```

I'll try and find a minute to add this to the README.md and submit a PR in the next couple days but wanted to capture this as an issue to explain it first.

timorkal:
Please +1

yywing:
Is it work?  My test is not work.
https://github.com/docker/buildx/issues/459

ksurl:
does this work for different registries in the same command? e.g. tag and push to docker hub and github container registry

`docker buildx build . --platform=linux/amd64 -t user/image -t ghcr.io/user/image --push`

immanuelfodor:
Yes, it does, I'm actively using it in my CI to publish to an internal registry and DockerHub simultaneously: https://github.com/immanuelfodor/kube-karp/blob/master/.drone.yml#L59

crazy-max:
https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#options

dudicoco:
@crazy-max according to the docs `--tag` accepts a stringArray, however when attempting to pass multiple tags the following error occures:
```sh
$ docker buildx build -t latest,test
error: "docker buildx build" requires exactly 1 argument.
```

It seems that currently it is only possible to pass multiple `-t` arguments for each tag, even though the docs state otherwise.

Can you please explain this behaviour?

crazy-max:
@dudicoco 

> `$ docker buildx build -t latest,test`

[...]

PR #717, opened 2021-08-06
Document Option Types for stringSlice and stringArray

jonstelly:
Call out differences between stringSlice and stringArray.  Clarification for #396


tonistiigi:
Looks like `--allow` and other flags also handle multiple instances fine.

crazy-max:
Like @tonistiigi said, `StringSlice` flags take comma-separated value as arguments and split them accordingly. For example `--ss="v1,v2" --ss="v3"` will result in `[]string{"v1", "v2", "v3"}`.

Issue #183, opened 2019-10-30
docker buildx bake --print does not output default targets


Puneeth-n:
I have a simple docker-compose file with two services having build properties.

`docker buildx bake` builds both the images. However, 

`docker buildx bake --print` prints the equivalent `docker-bake.json` except the default targets

tonistiigi:
Sorry, I don't understand this report.

Please follow up with source files, commands you are running, expected results and actual results.

Puneeth-n:
Sorry for not being clear.

So, I have a `docker-compose.yml` file:

```yaml
# vim: set ts=2 sw=2 sts=2 et :

version: '3.7'

services:
  addon:
    build:
      context: .
      dockerfile: ./Dockerfile
      args:
        CT_ECR: foo
        CT_TAG: bar
    image: ct-addon:bar
    environment:
      - NODE_ENV=test
      - AWS_ACCESS_KEY_ID=dummy
      - AWS_SECRET_ACCESS_KEY=dummy
    command: tail -f /dev/null

  aws:
    build:
      context: ../../../tools/mock_servers/fake_aws
      dockerfile: ./Dockerfile
      args:
        CT_ECR: foo
        CT_TAG: bar
    image: ct-fake-aws:bar
```

I want convert this `docker-compose.yml` file into a `docker-bake.json` file. so I ran: 
```shell
DOCKER_CLI_EXPERIMENTAL=enabled docker buildx bake --print > docker-bake.json
```

```shell
DOCKER_CLI_EXPERIMENTAL=enabled docker buildx bake -f docker-bake.json
```

and I got an error **`failed to find target default`**

Puneeth-n:
Now come to think of it, https://github.com/docker/buildx#--print specifies that it should be used to display build targets.

tonistiigi:
Yes, this was not the expected behavior of `--print` that is used to show the current active configuration, not to generate a build file. To make the JSON file usable you would need to add `"group": {"default": ["addon", "aws"]}` to it.

This is an interesting use-case though. It might make sense for us to implicitly create a `default` group containing all targets in none was specified.

Puneeth-n:
Yes. We have lots of `docker-compose.yml` files that we use to build and runs tests. However, `docker-compose.yml` wasn't actually made to builld docker images. Now, we have the `bake` command which lets us build docker images efficiently. I am just afraid that we end up with tons of docker related files.

[...]

PR #721, opened 2021-08-08
bake: `x-bake` extension field with compose

crazy-max:
Ref. compose-spec/compose-spec#120 https://github.com/docker/buildx/issues/183#issuecomment-894698860

Might also need to update the [compose-cli build options](https://github.com/docker/compose-cli/blob/fff0ed2bcd4660454f83e31217bf19ebdf282e4b/pkg/compose/build.go#L255-L288) to handle `x-bake` in a follow-up.

cc @chris-crone 

### DIFF
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -80,13 +80,15 @@ func ParseCompose(dt []byte) (*Config, error) {
 					return val, ok
 				})),
 				CacheFrom: s.Build.CacheFrom,
-				// TODO: add platforms
+			}
+			if err = t.composeExtTarget(s.Build.Extensions); err != nil {
+				return nil, err
 			}
 			if s.Build.Target != "" {
 				target := s.Build.Target
 				t.Target = &target
 			}
-			if s.Image != "" {
+			if len(t.Tags) == 0 && s.Image != "" {
 				t.Tags = []string{s.Image}
 			}
 			c.Targets = append(c.Targets, t)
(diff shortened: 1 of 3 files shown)


tonistiigi:
Tags should be read from regular service tags. Cache-from also already exists.

crazy-max:
@tonistiigi 

> Tags should be read from regular service tags.

Tags are only handled as a single value through `image` field in compose, that's why it's available as an extension.

> Cache-from also already exists.

Ah yes [`cache_from` is available in the compose v2 spec](https://github.com/compose-spec/compose-spec/blob/master/build.md#cache_from).

dudicoco:
> @tonistiigi
> 
> > Tags should be read from regular service tags.
> 
> Tags are only handled as a single value through `image` field in compose afaik, that's why it's available as an extension here.
> 
> > Cache-from also already exists.
> 
> Ah yes [`cache_from` is available in the compose v2 spec](https://github.com/compose-spec/compose-spec/blob/master/build.md#cache_from).

Correct me if i'm wrong, but I believe the compose spec `cache_from` doesn't support all of the buildx cache features such as cache from local.

bake/compose.go
```diff
@@ -70,7 +79,7 @@ func ParseCompose(dt []byte) (*Config, error) {
 				dockerfilePathP = &dockerfilePath
 			}
 			g.Targets = append(g.Targets, s.Name)
-			t := &Target{
+			err = mergo.Merge(t, &Target{
```
tonistiigi, line 82: There is a custom `Merge` method for targets.

bake/compose.go
```diff
@@ -111,3 +119,74 @@ func flatten(in compose.MappingWithEquals) compose.Mapping {
...
+			switch key {
+			case "tags":
+				if res, k := val.(string); k {
+					t.Tags = append(t.Tags, res)
+				} else {
+					for _, res := range val.([]interface{}) {
+						t.Tags = append(t.Tags, res.(string))
+					}
+				}
+			case "cache-to":
```
tonistiigi, line 138: you can restore `cache-from` but make sure it overrides the general field if set.

bake/compose.go
crazy-max: I've simplified the behavior. Should be ok now.

tonistiigi (approved):

PR #735, opened 2021-08-18
Fix `--ssh` flag's documentation

johnhamelink:
The docs for the `--ssh` flag in the `buildx_build` file weren't rendering correctly, because the vertical bar characer (`|`) is used to define cells in a table in github-flavoured markdown.


tonistiigi:
This part of docs is generated. Therefore CI fails. https://github.com/docker/buildx/blob/master/docs/docsgen/generate.go#L162 should be the place to do the replacement. `hack/update-docs` `hack/validate-docs` when you are done.

tonistiigi (changes requested):
^

PR #739, opened 2021-08-19
Go 1.17

crazy-max:


Dockerfile
```diff
@@ -8,17 +8,9 @@ FROM docker:$DOCKERD_VERSION AS dockerd-release
 # xx is a helper for cross-compilation
 FROM --platform=$BUILDPLATFORM tonistiigi/xx@sha256:21a61be4744f6531cb5f33b0e6f40ede41fa3a1b8c82d5946178f80cc84bfc04 AS xx
 
-FROM --platform=$BUILDPLATFORM golang:1.16-alpine AS golatest
+FROM --platform=$BUILDPLATFORM golang:1.17-alpine AS golatest
```
thaJeztah, line 11: Should we have a test-matrices to test `current` and `n-1` versions of go?

also see https://github.com/docker/cli/pull/3258#issuecomment-900332386

go.mod
```diff
@@ -54,6 +54,90 @@ require (
...
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.21.0 // indirect
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.0.0-RC1 // indirect
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.0.0-RC1 // indirect
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.0.0-RC1 // indirect
+	go.opentelemetry.io/otel/sdk v1.0.0-RC1 // indirect
+	go.opentelemetry.io/proto/otlp v0.9.0 // indirect
+	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2 // indirect
+	golang.org/x/net v0.0.0-20210226172049-e18ecbb05110 // indirect
+	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d // indirect
+	golang.org/x/sys v0.0.0-20210426230700-d19ff857e887 // indirect
```
AkihiroSuda, line 124: Go 1.17 on macOS requires https://github.com/golang/sys/commit/a76c4d0a0096537dc565908b53073460d96c8539 or later

Dockerfile
crazy-max: > Should we have a test-matrices to test `current` and `n-1` versions of go?

Good idea I will add a dedicated matrix in a follow-up.

> also see [docker/cli#3258 (comment)](https://github.com/docker/cli/pull/3258#issuecomment-900332386)

Ah yes buildx is rebuilt through the packaging pipeline. Forgot about that.

AkihiroSuda (approved):

Dockerfile
tonistiigi: I'm not sure we need a matrix. This is a cli tool. Good pattern to leave them configurable though.

go.mod
```diff
@@ -54,6 +54,90 @@ require (
 	k8s.io/client-go v0.20.6
 )
 
+require (
```
tonistiigi, line 57: Don't like this double require here. Unless anyone has a reason why this is good it should be removed.

There is also https://github.com/golang/go/issues/46144 issue. I'm not sure if it affects projects importing buildx. If it does then we should wait with updating gomod. It doesn't seem to buy us anything anyway as the removed go.mod files do not affect binary size.

crazy-max: > Don't like this double require here. Unless anyone has a reason why this is good it should be removed.

With this second require we can see all transitively-imported packages so I tend to like it but I'm not against to merge.

> There is also golang/go#46144 issue.

:+1:

tonistiigi: Except there are a lot of indirects in the first block as well. Is there any guidance from go project for how this should be dealt with?

crazy-max: No guidance afaik so I guess we can merge them.

morlay:
A extra topic, could we  remove the whole `vendor` folder too?

tonistiigi:
> A extra topic, could we remove the whole vendor folder too?

No, I'm against that. Dependencies should tracked as other source code for reviews, edits, debugging etc.

crazy-max:
@tonistiigi PTAL, I'll leave at 1.16 in go.mod for now to avoid https://github.com/golang/go/issues/46144 in other projects.

AkihiroSuda (approved):

tonistiigi (approved):

go.mod
tonistiigi: https://github.com/moby/buildkit/pull/2331

Issue #621, opened 2021-05-28
(feature request) `docker buildx build --quiet` is currently not implemented.


wimax-grapl:
(Semi-related: https://github.com/docker/buildx/issues/401)

Right now, buildx supports `--quiet` strictly for `docker build` compatibility - but it doesn't actually quiet builds, it just prints out 
```
quiet currently not implemented
```
https://github.com/docker/buildx/blob/4fec647b9d8f34f8569141124d8462c912858144/commands/build.go#L83

Would be great to get this on the roadmap :)

wimax-grapl:
Oh, and here is a comment I found that proposes how to implement `--quiet`:
https://github.com/docker/buildx/pull/403#issuecomment-710757042

crazy-max:
`--quiet` was implemented in v0.7.0. It maps to `--progress=quiet` and prints only the image ID on success.

PR #740, opened 2021-08-19
Implement `--quiet` support

bossmc:
See #621.  I needed this in order to script logic of the following structure (where I didn't want to use a temporary file on disk for `--iidfile`):

```
$ image=$(docker build --quiet .)
$ docker run $image
```

This worked _unless_ someone had installed `buildx` in place of `docker build` (as I had 😛 ) in which case the `image` was an empty string and the `docker run` failed catastrophically.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -54,7 +54,6 @@ type Options struct {
 	BuildArgs   map[string]string
 	Pull        bool
 	ImageIDFile string
-	Quiet       bool
 	ExtraHosts  []string
 	NetworkMode string
 
@@ -685,9 +684,6 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 				respMu.Unlock()
 				if len(res) == 1 {
 					digest := res[0].ExporterResponse["containerimage.digest"]
-					if opt.Quiet {
-						fmt.Println(digest)
-					}
 					if opt.ImageIDFile != "" {
 						return ioutil.WriteFile(opt.ImageIDFile, []byte(digest), 0644)
 					}
@@ -717,9 +713,6 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 							dt, desc, err := itpull.Combine(ctx, names[0], descs)
 							if err != nil {
 								return err
-							}
-							if opt.Quiet {
-								fmt.Println(desc.Digest)
 							}
 							if opt.ImageIDFile != "" {
 								return ioutil.WriteFile(opt.ImageIDFile, []byte(desc.Digest), 0644)
--- a/build/build.go
+++ b/build/build.go
@@ -54,6 +54,7 @@ type Options struct {
 	BuildArgs   map[string]string
 	Pull        bool
 	ImageIDFile string
+	Quiet       bool
 	ExtraHosts  []string
 	NetworkMode string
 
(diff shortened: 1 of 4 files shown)


build/build.go
```diff
@@ -683,8 +684,12 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 				resp[k] = res[0]
 				respMu.Unlock()
 				if len(res) == 1 {
+					digest := res[0].ExporterResponse["containerimage.digest"]
+					if opt.Quiet {
+						fmt.Println(digest)
```
tonistiigi, line 689: build pkg should not write to stdio. Read the value from the return params in command pkg.

bossmc: Yeah, I thought this felt wrong - I'll look at using the return value and update as you suggest.

bossmc:
@tonistiigi - Thanks for the review, I've updated the change (to pass the digest back to the main command driver and manipulate `stdout` there).

Possibly this is slightly presumptuous - it assumes that the target is called `default` and that the `Exporter` will add a `containerimage.digest` field.  The former is already assumed earlier in `buildTargets` and the latter is assumed in `build.Build` so I think these are safe assumptions/guaranteed.

commands/build.go
```diff
@@ -108,6 +104,12 @@ func runBuild(dockerCli command.Cli, in buildOptions) (err error) {
 		pull = *in.pull
 	}
 
+	if in.quiet && in.progress != "auto" && in.progress != "quiet" {
+		return errors.Errorf("progress=%s and quiet cannot be used together", in.progress)
```
tonistiigi, line 108: I wonder if this should be an error. Maybe `-q` just overrides. Thinking of scripts where different components add their flags.

tonistiigi (approved):

build/build.go
```diff
@@ -683,8 +683,9 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 				resp[k] = res[0]
 				respMu.Unlock()
 				if len(res) == 1 {
+					digest := res[0].ExporterResponse["containerimage.digest"]
 					if opt.ImageIDFile != "" {
```
AkihiroSuda, line 687: Maybe error out if digest is empty

commands/build.go
bossmc: My thinking was that explicitly asking for "fancy output"/"streamed output" and "no output" was user-error and should be rejected, can a component add a new `--progress` type?.

I allowed the user to pass `--progress=quiet -q` which is (possibly surprisingly) different to passing `--progress=quiet` on its own (the former doesn't print the digest at the end), since that's not contradictory.

build/build.go
bossmc: The previous code would have written an empty file to `--iidfile` in this case, changing that to an error would be an externally visible change.

In fact, looking at the code now, none of the changes in this file are actually needed, they're just a refactor due to me originally printing the digest here (and hence pulling the digest out for printing/writing to the `--iidfile`).

AkihiroSuda (approved):

commands/build.go
tonistiigi: I thought for a case where `--progress=plain` is used to not detect tty and then another part does not want progress at all.

build/build.go
tonistiigi: Maybe best in this case would be to always set `-o type=image` if image id is to be exported, either with `iidfile`, or `q`. And error if another conflicting output is already set. This can be follow-up though as strictly related.

tonistiigi (approved):

urielsegura03 (approved):

PR #746, opened 2021-08-24
Bake workflow

crazy-max:
Experiment to switch to buildx bake workflow in this repo.

* [x] Dockerized authors script and regenerate AUTHORS
* [x] Vendoring
* [x] Lint
* [x] Docs (will make a follow-up pr to use [cli-docs-tool](https://github.com/docker/cli-docs-tool))
* [x] Binaries
* [x] Release
* [x] Test
* [x] Add bake definition
* [x] Update GitHub Actions workflows


crazy-max:
Build looks faster if I use a single job on GitHub Actions.

Previously ~34m:
image

Now ~17m:
image

crazy-max:
PTAL @thaJeztah @tonistiigi 

Would need to change the status check in branche protection rules settings to:
* Include `build` and `validate (docs-validate)`
* Remove `binaries`, `cross`
* Switch `validate (validate-vendor)` with `validate (vendor-validate)`

Makefile
```diff
@@ -1,40 +1,66 @@
+ifeq (, $(shell docker buildx version))
```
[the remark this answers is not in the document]
crazy-max: @tonistiigi As discussed I've moved docs/vendor scripting from Makefile in dedicated files under `./hack`.

In a follow-up we could start thinking about adding an attribute to the local exporter via [`filesync.CopyToCaller`](https://github.com/moby/buildkit/blob/master/exporter/local/export.go#L121) on BuildKit to be able to cleanup `dest` beforehand as discussed. As a quick thought:

* `--output type=local,dest=path/to/output-dir,mode=copy` (copy files/folders, current behavior)
* `--output type=local,dest=path/to/output-dir,mode=mirror` (mirror, delete dest files/folders that no longer exist in source)
* `--output type=local,dest=path/to/output-dir,mode=backup` (backup files that may already exist in dest, `.~<numbered>~` will be the suffix. Each backup file will have a version number added to the filename. e.g. `path/to/output-dir/myfile.txt.~1~`. Pretty much like `cp --backup=numbered`)

Might need some changes upstream on https://github.com/tonistiigi/fsutil.

crazy-max:
PTAL @tonistiigi

docker-bake.hcl
```diff
@@ -0,0 +1,128 @@
...
+  targets = ["lint", "vendor-validate", "docs-validate"]
+}
+
+target "lint" {
+  inherits = ["_common"]
+  dockerfile = "./hack/dockerfiles/lint.Dockerfile"
+  output = ["type=cacheonly"]
+}
+
+target "vendor-validate" {
```
tonistiigi, line 37: I know you are trying to group things together but this order "vendor-validate" looks so unnatural to me compared to "validate-vendor". I feel like I would need to check this file every time for the target name if we leave it like this.

docker-bake.hcl
```diff
@@ -0,0 +1,128 @@
...
+}
+variable "BIN_OUT" {
+  default = "./bin"
+}
+variable "RELEASE_OUT" {
+  default = "./release-out"
+}
+
+// Special target: https://github.com/docker/metadata-action#bake-definition
+target "docker-metadata-action" {
```
tonistiigi, line 12: No changes are needed. We already discussed it probably but in this context, a name like "_ci_config", "_ci_inputs" would make more sense. Doesn't need to be builtin if metadata-action allows override in github yaml.

hack/dockerfiles/docs.Dockerfile
```diff
@@ -13,19 +13,28 @@ RUN apk add --no-cache rsync git
 WORKDIR /src
 COPY --from=docsgen /out/docsgen /usr/bin
 RUN --mount=target=/context \
-  --mount=target=.,type=tmpfs,readwrite  \
-  rsync -a /context/. . && \
-  docsgen && \
-  mkdir /out && cp -r docs/reference /out
+  --mount=target=.,type=tmpfs,readwrite <<EOT
```
tonistiigi, line 16: I wonder why this readwrite is here?

docker-bake.hcl
crazy-max: Yes, I was wondering about that too. Sgtm to use `validate-xxx`.

hack/dockerfiles/docs.Dockerfile
crazy-max: I think it was added to sync `/context` with the generated docs.

docker-bake.hcl
crazy-max: > We already discussed it probably but in this context, a name like "_ci_config", "_ci_inputs" would make more sense.

Named `meta-helper` like in binfmt repo.

> Doesn't need to be builtin if metadata-action allows override in github yaml.

Wdym? Remove this target and let the metadata-action handles the inheritance in `image` and `image-cross` targets with smth like `--set image.inherits.push(meta-helper)`? `push()` would add a value to the existing `inherits` slice.

hack/dockerfiles/docs.Dockerfile
tonistiigi: I mean what tmpfs is not readwrite already?

docker-bake.hcl
tonistiigi: Just meant `bake-target: _ci_inputs` . Didn't know metadata-action already allowed setting the name.

hack/dockerfiles/docs.Dockerfile
crazy-max: Indeed, tmpfs is readwrite by default

tonistiigi (approved):

PR #748, opened 2021-08-24
Built-in variable `BAKE_LOCAL_PLATFORM`

crazy-max:
Add built-in variable `BAKE_LOCAL_PLATFORM` that returns the current platform's default platform specification (e.g. `linux/amd64`).

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/containerd/containerd/platforms"
 	"github.com/docker/buildx/bake"
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/util/progress"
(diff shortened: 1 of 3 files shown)


tonistiigi:
What's the benefit of this compared to "local" that is accepted by `docker build(x)` ? Also, "platform" is not a good name for this. Eg. build's default platform is the  platform of the server that is definitely more common. Building for a local CLI system is more like an edge case and only works reliably in cross-compile dockerfiles as most users are on macos/windows..

crazy-max:
@tonistiigi 

> Could also be builtin var. We already added one for remotes and want to add another for `--install`. Probably more in the future.

Yes a builtin var looks better. `BAKE_LOCAL_PLATFORM` lgty?

tonistiigi (approved):

PR #751, opened 2021-08-27
container-driver: fix volume destination for cache

thaJeztah:
fixes https://github.com/docker/buildx/issues/749

The container-driver creates a Linux container (as there currently isn't a
Windows version of buildkitd). However, the defaults are platform specific.

Buildx was using the defaults from the buildkit `util/appdefault' package,
which resulted in Buildx running on a Windows client to create a Linux
container that used the Windows location, which causes it to fail:

    invalid mount config for type "volume": invalid mount path: 'C:/ProgramData/buildkitd/.buildstate' mount path must be absolute

This patch hard-codes the destination to the default Linux path.

PR #752, opened 2021-08-28
[v0.6] container-driver: fix volume destination for cache

tonistiigi:
pick #751

The container-driver creates a Linux container (as there currently isn't a
Windows version of buildkitd). However, the defaults are platform specific.

Buildx was using the defaults from the buildkit `util/appdefault' package,
which resulted in Buildx running on a Windows client to create a Linux
container that used the Windows location, which causes it to fail:

    invalid mount config for type "volume": invalid mount path: 'C:/ProgramData/buildkitd/.buildstate' mount path must be absolute

This patch hard-codes the destination to the default Linux path.

Signed-off-by: Sebastiaan van Stijn <github@gone.nl>
(cherry picked from commit 93867d02f0a989f084e54d7133371f583a4d3c5e)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,11 +5,13 @@ on:
   push:
     branches:
       - 'master'
+      - 'v[0-9]*'
     tags:
       - 'v*'
   pull_request:
     branches:
       - 'master'
+      - 'v[0-9]*'
 
 env:
   REPO_SLUG: "docker/buildx-bin"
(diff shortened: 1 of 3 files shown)


thaJeztah:
ah! didn't realize we started a branch (thought we still did 0.6 from master)

I should probably have a look if we need to get some of the docs changes into the branch (but that can wait)

https://github.com/docker/buildx/compare/v0.6...master

thaJeztah (approved):
LGTM

PR #747, opened 2021-08-24
Generate YAML doc

crazy-max:
Use [cli-docs-tool](https://github.com/docker/cli-docs-tool) to generate Markdown and now YAML docs.

@thaJeztah I have not used a sub gomod to avoid copy-pasting "require(...)" from the main one. `github.com/docker/cli-docs-tool` will not be used by `./cmd/buildx` as shown with the command below:

```shell
$ cd ./cmd/buildx/
$ go list -deps -f '{{define "M"}}{{.Path}}@{{.Version}}{{end}}{{with .Module}}{{if not .Main}}{{if .Replace}}{{template "M" .Replace}}{{else}}{{template "M" .}}{{end}}{{end}}{{end}}' | sort -u
cloud.google.com/go@v0.81.0
github.com/agext/levenshtein@v1.2.1
github.com/agl/ed25519@v0.0.0-20170116200512-5312a6153412
github.com/apparentlymart/go-cidr@v1.0.1
github.com/apparentlymart/go-textseg/v12@v12.0.0
github.com/beorn7/perks@v1.0.1
github.com/cenkalti/backoff/v4@v4.1.1
github.com/cespare/xxhash/v2@v2.1.1
github.com/compose-spec/compose-go@v0.0.0-20210729195839-de56f4f0cb3c
github.com/containerd/console@v1.0.2
github.com/containerd/containerd@v1.5.5
github.com/containerd/continuity@v0.1.0
github.com/containerd/typeurl@v1.0.2
github.com/davecgh/go-spew@v1.1.1
github.com/distribution/distribution/v3@v3.0.0-20210316161203-a01c71e2477e
github.com/docker/cli@v20.10.3-0.20210702143511-f782d1355eff+incompatible
github.com/docker/compose-on-kubernetes@v0.4.19-0.20190128150448-356b2919c496
github.com/docker/distribution@v2.7.1+incompatible
github.com/docker/docker-credential-helpers@v0.6.4-0.20210125172408-38bea2ce277a
github.com/docker/docker@v20.10.3-0.20210609100121-ef4d47340142+incompatible
github.com/docker/go-connections@v0.4.0
github.com/docker/go-metrics@v0.0.1
github.com/docker/go-units@v0.4.0
github.com/docker/go@v1.5.1-1.0.20160303222718-d30aec9fd63c
github.com/docker/spdystream@v0.0.0-20181023171402-6480d4af844c
github.com/fvbommel/sortorder@v1.0.1
github.com/go-logr/logr@v0.2.0
[...]

PR #762, opened 2021-09-08
Remove generated YAML docs

crazy-max:
As discussed in #747, remove generated YAML docs from this repository but keep bake targets to be able to trigger YAML generation from [docker/docker.github.io](https://github.com/docker/docker.github.io/) in a follow-up.


tonistiigi (approved):
In CI we can make sure that yaml generation passes cleanly. So there are no surprises when something breaks.

thaJeztah (approved):
LGTM

> but keep bake targets to be able to trigger YAML generation from docker/docker.github.io in a follow-up.

Perhaps we can make CI push an image to Docker Hub that contains the docs ("latest", "release branch" and/or "tag") that we can use in the docs repository to either copy the files from when building the docs, or to update the docs there

crazy-max:
> Perhaps we can make CI push an image to Docker Hub that contains the docs ("latest", "release branch" and/or "tag") that we can use in the docs repository to either copy the files from when building the docs, or to update the docs there

Yeah we can do that too with a `docker/buildx-docs` repo on the registry but as you said Netlify needs the files on the repo for preview so I guess the remote bake definition is our best shot atm: https://github.com/docker/docker.github.io/pull/13488

PR #773, opened 2021-09-20
bake: fix using push override with output definition

tonistiigi:
Currently, using `bake --push` will replace the target definition with `type=registry`. This means that if the definition defined an output then it gets cleared and completely overwritten.

I think the basis of this refactor can also be used to implement additional operators like `--set key+=value` in the future.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -2,6 +2,7 @@ package bake
 
 import (
 	"context"
+	"encoding/csv"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -27,6 +28,11 @@ var gitURLPathWithFragmentSuffix = regexp.MustCompile(`\.git(?:#.+)?$`)
 type File struct {
 	Name string
 	Data []byte
+}
+
+type Override struct {
+	Key   string
+	Value string
 }
 
 func defaultFilenames() []string {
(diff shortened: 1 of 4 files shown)


crazy-max (approved):
LGTM

> I think the basis of this refactor can also be used to implement additional operators like `--set key+=value` in the future.

Wonder if we could also add this kind of logic in the HCL def like:

```hcl
# docker-bake.hcl
target "webapp" {
    tags = ["username/webapp:latest"]
}
```

```hcl
# docker-bake2.hcl
target "webapp" {
    extends {
        tags = ["ghcr.io/username/webapp:latest"]
    }
}
```

tonistiigi:
@crazy-max With referencing targets you could do

```
target "webapp" {
    tags = concat(target.webapp.tags, ["ghcr.io/username/webapp:latest"])
}
```

kevbook:
I am also trying to merge tags, but I am getting the following error with this syntax. Any help would be appreciated (no rush though).

```hcl
target "webapp" {
    tags = concat(target.baseimg.tags, ["ghcr.io/username/webapp:latest"])
}

ERROR: build-pipeline.hcl: Unknown variable; There is no variable named "target"., and 1 other diagnostic(s)
```

PR #774, opened 2021-09-21
Driver e2e tests

crazy-max:
Follow-up #691 

This workflow handles multi-matrix jobs with:
* buildkit image (latest and master)
* buildkit cfg
* driver (docker, container and kubernetes)
* multi-node (restrict to kubernetes atm)

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,100 @@
+name: e2e
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'master'
+      - 'v[0-9]*'
+  pull_request:
+    branches:
+      - 'master'
+      - 'v[0-9]*'
+
+jobs:
+  driver:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        driver:
+          - docker
+          - docker-container
+          - kubernetes
+        buildkit:
+          - moby/buildkit:buildx-stable-1
+          - moby/buildkit:master
+        buildkit-cfg:
+          - bkcfg-false
+          - bkcfg-true
+        multi-node:
+          - mnode-false
+          - mnode-true
+        platforms:
+          - linux/amd64,linux/arm64
+        include:
+          - driver: kubernetes
+            driver-opt: qemu.install=true
+        exclude:
+          - driver: docker
+            multi-node: mnode-true
+          - driver: docker
+            buildkit-cfg: bkcfg-true
+          - driver: docker-container
+            multi-node: mnode-true
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        if: matrix.driver == 'docker' || matrix.driver == 'docker-container'
+      -
+        name: Install buildx
+        run: |
+          make install
+          docker buildx version
+      -
+        name: Init env vars
+        run: |
+          # BuildKit cfg
+          if [ "${{ matrix.buildkit-cfg }}" = "bkcfg-true" ]; then
+            cat > "/tmp/buildkitd.toml" <<EOL
+          [worker.oci]
+            max-parallelism = 2
+          EOL
+            echo "BUILDKIT_CFG=/tmp/buildkitd.toml" >> $GITHUB_ENV
+          fi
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+          - 0
+          - 1
+        multi-node:
+          - 0
+          - 1
+        platforms:
+          - linux/amd64,linux/arm64
+        include:
+          - driver: kubernetes
+            driver-opt: qemu.install=true
```
crazy-max, line 37: @morlay `driver-opt` can be used for a specific driver like `qemu.install=true` for kubernetes.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+        driver:
+          - docker
+          - docker-container
+          - kubernetes
+        buildkit:
+          - moby/buildkit:buildx-stable-1
+          - moby/buildkit:master
+        buildkit-cfg:
+          - 0
+          - 1
```
crazy-max, line 29: Support for `--config` is also tested for kubernetes and container drivers.

hack/test-driver
```diff
@@ -0,0 +1,99 @@
...
+RUN echo "I am running on \$BUILDPLATFORM, building for \$TARGETPLATFORM" > /log
+FROM busybox
+COPY --from=build /log /log
+RUN cat /log
+RUN uname -a
+EOL
+
+# build
+${BUILDX_CMD} build ${buildPlatformFlag} \
+  --output="type=cacheonly" \
```
crazy-max, line 97: For now I use `type=cacheonly` as output but in the future we could push to a local registry using [GHA services](https://github.com/docker/build-push-action/blob/375f72aff65b645c51f122f04a776df8416b222b/.github/workflows/ci.yml#L153-L157) or private registries like we do in [build-push-action](https://github.com/docker/build-push-action/blob/master/.github/workflows/e2e.yml).

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+      - 'master'
+      - 'v[0-9]*'
+  pull_request:
+    branches:
+      - 'master'
+      - 'v[0-9]*'
+
+jobs:
+  driver:
+    runs-on: ubuntu-latest
```
AkihiroSuda, line 16: nit: ubuntu-20.04 might be slightly better for reproducibility

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+      fail-fast: false
+      matrix:
+        driver:
+          - docker
+          - docker-container
+          - kubernetes
+        buildkit:
+          - moby/buildkit:buildx-stable-1
+          - moby/buildkit:master
+        buildkit-cfg:
```
AkihiroSuda, line 27: better to use a form like `buildkit-cfg: ["buildkit-cfg-true", "buildkit-cfg-false"]`
(Please see the screenshot in my other comment)

crazy-max: sgtm

crazy-max:
PTAL @AkihiroSuda @morlay

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,99 @@
...
+            buildkit-cfg: bkcfg-true
+          - driver: docker-container
+            multi-node: mnode-true
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
```
morlay, line 51: This step may only when `driver` is `docker` or `docker-container`

could be

```
uses: docker/setup-qemu-action@v1
if:  ${{ matrix.driver == 'docker' && matrix.driver == 'docker-container'  }}
```

AkihiroSuda (approved):

tonistiigi (approved):
Let's merge this but as a follow-up, I'd like `./test-driver` to run inside a container. So it is not specific to github actions but can be invoked from any system.

`./hack/shell` already creates such a container today. I tried running `./test-driver` inside that container and it worked for both docker and container driver. It doesn't work for k3s atm as it is not installed but I don't see a big reason why it couldn't be.

We can also discuss in later if it makes sense to use bats or go tests for this but ok for a starter.

PR #782, opened 2021-09-29
docker-container: place build containers in a separate cgroup

djs55:
Previously build containers created by the `docker-container` driver were in the default parent cgroup, along with non-build containers. This made it hard to apply resource limits to all the builds on a machine.

This PR adds a `--driver-opt cgroup-parent=CGROUP` to allow the cgroup to be customised. A default value of `/docker/buildx` is  set.

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -38,10 +38,11 @@ const (
 
 type Driver struct {
 	driver.InitConfig
-	factory driver.Factory
-	netMode string
-	image   string
-	env     []string
+	factory      driver.Factory
+	netMode      string
+	image        string
+	cgroupParent string
+	env          []string
 }
 
 func (d *Driver) IsMobyDriver() bool {
@@ -124,6 +125,11 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 		}
 		if d.netMode != "" {
 			hc.NetworkMode = container.NetworkMode(d.netMode)
+		}
+		if info, err := d.DockerAPI.Info(ctx); err == nil && info.CgroupDriver == "cgroupfs" {
+			if d.cgroupParent != "" {
+				hc.CgroupParent = d.cgroupParent
+			}
 		}
 		_, err := d.DockerAPI.ContainerCreate(ctx, cfg, hc, &network.NetworkingConfig{}, nil, d.Name)
 		if err != nil {
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -127,6 +127,9 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 			hc.NetworkMode = container.NetworkMode(d.netMode)
 		}
 		if info, err := d.DockerAPI.Info(ctx); err == nil && info.CgroupDriver == "cgroupfs" {
+			// Place all buildkit containers inside this cgroup by default so limits can be attached
+			// to all build activity on the host.
+			hc.CgroupParent = "/docker/buildx"
 			if d.cgroupParent != "" {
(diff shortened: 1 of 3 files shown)


thaJeztah:
I noticed that the buildkit code embedded in dockerd also does some setting up of a parent cgroup (originally added (extracted to a function later) in https://github.com/moby/moby/commit/d52485c2f901e62e9a26445e230ec3f6662fa276);

https://github.com/moby/moby/blob/c4040417b6fe21911dc7ab5e57db27519dd44a6a/cmd/dockerd/daemon.go#L293

```go
	cgroupParent := newCgroupParent(config)
```

https://github.com/moby/moby/blob/2b70006e3bfa492b8641ff443493983d832955f4/cmd/dockerd/daemon_unix.go#L125

```go
func newCgroupParent(config *config.Config) string {
	cgroupParent := "docker"
	useSystemd := daemon.UsingSystemd(config)
	if useSystemd {
		cgroupParent = "system.slice"
	}
	if config.CgroupParent != "" {
		cgroupParent = config.CgroupParent
	}
	if useSystemd {
		cgroupParent = cgroupParent + ":" + "docker" + ":"
	}
	return cgroupParent
}
```

- Are those used for the same purpose? (or is the above only used for containers used _during_ build? (I _think_ it may be only for build-containers, and therefore is set to match `dockerd --cgroup-parent` option)
- The code above looks to be taking systemd cgroups into account; does that mean we need to set a _different_ default depending on wether the daemon that runs the container uses systemd or not?


For the last bullet, I'm looking at the daemon code that handles (custom) parent cgroups for containers, which seems to confirm that; https://github.com/moby/moby/blob/306fa44b7ca59282dc8695e6d169c5b25698d0cb/daemon/daemon_unix.go#L709-L714

```go
if hostConfig.CgroupParent != "" && UsingSystemd(daemon.configStore) {
	// CgroupParent for systemd cgroup should be named as "xxx.slice"
	if len(hostConfig.CgroupParent) <= 6 || !strings.HasSuffix(hostConfig.CgroupParent, ".slice") {
		return warnings, fmt.Errorf("cgroup-parent for systemd cgroup should be a valid slice named as \"xxx.slice\"")
	}
}
```

Wondering how to solve that

-  `buildx` could inspect the daemon's configuration before creating the container (and use format "A" or "B")
- some special value (specific for the `buildx builder` case) to pass, and have the daemon perform the check and pick the right format to use
- (not sure if possible) some "uniform" format that the client can use, and the daemon to convert it to the right format for the situation (this would benefit "any" container, but A) not sure if possible, and B) not sure if it's useful; setting a custom parent cgroup usually would mean that that cgroup has to be created up front?)

tonistiigi (approved):

driver/docker-container/driver.go
```diff
@@ -121,10 +122,18 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
...
+				// Place all buildkit containers inside this cgroup so limits can be attached
+				// to all build activity on the host.
+				CgroupParent: "/docker/buildx",
+			},
 		}
 		if d.netMode != "" {
 			hc.NetworkMode = container.NetworkMode(d.netMode)
 		}
+		if d.cgroupParent != "" {
+			hc.CgroupParent = d.cgroupParent
```
AkihiroSuda, line 135: Probably, this should be set only when  `d.DockerAPI.Info.CgroupDriver == “cgroupfs”`

djs55:
Thanks, I've pushed an update which gates the setting on `CgroupDriver == "cgroupfs"` so hopefully it doesn't break anything if using systemd. I must admit I'm not very familiar with systemd, I need to spend some time playing with it.

AkihiroSuda (approved):

PR #783, opened 2021-09-29
bake: restore previous override merge behavior

tonistiigi:
This broke with #773. The added test works with pre-#773 as well

For array fields, overrides are merged together
but override is not merged with the target. If merging
with the target is desired we can add support for
overrides with += operator in the future.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -31,8 +31,8 @@ type File struct {
 }
 
 type Override struct {
-	Key   string
-	Value string
+	Value    string
+	ArrValue []string
 }
 
 func defaultFilenames() []string {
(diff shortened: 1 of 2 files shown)


bake/bake.go
```diff
@@ -503,21 +509,21 @@ func (t *Target) AddOverrides(overrides []Override) error {
 			}
 			t.Labels[keys[1]] = value
 		case "tags":
-			t.Tags = append(t.Tags, value)
+			t.Tags = o.ArrValue
 		case "cache-from":
 			t.CacheFrom = append(t.CacheFrom, value)
```
tonistiigi, line 514: ^

crazy-max (approved):
Works fine on my side

PR #785, opened 2021-09-29
Fix --driver flag usage markdown output

crazy-max:
Also render value as code for other flags in markdown.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -277,30 +277,30 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 	flags := cmd.Flags()
 
-	flags.BoolVar(&options.exportPush, "push", false, "Shorthand for --output=type=registry")
-	flags.BoolVar(&options.exportLoad, "load", false, "Shorthand for --output=type=docker")
+	flags.BoolVar(&options.exportPush, "push", false, "Shorthand for `--output=type=registry`")
+	flags.BoolVar(&options.exportLoad, "load", false, "Shorthand for `--output=type=docker`")
 
-	flags.StringArrayVarP(&options.tags, "tag", "t", []string{}, "Name and optionally a tag in the 'name:tag' format")
+	flags.StringArrayVarP(&options.tags, "tag", "t", []string{}, "Name and optionally a tag (format: `name:tag`)")
 	flags.SetAnnotation("tag", "docs.external.url", []string{"https://docs.docker.com/engine/reference/commandline/build/#tag-an-image--t"})
 	flags.StringArrayVar(&options.buildArgs, "build-arg", []string{}, "Set build-time variables")
 	flags.SetAnnotation("build-arg", "docs.external.url", []string{"https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg"})
 
-	flags.StringVarP(&options.dockerfileName, "file", "f", "", "Name of the Dockerfile (Default is 'PATH/Dockerfile')")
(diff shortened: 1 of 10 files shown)


commands/build.go
```diff
@@ -277,30 +277,30 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
...
 
-	flags.StringVarP(&options.dockerfileName, "file", "f", "", "Name of the Dockerfile (Default is 'PATH/Dockerfile')")
+	flags.StringVarP(&options.dockerfileName, "file", "f", "", "Name of the Dockerfile (default: `PATH/Dockerfile`)")
 	flags.SetAnnotation("file", "docs.external.url", []string{"https://docs.docker.com/engine/reference/commandline/build/#specify-a-dockerfile--f"})
 
 	flags.StringArrayVar(&options.labels, "label", []string{}, "Set metadata for an image")
 
-	flags.StringArrayVar(&options.cacheFrom, "cache-from", []string{}, "External cache sources (eg. user/app:cache, type=local,src=path/to/dir)")
-	flags.StringArrayVar(&options.cacheTo, "cache-to", []string{}, "Cache export destinations (eg. user/app:cache, type=local,dest=path/to/dir)")
+	flags.StringArrayVar(&options.cacheFrom, "cache-from", []string{}, "External cache sources (e.g. `user/app:cache`, `type=local,src=path/to/dir`)")
```
thaJeztah, line 293: I think it's common to put a comma after `e.g.`, so

```suggestion
	flags.StringArrayVar(&options.cacheFrom, "cache-from", []string{}, "External cache sources (e.g., `user/app:cache`, `type=local,src=path/to/dir`)")
```

commands/create.go
```diff
@@ -221,7 +223,7 @@ func createCmd(dockerCli command.Cli) *cobra.Command {
 	flags := cmd.Flags()
 
 	flags.StringVar(&options.name, "name", "", "Builder instance name")
-	flags.StringVar(&options.driver, "driver", "", fmt.Sprintf("Driver to use (available: %v)", drivers))
+	flags.StringVar(&options.driver, "driver", "", fmt.Sprintf("Driver to use (available: `%v`)", drivers))
```
thaJeztah, line 226: Not a stopper for this PR, but wondering if we should reformat the `drivers` to the same (comma-separated) format we use for some other options (e.g., `--progress` shows: `(auto, plain, tty)`).

commands/build.go
```diff
@@ -277,30 +277,30 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 	flags := cmd.Flags()
 
-	flags.BoolVar(&options.exportPush, "push", false, "Shorthand for --output=type=registry")
-	flags.BoolVar(&options.exportLoad, "load", false, "Shorthand for --output=type=docker")
+	flags.BoolVar(&options.exportPush, "push", false, "Shorthand for `--output=type=registry`")
+	flags.BoolVar(&options.exportLoad, "load", false, "Shorthand for `--output=type=docker`")
 
-	flags.StringArrayVarP(&options.tags, "tag", "t", []string{}, "Name and optionally a tag in the 'name:tag' format")
+	flags.StringArrayVarP(&options.tags, "tag", "t", []string{}, "Name and optionally a tag (format: `name:tag`)")
```
thaJeztah, line 283: We should update the `docker` cli descriptions to have the same (for when `DOCKER_BUILDKIT=0`)

thaJeztah (approved):
LGTM, but left some thoughts/sugestions

commands/create.go
crazy-max: Sure will do that

tonistiigi (approved):

Issue #567, opened 2021-03-23
Builder config for ssh certs or for insecure remote registry pull


suityou01:
My command

`docker buildx build --platform linux/arm/64  -t 192.168.1.5:443/bpmcore_openid_provider_storage:2 . --load`

Is failing with 

> error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: failed to do request: Head https://192.168.1.5:443/v2/cbslabs/rejson/arm64/manifests/1: x509: certificate signed by unknown authority

How can I configure an insecure remote registry for the builder?

tonistiigi:
Configure registry exceptions in buildkit config file https://github.com/moby/buildkit/blob/master/docs/buildkitd.toml.md and set it with `--config` on `buildx create`

cldfzn:
Is there any useful and complete documentation on setting up buildkit with non-public certs?

Issue #455, opened 2020-11-26
Don't push image on private registry


sminodonte:
I run this command:

docker buildx build --platform linux/arm64 -t <IP:_PRIVATE_REGISTRY>:5000/i<IMAGE_NAME>:latest . --push

and the result is:


```
WARN[0000] invalid non-bool value for BUILDX_NO_DEFAULT_LOAD:  
[+] Building 5.1s (11/11) FINISHED                                                                                                                                                                                                                                                                                                                                          
 => [internal] load .dockerignore                                                                                                                                                                                                                                                                                                                                      0.1s
 => => transferring context: 2B                                                                                                                                                                                                                                                                                                                                        0.0s
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                                                                                                                   0.1s
 => => transferring dockerfile: 60B                                                                                                                                                                                                                                                                                                                                    0.0s
 => [internal] load metadata for docker.io/library/node:10.15.3-alpine                                                                                                                                                                                                                                                                                                 2.7s
 => [1/6] FROM docker.io/library/node:10.15.3-alpine@sha256:aa28f3b6b4087b3f289bebaca8d3fb82b93137ae739aa67df3a04892d521958e                                                                                                                                                                                                                                           0.0s
 => => resolve docker.io/library/node:10.15.3-alpine@sha256:aa28f3b6b4087b3f289bebaca8d3fb82b93137ae739aa67df3a04892d521958e                                                                                                                                                                                                                                           0.0s
 => [internal] load build context                                                                                                                                                                                                                                                                                                                                      1.9s
 => => transferring context: 1.01MB                                                                                                                                                                                                                                                                                                                                    1.8s
 => CACHED [2/6] WORKDIR /app                                                                                                                                                                                                                                                                                                                                          0.0s
 => CACHED [3/6] COPY package*.json ./                                                                                                                                                                                                                                                                                                                                 0.0s
 => CACHED [4/6] RUN apk add --no-cache make gcc g++ python zeromq-dev                                                                                                                                                                                                                                                                                                 0.0s
 => CACHED [5/6] COPY . .                                                                                                                                                                                                                                                                                                                                              0.0s
 => CACHED [6/6] RUN npm install                                                                                                                                                                                                                                                                                                                                       0.0s
 => ERROR exporting to image                                                                                                                                                                                                                                                                                                                                           0.4s
 => => exporting layers                                                                                                                                                                                                                                                                                                                                                0.0s
[...]

Issue #306, opened 2020-06-12
uildx push to self-signed registry fails using docker-container driver


Maddog2050:

tonistiigi:
Atm you would need to use `docker cp` to copy in your tls certs as `--config` to point to them.

PR #787, opened 2021-09-29
Copy CA and user TLS registries certs to the container

crazy-max:
Fixes https://github.com/docker/build-push-action/issues/436
Fixes https://github.com/moby/buildkit/issues/1055
Fixes https://github.com/docker/buildx/issues/567
Fixes https://github.com/docker/buildx/issues/80
Fixes https://github.com/docker/buildx/issues/306
Fixes https://github.com/docker/buildx/issues/455

Follow-up moby/buildkit#2361 and moby/buildkit#2377

Since https://github.com/moby/buildkit/pull/1410 we can provide rootCA and keyPairs for registries but we have to copy those certs in the container after the builder is created.

This PR allows to copy certificates to the container when a `docker-container` builder is created by reading the buildkitd toml configuration. Wonder if this should be opt-in or not for security reasons.

Same logic might be used upstream for `buildctl` https://github.com/moby/buildkit/issues/1050.


driver/docker-container/driver.go
```diff
@@ -196,6 +200,126 @@ func (d *Driver) copyLogs(ctx context.Context, l progress.SubLogger) error {
...
+
+	// Create a temp folder and copy all certs to it with copyFileParents
+	// to keep the exact same folder structure as defined in BuildKit config.
+	// It's required to do that here because we cannot copy a file in a folder
+	// that does not exist in the container with the CopyToContainer API.
+	tmpRegCerts, err := os.MkdirTemp("", "basedir-certs")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpRegCerts)
```
crazy-max, line 299: Is there a better way to do this with Docker API (cc @tonistiigi @thaJeztah)? Would also be better to use a tmpfs for security.

tonistiigi: I think we should try to rewrite the cert paths in config and reencode it so the certs always point to `/etc/buildkit/certs` or something. The only issue would be making sure that if buildkit config gets updated it still works. Maybe parse it to https://pkg.go.dev/github.com/pelletier/go-toml#Tree

crazy-max: > I think we should try to rewrite the cert paths in config and reencode it so the certs always point to `/etc/buildkit/certs` or something.

Yes that could do it. I wonder if an extension of the root ca and key pairs fields in buildkitd config to allow base64 encoded content would not be a better alternative.

driver/docker-container/driver.go
```diff
@@ -205,6 +213,211 @@ func (d *Driver) copyLogs(ctx context.Context, l progress.SubLogger) error {
 	return rc.Close()
 }
 
+// copyToContainer is based on the implementation from docker/cli
+// https://github.com/docker/cli/blob/master/cli/command/container/cp.go
+func (d *Driver) copyToContainer(ctx context.Context, srcPath string, dstPath string) error {
```
tonistiigi, line 218: Could we reuse some of this code in k8s driver? cc @morlay

morlay: it couldn't.

In k8s, we have to read file data and store it in ConfigMap or Secrets, and mount it into running containter.

I prefix add an interface for Driver to add files, like

```
inteface Driver {
   AddFile(mountPath string, data []byte, secure bool)
}
```
Then diffrent driver to do the mounting

tonistiigi: Actually, this part might be fine to be driver-specific but the part that modifies the toml file and copies certs under `.docker` should not be specific to the driver but shared.

morlay: So let driver have ability to add file will be useful.

if the driver not need those files,  it could skip `AddFile`

But driver like `docker-container` and `kubernetes` need, it could be easy to `AddFile` with diffirent ways.

morlay: At least, `copyBuildKitConfToContainer` is needed to `kubernetes` driver too.

crazy-max: > should not be specific to the driver but shared.

Yes, `copyBuildKitConfToContainer` could be an utility func that would be called for any driver that uses `ConfigFile` and returns the handle for `/etc/buildkit` conf. It could then be used with `copyToContainer` for the `container` driver and for `kubernetes` I think a `persistenceVolume` and `persistenceVolumeClaims`.

morlay: @crazy-max 

`kubernetes` will be use the `configMap`/`secret` with file contents, because the container should be scheduled to any cluster node. 

Anyway. If we want to added this feature to support `container` and `kubernetes` both, consider my proposal.
Then I could help upgrade `kubernetes` driver to support this feature too.

tonistiigi: I don't think `AddFile()` would work a iiuc `/etc/buildkit` is already mounted in k8s so you can't just pick a path like you  can in container driver.

So shared parts: copy files under `.docker`. Parse toml and create correct config file. Provide list of config + certs to driver

Separate by driver:
- container: copy to container
- kubernetes: fill in configmap https://github.com/docker/buildx/commit/c76b5eac0398a6a186de09318d055fb495a6c192#diff-ad9f131660b45257e1b16c6d5be47cfc1e4bb74f0322fbc6619afedba317288cR119

tonistiigi: @morlay We need to transform the paths because they are paths in random places of the host system that do not work inside container. I don't understand why you have two volumes in your example when they both modify `/etc/buildkit` . I think we should just have one config mount in `/etc/buildkit` and put everything there.

morlay: > we can't copy a file in a folder that does not exist in the container with the CopyToContainer Docker API.

Yes, but we could write to temp files before CopyToContainer in `AddFile`

If worry about copy cost. we could design the API like https://docs.min.io/docs/golang-client-api-reference.html#PutObject

```go
type PutFileOptions struct {
   Secure bool
}

type Driver interface {
   PutFile(ctx context.Context, containerAbsPath string, r io.Reader, putOpts PutFileOptions) error
}
```

tonistiigi: So configmap can not contain subdirectories?

morlay: @tonistiigi Yes. `/` is not allowed in ConfigMap key. 

For subdirectories, we have to rename the absolute path, and resolve it back with `volumes` and `volumeMounts` like my example did.

This why I prefer have a `PutFile()` interface to make driver could do mounting files,
Then we could just convert and mount files for diffrent drivers in `PutFile()`.

PR #790, opened 2021-10-01
build: add shm-size support

### LINKED ISSUE #418: --shm-size in docker build for Mac is not accepting the values indicated
- [x] I have tried with the latest version of my channel (Stable)
  - [x] I have uploaded Diagnostics
  - Diagnostics ID: 80E045E9-84C3-4BA1-A4EE-2A443EA5AFF8/20201016192548

### Expected behavior
`--shm-size=XXX` should set /dev/shm to the desired value

### Actual behavior
any value for `--shm-size` sets the value to 64M, only

### Information
  - macOS Version: macOS Catalina 10.15.7
  - docker for mac version: 2.4.0.0 (48506) Stable

### Steps to reproduce the behavior
  1. Use the sample Dockerfile, that stores the allocated size inside /dev/shm during build time in a temp file that can be read on the container when running
```
FROM busybox
RUN df -kh /dev/shm > /tmp/output
CMD cat /tmp/output
```
  2. Build the image passing the `--shm-size` tag
```
docker build --shm-size=1g -t test .
[+] Building 1.6s (6/6) FINISHED
=> [internal] load build definition from Dockerfile
=> => transferring dockerfile: 111B
=> [internal] load .dockerignore
=> => transferring context: 2B
=> [internal] load metadata for docker.io/library/busybox:latest
=> CACHED [1/2] FROM docker.io/library/busybox@sha256:a9286defaba7b3a519d585ba0e37d0b2cbee74ebfe590960b0b1d6a5e97d1e1d
=> [2/2] RUN df -kh /dev/shm > /tmp/output
=> exporting to image
=> => exporting layers
=> => writing image sha256:2f450e57281dca99454df4ad8becd0fcb2e1153470b5ed365aa946042d001605
=> => naming to docker.io/library/test
```
  3. Run the generated image, which will print the size of /dev/shm stored in the temporary text file during build
```
docker run test
Filesystem                Size      Used Available Use% Mounted on
shm                      64.0M         0     64.0M   0% /dev/shm
```

Tried with different values for the `--shm-size` parameter, but in all cases only 64M is allocated to /dev/shm.

crazy-max:
Follow-up moby/buildkit#2384
Fixes #418 

vendor: https://github.com/moby/buildkit/compare/ba673bbdab4f...8778943ac3da

```dockerfile
# Dockerfile
FROM busybox
RUN df -kh /dev/shm
```

```shell
$ docker buildx build --no-cache --shm-size 128m .
#6 [2/2] RUN df -kh /dev/shm
#6 0.170 Filesystem                Size      Used Available Use% Mounted on
#6 0.170 shm                     128.0M         0    128.0M   0% /dev/shm
#6 DONE 0.2s
```


tonistiigi:
I'm not sure about adding this to bake atm. I think we need to think through what to do with this and similar fields like ulimit, memory limits etc and if they should be grouped together under one field in bake. Network=host isn't currently in bake either.

tonistiigi:
> I'm not sure about adding this to bake atm. I think we need to think through what to do with this and similar fields like ulimit, memory limits etc and if they should be grouped together under one field in bake. Network=host isn't currently in bake either

Also, in a lot of cases, these fields make more sense in a dockerfile, per command.

crazy-max:
> I'm not sure about adding this to bake atm. I think we need to think through what to do with this and similar fields like ulimit, memory limits etc and if they should be grouped together under one field in bake. Network=host isn't currently in bake either.

Yes we can do that in a follow-up for bake sure. Maybe a `resources` field object in the HCL definition would be ok in this case.

> Also, in a lot of cases, these fields make more sense in a dockerfile, per command.

While working on the gateway impl for shm-size, I had started to add `RUN --shm=size=128m` but it might require some changes related to what you said about resources.

thaJeztah:
> Also, in a lot of cases, these fields make more sense in a dockerfile, per command.

Yes, was slightly wondering this as well. Perhaps we should look when/why it was added in the classic builder, and what the exact use-case was.

> While working on the gateway impl for shm-size, I had started to add RUN --shm=size=128m but it might require some changes related to what you said about resources.

In the end, `shm` is effectively just a `tmpfs`. Wondering if adding a `size` option to `--mount` would be an alternative (`--mount=type=tmpfs.target=/dev/shm,size=xxx`)
I recall there were similar discussions about `docker service` not having a `--shm-size` option.

tonistiigi:
@thaJeztah That's a good point

@crazymax I wonder if we should revert the llb part in https://github.com/moby/buildkit/pull/2384 and instead add `size`. They are different in containerd but makes little sense to actually keep them separate here. In runtime specs it is handled as a regular mount as well.

crazy-max:
> I wonder if we should revert the llb part in [moby/buildkit#2384](https://github.com/moby/buildkit/pull/2384) and instead add `size`.

@tonistiigi Yeah would make sense to change that in llb but keep the frontend attr specific to shm-size for the whole build right?

tonistiigi:
> but keep the frontend attr specific to shm-size for the whole build right?

yes

crazy-max:
@tonistiigi Bake impl has been removed.

thaJeztah:
(relates to https://github.com/moby/moby/issues/40379)

build/build.go
```diff
@@ -556,6 +557,11 @@ func toSolveOpt(ctx context.Context, d driver.Driver, multiDriver bool, opt Opti
 	}
 	so.FrontendAttrs["add-hosts"] = extraHosts
 
+	// setup shm size
+	if opt.ShmSize.Value() > 0 {
+		so.FrontendAttrs["shm-size"] = strconv.FormatInt(opt.ShmSize.Value()/1024, 10)
```
tonistiigi, line 562: I wonder why the buildkit api is KB if the flag is not

crazy-max: Hum yes it should be given in bytes on BuildKit.

tonistiigi (approved):

PR #810, opened 2021-10-20
build: warning msg on deprecated flags

crazy-max:
Relates to https://github.com/moby/moby/issues/40379

Add warning message for deprecated flags via pflag annotations:

```shell
$ docker buildx build --load --squash --cgroup-parent /build .
WARN cgroup-parent is not implemented.
WARN squash flag was never taken out of experimental and is deprecated with BuildKit. You should squash inside build using the multi-stage feature for efficiency.
#1 transferring dockerfile: 2.94kB 0.1s done
#1 DONE 0.1s
```

* logrus formatter has been changed to avoid prefix like `WARN[000]`.
* long description has been added to the root command.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -546,7 +546,7 @@ func toSolveOpt(ctx context.Context, d driver.Driver, multiDriver bool, opt Opti
 		so.FrontendAttrs["force-network-mode"] = opt.NetworkMode
 	case "", "default":
 	default:
-		return nil, nil, errors.Errorf("network mode %q not supported by buildkit", opt.NetworkMode)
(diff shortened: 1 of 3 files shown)


tonistiigi (commented):
Squash/Security-opt/isolation should be deprecated. Replacement for squash is to squash inside build (eg. in Dockerfile with additional stage). For Security-opt, `RUN --security=insecure` should be used. For `--net=custom` alternative is to set it on `buildx create`. I think this one could remain error.

For cgroup-parent I'm not sure. Maybe let's just implement it although it is a bad design. (edit: maybe implement and leave it as hidden?)

From quick overview, the rest can remain hidden as they were. No point to show these messages for flags that don't change build outcome and when we are not sure if they will fit in design or not.

For formatting, no capitalized messages. The logrus default seems fine to me.

PR #813, opened 2021-10-26
disable logrus timestamp logging

crazy-max:
`WARN[0000] ...` > `WARN ...`


crazy-max:
#924

PR #824, opened 2021-11-04
store snapshot of config files on create

tonistiigi:
Files can be reused when container needs to be booted again.

The second commit enables config files for k8s. @morlay PTAL. If something is wrong we can remove the second commit or you can push a fix to this PR.

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -225,7 +225,7 @@ func driversForNodeGroup(ctx context.Context, dockerCli command.Cli, ng *store.N
 					}
 				}
 
-				d, err := driver.GetDriver(ctx, "buildx_buildkit_"+n.Name, f, dockerapi, dockerCli.ConfigFile(), kcc, n.Flags, n.ConfigFile, n.DriverOpts, n.Platforms, contextPathHash)
+				d, err := driver.GetDriver(ctx, "buildx_buildkit_"+n.Name, f, dockerapi, dockerCli.ConfigFile(), kcc, n.Flags, n.Files, n.DriverOpts, n.Platforms, contextPathHash)
 				if err != nil {
 					di.Err = err
 					return nil
@@ -360,7 +360,7 @@ func getDefaultDrivers(ctx context.Context, dockerCli command.Cli, defaultOnly b
 		}
 	}
 
-	d, err := driver.GetDriver(ctx, "buildx_buildkit_default", nil, dockerCli.Client(), dockerCli.ConfigFile(), nil, nil, "", nil, nil, contextPathHash)
+	d, err := driver.GetDriver(ctx, "buildx_buildkit_default", nil, dockerCli.Client(), dockerCli.ConfigFile(), nil, nil, nil, nil, nil, contextPathHash)
 	if err != nil {
 		return nil, err
 	}
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -7,6 +7,8 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"path"
+	"path/filepath"
 	"time"
 
 	"github.com/docker/buildx/driver"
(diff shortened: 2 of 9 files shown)


morlay:
@tonistiigi I prefer use single one ConfigMap

if we could multi ConfigMap.
When next deployment with less configurations will not delete some ConfigMap
https://github.com/docker/buildx/blob/master/driver/kubernetes/driver.go#L168-L182

Example
```
# first release
/etc/buildkit
/etc/buildkit/certs   

# second release
# the dir list is from new config content, once certs not includes, /etc/buildkit/certs will not includes.
/etc/buildkit
# ConfigMap for /etc/buildkit/certs will be remain in k8s
```

**If still keep multiple ConfigMaps, need to add label in ConfigMaps, and List by label before delete**


For single one ConfigMap. could be:

```golang
configFiles := []struct {
	name    string
	absPath string
}{}

d.Spec.Template.Spec.Containers[0].VolumeMounts = make([]corev1.VolumeMount, len(configFiles))
d.Spec.Template.Spec.Volumes = make([]corev1.Volume, len(configFiles))

for i, cf := range configFiles {
	d.Spec.Template.Spec.Containers[0].VolumeMounts[i] = corev1.VolumeMount{
		Name:      "config-" + cf.name,
		MountPath: cf.name,
		SubPath:   filepath.Base(cf.absPath),
	}

	d.Spec.Template.Spec.Volumes[i] = corev1.Volume{
		Name: "config-" + cf.name,
		VolumeSource: corev1.VolumeSource{
			ConfigMap: &corev1.ConfigMapVolumeSource{
				LocalObjectReference: corev1.LocalObjectReference{
					Name: "config", // from same ConfigMap
				},
				Items: []corev1.KeyToPath{
					{
						Key: cf.name,
						Path: filepath.Base(cf.absPath),
					},
				},
			},
		}}
}
```

tonistiigi:
I'd like to keep the file structure the same as container driver(+rest in future) and reuse the same code.

I didn't quite get in what case the second release with config and no certs can happen. Also not sure how they remain mounted if they are removed from container spec. On deletion cleanup by id should still work. I added incrementing numbers to names.

tonistiigi:
If you think deletion by a label is safer feel free to add commit for that.

morlay:
@tonistiigi 
Sure. This PR could be ok to merge.
I may add the deletion by a label later today if I have time.

crazy-max:
@tonistiigi Could the spec be a map where we also define the base dir. So instead of:

```json
{
  "Nodes": [
    {
      "Files": {
        "buildkitd.toml": "...",
        "certs/docker.io/cert.pem": "...",
        "certs/docker.io/key.pem": "...",
        "certs/docker.io/myca.pem": "...",
        "certs/ghcr.io/ghcr.pem": "...",
        "certs/ghcr.io/ghcr_cert.pem": "...",
        "certs/ghcr.io/ghcr_key.pem": "..."
      }
    }
  ],
}
```

we have:

```json
{
  "Nodes": [
    {
      "FS": [
        {
          "BaseDir": "/etc/buildkitd",
          "Files": {
            "buildkitd.toml": "...",
            "certs/docker.io/cert.pem": "...",
            "certs/docker.io/key.pem": "...",
            "certs/docker.io/myca.pem": "..."
          }
        }
      ]
    }
  ]
}
```

or

```json
{
  "Nodes": [
    {
      "Files": [
        {
          "BaseDir": "/etc/buildkitd",
          "File": "buildkitd.toml",
          "Content": "..."
        },
        {
          "BaseDir": "/etc/buildkitd",
          "File": "certs/docker.io/cert.pem",
          "Content": "..."
        },
        {
          "BaseDir": "/etc/buildkitd",
          "File": "certs/docker.io/key.pem",
          "Content": "..."
        },
        {
          "BaseDir": "/etc/buildkitd",
          "File": "certs/docker.io/myca.pem",
          "Content": "..."
        },
        {
          "BaseDir": "/etc/buildkitd",
          "File": "certs/ghcr.io/ghcr.pem",
          "Content": "..."
        },
        {
          "BaseDir": "/etc/buildkitd",
          "File": "certs/ghcr.io/ghcr_cert.pem",
          "Content": "..."
        },
        {
          "BaseDir": "/etc/buildkitd",
          "File": "certs/ghcr.io/ghcr_key.pem",
          "Content": "..."
        },
      ]
    }
  ],
}
```

to be able to handle other use cases in the future outside buildkitd config?

tonistiigi:
@crazy-max It could be though that we want to change the etc dir in the future. Eg. Rootless could be one case

crazy-max (approved):

PR #826, opened 2021-11-04
build: fix setting iidfile with multi-node push

tonistiigi:
I can't see any reason for this early return. It breaks using iidfile together with push on multi-node case.


crazy-max (approved):
LGTM, should fix https://github.com/docker/build-push-action/issues/386

PR #830, opened 2021-11-08
ls: add format, quiet and no-trunc opts

### LINKED ISSUE #325: Add --format support for command output
It will be good to have `--format` support like [docker does](https://docs.docker.com/config/formatting/) for info commands.
At least for `docker buildx ls`.
To get easy way to check "is desired builder already created", "are there a builder for desired platform", "can node NODE1 build for i386 platform" etc.

### LINKED ISSUE #380: Add a --quiet --no-trunc option to docker buildx ls
Hello, I'm making a python lib that wraps the docker CLI commands (similar to docker-py, but I use the cli binary, so it's possible to use buildx for example, unlike docker-py). 

When writing the lib, I noticed that it was harder to work with docker buildx than with "docker container" or "docker volume", or "docker network". This is because the output is much different. Especially, what's interesting to me is the "--quiet" option which is available in "docker container ls" or "docker image ls". Without it, I have to do some complicated parsing and it's quite a pain.

As an example:

```
$ docker image ls --quiet --no-trunc
sha256:884ad2927daef99a5df55f19a1879a56b7c0de57962198918b14fd59e0f5501f
sha256:9999ebee41a37c1b721a55c4194e478985f3ec71345aa1a4d176b2a2caff652a
sha256:a29000ba32ad94cc954a598800cc19382abf5bd2e4eb9efb452ac25800aa85ea
sha256:e857357de1d3f939650d88affb23b0a7dede15a9655ff28b76b68d23906b1dc0
sha256:7682a46a212a122c13610420de3f101121f7ec697b349a8356894e4461bf50bd
sha256:16470a424dfdef1ae9031d2fe597641600d723f1d6e66b5956fab02c0377c44b
sha256:7e7796b33525a31280bb107287e131784dfbcdccbad5f233e0c86b9d7075f0d3
sha256:66f5db10c13450e1125b95ad25fb5e59f15a8c499cf14b3dc1b8c6997090a1e4
sha256:8ac321c02ca4295664a66d031728257409619d6f15a90edccf3e4243f1961ddd
sha256:be039c2175a7241a8eb4d70ac2bc5f174c7c3bb2851bc38f5eaa184d525a2db8
sha256:0f19d1eb075a91a9797694640824d5012c5fb24cc2cf479153bc3d2861758a33
sha256:5d6bfa2e7743dae8f01c43c70c734fa7e910d6ec69fddb01625fdcaf6fa0235
```

```
$ docker network ls --quiet --no-trunc
dcc3a6eb9f6f1cbca25c7c11b986ba837dc3438facf8dd60632e64fff6e673f5
ffb3f184cd0a2077f75a507320a9613eec135dc6cb234340ea924d576215e96e
1d41d84573897ba7df0dc8fd48bc142d7673362bd8bee870e7bf4f0989b6883d
e22930c1044733530f41ad70df4731b68f196d398064d0a93f7d41e89b2c747e
```

```
$ docker buildx ls
NAME/NODE                DRIVER/ENDPOINT             STATUS   PLATFORMS
blissful_swartz          docker-container
  blissful_swartz0       unix:///var/run/docker.sock inactive
charming_chebyshev       docker-container
  charming_chebyshev0    unix:///var/run/docker.sock inactive
charming_visvesvaraya    docker-container
  charming_visvesvaraya0 unix:///var/run/docker.sock inactive
vigorous_bell            docker-container
  vigorous_bell0         unix:///var/run/docker.sock inactive
default *                docker
  default                default                     running  linux/amd64, linux/386
```

Which is much harder to parse for a third party lib.

### LINKED ISSUE #764: docker buildx ls / docker builder ls output for failures could use some TLC
For testing-purposes I had various "broken" contexts on my machine. As a result, trying to run `docker buildx ls` (or `docker builder ls`) produces a very "messy" / "confusing" output. While this may not be a standard scenario, I think we should try to make this more elegant.

```bash
docker buildx ls
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseRead: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseRead: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseRead: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed
NAME/NODE         DRIVER/ENDPOINT             STATUS   PLATFORMS
serene_mcnulty    docker-container
  serene_mcnulty0 unix:///var/run/docker.sock inactive
broken-ssh        docker-container
  broken-ssh      broken-ssh                  Get "http://docker.example.com/v1.24/containers/buildx_buildkit_broken-ssh/json": context deadline exceeded
desktop-linux     docker
  desktop-linux   desktop-linux               running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
dind              docker-container
  dind            dind                        Cannot connect to the Docker daemon at tcp://localhost:2375. Is the docker daemon running?
hackdoor1         docker-container
  hackdoor1       hackdoor1                   error during connect: Get "http://docker.example.com/v1.24/containers/buildx_buildkit_hackdoor1/json": command [ssh -- old-prod.corp1.citigroup.com docker system dial-stdio] has exited with exit status 255, please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=ssh: Could not resolve hostname old-prod.corp1.citigroup.com: nodename nor servname provided, or not known

master             docker-container
  master           master           Cannot connect to the Docker daemon at tcp://localhost:2375. Is the docker daemon running?
minecraft-test     docker
  minecraft-test   minecraft-test   running linux/amd64, linux/386
my-context         docker-container
  my-context       my-context       error during connect: Get "http://myserver:2376/v1.24/containers/buildx_buildkit_my-context/json": dial tcp: lookup myserver on 8.8.8.8:53: no such host
my-remote-engine   docker
  my-remote-engine my-remote-engine running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
swarm-test-01      docker
  swarm-test-01    swarm-test-01    running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
default *          docker
  default          default          running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```

There also looks to be a really long timeout in place (this may be in the `docker` cli, not sure):

```bash
time docker buildx ls
...
________________________________________________________
Executed in   21.12 secs   fish           external
   usr time  438.13 millis  144.00 micros  437.99 millis
   sys time  526.49 millis  936.00 micros  525.55 millis
```

Wondering if we would be able to print information "when it arrives". Currently, no output is generated until the command times out. Perhaps we can either;

- print the builder once we're able to connect (adding lines to the output when they arrive)
- print all builders, but show "pending" (e.g.) and refresh the lines once we were able to connect.

So something like:

```
NAME/NODE         DRIVER/ENDPOINT             STATUS   PLATFORMS
serene_mcnulty    docker-container
  serene_mcnulty0 unix:///var/run/docker.sock inactive
broken-ssh        docker-container
  broken-ssh      broken-ssh                  pending
desktop-linux     docker
  desktop-linux   desktop-linux               running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
dind              docker-container
  dind            dind                        pending
```

...

```
NAME/NODE         DRIVER/ENDPOINT             STATUS   PLATFORMS
serene_mcnulty    docker-container
  serene_mcnulty0 unix:///var/run/docker.sock inactive
broken-ssh        docker-container
  broken-ssh      broken-ssh                  unknown
desktop-linux     docker
  desktop-linux   desktop-linux               running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
dind              docker-container
  dind            dind                        unknown


Failed to get status for 'broken-ssh': Get "http://docker.example.com/v1.24/containers/buildx_buildkit_broken-ssh/json": context deadline exceeded
Failed to get status for 'dind': Cannot connect to the Docker daemon at tcp://localhost:2375. Is the docker daemon running?
```

crazy-max:
Fixes #325
Fixes #380
Fixes #764

Adds `filter`, `format` and `no-trunc` options for `ls` command:

```
$ docker buildx ls
NAME/NODE             DRIVER/ENDPOINT                                             STATUS     BUILDKIT   PLATFORMS
builder               docker-container
 \_ builder0           \_ unix:///var/run/docker.sock                             running    7615120    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le,…
 \_ mac-mini-m1        \_ tcp://mac-mini-m1.home.foo.com:2376                     stopped               linux/arm64*
builder2*             docker-container
 \_ builder20          \_ unix:///var/run/docker.sock                             running    7615120    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le,…
remote-builder        remote
 \_ remote-builder0    \_ docker-container://buildx_buildkit_dk-remote-builder0   inactive
 \_ aws_graviton2      \_ tcp://10.0.0.1:1234                                     running               darwin/arm64*, linux/arm64*, linux/arm/v5*, linux/arm/v6*, linux/arm/v7*, windows/arm64*
 \_ linuxone_s390x     \_ tcp://10.0.0.2:1234                                     running               linux/s390x*
default               docker
 \_ default            \_ default                                                 running    20.10.21   linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386,…
desktop-linux         docker
 \_ desktop-linux      \_ desktop-linux                                           error
desktop-windows       docker
 \_ desktop-windows    \_ desktop-windows                                         error
docker1903            docker
 \_ docker1903         \_ docker1903                                              running    19.03.15   linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386,…
docker2010            docker
 \_ docker2010         \_ docker2010                                              running    20.10.21   linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386,…

Failed to get status for desktop-linux (desktop-linux): protocol not available
Failed to get status for desktop-windows (desktop-windows): protocol not available
```

```
$ docker buildx ls --no-trunc
NAME/NODE             DRIVER/ENDPOINT                                             STATUS     BUILDKIT   PLATFORMS
builder               docker-container
 \_ builder0           \_ unix:///var/run/docker.sock                             running    7615120    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
 \_ mac-mini-m1        \_ tcp://mac-mini-m1.home.foo.com:2376                     stopped               linux/arm64*
builder2*             docker-container
 \_ builder20          \_ unix:///var/run/docker.sock                             running    7615120    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
remote-builder        remote
 \_ remote-builder0    \_ docker-container://buildx_buildkit_dk-remote-builder0   inactive
 \_ aws_graviton2      \_ tcp://10.0.0.1:1234                                     running               darwin/arm64*, linux/arm64*, linux/arm/v5*, linux/arm/v6*, linux/arm/v7*, windows/arm64*
 \_ linuxone_s390x     \_ tcp://10.0.0.2:1234                                     running               linux/s390x*
default               docker
 \_ default            \_ default                                                 running    20.10.21   linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
desktop-linux         docker
 \_ desktop-linux      \_ desktop-linux                                           error
desktop-windows       docker
 \_ desktop-windows    \_ desktop-windows                                         error
docker1903            docker
 \_ docker1903         \_ docker1903                                              running    19.03.15   linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
docker2010            docker
 \_ docker2010         \_ docker2010                                              running    20.10.21   linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6

Failed to get status for desktop-linux (desktop-linux): protocol not available
Failed to get status for desktop-windows (desktop-windows): protocol not available
```

```
$ docker buildx ls --quiet
builder
builder2
remote-builder
default
desktop-linux
desktop-windows
docker1903
docker2010
```

```
$ docker buildx ls --format json
{"Name":"builder","Driver":"docker-container","Dynamic":false,"Nodes":[{"Name":"builder0","Endpoint":"unix:///var/run/docker.sock","Platforms":["linux/amd64","linux/amd64/v2","linux/amd64/v3","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/mips64le","linux/mips64","linux/arm/v7","linux/arm/v6"],"Flags":["--debug","--allow-insecure-entitlement","security.insecure","--allow-insecure-entitlement","network.host"],"DriverOpts":{"env.BUILDKIT_STEP_LOG_MAX_SIZE":"10485760","env.BUILDKIT_STEP_LOG_MAX_SPEED":"10485760","env.JAEGER_TRACE":"localhost:6831","image":"moby/buildkit:master","network":"host"},"Status":"running","Version":"7615120"},{"Name":"mac-mini-m1","Endpoint":"tcp://mac-mini-m1.home.foo.com:2376","DriverOpts":{"image":"moby/buildkit:master"},"Status":"stopped"}]}
{"Name":"builder2","Driver":"docker-container","Dynamic":false,"Nodes":[{"Name":"builder20","Endpoint":"unix:///var/run/docker.sock","Platforms":["linux/amd64","linux/amd64/v2","linux/amd64/v3","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/mips64le","linux/mips64","linux/arm/v7","linux/arm/v6"],"Flags":["--debug","--allow-insecure-entitlement","security.insecure","--allow-insecure-entitlement","network.host"],"DriverOpts":{"env.BUILDKIT_STEP_LOG_MAX_SIZE":"10485760","env.BUILDKIT_STEP_LOG_MAX_SPEED":"10485760","env.JAEGER_TRACE":"localhost:6831","image":"moby/buildkit:master","network":"host"},"Status":"running","Version":"7615120"}]}
{"Name":"remote-builder","Driver":"remote","Dynamic":false,"Nodes":[{"Name":"remote-builder0","Endpoint":"docker-container://buildx_buildkit_dk-remote-builder0","Status":"inactive"},{"Name":"aws_graviton2","Endpoint":"tcp://10.0.0.2:1234","Platforms":["linux/arm64","linux/arm/v7","linux/arm/v6"],"DriverOpts":{"cacert":"/home/crazy/.certs/aws_graviton2/ca.pem","cert":"/home/crazy/.certs/aws_graviton2/cert.pem","key":"/home/crazy/.certs/aws_graviton2/key.pem"},"Status":"running"},{"Name":"linuxone_s390x","Endpoint":"tcp://10.0.0.1:1234","Platforms":["linux/s390x"],"DriverOpts":{"cacert":"/home/crazy/.certs/linuxone_s390x/ca.pem","cert":"/home/crazy/.certs/linuxone_s390x/cert.pem","key":"/home/crazy/.certs/linuxone_s390x/key.pem"},"Status":"running"}]}
{"Name":"default","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"default","Endpoint":"default","Platforms":["linux/amd64","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/arm/v7","linux/arm/v6"],"Status":"running","Version":"20.10.21"}]}
{"Name":"desktop-linux","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"desktop-linux","Endpoint":"desktop-linux","Status":"error","Err":"protocol not available"}]}
{"Name":"desktop-windows","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"desktop-windows","Endpoint":"desktop-windows","Status":"error","Err":"protocol not available"}]}
{"Name":"docker1903","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"docker1903","Endpoint":"docker1903","Platforms":["linux/amd64","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/arm/v7","linux/arm/v6"],"Status":"running","Version":"19.03.15"}]}
{"Name":"docker2010","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"docker2010","Endpoint":"docker2010","Platforms":["linux/amd64","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/arm/v7","linux/arm/v6"],"Status":"running","Version":"20.10.21"}]}
```

```
$ docker buildx ls --format raw
name: builder
driver: docker-container
nodes:

  name: builder0
  endpoint: unix:///var/run/docker.sock
  status: running
  buildkit: v0.10.6
  platforms: linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6

  name: mac-mini-m1
  endpoint: tcp://mac-mini-m1.home.foo.com:2376
  status: running
  buildkit: v0.10.6
  platforms: linux/arm64*, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6

name: builder2*
driver: docker-container
nodes:

  name: builder20
  endpoint: unix:///var/run/docker.sock
  status: running
  buildkit: v0.10.6
  platforms: linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6

name: remote-builder
driver: remote
nodes:

  name: remote-builder0
  endpoint: docker-container://buildx_buildkit_dk-remote-builder0
  status: inactive

  name: aws_graviton2
  endpoint: tcp://10.0.0.1:1234
  status: running
  platforms: darwin/arm64*, linux/arm64*, linux/arm/v5*, linux/arm/v6*, linux/arm/v7*, windows/arm64*

  name: linuxone_s390x
  endpoint: tcp://10.0.0.2:1234
  status: running
  platforms: linux/s390x*

name: default
driver: docker
nodes:

  name: default
  endpoint: default
  status: running
  buildkit: 20.10.21
  platforms: linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6

name: desktop-linux
error: protocol not available

name: desktop-windows
error: protocol not available

name: docker1903
driver: docker
nodes:

  name: docker1903
  endpoint: docker1903
  status: running
  buildkit: 19.03.15
  platforms: linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6

name: docker2010
driver: docker
nodes:

  name: docker2010
  endpoint: docker2010
  status: running
  buildkit: 20.10.21
  platforms: linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```


bake/remote.go
```diff
@@ -6,8 +6,8 @@ import (
 	"context"
 	"strings"
 
-	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/driver"
+	"github.com/docker/buildx/driver/driverutil"
```
tonistiigi, line 10: The types in this pkg should just go to `driver` pkg. I think the functions fit the `storeutil` pkg. Can be in a separate file there.

driver/driverutil/driverutil.go
```diff
@@ -0,0 +1,217 @@
...
+
+	if err := eg.Wait(); err != nil {
+		return nil, err
+	}
+
+	return dis, nil
+}
+
+// FilterAvailable returns only available drivers
+func FilterAvailable(drivers []Info) ([]Info, error) {
```
tonistiigi, line 202: This one can go to `driver` pkg with the types.

commands/bake.go
```diff
@@ -90,7 +92,7 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions) (err error
 		}
 	}()
 
-	dis, err := getInstanceOrDefault(ctx, dockerCli, in.builder, contextPathHash)
+	dis, err := driverutil.GetByNameOrDefault(ctx, dockerCli, in.builder, contextPathHash)
```
tonistiigi, line 95: for at least some of them I don't see why they have moved and are public. They are still called from only one package.

commands/create.go
```diff
@@ -135,15 +138,14 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
 		}
 	} else {
 		if len(args) > 0 {
-			ep, err = validateEndpoint(dockerCli, args[0])
+			ep, err = dockerutil.ValidateEndpoint(dockerCli, args[0])
```
tonistiigi, line 141: let's not make unnecessary new public api. This function is called in exactly one place. If you want to move some function definition to separate files then that is fine(but should probably be in a separate refactoring PR).

util/dockerutil/docker.go
```diff
@@ -0,0 +1,87 @@
...
+
+	h, err := opts.ParseHost(true, ep)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to parse endpoint %s", ep)
+	}
+	return h, nil
+}
+
+// ClientForEndpoint returns a docker client for an endpoint
+func ClientForEndpoint(dockerCli command.Cli, name string) (client.APIClient, error) {
```
tonistiigi, line 48: I think this fits fine in storeutil as well(if there is any reason for them to be public at all). The imports are same.

util/formatter/formatter.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// Format keys used to specify certain kinds of output formats
+const (
+	FormatTable = "table"
+	FormatJSON  = "json"
+)
+
+// Print prints formatted output in different formats
+func Print(o interface{}, format string, w io.Writer, t func(t *tablewriter.Table)) error {
```
tonistiigi, line 24: This does not look consistent with docker cli. There is no `--format json` but `{{json .}}`. And we don't import this "tablewriter" package.

commands/ls.go
```diff
@@ -3,142 +3,150 @@ package commands
...
-	if err != nil {
+	timeoutCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
+	defer cancel()
+
+	progressMode := progress.PrinterModeAuto
+	if in.format != formatter.FormatTable {
+		progressMode = progress.PrinterModeQuiet
+	}
+	pw := progress.NewPrinter(context.Background(), os.Stderr, progressMode)
+	if err = bc.LoadBuildersData(timeoutCtx, builders, pw); err != nil {
```
tonistiigi, line 96: I don't like this progress on `ls`. It is confusing because the user does not want to load or build anything. Yes, in some cases it may take time to list builders but that is the same with any other command.

If we want to improve this then a simple line like "Checking builders online status...` after a couple of seconds timeout would do. Maybe cancelable with `^D` (shows status as "timeout"). But it has nothing to do with filters.

commands/ls.go
```diff
@@ -3,142 +3,150 @@ package commands
...
-			if err != "" {
-				fmt.Fprintf(w, "  %s\t%s\t%s\n", n.Name, n.Endpoint, err)
+		}
+	}); err != nil {
+		return err
+	}
+
+	if hasError && in.format == formatter.FormatTable {
+		_, _ = fmt.Fprintf(dockerCli.Out(), "\n")
+		for _, b := range fbuilders {
```
tonistiigi, line 129: These should be consistently formatted.

```
2 error while checking builder status:
<builder/node>: error message
<builder/node>: error message
```

Again, would prefer to not squash all these different things together.

thaJeztah:
I'll try to have a closer look, but some quick thoughts

> For the table format (default), errors are now displayed below the table to avoid a messy output as suggested by @thaJeztah in #764.

Hm.. yes, so I suggested that indeed. I think my initial thought there was that the problem in the current output was that the error message took the place of the `STATUS` column, so the `STATUS` columnn either had a "known" status, or contained an error message; the `WARN[0020] commandConn.CloseWrite: commandconn: failed to wait: signal: killed` lines definitely felt out of place though.

Looking at your example output, I'm now wondering if (instead) it would make sense to add an `ERROR` column; this would still allop the error to be grouped with the builder, but the `STATUS` column to show more useful information.

Similar to `docker stack ps`;

```bash
docker service create --name foobar nginx:alpine
docker service ps foobar

ID             NAME       IMAGE          NODE             DESIRED STATE   CURRENT STATE            ERROR     PORTS
a9b0sgzs04ej   foobar.1   nginx:alpine   docker-desktop   Running         Running 11 seconds ago

docker service create --name failing --detach busybox sh -c 'sleep 2; exit 1'
docker service ps failing
ID             NAME            IMAGE            NODE             DESIRED STATE   CURRENT STATE          ERROR                       PORTS
jskpciju2hpw   failing.1       busybox:latest   docker-desktop   Ready           Ready 1 second ago
mfl1c0dr88wf    \_ failing.1   busybox:latest   docker-desktop   Shutdown        Failed 1 second ago    "task: non-zero exit (1)"
wmg9tn7w6btk    \_ failing.1   busybox:latest   docker-desktop   Shutdown        Failed 9 seconds ago   "task: non-zero exit (1)"
```

I _think_ `docker stack ps` by default truncates output (to keep a shorter line length), so to view the full error message, the `--no-trunc` option may be needed in some cases.

> A progress output has also been added but we could hide it if you prefer.

Yes, I'm a bit on the fence on that one. That said, "it's tricky" `docker service ps` is of course handled server-side, which allows it to keep state, and present the current results without delay, whereas `docker buildx ls` will have to connect to all builders every time it's called.

Something "in between" would be to have a way to stay "attached" (like `docker stats`); print all builders, and refresh their status once we get those results, but that may feel a bit foreign as well (compared to other `ps` / `ls` commands).

Overall, I'm wondering if we could find some middle-ground where `docker buildx ls` / `docker builder ls` would provide a quick overview of the builders that are defined (perhaps skipping actually checking status), as currently a single offline builder blocks / slows down the output, which makes it quite cumbersome to use.

I'm also looking if we can format the output more so that the "essential" information can be presented as "one row per builder", instead of the staggered (multiple lines per builder) output, which makes it hard to see "at a glance" what nodes I have, and to find the information.

Currently we have;

- `Name`
- `Node`
- `Driver`
- `Endpoint`
- `Status`
- `Platforms`
- (`Error`)

Of those;

- `Name`, `Driver`, and (possibly) `Platforms` are useful information (although `Platforms` can get wieldy easily).
- `Status` will be useful, but is the tricky one (see discussion above)
- I'm a bit on the fence on `Node` and `Endpoint`; I guess the information is specifically useful if a builder uses multiple nodes, but may not always be needed to get a quick overview? Depending on how builders are used, giving a sensible name to your nodes would help. Perhaps we need a `--verbose` option to expand builders to include the nodes that are part of it.

crazy-max:
@thaJeztah Switched to cli formatter. See updated PR description.

builder/node.go
```diff
@@ -163,6 +166,45 @@ func (b *Builder) LoadNodes(ctx context.Context, withData bool) (_ []Node, err e
...
+		err = strings.TrimSpace(n.Err.Error())
+	}
+	var pp []string
+	for _, p := range n.Platforms {
+		pp = append(pp, platforms.Format(p))
+	}
+	return json.Marshal(struct {
+		Name        string
+		Endpoint    string
+		Platforms   []string          `json:",omitempty"`
```
thaJeztah, line 186: I know we haven't been very consistent in these, but wondering if we should include an explicit name to use in the json output; wdyt? i.e., if we explicitly want the output to be starting with an uppercase;

```suggestion
		Platforms   []string          `json:"Platforms,omitempty"`
```

builder/node.go
```diff
@@ -163,6 +166,45 @@ func (b *Builder) LoadNodes(ctx context.Context, withData bool) (_ []Node, err e
...
+	var err string
+	if n.Err != nil {
+		status = "error"
+		err = strings.TrimSpace(n.Err.Error())
+	}
+	var pp []string
+	for _, p := range n.Platforms {
+		pp = append(pp, platforms.Format(p))
+	}
+	return json.Marshal(struct {
```
thaJeztah, line 183: I was initially wondering if we should have an actual type for this; possibly if we need specific formatting functions for printing, we could define those on the type.

Then I wondered what we needed the ad-hoc type for (was it because of recursion happening during MarshalJSON?); but then noticed the platforms handling above.

Do you know what specifically was needed to use the `platforms.Format()`? Out of curiousity, I tried what happened if we use `[]ocispecs.Platform` as-is, and (at a glance :warning:), the output seems to look fine.

This is still using the ad-hoc type, but updating it to use OCI platform;

```patch
diff --git a/builder/node.go b/builder/node.go
index 665413f7..2954b153 100644
--- a/builder/node.go
+++ b/builder/node.go
@@ -5,7 +5,6 @@ import (
        "encoding/json"
        "strings"

-       "github.com/containerd/containerd/platforms"
        "github.com/docker/buildx/driver"
        ctxkube "github.com/docker/buildx/driver/kubernetes/context"
        "github.com/docker/buildx/store"
@@ -176,25 +175,21 @@ func (n *Node) MarshalJSON() ([]byte, error) {
                status = "error"
                err = strings.TrimSpace(n.Err.Error())
        }
-       var pp []string
-       for _, p := range n.Platforms {
-               pp = append(pp, platforms.Format(p))
-       }
        return json.Marshal(struct {
                Name        string
                Endpoint    string
-               Platforms   []string          `json:",omitempty"`
-               Flags       []string          `json:",omitempty"`
-               DriverOpts  map[string]string `json:",omitempty"`
-               Files       map[string][]byte `json:",omitempty"`
-               Status      string            `json:",omitempty"`
-               ProxyConfig map[string]string `json:",omitempty"`
-               Version     string            `json:",omitempty"`
-               Err         string            `json:",omitempty"`
+               Platforms   []ocispecs.Platform `json:",omitempty"`
+               Flags       []string            `json:",omitempty"`
+               DriverOpts  map[string]string   `json:",omitempty"`
+               Files       map[string][]byte   `json:",omitempty"`
+               Status      string              `json:",omitempty"`
+               ProxyConfig map[string]string   `json:",omitempty"`
+               Version     string              `json:",omitempty"`
+               Err         string              `json:",omitempty"`
        }{
                Name:        n.Name,
                Endpoint:    n.Endpoint,
-               Platforms:   pp,
+               Platforms:   n.Platforms,
                Flags:       n.Flags,
                DriverOpts:  n.DriverOpts,
                Files:       n.Files,
```

thaJeztah: With the above, the output _seems_ to be fine (but perhaps there's specific corner-cases we're not accounting for?)

```bash
./bin/build/docker-buildx ls
NAME/NODE           DRIVER/ENDPOINT     STATUS    BUILDKIT                           PLATFORMS
default *           docker
 \_ default          \_ default         running   22.06.0-beta.0-902-g2708be0db4.m   linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x,…
desktop-linux       docker
 \_ desktop-linux    \_ desktop-linux   running   22.06.0-beta.0-902-g2708be0db4.m   linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x,…
foo                 docker
 \_ foo              \_ foo             error

Failed to get status for foo (foo): Unix socket path "/var/folders/6f/tz5jf4nn1_n5jb0ctrrw5p2w0000gn/T/TestBuildWithBuildercustom_context67217891/001/docker.sock" is too long
```

👆 😂 Ah, fun; looks like I ran some integration test on my machine, and macOS "temp" directories are symlinks, and the _actual_ location looks to be too long.


I also tried outputing only the `.Platforms`, which also seems to work, although (not sure where those come from), it looks like there's empty lines in between each row 🤔 

```bash
 ./bin/build/docker-buildx ls --format '{{.Platforms}}'

linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/mips64le, linux/mips64

linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/mips64le, linux/mips64


```

thaJeztah: > but perhaps there's specific corner-cases we're not accounting for?

Oh! Perhaps this was for the JSON format? As without it, it would render the platform as a struct? (Wondering if perhaps we should actually keep that 🤔)


```bash
 ./bin/build/docker-buildx ls --format json
{"Name":"default","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"default","Endpoint":"default","Platforms":[{"architecture":"arm64","os":"linux"},{"architecture":"amd64","os":"linux"},{"architecture":"amd64","os":"linux","variant":"v2"},{"architecture":"riscv64","os":"linux"},{"architecture":"ppc64le","os":"linux"},{"architecture":"s390x","os":"linux"},{"architecture":"mips64le","os":"linux"},{"architecture":"mips64","os":"linux"}],"Status":"running","Version":"22.06.0-beta.0-902-g2708be0db4.m"}]}
{"Name":"desktop-linux","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"desktop-linux","Endpoint":"desktop-linux","Platforms":[{"architecture":"arm64","os":"linux"},{"architecture":"amd64","os":"linux"},{"architecture":"amd64","os":"linux","variant":"v2"},{"architecture":"riscv64","os":"linux"},{"architecture":"ppc64le","os":"linux"},{"architecture":"s390x","os":"linux"},{"architecture":"mips64le","os":"linux"},{"architecture":"mips64","os":"linux"}],"Status":"running","Version":"22.06.0-beta.0-902-g2708be0db4.m"}]}
{"Name":"foo","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"foo","Endpoint":"foo","Status":"error","Err":"Unix socket path \"/var/folders/6f/tz5jf4nn1_n5jb0ctrrw5p2w0000gn/T/TestBuildWithBuildercustom_context67217891/001/docker.sock\" is too long"}]}
```

thaJeztah: Oh, and I guess it doesn't show the `default` / `native` platform; (`linux/arm64*` vs `linux/arm64`); 

edit: ⚠️ actually, that's not it (where is that added?)

If we really want a string presentation in all cases, we could have a local type that wraps `oci.Platform` and implements `MarshalText`;

```patch
diff --git a/builder/node.go b/builder/node.go
index 665413f7..5ee00381 100644
--- a/builder/node.go
+++ b/builder/node.go
@@ -166,6 +166,12 @@ func (b *Builder) LoadNodes(ctx context.Context, withData bool) (_ []Node, err e
        return b.nodes, nil
 }

+type Platform ocispecs.Platform
+
+func (p *Platform) MarshalText() ([]byte, error) {
+       return []byte(platforms.Format(ocispecs.Platform(*p))), nil
+}
+
 func (n *Node) MarshalJSON() ([]byte, error) {
        var status string
        if n.DriverInfo != nil {
@@ -176,14 +182,14 @@ func (n *Node) MarshalJSON() ([]byte, error) {
                status = "error"
                err = strings.TrimSpace(n.Err.Error())
        }
-       var pp []string
+       var pp []Platform
        for _, p := range n.Platforms {
-               pp = append(pp, platforms.Format(p))
+               pp = append(pp, Platform(p))
        }
        return json.Marshal(struct {
                Name        string
                Endpoint    string
-               Platforms   []string          `json:",omitempty"`
+               Platforms   []Platform        `json:",omitempty"`
                Flags       []string          `json:",omitempty"`
                DriverOpts  map[string]string `json:",omitempty"`
                Files       map[string][]byte `json:",omitempty"`
```

jedevc:
For `buildx ls`, we should follow similar conventions to `buildx inspect`. i.e. the following should be the same:

```
$ docker buildx inspect desktop-linux
Name:   desktop-linux
Driver: docker

Nodes:
Name:      desktop-linux
Endpoint:  desktop-linux
Status:    running
Buildkit:  20.10.21
Platforms: linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6

$ buildx ls --format raw
...
name: desktop-linux
driver: docker

  name: desktop-linux
  endpoint: desktop-linux
  status: running
  buildkit: 20.10.21
  platforms: linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```

I think we should stick to the title-case naming of inspect, and follow the status-quo there. However, I think maybe we could indent the `Nodes` of inspect, and keep the nodes key as suggested by @thaJeztah [here](https://github.com/docker/buildx/pull/830#issuecomment-1346309541).

`buildx ls` should also take at least one argument (ideally, arbitrary), which should do the job of filtering, similar to how `docker image ls` operates:

```
$ docker image ls registry
REPOSITORY   TAG       IMAGE ID       CREATED       SIZE
registry     2         81c944c2288b   4 weeks ago   24.1MB
```

---

Slightly unrelated, since this has been an issue for a while - how do we determine the size of the table? The produced table is quite large, and now that we're shortening the list of platforms, I was wondering if maybe we could consider setting the default size of the table to the width of the terminal window if that info is accessible? It can be kinda frustrating to view in smaller terminals as is, though it's mostly a separate issue, since it has been like this for a while.

builder/node.go
crazy-max: > Do you know what specifically was needed to use the `platforms.Format()`?

This is for the `--format json` so we are consistent with table / raw result.

> I also tried outputing only the `.Platforms`, which also seems to work, although (not sure where those come from), it looks like there's empty lines in between each row 🤔

Blank lines are the builders which don't carry platforms (only nodes). I'm not sure how we can handle this with custom go template as builders and nodes are really different.

> Oh, and I guess it doesn't show the `default` / `native` platform; (`linux/arm64*` vs `linux/arm64`);

`*` on platform means user enforces this platform when creating the builder or appending a node (`create --platform linux/amd64`) so it will build on this node if it matches the target platform. Atm table, raw format shows supported and preferred platforms but json format doesn't (maybe it should?).

crazy-max:
> For `buildx ls`, we should follow similar conventions to `buildx inspect`. i.e. the following should be the same:

I think it should be the other way around to be consistent with docker/cli like:

```console
$ docker image ls --format raw
repository: buildkit-tests
tag: latest
image_id: 86c59caf767c
created_at: 2022-12-13 07:58:20 +0100 CET
virtual_size: 1.65GB

repository: <none>
tag: <none>
image_id: d57078eda7ec
created_at: 2022-12-13 07:53:45 +0100 CET
virtual_size: 1.65GB

repository: <none>
tag: <none>
image_id: fdc8939962ce
created_at: 2022-12-13 07:47:31 +0100 CET
virtual_size: 1.65GB
```

I want to also add the same flags for the `inspect` command so we can get rid of this: https://github.com/docker/setup-buildx-action/blob/39a1a82492fd1ad19af19d61b5f748e4cb6cd1af/src/buildx.ts#L111-L182 :sweat_smile:

> `buildx ls` should also take at least one argument (ideally, arbitrary), which should do the job of filtering, similar to how `docker image ls` operates:

sgtm for a follow-up as well

> Slightly unrelated, since this has been an issue for a while - how do we determine the size of the table? The produced table is quite large, and now that we're shortening the list of platforms, I was wondering if maybe we could consider setting the default size of the table to the width of the terminal window if that info is accessible? It can be kinda frustrating to view in smaller terminals as is, though it's mostly a separate issue, since it has been like this for a while.

Depends if docker/cli provides this, I'm not sure.

thaJeztah:
> Slightly unrelated, since this has been an issue for a while - how do we determine the size of the table?

I'm still a bit on the fence how relevant showing _all_ platforms is in the default presentation?

- Do we know how it's used?
- Should it default to show the default (native) platform only?
- ^^ and only show the emulated platforms on `inspect` ?

In general, for the "overview" I (personally) feel like it's "just to get a quick list of builders", so that I ca

- "pick" one (`docker builder use` / `docker buildx use`) 
- have a quick overview of builders that may need to be updated (seeing an older version in the ~`VERSION`~ `BUILDKIT` column
- manage them (`docker builder rm` /`docker buildx rm` etc.). Showing further details

Within that context, _perhaps_ "native" platform is relevant (I want to use a native `arm64` builder), but emulated platforms may be less relevant (but we could add a `--filter platform=....` option if I want to make a selection of builders that _can_ support a specific platform).

jsumners:
Is there any alternative to this for getting `buildx ls` results in a parseable format? It looks like this isn't going to available any time soon.

invisiblepancake (approved):

invisiblepancake (approved):

invisiblepancake (approved):

PR #840, opened 2021-11-13
chore: create packages

crazy-max:
Follow-up to our discussion about releasing buildx as a separate package.

I first started to implement this on the [docker-ce-packaging](https://github.com/docker/docker-ce-packaging) repo but the [Go version](https://github.com/docker/docker-ce-packaging/blob/8bbb124d84a94765592fcb4aeef66982daf2d69d/common.mk#L19) is not the one we want for buildx. I think we should first merge docker/cli#3258.

Creating packages here allows for easier maintenance. In a follow-up we could upload pkgs and update repodata on `download.docker.com` (cc @StefanScherer).

Test it:

```console
$ docker buildx bake "https://github.com/crazy-max/buildx.git#deb-pkg" pkg
```

```console
$ dpkg-deb -I bin/docker-buildx_0.7.0~9-g1250d945.m_amd64.deb 
 new Debian package, version 2.0.
 size 12219872 bytes: control archive=419 bytes.
       1 bytes,     0 lines      conffiles
     306 bytes,    10 lines      control
      81 bytes,     1 lines      md5sums
 Package: docker-buildx
 Version: 0:0.7.0~9-g1250d945.m
 Section: default
 Priority: optional
 Architecture: amd64
 Maintainer: Docker <support@docker.com>
 Vendor: Docker
 Installed-Size: 46008
 Homepage: https://github.com/docker/buildx
 Description: Docker Buildx plugin extends build capabilities with BuildKit.
```

```console
$ rpm -qip bin/docker-buildx-0.7.0~9-g1250d945.m.x86_64.rpm 
Name        : docker-buildx
Epoch       : 0
Version     : 0.7.0~9-g1250d945.m
Release     : 1
Architecture: x86_64
Install Date: (not installed)
Group       : Tools/Docker
Size        : 47112192
License     : Apache-2.0
Signature   : (none)
Source RPM  : docker-buildx-0.7.0~9-g1250d945.m-1.src.rpm
Build Date  : Thu Dec 16 11:31:29 2021
Build Host  : buildkitsandbox
Relocations : (not relocatable)
Packager    : Docker <support@docker.com>
Vendor      : Docker
URL         : https://github.com/docker/buildx
Summary     : Docker Buildx plugin extends build capabilities with BuildKit.
Description :
Docker Buildx plugin extends build capabilities with BuildKit.
```

cc @thaJeztah @tonistiigi

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,15 +18,14 @@ env:
   RELEASE_OUT: "./release-out"
 
 jobs:
-  build:
+  test:
     runs-on: ubuntu-latest
     steps:
       -
         name: Checkout
         uses: actions/checkout@v2
-      -
-        name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        with:
+          fetch-depth: 0
       -
         name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
@@ -39,6 +38,22 @@ jobs:
         uses: codecov/codecov-action@v2
         with:
           file: ./coverage/coverage.txt
+
+  build:
+    runs-on: ubuntu-latest
+    needs: test
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
       -
         name: Build binaries
         run: |
@@ -87,6 +102,29 @@ jobs:
         with:
           draft: true
           files: ${{ env.RELEASE_OUT }}/*
+
+  pkg:
+    runs-on: ubuntu-latest
+    needs: test
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          driver-opts: image=${{ env.REPO_SLUG_ORIGIN }}
+      -
+        name: Build packages
+        run: |
+          make pkg-cross
+          tree ./bin
 
   buildkit-edge:
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,8 @@ on:
       - 'v[0-9]*'
 
 env:
-  REPO_SLUG: "docker/buildx-bin"
+  BIN_REPO_SLUG: "docker/buildx-bin"
+  PKG_REPO_SLUG: "docker/buildx-pkg"
   RELEASE_OUT: "./release-out"
 
 jobs:
@@ -71,7 +72,7 @@ jobs:
         uses: docker/metadata-action@v3
         with:
           images: |
-            ${{ env.REPO_SLUG }}
+            ${{ env.BIN_REPO_SLUG }}
           tags: |
             type=ref,event=branch
             type=ref,event=pr
@@ -113,6 +114,18 @@ jobs:
         with:
           fetch-depth: 0
       -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            ${{ env.PKG_REPO_SLUG }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+          bake-target: meta-helper
+      -
         name: Set up QEMU
         uses: docker/setup-qemu-action@v1
       -
@@ -121,10 +134,22 @@ jobs:
         with:
           driver-opts: image=${{ env.REPO_SLUG_ORIGIN }}
       -
-        name: Build packages
-        run: |
-          make pkg-cross
-          tree ./bin
+        name: Login to DockerHub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Build and push image
+        uses: docker/bake-action@v1
+        with:
+          files: |
+            ./docker-bake.hcl
+            ${{ steps.meta.outputs.bake-file }}
+          targets: pkg-cross
+          set: |
+            *.output=type=image,push=${{ github.event_name != 'pull_request' }}
 
   buildkit-edge:
     runs-on: ubuntu-latest
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
-# syntax=docker/dockerfile:1.3
+# syntax=docker/dockerfile:1.3-labs
 
 ARG GO_VERSION=1.17
+ARG NFPM_VERSION=v2.13.0
 ARG DOCKERD_VERSION=20.10.8
 
 FROM docker:$DOCKERD_VERSION AS dockerd-release
 
 # xx is a helper for cross-compilation
-FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.0.0 AS xx
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.1.0 AS xx
 
 FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS golatest
 
@@ -16,24 +17,24 @@ RUN apk add --no-cache file git
 ENV GOFLAGS=-mod=vendor
 WORKDIR /src
 
-FROM gobase AS buildx-version
+FROM gobase AS version
 RUN --mount=target=. \
   PKG=github.com/docker/buildx VERSION=$(git describe --match 'v[0-9]*' --dirty='.m' --always --tags) REVISION=$(git rev-parse HEAD)$(if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi); \
   echo "-X ${PKG}/version.Version=${VERSION} -X ${PKG}/version.Revision=${REVISION} -X ${PKG}/version.Package=${PKG}" | tee /tmp/.ldflags; \
   echo -n "${VERSION}" | tee /tmp/.version;
 
-FROM gobase AS buildx-build
+FROM gobase AS build
 ENV CGO_ENABLED=0
 ARG LDFLAGS="-w -s"
 ARG TARGETPLATFORM
 RUN --mount=type=bind,target=. \
   --mount=type=cache,target=/root/.cache \
   --mount=type=cache,target=/go/pkg/mod \
-  --mount=type=bind,source=/tmp/.ldflags,target=/tmp/.ldflags,from=buildx-version \
+  --mount=type=bind,source=/tmp/.ldflags,target=/tmp/.ldflags,from=version \
   set -x; xx-go build -ldflags "$(cat /tmp/.ldflags) ${LDFLAGS}" -o /usr/bin/buildx ./cmd/buildx && \
   xx-verify --static /usr/bin/buildx
 
-FROM buildx-build AS test
+FROM build AS test
 RUN --mount=type=bind,target=. \
   --mount=type=cache,target=/root/.cache \
   --mount=type=cache,target=/go/pkg/mod \
@@ -44,13 +45,13 @@ FROM scratch AS test-coverage
 COPY --from=test /tmp/coverage.txt /coverage.txt
 
 FROM scratch AS binaries-unix
-COPY --from=buildx-build /usr/bin/buildx /
+COPY --from=build /usr/bin/buildx /
 
 FROM binaries-unix AS binaries-darwin
 FROM binaries-unix AS binaries-linux
 
 FROM scratch AS binaries-windows
-COPY --from=buildx-build /usr/bin/buildx /buildx.exe
+COPY --from=build /usr/bin/buildx /buildx.exe
 
 FROM binaries-$TARGETOS AS binaries
 
@@ -59,11 +60,33 @@ FROM --platform=$BUILDPLATFORM alpine AS releaser
 WORKDIR /work
 ARG TARGETPLATFORM
 RUN --mount=from=binaries \
-  --mount=type=bind,source=/tmp/.version,target=/tmp/.version,from=buildx-version \
+  --mount=type=bind,source=/tmp/.version,target=/tmp/.version,from=version \
   mkdir -p /out && cp buildx* "/out/buildx-$(cat /tmp/.version).$(echo $TARGETPLATFORM | sed 's/\//-/g')$(ls buildx* | sed -e 's/^buildx//')"
 
 FROM scratch AS release
 COPY --from=releaser /out/ /
+
+# Pkg
+FROM --platform=$BUILDPLATFORM goreleaser/nfpm:${NFPM_VERSION} AS nfpm
+FROM --platform=$BUILDPLATFORM alpine AS build-pkg
+RUN apk add --no-cache bash file git zip tar
+WORKDIR /work
+COPY --from=xx / /
+COPY --from=build /usr/bin/buildx /usr/bin/buildx
+ARG TARGETPLATFORM
+ARG PKG_TYPES
+ARG PKG_APK_RELEASES
+ARG PKG_DEB_RELEASES
+ARG PKG_RPM_RELEASES
+ARG PKG_VENDOR
+ARG PKG_PACKAGER
+RUN --mount=source=.,target=/src \
+  --mount=from=nfpm,source=/usr/bin/nfpm,target=/usr/bin/nfpm \
+  --mount=source=/tmp/.version,target=/tmp/.version,from=version \
+  VERSION=$(cat /tmp/.version) NFPM_CONFIG=/src/hack/nfpm.yml OUTPUT=/build /src/hack/create-pkg
+
+FROM scratch AS pkg
+COPY --from=build-pkg /build /
 
 # Shell
 FROM docker:$DOCKERD_VERSION AS dockerd-release
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -180,7 +180,7 @@ variable "PKG_PACKAGER" {
 # PKG_TYPES=deb PKG_DEB_RELEASES=debian11 docker buildx bake pkg
 # docker buildx bake --set *.platform=windows/amd64 --set *.output=./bin pkg
 target "pkg" {
-  inherits = ["binaries"]
+  inherits = ["meta-helper", "binaries"]
   args = {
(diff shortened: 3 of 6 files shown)


Dockerfile
```diff
@@ -59,12 +59,59 @@ FROM --platform=$BUILDPLATFORM alpine AS releaser
...
 FROM scratch AS release
 COPY --from=releaser /out/ /
 
+# Deb
+FROM --platform=$BUILDPLATFORM debian:bullseye-slim AS build-deb
+RUN apt-get update && apt-get install -y debhelper dh-make
+WORKDIR /build
+COPY --from=xx / /
+COPY --from=build /usr/bin/buildx ./docker-buildx-standalone/usr/bin/
+COPY --from=build /usr/bin/buildx ./docker-buildx-plugin/usr/libexec/docker/cli-plugins/docker-buildx
```
tonistiigi, line 74: I don't think we need two packages. We can just put the binary at `/usr/libexec/docker/cli-plugins/docker-buildx` and symlink it to `/usr/bin/buildx`.

Dockerfile
```diff
@@ -59,12 +59,59 @@ FROM --platform=$BUILDPLATFORM alpine AS releaser
...
+ARG TARGETPLATFORM
+RUN --mount=type=bind,source=/tmp/.version,target=/tmp/.version,from=version <<EOT
+#!/usr/bin/env bash
+set -e
+version=$(cat /tmp/.version)
+if [[ $version =~ ^[a-f0-9]{7}$ ]]; then
+  version="v0.0.0+${version}"
+fi
+mkdir -p ./docker-buildx-standalone/DEBIAN
+cat > ./docker-buildx-standalone/DEBIAN/control <<EOL
```
tonistiigi, line 84: That seems a bit too much data even for heredocs. Maybe create a template file under contrib or dist.

crazy-max: Switched to nfpm so we can handle deb and rpm packages with the same config. Maybe apk is not necessary as buildx is already available for alpine: https://pkgs.alpinelinux.org/packages?name=*buildx*&branch=v3.14

docker-bake.hcl
```diff
@@ -122,12 +122,38 @@ target "binaries-cross" {
   ]
 }
 
-target "release" {
+target "pkg" {
+  inherits = ["binaries"]
+  args = {
+    PKG = "apk,deb,rpm"
```
crazy-max, line 128: Maybe apk is not necessary as buildx is already available for alpine: https://pkgs.alpinelinux.org/packages?name=*buildx*&branch=v3.14

Dockerfile
```diff
@@ -59,12 +60,44 @@ FROM --platform=$BUILDPLATFORM alpine AS releaser
...
+set -e
+version=$(cat /tmp/.version)
+if [[ $version =~ ^[a-f0-9]{7}$ ]]; then
+  version="v0.0.0+${version}"
+fi
+for pkg in ${PKG//,/ }
+do
+  arch=$(xx-info march)
+  [[ "$pkg" = "apk" ]] && arch=$(xx-info alpine-arch)
+  [[ "$pkg" = "deb" ]] && arch=$(xx-info debian-arch)
```
crazy-max, line 91: needs https://github.com/tonistiigi/xx/pull/52 to handle rhel archs properly.

docker-bake.hcl
tomalok: fwiw, here's the `APKBUILD` that we use.  we regularly rebuild for new `go` versions, CVEs, etc.

https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/docker-cli-buildx/APKBUILD

crazy-max: @tomalok sgtm thanks for your feedback

docker-bake.hcl
```diff
@@ -122,12 +122,38 @@ target "binaries-cross" {
   ]
 }
 
-target "release" {
+target "pkg" {
+  inherits = ["binaries"]
+  args = {
+    PKG = "deb,rpm"
```
crazy-max, line 128: @tonistiigi As discussed only create deb and rpm pkgs

hack/nfpm.yml
```diff
@@ -0,0 +1,30 @@
+name: docker-buildx
```
crazy-max, line 1: Do we have a convention for package names? I see docker scan is [named `docker-scan-plugin`](https://github.com/docker/docker-ce-packaging/blob/8bbb124d84a94765592fcb4aeef66982daf2d69d/deb/common/control#L105). Maybe we should do the same here with `docker-buildx-plugin`? cc @thaJeztah

tonistiigi: Based on https://github.com/docker/cli/pull/3314#issuecomment-945667368 we are distancing from "plugin" lingo so would rather avoid it. The bundled tools by Docker org that make up "docker" are better referred as components. Compose is compose, not compose-plugin. Same for buildx.

hack/nfpm.yml
```diff
@@ -0,0 +1,30 @@
...
+platform: linux
+
+version: ${VERSION}
+epoch: 0
+
+section: default
+
+maintainer: Docker <support@docker.com>
+description: Docker Buildx plugin extends build capabilities with BuildKit.
+vendor: Docker, Inc
```
AkihiroSuda, line 16: Maybe the vendor name should be provided via `--build-arg`?

crazy-max: Hum yeah indeed, thanks for your feedback

crazy-max: needs https://github.com/goreleaser/nfpm/pull/406 to expand env for vendor field or we could sed but that's kinda dirty.

.github/workflows/build.yml
```diff
@@ -46,6 +46,10 @@ jobs:
         name: Build binaries
         run: |
           make release
+      -
+        name: Build packages
+        run: |
+          make pkg
```
crazy-max, line 52: Packages will not be included in GitHub Releases (only build). Instead we want to use the remote bake definition from docker-ce-packaging repo.

thaJeztah:
So IIUC, the idea is for the buildx repository to provide the "recipe" for building the `.deb` and `.rpm` packages, and to update the `docker-ce-packaging` pipeline, to run it as part of releases, correct?

I had a look at what it produces, and noticed that this currently produces static binaries (packaged as `.deb` / `.rpm`).

```bash
ldd /usr/libexec/docker/cli-plugins/docker-buildx
	linux-vdso.so.1 (0x00007fffc1dc7000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fa7e9f0a000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa7e9b19000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa7ea129000)
```

I know the current buildx binaries we ship in the `docker-ce-cli` packages are dynamically linked, and perhaps that's not an issue, but I do recall that @justincormack didn't want to be using static binaries (and I also recall some binaries could not be shipped as static binaries on Docker Desktop, related to some GPL code, but not sure if that's the case for these).

As to naming of the package, I know we currently use `docker-scan-plugin` (for `docker scan`); given that these debs/rpms _do_ install a cli plugin, I'm inclined to name them accordingly; both for consistency, and  because I'm not sure if there's much value in "obscuring" the fact that it is a plugin. I'll leave that choice to @justincormack though.

thaJeztah:
> Depending on whether or not these will be dynamically linked, we need to either build for each distro+distro-version+architecture, or build once, and duplicate the packages for each one.

Hmmm.. so "build once and copy" won't work, as the version (including distro etc) is also part of the package's metadata, so I guess we'll have to build for each distro (or at least _package_ for each distro). Of course; not sure if that's an actual issue if the binary is exactly the same for all of them (i.e., upgrading distro would be fine with keeping the "old" package).

Here's what's in the `control` file in side the package (compared between the buildx package and the docker-ce-cli) - I haven't checked yet for RPMs;

```bash
cat ./bin/docker-buildx_0.7.0~9-g1250d945_amd64/control/control
Package: docker-buildx
Version: 0:0.7.0~9-g1250d945
Section: default
Priority: optional
Architecture: amd64
Maintainer: Docker <support@docker.com>
Vendor: Docker, Inc
Installed-Size: 51408
Homepage: https://github.com/docker/buildx
Description: Docker Buildx plugin extends build capabilities with BuildKit.
```


```bash
cat ./bin/docker-ce-cli_20.10.11~3-0~ubuntu-focal_amd64/control/control
Package: docker-ce-cli
Source: docker-ce
Version: 5:20.10.11~3-0~ubuntu-focal
Architecture: amd64
Maintainer: Docker <support@docker.com>
Installed-Size: 153865
Depends: libc6 (>= 2.4)
Recommends: docker-scan-plugin
Conflicts: docker (<< 1.5~), docker-engine, docker-engine-cs, docker.io, lxc-docker, lxc-docker-virtual-package
Breaks: docker-ce (<< 5:0)
Replaces: docker-ce (<< 5:0)
Section: admin
Priority: optional
Homepage: https://www.docker.com
Description: Docker CLI: the open-source application container engine
 Docker is a product for you to build, ship and run any application as a
 lightweight container
 .
 Docker containers are both hardware-agnostic and platform-agnostic. This means
 they can run anywhere, from your laptop to the largest cloud compute instance and
 everything in between - and they don't require you to use a particular
 language, framework or packaging system. That makes them great building blocks
 for deploying and scaling web apps, databases, and backend services without
 depending on a particular stack or provider.
```

hack/nfpm.yml
```diff
@@ -0,0 +1,30 @@
...
+platform: linux
+
+version: ${VERSION}
+epoch: 0
+
+section: default
+
+maintainer: Docker <support@docker.com>
+description: Docker Buildx plugin extends build capabilities with BuildKit.
+vendor: ${VENDOR}
```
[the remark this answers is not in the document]
crazy-max: > Do you know if `maintainer` sets `Packager` in the RPM packages?

Yes it is: https://github.com/goreleaser/nfpm/blob/1de91949f55eae477f999e45171586a26b702109/rpm/rpm.go#L231

For deb or apk, maintainer is associated with the right field, [`maintainer` for apk](https://github.com/goreleaser/nfpm/blob/1de91949f55eae477f999e45171586a26b702109/apk/apk.go#L553) and [`Maintainer` for deb](https://github.com/goreleaser/nfpm/blob/1de91949f55eae477f999e45171586a26b702109/deb/deb.go#L676) so I guess that's a mistake on NFPM to use that field as Packager for rpm :(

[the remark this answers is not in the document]
crazy-max: I will open a PR on NFPM repo to switch between vendor/packager.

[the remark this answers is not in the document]
crazy-max: We might be able to add an override on NFPM repo for this case.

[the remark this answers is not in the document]
crazy-max: PR has been merged and `PACKAGER` arg has been added. Here is the result:

```console
$ rpm -qip bin/docker-buildx-0.7.0~9-g1250d945.m.x86_64.rpm 
Name        : docker-buildx
Epoch       : 0
Version     : 0.7.0~9-g1250d945.m
Release     : 1
Architecture: x86_64
Install Date: (not installed)
Group       : Tools/Docker
Size        : 47112192
License     : Apache-2.0
Signature   : (none)
Source RPM  : docker-buildx-0.7.0~9-g1250d945.m-1.src.rpm
Build Date  : Thu Dec 16 11:31:29 2021
Build Host  : buildkitsandbox
Relocations : (not relocatable)
Packager    : Docker <support@docker.com>
Vendor      : Docker
URL         : https://github.com/docker/buildx
Summary     : Docker Buildx plugin extends build capabilities with BuildKit.
Description :
Docker Buildx plugin extends build capabilities with BuildKit.
```

crazy-max:
@thaJeztah 

> So IIUC, the idea is for the buildx repository to provide the "recipe" for building the `.deb` and `.rpm` packages, and to update the `docker-ce-packaging` pipeline, to run it as part of releases, correct?

Yes that's correct

> I had a look at what it produces, and noticed that this currently produces static binaries (packaged as `.deb` / `.rpm`).
> 
> ```shell
> ldd /usr/libexec/docker/cli-plugins/docker-buildx
> 	linux-vdso.so.1 (0x00007fffc1dc7000)
> 	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fa7e9f0a000)
> 	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa7e9b19000)
> 	/lib64/ld-linux-x86-64.so.2 (0x00007fa7ea129000)
> ```
> 
> I know the current buildx binaries we ship in the `docker-ce-cli` packages are dynamically linked, and perhaps that's not an issue, but I do recall that @justincormack didn't want to be using static binaries (and I also recall some binaries could not be shipped as static binaries on Docker Desktop, related to some GPL code, but not sure if that's the case for these).

Not sure but I think it depends on the situation. For a cli plugin like buildx it doesn't make much sense and in the contrary adds a lot of overhead imo.

> As to naming of the package, I know we currently use `docker-scan-plugin` (for `docker scan`); given that these debs/rpms _do_ install a cli plugin, I'm inclined to name them accordingly; both for consistency, and because I'm not sure if there's much value in "obscuring" the fact that it is a plugin. I'll leave that choice to @justincormack though.

Sure let me know if it needs to be changed: https://github.com/crazy-max/buildx/blob/d24a0cdae3563b04d7f1a4e3cee93efa0605b2c1/hack/nfpm.yml#L1

> * We need to double-check if there's differences in the `.deb` / `.rpm` metadata that may be different between distro versions. Admitted; the install for this plugin is more straightforward than (e.g.) `dockerd` or `containerd`, but I know at some point our release-pipeline tried to build packages using a single distro, and there were incompatibilities between `.deb` packages built for newer distros when installed on older versions of the distro (see

For a cli plugin pkg which only "extracts" without using any triggers or pre(post)install scripts it doesn't seem we can hit this case.

> * or build once, and duplicate the packages for each one.

Don't think it would work as the distro ver would be included in the release.

> * Have a way to generate the right filename for each distro/distro-version

We can follow the Debian version scheme sure. Smth like `docker-buildx_0.7.0~deb11u0` or `docker-buildx_0.7.0-el8` for rpm pkgs.

> Given that `buildx` is currently stashed in the `docker-ce-cli` package, and this will move it to a separate package, we need to add the correct metadata for this, so that they cannot be installed side-by-side (which would overwrite each other's binaries)

Good point

> Hmmm.. so "build once and copy" won't work, as the version (including distro etc) is also part of the package's metadata, so I guess we'll have to build for each distro (or at least _package_ for each distro). Of course; not sure if that's an actual issue if the binary is exactly the same for all of them (i.e., upgrading distro would be fine with keeping the "old" package).

I can add a specific ARG to change the release based on the distro.

> Here's what's in the `control` file in side the package (compared between the buildx package and the docker-ce-cli) - I haven't checked yet for RPMs;

### deb

```console
$ dpkg-deb -I bin/docker-buildx_0.7.0~9-g1250d945.m_amd64.deb 
 new Debian package, version 2.0.
 size 12219872 bytes: control archive=419 bytes.
       1 bytes,     0 lines      conffiles
     306 bytes,    10 lines      control
      81 bytes,     1 lines      md5sums
 Package: docker-buildx
 Version: 0:0.7.0~9-g1250d945.m
 Section: default
 Priority: optional
 Architecture: amd64
 Maintainer: Docker <support@docker.com>
 Vendor: Docker
 Installed-Size: 46008
 Homepage: https://github.com/docker/buildx
 Description: Docker Buildx plugin extends build capabilities with BuildKit.
```

Compared with docker-scan-plugin pkg:

```console
$ dpkg-deb -I bin/docker-scan-plugin_0.12.0~debian-bullseye_amd64.deb 
 new Debian package, version 2.0.
 size 3518216 bytes: control archive=556 bytes.
     340 bytes,    11 lines      control
     164 bytes,     2 lines      md5sums
 Package: docker-scan-plugin
 Source: docker-ce (5:20.10.12~3-0~debian-bullseye)
 Version: 0.12.0~debian-bullseye
 Architecture: amd64
 Maintainer: Docker <support@docker.com>
 Installed-Size: 12946
 Enhances: docker-ce-cli
 Section: admin
 Priority: optional
 Homepage: https://github.com/docker/scan-cli-plugin
 Description: Docker scan cli plugin.
```

## rpm

```console
$ rpm -qip bin/docker-buildx-0.7.0~9-g1250d945.m.x86_64.rpm 
Name        : docker-buildx
Epoch       : 0
Version     : 0.7.0~9-g1250d945.m
Release     : 1
Architecture: x86_64
Install Date: (not installed)
Group       : Tools/Docker
Size        : 47112192
License     : Apache-2.0
Signature   : (none)
Source RPM  : docker-buildx-0.7.0~9-g1250d945.m-1.src.rpm
Build Date  : Thu Dec 16 11:31:29 2021
Build Host  : buildkitsandbox
Relocations : (not relocatable)
Packager    : Docker <support@docker.com>
Vendor      : Docker
URL         : https://github.com/docker/buildx
Summary     : Docker Buildx plugin extends build capabilities with BuildKit.
Description :
Docker Buildx plugin extends build capabilities with BuildKit.
```

Compared with docker-scan-plugin pkg:

```console
$ rpm -qip bin/docker-scan-plugin-0.12.0-3.el8.x86_64.rpm 
warning: bin/docker-scan-plugin-0.12.0-3.el8.x86_64.rpm: Header V4 RSA/SHA512 Signature, key ID 621e9f35: NOKEY
Name        : docker-scan-plugin
Epoch       : 0
Version     : 0.12.0
Release     : 3.el8
Architecture: x86_64
Install Date: (not installed)
Group       : Tools/Docker
Size        : 13247224
License     : ASL 2.0
Signature   : RSA/SHA512, Mon Dec 13 15:21:53 2021, Key ID c52feb6b621e9f35
Source RPM  : docker-scan-plugin-0.12.0-3.el8.src.rpm
Build Date  : Mon Dec 13 12:45:11 2021
Build Host  : c4ca28c21cdf
Relocations : (not relocatable)
Packager    : Docker <support@docker.com>
Vendor      : Docker
URL         : https://github.com/docker/scan-cli-plugin/
Summary     : Docker Scan plugin for the Docker CLI
Description :
Docker Scan plugin for the Docker CLI.
```

crazy-max:
@thaJeztah 

> > Hmmm.. so "build once and copy" won't work, as the version (including distro etc) is also part of the package's metadata, so I guess we'll have to build for each distro (or at least _package_ for each distro). Of course; not sure if that's an actual issue if the binary is exactly the same for all of them (i.e., upgrading distro would be fine with keeping the "old" package).
> 
> I can add a specific ARG to change the release based on the distro.

I added the possibility to define a release version for each distro. See https://github.com/crazy-max/buildx/blob/23d8eaffa41e2748b0b7894bb239290352745c4b/docker-bake.hcl#L10-L15

crazy-max:
@thaJeztah Last commit publish pkgs to `docker/buildx-pkg`. Will look like this: https://hub.docker.com/r/crazymax/buildx-pkg/tags

image

Structure inside the scratch image:

image

crazy-max:
Changed pkg name to `docker-buildx-plugin` as discussed. PTAL @thaJeztah @tonistiigi.

PR #852, opened 2021-11-19
Add an flag to buildx rm to keep the buildkitd daemon running

### LINKED ISSUE #794: Docker buildx rm always deletes buildkit Kubernetes deployments
Hello,

We are currently using `buildx` with the Kubernetes driver to execute image builds on Kubernetes pods. We are managing the deployments separately and we are using the `--name` flag to register these deployments into the `buildx` context.
Our issue is that when we remove the `buildx` context with `docker buildx rm`, it also deletes the Kubernetes deployments, which we don't want to do in our case.

For people that are managing a shared pool of buildkit pods and run builds from multiple `buildx` clients, the current behaviour does not look very safe given that no one can remove the `buildx` context from their machine without deleting the entire pool of pods. It would be nice to specify at the creation that a `buildx` context is using already existing Kubernetes deployments, and not delete those deployments when deleting the `buildx` context.

However, I can understand that changing the current behaviour might be tricky. If that is not possible, a flag like `--do-not-delete-deployments` to pass to `docker buildx rm` would allow us to delete our `buildx` contexts safely.

If the idea looks good for you, we are happy to provides help implementing this feature into buildkit.

zaymat:
This PR adds the flag  `--keep-buildkitd` to the `buildx rm` command to preserve the buildkitd daemon after the buildx context is deleted.

This flag can be useful when someone uses the `kubernetes` driver and manages the `buildkit` deployment outside of `buildx`.

```bash
docker buildx rm --help

  Usage:  docker buildx rm [NAME]
  
  Remove a builder instance
  
  Options:
        --builder string   Override the configured builder instance
        --keep-buildkitd   Keep the buildkitd daemon running
        --keep-state       Keep BuildKit state
```

Fixes #794

Signed-off-by: Mayeul Blanzat <mayeul.blanzat@datadoghq.com>

### DIFF
--- a/commands/rm.go
+++ b/commands/rm.go
@@ -12,8 +12,9 @@ import (
 )
 
 type rmOptions struct {
-	builder   string
-	keepState bool
+	builder    string
+	keepState  bool
+	keepDaemon bool
 }
 
 func runRm(dockerCli command.Cli, in rmOptions) error {
@@ -30,7 +31,7 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
 		if err != nil {
 			return err
 		}
-		err1 := rm(ctx, dockerCli, ng, in.keepState)
+		err1 := rm(ctx, dockerCli, ng, in.keepState, in.keepDaemon)
 		if err := txn.Remove(ng.Name); err != nil {
 			return err
 		}
@@ -42,7 +43,7 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
 		return err
 	}
 	if ng != nil {
-		err1 := rm(ctx, dockerCli, ng, in.keepState)
+		err1 := rm(ctx, dockerCli, ng, in.keepState, in.keepDaemon)
 		if err := txn.Remove(ng.Name); err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


commands/rm.go
```diff
@@ -70,23 +71,28 @@ func rmCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.BoolVar(&options.keepState, "keep-state", false, "Keep BuildKit state")
+	flags.BoolVar(&options.keepBuildkitd, "keep-buildkitd", false, "Keep the buildkitd daemon running")
```
crazy-max, line 74: I think `--keep-instance` would be a better name.

tonistiigi: @crazy-max maybe `--keep-daemon`? "Instance" is not very descriptive either as anything can have "an instance".

crazy-max: yes `--keep-daemon` sgtm :+1:

zaymat: `--keep-daemon` sgtm too 👍 

I'll make the change

docs/reference/buildx_rm.md
```diff
@@ -34,7 +34,7 @@ Same as [`buildx --builder`](buildx.md#builder).
 Keep BuildKit state, so it can be reused by a new builder with the same name.
 Currently, only supported by the [`docker-container` driver](buildx_create.md#driver).
 
-### <a name="keep-buildkitd"></a> Keep the buildkitd daemon running (--keep-buildkitd)
+### <a name="keep-daemon"></a> Keep the buildkitd daemon running (--keep-daemon)
 
 Keep the buildkitd daemon running after the buildx context is removed. This is useful is when you manage buildkitd daemons and buildx contexts independently.
```
crazy-max, line 39: `This is useful is when` > `This is useful when`

crazy-max (changes requested):

crazy-max (changes requested):
Please also squash the commits. https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md#review

zaymat:
> Please also squash the commits. https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md#review

Sure 👍  Just did it

crazy-max (approved):
LGTM

tonistiigi (approved):

PR #854, opened 2021-11-20
imagetools inspect: add --format flag

crazy-max:
follow-up https://github.com/docker/roadmap/issues/243#issuecomment-974385422

adds `--format` flag to `imagetools inspect` command. Currently supported values are `manifest` (default and current behavior) and `buildinfo`.

`buildinfo` is the new one to retrieve [build information](https://github.com/moby/buildkit/blob/master/docs/build-repro.md) for an image in the registry.

```console
$ docker buildx imagetools inspect --format buildinfo crazymax/buildkit-buildsources-test:buildx
Name:    docker.io/crazymax/buildkit-buildsources-test:buildx

Sources:
  Type: docker-image
  Ref:  docker.io/docker/buildx-bin:0.6.1@sha256:a652ced4a4141977c7daaed0a074dcd9844a78d7d2615465b12f433ae6dd29f0
  Pin:  sha256:a652ced4a4141977c7daaed0a074dcd9844a78d7d2615465b12f433ae6dd29f0

  Type: docker-image
  Ref:  docker.io/library/alpine:3.13
  Pin:  sha256:026f721af4cf2843e07bba648e158fb35ecc876d822130633cc49f707f0fc88c

  Type: docker-image
  Ref:  docker.io/moby/buildkit:v0.9.0
  Pin:  sha256:8dc668e7f66db1c044aadbed306020743516a94848793e0f81f94a087ee78cab

  Type: docker-image
  Ref:  docker.io/tonistiigi/xx@sha256:21a61be4744f6531cb5f33b0e6f40ede41fa3a1b8c82d5946178f80cc84bfc04
  Pin:  sha256:21a61be4744f6531cb5f33b0e6f40ede41fa3a1b8c82d5946178f80cc84bfc04

  Type: git
  Ref:  https://github.com/crazy-max/buildkit-buildsources-test.git#master
  Pin:  c7f9447d38471b486e248857a89143776d27096c

  Type: http
  Ref:  https://raw.githubusercontent.com/moby/moby/master/README.md
  Pin:  sha256:419455202b0ef97e480d7f8199b26a721a417818bc0e2d106975f74323f25e6c
```

cc @JamieMagee


util/imagetools/inspect.go
```diff
@@ -47,14 +51,19 @@ func (r *Resolver) resolver() remotes.Resolver {
...
 		},
-		Client: tracing.DefaultClient,
 	})
 }
 
 func (r *Resolver) Resolve(ctx context.Context, in string) (string, ocispec.Descriptor, error) {
+	// needs to discard containerd logger otherwise it will print
+	// unnecessary info during image reference resolution like
+	// https://github.com/containerd/containerd/blob/1a88cf5242445657258e0c744def5017d7cfb492/remotes/docker/resolver.go#L288
+	ctx = bxctx.WithLoggerDiscarded(ctx)
```
crazy-max, line 65: needs to discard containerd logger otherwise it will print unnecessary info during image reference resolution:

```console
$ docker buildx imagetools inspect alpinefoo
INFO[0001] trying next host                              error="pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed" host=registry-1.docker.io
error: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

now:

```console
$ docker buildx imagetools inspect alpinefoo
error: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

crazy-max:
In addition, wondering if we could add another flag to display buildinfo from a metadata file generated by the [build command](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#options).

JamieMagee:
@crazy-max Thanks for this!

What's the typical release cadence for `buildx`? And when could I expect to see this functionality in a docker cli or docker ce release?

crazy-max:
@tonistiigi As discussed, `--format` flag has been added to the `imagetools inspect` command to handle build information. In a follow-up we could add a `config` format to display the image configuration.

@JamieMagee 

> What's the typical release cadence for `buildx`? And when could I expect to see this functionality in a docker cli or docker ce release?

The `buildinfo` feature is only available in the main branch of the BuildKit repo atm. Not sure when the first 0.10 RC of BuildKit will be released. I think @tonistiigi and @sipsma want to consolidate the `MergeOp` feature and provide `COPY` with merge semantics.

About buildx, there will be a patch release ([0.7.1](https://github.com/docker/buildx/milestone/10)) to fix the current issues. This PR will probably be available in the next minor release but you can already [download the artifacts in staging](https://github.com/docker/buildx/actions/runs/1494644633) from the GitHub workflow if you want to try it.

tonistiigi:
I'd expect the template format like other commands. Eg. `{{json .Config}}`, `{{json .BuildInfo}}`

commands/imagetools/inspect.go
```diff
@@ -85,7 +134,8 @@ func inspectCmd(dockerCli command.Cli, rootOpts RootOptions) *cobra.Command {
 	}
 
 	flags := cmd.Flags()
-	flags.BoolVar(&options.raw, "raw", false, "Show original JSON manifest")
+	flags.StringVar(&options.format, "format", "manifest", "Format the output")
```
AkihiroSuda, line 137:
```suggestion
	flags.StringVar(&options.format, "format", "manifest", "Format the output (\"manifest\", or \"buildinfo\")")
```

crazy-max:
> I'd expect the template format like other commands. Eg. `{{json .Config}}`, `{{json .BuildInfo}}`

Agree but then `--raw` should be removed.

tonistiigi:
`--raw` is not the same as `{{json .}}`. It is the data that can be used to validate the checksum. We could move it to another field but it does have a very specific use case.

crazy-max:
Ok so `{{json .}}` would print manifest + config + buildinfo right? Currently we display the manifest in [raw format](https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/cli/command/formatter/formatter.go#L17) so I would guess the default value for format would be `{{raw .Manifest}}`. I guess we should implement a formatter like the one from #830. WDYT?

tonistiigi:
`{{json .}}` printing all in json format, `{{.buildinfo}}` printing the buildinfo block. `{{json .config}}` printing the config as json would seem logical to me. `{{.manifest}}` I guess is the same but a bit weird that it is either index or manifest.

For `--raw` I think we can make it just incompatible with `--format`. Theoretically `--raw --format `{{json .config}}` could print the raw value of config, but that seems too special case to be worth of a custom implementation.

Is the config taking the first item from the index atm?

crazy-max:
@tonistiigi 

Sorry didn't see your comment!

> `{{json .}}` printing all in json format, `{{.buildinfo}}` printing the buildinfo block. `{{json .config}}` printing the config as json would seem logical to me. `{{.manifest}}` I guess is the same but a bit weird that it is either index or manifest.
> 
> For `--raw` I think we can make it just incompatible with `--format`. Theoretically `--raw --format `{{json .config}}` could print the raw value of config, but that seems too special case to be worth of a custom implementation.

Ok sgtm

> Is the config taking the first item from the index atm?

Atm I don't handle config, only buildinfo and manifest.

tonistiigi (commented):
```
» docker buildx imagetools inspect --raw alpine
error: format and raw cannot be used together
```

util/imagetools/printers.go
```diff
@@ -1,72 +1,240 @@
...
+	case images.MediaTypeDockerSchema2Manifest, ocispecs.MediaTypeImageManifest:
+		manifest = p.manifest
+	case images.MediaTypeDockerSchema2ManifestList, ocispecs.MediaTypeImageIndex:
+		manifest = p.index
+	}
+	v := struct {
+		Name      string               `json:"name,omitempty"`
+		Manifest  interface{}          `json:"manifest,omitempty"`
+		Config    ocispecs.Image       `json:"config,omitempty"`
+		BuildInfo binfotypes.BuildInfo `json:"buildinfo,omitempty"`
```
tonistiigi, line 127: This omitempty does not work if it is not a pointer

tonistiigi (commented):
For multi-platform this breaks:

```
docker buildx imagetools inspect  mysql
error: no match for platform in manifest sha256:0962b771c2398c6dcddbbe77b3cf6658408396229b612035d938fb7c8d11c23c: not found
```

tonistiigi:
As a follow-up, looks like some pkg reorganization is needed as these functions definitely should not require importing `ttrpc` etc.

docs/reference/buildx_imagetools_inspect.md
```diff
@@ -45,7 +64,492 @@ Manifests:
...
 Same as [`buildx --builder`](buildx.md#builder).
 
+### <a name="format"></a> Format the output (--format)
+
+Format the output using the given Go template. Defaults to `{{.Manifest}}` if
+unset. Following fields are available:
+
+* `.Name`: provides the reference of the image
+* `.Manifest`: provides manifest or manifest list
+* `.Image`: provides the image config as `application/vnd.oci.image.config.v1+json` mediatype
```
tonistiigi, line 74: what does this "as mediatype" mean?

crazy-max: Yes that's it I will remove that. Was taken from https://github.com/opencontainers/image-spec/blob/1f308fd27264a52f6a089202b1ff9dc15ff27398/specs-go/v1/config.go#L81

tonistiigi (approved):

PR #860, opened 2021-11-23
build: add no-cache-filter

tonistiigi:
Allows specifying the `no-cache` per stage. Partially implements https://github.com/moby/buildkit/issues/1213

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -524,7 +524,7 @@ type Target struct {
 	Pull             *bool             `json:"pull,omitempty" hcl:"pull,optional"`
 	NoCache          *bool             `json:"no-cache,omitempty" hcl:"no-cache,optional"`
 	NetworkMode      *string           `json:"-" hcl:"-"`
-
+	NoCacheFilter    []string          `json:"no-cache-filter,omitempty" hcl:"no-cache-filter,optional"`
 	// IMPORTANT: if you add more fields here, do not forget to update newOverrides and README.
 }
 
@@ -536,6 +536,7 @@ func (t *Target) normalize() {
 	t.CacheFrom = removeDupes(t.CacheFrom)
 	t.CacheTo = removeDupes(t.CacheTo)
 	t.Outputs = removeDupes(t.Outputs)
+	t.NoCacheFilter = removeDupes(t.NoCacheFilter)
 
 	for k, v := range t.Contexts {
 		if v == "" {
@@ -608,6 +609,9 @@ func (t *Target) Merge(t2 *Target) {
 	if t2.NetworkMode != nil {
 		t.NetworkMode = t2.NetworkMode
 	}
+	if t2.NoCacheFilter != nil { // merge
+		t.NoCacheFilter = append(t.NoCacheFilter, t2.NoCacheFilter...)
+	}
 	t.Inherits = append(t.Inherits, t2.Inherits...)
 }
 
@@ -666,6 +670,8 @@ func (t *Target) AddOverrides(overrides map[string]Override) error {
 				return errors.Errorf("invalid value %s for boolean key no-cache", value)
 			}
 			t.NoCache = &noCache
+		case "no-cache-filter":
+			t.NoCacheFilter = o.ArrValue
 		case "pull":
 			pull, err := strconv.ParseBool(value)
 			if err != nil {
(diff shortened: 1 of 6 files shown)


crazy-max:
Should be added to compose with the `x-bake` extension too: https://github.com/docker/buildx/blob/5c2b9bbfc5f9ef485781b5165131c4d13f21fcc9/bake/compose.go#L119

crazy-max (approved):

thaJeztah:
Silly (?) idea; would it work if, instead of adding an additional flag, we would allow `--no-cache` to take these options?

   --no-cache (no args) same as today (disable all caches)
    --no-cache <filter options> (what's implemented in this PR

tonistiigi:
@thaJeztah No, I don't think this is possible. Even with some crazy hack (that is not worth the maintenance cost) to flag parsing, it is ambiguous to positional args in some cases.

Theoretically, if we wanted we could deprecate the old flag and add something that signifies "all stages" to new flag. But I think the current flag is already too common.

tonistiigi:
@thaJeztah sgty?

thaJeztah:
> Even with some crazy hack (that is not worth the maintenance cost) to flag parsing, it is ambiguous to positional args in some cases.

Hm.. yeah, it may be tricky to implement it (with the positional args). I was curious if we wouldn't already be in this situation with `--no-cache` (because boolean flags also can accept `true / false` as argument);

```bash
mkdir true
echo "FROM busybox" > true/Dockerfile

docker build -t foo true
[+] Building 0.1s (5/5) FINISHED
...


docker build -t foo --no-cache true
[+] Building 0.1s (5/5) FINISHED
...
```

But it looks like (unlike for other type of flags?) boolean flags _require_ values to be passed with `=` ? (TIL)

```bash
docker build -t foo --no-cache true true
"docker build" requires exactly 1 argument.
See 'docker build --help'.

docker build -t foo --no-cache=true true
[+] Building 0.1s (5/5) FINISHED
...
```

thaJeztah:
So the "short answer" is that I don't really like `--no-cache-filter`. It's quite verbose for a flag name, it overlaps with `--no-cache`, and (imo) it would still be nicer if it would be possible with the existing flag, but yes, I see how it may be a bit of a challenge to get it right.


As to deprecating `--no-cache`; I'm not sure if we'd be able to pull that off; it's been around for a long time, and I doubt we'd be able to remove it without it being a large breaking change.

Irregardless of what we choose in that respect, I think we should consider making them mutually exclusive, and print a conflict error if both is used. Looking at (e.g.) this example from https://github.com/moby/buildkit/issues/1213:

> `--no-cache --no-cache-filter !target1,!target2` invalidates cache for all targets but `target1` and `target2` (less common usecase)

Looking at that example, I don't think the `--no-cache` flag adds any value there, and can be deduced from the `--no-cache-filter` flag alone (`--no-cache-filter` implicitly says; "use cache, (except) for ...")

thaJeztah:
One more thing I observed in the examples from that ticket;

> `--no-cache-filter mount:id1` invalidates cache for the cache mount with id id1 (defined with RUN `--mount=type=cache,id=id1,...`)

Less of a strong opinion on that, but I recall at some point, we "deprecated" some syntaxes that used a colon (`:`) as separator (e.g. `--security-opt seccomp:unconfined`) in favor of using `=` (`--security-opt seccomp=unconfined`). I'm wondering if we should do the same here, so that it would become:

```bash
--no-cache-filter mount=id1
```

Also wondering; if the above is to "use all cache mounts, except for mount `id1`", would there also be a reverse? ("only use cache mount `id1`"), and what that would look like?

```bash
--no-cache-filter !mount:id1
```

If we'd go for `=`, that could become

```bash
--no-cache-filter mount!=id1
```

(which feels the 'natural' thing to do), but if we want parsing to be handled using the "csv" notation used elsewhere, I guess the "correct" way would be;

```bash
--no-cache-filter mount=!id1
```

Which, hm, would be a bit awkward as well

(overall, the double negative is a bit hard to grasp; `--no-cache` combined with exclusions is easy to get "wrong")

tonistiigi:
@thaJeztah What changes are you asking? I can't implement `!` or mount-id atm as there is no support for either in the current buildkit. While stage based no-cache has always been supported in buildkit but never exposed to docker.

I don't think `--no-cache` can be deduced for all cases with exclusions. What does `foo,!bar` mean?

tonistiigi:
ping @thaJeztah

crazy-max:
I have a [use case](https://github.com/crazy-max/docker-cli/blob/cb0968d143280584a53ec388e9fbad9d31cf69b4/docker-bake.hcl#L93-L105) for that one :slightly_smiling_face:

thaJeztah:
As discussed;

- perhaps good to (for now) produce an error if both flags are set
- if possible; add docs to describe the behavior (including the "typo" case?  😉)

tonistiigi:
@thaJeztah PTAL

crazy-max (approved):
LGTM

taranlu-houzz:
@tonistiigi Hello, not sure where the right place to ask this is, but I am unclear on what to provide as an argument to the `--no-cache-filter` flag. For example, if I want to skip one specific step (e.g. 12/14), what would the argument need to be? I tried several things, but none of them seemed to work (at least, the output from `docker build --progress plain` always showed the step as `#CACHED`).

tonistiigi:
@taranlu-houzz you need to move your step to a separate stage

```
FROM alpine AS base
RUN first
RUN second
```

to skip second:

```
FROM alpine AS base
RUN first
FROM base AS mid
RUN second
```

`--no-cache-filter mid`

taranlu-houzz:
Hmm, I see. Okay, I will try that out. Is this documented anywhere? I am new-ish to `docker` and I couldn't really find any examples online of how to use `--no-cache-filter`.

Issue #179, opened 2019-10-26
Add entitlements to bake


tonistiigi:
Entitlements are not currently enabled in bake. You can do `buildx build --allow security.insecure` but you can't set the same in a file for bake.

Setting this should be allowed with a two-step process. A target can request that entitlement is required, but that doesn't automatically mean that any value should be directly sent to the daemon as validation of the authority of the request. The model of entitlements requires that the user explicitly allows a possibly insecure option, and running bake should still be considered somewhat secure.

So instead, if the user is building a bake target that has requests an insecure entitlement, they would need to pass the `--allow` flag with the bake command, or possibly if they don't, they get prompted.

For example:

```
Error: target "foo" requires entitlement "security.insecure"

Pass "--allow security.insecure" to grant the requested privileges.

Your full command with requested privileges:

docker buildx bake --allow security.insecure release
```

The benefit of defining the requirements in the files is that we can show nice message (or prompt) and we can error right away, not on failure for a specific build. We can also show all the errors together as a group.

Not only build entitlements should get protected. There are some other properties (some of which we already support) that should be protected as well.

### These are:

- File paths:
Context directory, Dockerfile path, the output path for local/tar exporter, the output path for local cache exporter, secret and ssh key paths.

- Images:
Image reference (name), exported cache reference, push action

- SSH socket

Most of these options have values that are allowed by default. For the file paths, all paths pointing inside the working directory should be allowed by default, but if you set a Dockerfile path to `/etc/passwd/` in a bake file, it should not be allowed unless you pass `--allow fs.read=/etc/`. 

The same goes for images. It is expected that you can create a new image by default, but it is possibly harmful to allow bake to push an image automatically. A tricky case is protecting certain image names. I think we should allow tagging images as `docker.io/username/foo` but not just `foo` because the user does not own the `docker.io/library` namespace. So a bake shouldn't be used to rewrite an image like "ubuntu" without the user noticing. This is somewhat controversial as it would probably break some compose files. Otoh, it would motivate users to avoid this harmful behavior of not properly naming images. 

Proposed values for `bake --allow`:

```
network.host
security.insecure
fs.read=<path>
fs.write=<path>
fs=<path>
image.push=<pattern>
image.load=<pattern>
image.pull=<pattern> // requires buildkit change
image=<pattern>
ssh
```

[...]

PR #863, opened 2021-11-25
Add NetworkMode to bake target

### LINKED ISSUE #848: buildx bake not honouring docker-compose build network mode host
Assuming I have the below, and a docker container running on my machine that exposes port `3000`

**docker-compose.yml**
```
version: '3.2'

services:
  app:
    image: "test-boolean2"
    restart: always
    user: root
    build:
      context: .
      network:
        host
      dockerfile: Dockerfile
```

**Dockerfile**
```
FROM centos:8

RUN curl http://localhost:3000
```

- `docker-compose build` succeeds
- `docker buildx bake` succeeds when using `docker-container` driver with `--driver-opt network=host`
- `docker buildx build --network=host .` succeeds when using the `default` builder/`docker` driver

However `docker buildx bake` fails when using the `default` builder as it can't reach `localhost:3000`. It feels as though perhaps `buildx bake` might not be using `network: host` specified in the `docker-compose.yml` correctly?

zachary-povey:
Allows specification of network mode in a bake target.

Fixes #848

Signed-off-by: Zachary Povey <zachary.povey@autotrader.co.uk>

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -567,8 +567,6 @@ func (t *Target) AddOverrides(overrides map[string]Override) error {
 					}
 				}
 			}
-		case "network":
-			t.NetworkMode = &o.Value
 		default:
 			return errors.Errorf("unknown key: %s", keys[0])
 		}
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -426,7 +426,7 @@ type Target struct {
 	Outputs          []string          `json:"output,omitempty" hcl:"output,optional"`
 	Pull             *bool             `json:"pull,omitempty" hcl:"pull,optional"`
 	NoCache          *bool             `json:"no-cache,omitempty" hcl:"no-cache,optional"`
-	NetworkMode      *string           `json:"network,omitempty" hcl:"network,optional"`
+	NetworkMode      *string
 
 	// IMPORTANT: if you add more fields here, do not forget to update newOverrides and README.
 }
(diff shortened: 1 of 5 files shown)


tonistiigi:
I wonder if we should leave it only for compose files for now. Using `build --network` is discouraged in buildx. Instead you are supposed to use `RUN --network` inside the dockerfile https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#network-modes-run---networknonehostdefault and allow network entitlement on build. Proposal for entitlements support in bake is in https://github.com/docker/buildx/issues/179

zachary-povey:
@tonistiigi - happy to change it to just compose files, should I do that? 

Mu use case for this is backwards compatibility with `docker-compose` anyway, it just seemed (from a naive perspective) if something was available to `build` it should be available via a `bake` target.

tonistiigi:
> happy to change it to just compose files, should I do that?

I think so. cc @crazy-max

crazy-max:
> 

Yes sgtm

zachary-povey:
Ok, thanks both :) 

@tonistiigi I've pushed a commit removing support from HCL/JSON files. Did you also want me to remove the ability to specify it via a cmd line override with `--set`?

bake/bake.go
```diff
@@ -563,6 +567,8 @@ func (t *Target) AddOverrides(overrides map[string]Override) error {
 					}
 				}
 			}
+		case "network":
+			t.NetworkMode = &o.Value
```
crazy-max, line 571: > Did you also want me to remove the ability to specify it via a cmd line override with `--set`?

Yes remove the override for this case.

zachary-povey: Ok that's done now 👍

zachary-povey:
@tonistiigi @crazy-max - think this is good to go now?

crazy-max (approved):
LGTM

PR #865, opened 2021-11-27
refactor: mutualize builder logic

crazy-max:
follow-up #830

Has been reviewed based on [your comments](https://github.com/docker/buildx/pull/830#pullrequestreview-811800194) @tonistiigi. After that we will be able to implement filter/format for some commands. I kept some kubernetes driver logic as it was before but it should be moved to its own factory.


errordeveloper (commented):
As we discussed, perhaps it would be good to make a separate PR for the first commit here that could merged soon :)

crazy-max:
closing this one, see https://github.com/docker/buildx/pull/1430

PR #881, opened 2021-12-10
bake: fix groups print

crazy-max:
fixes #877 

@tonistiigi Also found an odd case when a group and target can have the same name without erroring:

```hcl
group "foo" {
  targets = ["image"]
}
target "foo" {
  dockerfile = "bar"
}
target "image" {
  dockerfile = "test"
}
```

```console
$ docker buildx bake --print foo
{
  "group": {
    "default": {
      "targets": [
        "image"
      ]
    }
  },
  "target": {
    "image": {
      "context": ".",
      "dockerfile": "test"
    }
  }
}
```

Not sure if it's expected?

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #884, opened 2021-12-14
bake: fix group resolution

crazy-max:
follow-up #881

If a target has the same name as a group, it's resolution is skipped.

```hcl
group "foo" {
  targets = ["foo"]
}
target "foo" {
  dockerfile = "bar"
}
```

```
$ docker buildx bake foo --print
{
  "group": {
    "default": {
      "targets": [
        "foo"
      ]
    }
  },
  "target": {}
}
```

Now:

```
$ docker buildx bake foo --print
{
  "group": {
    "default": {
      "targets": [
        "foo"
      ]
    }
  },
  "target": {
    "foo": {
      "context": ".",
      "dockerfile": "bar"
    }
  }
}

### DIFF
--- a/bake/bake_test.go
+++ b/bake/bake_test.go
@@ -585,3 +585,69 @@ services:
 	require.Equal(t, "./Dockerfile", *m["addon"].Dockerfile)
 	require.Equal(t, "./aws.Dockerfile", *m["aws"].Dockerfile)
 }
+
+func TestReadTargetsSameGroupTarget(t *testing.T) {
+	t.Parallel()
+	ctx := context.TODO()
+
+	f := File{
+		Name: "docker-bake.hcl",
+		Data: []byte(`
+group "foo" {
+  targets = ["foo"]
+}
+target "foo" {
+  dockerfile = "bar"
+}
+target "image" {
+  output = ["type=docker"]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: bake/bake_test.go [hunk 1]
```diff
@@ -585,3 +585,69 @@ services:
...
 	require.Equal(t, "./aws.Dockerfile", *m["aws"].Dockerfile)
 }
+
+func TestReadTargetsSameGroupTarget(t *testing.T) {
+	t.Parallel()
+	ctx := context.TODO()
+
+	f := File{
+		Name: "docker-bake.hcl",
+		Data: []byte(`
```
errordeveloper, line 595: I find JSON a bit nicer for these kind of tests, as you can indent each line as much as needed ;)

However, this test suite in particular has lots of strings with YAML, JSON & HCL... maybe a cleanup is due to move these assets to separate files.

errordeveloper (approved):

errordeveloper:
I think the `group` and `target` functions seem a bit hard to read, maybe less ambiguous names could help, or perhaps these could be rewritten a little. Just an observation...

tonistiigi (approved):

PR #885, opened 2021-12-14
cli: add --all-inactive for rm command

### LINKED ISSUE #851: add buildx rm --all-inactive
Sometimes you find that there are too many builders you have forgotten about in `buildx ls` and it is a pain to remove them one by one in `buildx rm`.

`buildx rm --all-inactive` could look up all builders that are not in running state and prompt the user if they want to remove all of them together.

crazy-max:
fixes #851 

Add `--inactive` to be able to remove nodes that are not in running state. This will also remove builders that don't have any node left available after the process.

### DIFF
--- a/commands/rm.go
+++ b/commands/rm.go
@@ -2,23 +2,36 @@ package commands
 
 import (
 	"context"
+	"time"
 
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/store/storeutil"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/moby/buildkit/util/appcontext"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 )
 
 type rmOptions struct {
-	builder    string
-	keepState  bool
-	keepDaemon bool
+	builder     string
+	keepState   bool
+	keepDaemon  bool
+	allInactive bool
+	force       bool
 }
+
+const (
+	rmInactiveWarning = `WARNING! This will remove all builders that are not in running state. Are you sure you want to continue?`
+)
 
 func runRm(dockerCli command.Cli, in rmOptions) error {
 	ctx := appcontext.Context()
+
+	if in.allInactive && !in.force && !command.PromptForConfirmation(dockerCli.In(), dockerCli.Out(), rmInactiveWarning) {
+		return nil
+	}
 
 	txn, release, err := storeutil.GetStore(dockerCli)
 	if err != nil {
@@ -26,12 +39,16 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
 	}
 	defer release()
 
+	if in.allInactive {
+		return rmAllInactive(ctx, txn, dockerCli, in)
+	}
+
 	if in.builder != "" {
 		ng, err := storeutil.GetNodeGroup(txn, dockerCli, in.builder)
 		if err != nil {
 			return err
 		}
-		err1 := rm(ctx, dockerCli, ng, in.keepState, in.keepDaemon)
+		err1 := rm(ctx, dockerCli, in, ng)
 		if err := txn.Remove(ng.Name); err != nil {
 			return err
 		}
@@ -43,7 +60,7 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
 		return err
 	}
 	if ng != nil {
-		err1 := rm(ctx, dockerCli, ng, in.keepState, in.keepDaemon)
+		err1 := rm(ctx, dockerCli, in, ng)
 		if err := txn.Remove(ng.Name); err != nil {
 			return err
 		}
@@ -63,6 +80,9 @@ func rmCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.builder = rootOpts.builder
 			if len(args) > 0 {
+				if options.allInactive {
+					return errors.New("cannot specify builder name when --all-inactive is set")
+				}
 				options.builder = args[0]
 			}
 			return runRm(dockerCli, options)
@@ -72,11 +92,13 @@ func rmCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVar(&options.keepState, "keep-state", false, "Keep BuildKit state")
 	flags.BoolVar(&options.keepDaemon, "keep-daemon", false, "Keep the buildkitd daemon running")
+	flags.BoolVar(&options.allInactive, "all-inactive", false, "Remove all inactive builders")
+	flags.BoolVarP(&options.force, "force", "f", false, "Do not prompt for confirmation")
 
 	return cmd
 }
 
-func rm(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, keepState, keepDaemon bool) error {
+func rm(ctx context.Context, dockerCli command.Cli, in rmOptions, ng *store.NodeGroup) error {
 	dis, err := driversForNodeGroup(ctx, dockerCli, ng, "")
 	if err != nil {
 		return err
@@ -86,12 +108,12 @@ func rm(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, keepSta
 			continue
 		}
 		// Do not stop the buildkitd daemon when --keep-daemon is provided
-		if !keepDaemon {
+		if !in.keepDaemon {
 			if err := di.Driver.Stop(ctx, true); err != nil {
 				return err
 			}
 		}
-		if err := di.Driver.Rm(ctx, true, !keepState, !keepDaemon); err != nil {
+		if err := di.Driver.Rm(ctx, true, !in.keepState, !in.keepDaemon); err != nil {
 			return err
 		}
 		if di.Err != nil {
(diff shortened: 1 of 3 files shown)


commands/rm.go
```diff
@@ -100,3 +119,53 @@ func rm(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, keepSta
...
+	for _, b := range builders {
+		func(b *nginfo) {
+			eg.Go(func() error {
+				if err := loadNodeGroupData(ctx, dockerCli, b); err != nil {
+					return errors.Wrapf(err, "cannot load %s", b.ng.Name)
+				}
+				var kn []store.Node
+				for idx, n := range b.ng.Nodes {
+					d := b.drivers[idx]
+					if d.info == nil || d.info.Status != driver.Running {
```
crazy-max, line 147: Should be better to check for `driver.Inactive` instead and keep stopped or starting nodes?

commands/rm.go
```diff
@@ -100,3 +119,53 @@ func rm(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, keepSta
...
+						_ = d.di.Driver.Rm(ctx, true, true, true)
```
crazy-max, line 148: Should we forward `keepState` and `keepDaemon` options?

commands/rm.go
tonistiigi: `driver.Inactive` just seems initial state so unrelated in here. Should check nodes but iiuc you need to check all nodes and rm if none of them are running.

tonistiigi: I think logic is same for dynamic nodes @morlay

commands/rm.go
tonistiigi: I think so

commands/rm.go
```diff
@@ -72,6 +90,7 @@ func rmCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVar(&options.keepState, "keep-state", false, "Keep BuildKit state")
 	flags.BoolVar(&options.keepDaemon, "keep-daemon", false, "Keep the buildkitd daemon running")
+	flags.BoolVar(&options.inactive, "inactive", false, "Remove inactive builders")
```
tonistiigi, line 93: I prefer `all-inactive` that makes it clear that this is a different mode

commands/rm.go
```diff
@@ -2,30 +2,48 @@ package commands
...
+		return nil
+	}
+
 	txn, release, err := storeutil.GetStore(dockerCli)
 	if err != nil {
 		return err
 	}
 	defer release()
 
+	if in.inactive {
```
tonistiigi, line 43: Should check that no named args were added that would conflict with this.

commands/rm.go
crazy-max: Agree but more generally I think the CLI has a confusing design for some commands. For example the following would make more sense:

* `docker buildx du` > `docker buildx cache du`
* `docker buildx prune` > `docker buildx cache prune`
* `docker buildx rm --all-inactive` > `docker buildx prune`

commands/rm.go
crazy-max: > Should check nodes but iiuc you need to check all nodes and rm if none of them are running.

Yes already done: https://github.com/docker/buildx/pull/885/files#diff-aaef9e407115d8cec3e6bde548bd5542bedc65207d4a74e636d68174979a3dc8R157-R164

crazy-max: > `driver.Inactive` just seems initial state so unrelated in here.

Yes indeed, just checking node is not in running state seems good.

commands/rm.go
```diff
@@ -100,3 +122,56 @@ func rm(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, keepSta
...
+	eg, _ := errgroup.WithContext(ctx)
+	for _, b := range builders {
+		func(b *nginfo) {
+			eg.Go(func() error {
+				if err := loadNodeGroupData(ctx, dockerCli, b); err != nil {
+					return errors.Wrapf(err, "cannot load %s", b.ng.Name)
+				}
+				if b.ng.Dynamic {
+					return nil
+				}
```
crazy-max, line 149: I skip node group in dynamic node. Let me know if it sgty @morlay @AkihiroSuda

commands/rm.go
```diff
@@ -72,11 +93,12 @@ func rmCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVar(&options.keepState, "keep-state", false, "Keep BuildKit state")
 	flags.BoolVar(&options.keepDaemon, "keep-daemon", false, "Keep the buildkitd daemon running")
+	flags.BoolVar(&options.allInactive, "all-inactive", false, "Remove all inactive builders")
```
errordeveloper, line 96: I think adding `--force` to remove without prompt will make this inline with other commands that have an interactive prompt (like `docker system prune`) .

commands/rm.go
tonistiigi: Using `prune` for something other than build cache is quite confusing imho. Especially because it means completely different thing atm. For the possibility of cache subcommand we can discuss in another issue.

crazy-max: > Using `prune` for something other than build cache is quite confusing imho.

Yes can be confusing.

> For the possibility of cache subcommand we can discuss in another issue.

:+1:

crazy-max:
PTAL @tonistiigi @errordeveloper

commands/rm.go
```diff
@@ -62,7 +81,13 @@ func rmCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 		Args:  cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.builder = rootOpts.builder
+			if !options.allInactive && options.force {
+				return errors.New("cannot specify --force without --all-inactive")
```
tonistiigi, line 85: nit: Not sure if this error is needed as it doesn't really matter for the user's intent. I guess the current behavior is that single `rm` is "always forced". Maybe that would change in the future.

commands/rm.go
```diff
@@ -100,3 +127,56 @@ func rm(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, keepSta
...
+					return errors.Wrapf(err, "cannot load %s", b.ng.Name)
+				}
+				if b.ng.Dynamic {
+					return nil
+				}
+				var kn []store.Node
+				for idx, n := range b.ng.Nodes {
+					d := b.drivers[idx]
+					if d.info == nil || d.info.Status != driver.Running {
+						_ = d.di.Driver.Rm(ctx, true, !in.keepState, !in.keepDaemon)
```
tonistiigi, line 159: Could we still show these errors? They should not fail the removal of the builder but for the user these would probably be the most useful.

commands/rm.go
```diff
@@ -100,3 +127,56 @@ func rm(ctx context.Context, dockerCli command.Cli, ng *store.NodeGroup, keepSta
...
+						continue
+					}
+					kn = append(kn, n)
+				}
+				b.ng.Nodes = kn
```
tonistiigi, line 164: I don't think this is the behavior we want. We shouldn't reconfigure the builders. If all the nodes are inactive then remove the builder. If at least one of them is running then skip it.

commands/rm.go
crazy-max: > nit: Not sure if this error is needed as it doesn't really matter for the user's intent.

to be fair I was not sure about that one too. will remove this cond.

> I guess the current behavior is that single `rm` is "always forced".

yes that's the current behavior.

commands/rm.go
crazy-max: sure sgtm

commands/rm.go
crazy-max: ok I was not thinking like this for `rm --all-inactive`. for me it removes inactives builders and nodes so if a node is inactive in a builder it should be removed from this builder. This way we we only have active builders and nodes in our stack.

```
$ docker buildx ls
NAME/NODE     DRIVER/ENDPOINT                        STATUS   PLATFORMS
builder       docker-container
  builder0    unix:///var/run/docker.sock            inactive
  mac-mini-m1 tcp://sifive.home.com:2375      running  linux/rsicv64*, linux/riscv64
  sifive      tcp://mac-mini-m1.home.pl4gue.com:2376 Cannot connect to the Docker daemon at tcp://mac-mini-m1.home.com:2376. Is the docker daemon running?
builder2      docker-container
  builder20   unix:///var/run/docker.sock            inactive
desktop-linux                                        protocol not available
default *     docker
  default     default                                running                linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```

here `builder0` and `sifive` will never be removed because `mac-mini-m1` is active. I think the user would want `builder0` and `sifive` removed because they are inactive. maybe we could allow that if `--force` is being used?

tonistiigi: We have `--leave` for that use case where you want to reconfigure builders. The use case for `rm --all-inactive` was that there are cases where you have too much stuff under `buildx ls` and want to do some housekeeping. Now you would need to write some loop that repeatedly calls `buildx rm <buildername>`. In practice it is even harder as you can't determine the ones you want to keep. So in reality it is a bunch of copy-paste. `--all-inactive` helps you with this loop and triggers `rm` for all filtered builders as if you invoked `rm` for all of them separately.

Removing a node from a builder doesn't make sense in here as this is not what `rm` does. Your case would be more like `--leave-all-inactive` but I don't think that case is something that users hit nearly as often and you will also not have tens of inactive nodes under a builder where you would need a loop to clear them.

crazy-max: Ah yes completely forgot about the `--leave` flag!

tonistiigi (approved):

PR #887, opened 2021-12-15
docker-container: set UsernsMode only when needed

### LINKED ISSUE #561: Rootless Docker support (`container init caused: write sysctl key net.ipv4.ping_group_range: write /proc/sys/net/ipv4/ping_group_range: invalid argument: unknown`)
Is there support for using buildx command with a rootless docker daemon? I tried the default buildkit image and the rootless one but both raises this error while booting up buildkit:
```
error: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: write sysctl key net.ipv4.ping_group_range: write /proc/sys/net/ipv4/ping_group_range: invalid argument: unknown
```
As a workaround, I found out that manually running buildkit with rootless [config](https://github.com/moby/buildkit/blob/master/docs/rootless.md) worked

```sh
docker run --name buildx_buildkit_builder0 -d --security-opt seccomp=unconfined --security-opt apparmor=unconfined --device /dev/fuse moby/buildkit:rootless --oci-worker-no-process-sandbox
```

You just gotta make sure that the name of the container matches the name of the buildx builder. In this case, my builder was named "builder" so the buildkit container had to be named "buildx_buildkit_builder0"

AkihiroSuda:
Set `UsernsMode="host"` only when the daemon is running in userns-remapping mode.

Fix #561

The issue will be also fixed in moby/moby#43084 (Docker 20.10.13).
This buildx PR helps users of old releases of Docker.

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -37,6 +37,7 @@ const (
 type Driver struct {
 	driver.InitConfig
 	factory      driver.Factory
+	userNSRemap  bool // true if dockerd is running with userns-remap mode
 	netMode      string
 	image        string
 	cgroupParent string
@@ -112,7 +113,6 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 	if err := l.Wrap("creating container "+d.Name, func() error {
 		hc := &container.HostConfig{
 			Privileged: true,
-			UsernsMode: "host",
 			Mounts: []mount.Mount{
 				{
 					Type:   mount.TypeVolume,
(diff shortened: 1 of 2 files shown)


aarongable:
I locally rebuilt the buildx plugin with this commit applied and confirmed that this resolves the issue for me.

tonistiigi (approved):

jedevc:
@AkihiroSuda could we move this logic into the driver instead of having it in the factory, so that we check the property when creating the container? I'm happy to do it, just wanted to check if there was a reason it was done this way?

With #1206, if the Docker socket isn't available then the factory will fail, so we can't create new docker-container builders as in #1358 (ideally, this would only fail if we passed `--bootstrap` as well).

AkihiroSuda:
@jedevc SGTM 👌

PR #889, opened 2021-12-16
vendor: update buildkit to 539be170

tonistiigi:

Issue #850, opened 2021-11-18
Recent Change to Dockerignore Parsing


SlvrEagle23:
One of my GitHub Actions involves a build which uses a `.dockerignore` file that basically looks like this:

```
**
!web
!util/docker/web
!composer.json
```

The intent is to automatically ignore anything that isn't expressly specified later in the file.

Up until today, on my GitHub Actions, this worked without any issue. Today, however, I started getting failed builds:

```
#7 [2/8] COPY ./util/docker/web /bd_build/
#7 DONE 0.4s

#8 [3/8] RUN chmod a+x /bd_build/*.sh     && /bd_build/prepare.sh     && /bd_build/add_user.sh     && /bd_build/setup.sh     && /bd_build/cleanup.sh     && rm -rf /bd_build
#8 0.070 chmod: cannot access '/bd_build/*.sh': No such file or directory
```

Removing the `.dockerignore` file caused the problem to be resolved and the build to continue as normal. Something has changed in the last 24 hours with some part of the stack that enables building Docker images via Buildkit on GitHub Actions that has affected how the `.dockerignore` file is parsed.

My GitHub Actions report that my current version of buildkit is `moby/buildkit:buildx-stable-1 => buildkitd github.com/moby/buildkit v0.9.3 8d2625494a6a3d413e3d875a2ff7dd9b1ed1b1a9`

slavovojacek:
Had the same issue, everything good locally but GH actions were failing for us. Figured it was `.dockerignore` related, we did the below to fix.

```diff
  **/*
  !*.json
- !prisma
- !src
+ !prisma/**
+ !src/**
```

tonistiigi:
Could either of you post a full example?

crazy-max:
@SlvrEagle23 If you have a link to your repo that would be great!

tonistiigi:
@aaronlehmann This looks to be related to your `*` updates in patternmatcher. The vendor of buildkit&moby seems to be regression point https://github.com/docker/buildx/pull/775

aaronlehmann:
This is related to https://github.com/moby/moby/pull/42676

@tonistiigi: I want to confirm the expected behavior here. Given the `.dockerignore file`:

```
**
!util/docker/web
```

It sounds like we expect a file `util/docker/web/foo` *not* to be ignored because of the exclusion.

What if `.dockerignore` was this instead?

```
**
!util/docker/web
util/docker/web/foo
```

[...]

PR #894, opened 2021-12-22
[v0.7] revert dockerignore changes

tonistiigi:
Reverts
- https://github.com/tonistiigi/fsutil/commit/da5201e0b3af118e109157de2b76c0de57a6fff2
- https://github.com/tonistiigi/fsutil/commit/9d0ab16e32163800c6b463804cc217946e78a817

@aaronlehmann

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,7 +58,7 @@ require (
 replace (
 	github.com/docker/cli => github.com/docker/cli v20.10.3-0.20210702143511-f782d1355eff+incompatible
 	github.com/docker/docker => github.com/tonistiigi/docker v0.10.1-0.20211122204227-65a6f25dbca2
-	github.com/tonistiigi/fsutil => github.com/tonistiigi/fsutil v0.0.0-20211122210416-da5201e0b3af
+	github.com/tonistiigi/fsutil => github.com/tonistiigi/fsutil v0.0.0-20211222194517-22a581344001
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc => github.com/tonistiigi/opentelemetry-go-contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.0.0-20210714055410-d010b05b4939
 	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace => github.com/tonistiigi/opentelemetry-go-contrib/instrumentation/net/http/httptrace/otelhttptrace v0.0.0-20210714055410-d010b05b4939
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => github.com/tonistiigi/opentelemetry-go-contrib/instrumentation/net/http/otelhttp v0.0.0-20210714055410-d010b05b4939
--- a/go.sum
+++ b/go.sum
@@ -1201,8 +1201,8 @@ github.com/tommy-muehle/go-mnd v1.1.1/go.mod h1:dSUh0FtTP8VhvkL1S+gUR1OKd9ZnSaoz
 github.com/tommy-muehle/go-mnd v1.3.1-0.20200224220436-e6f9a994e8fa/go.mod h1:dSUh0FtTP8VhvkL1S+gUR1OKd9ZnSaozuI6r3m6wOig=
 github.com/tonistiigi/docker v0.10.1-0.20211122204227-65a6f25dbca2 h1:e6PyP4Gi3HZ9hSH9gGGGSaBh3EI+HZj5E4fgEV43ZCI=
 github.com/tonistiigi/docker v0.10.1-0.20211122204227-65a6f25dbca2/go.mod h1:v9W/4hjeg+54O4ffkt1Dnh7nOIE9uHi9F0W9clN3nTQ=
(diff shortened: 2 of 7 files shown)


aaronlehmann:
LGTM

crazy-max (commented):
I'm still reproducing #850: https://github.com/crazy-max/buildx-dockerignore-issue/runs/4607531429?check_suite_focus=true#step:4:136

aaronlehmann:
I think this revert was meant for #883 (but maybe I misunderstood).

#850 might need the changes in moby/moby#43037 and moby/moby#43047. Even though this code is no longer being used by fsutil with the reverts here, there is code in `pkg/archive/archive.go` using it - perhaps that's involved?

Alternatively, moby/moby#42676 could be reverted, but this would regress moby/moby#41433.

tonistiigi:
> Alternatively, moby/moby#42676 could be reverted

Then we would need to also revert https://github.com/tonistiigi/fsutil/pull/107 ? Is there a simpler fix that just doesn't call the new methods that it adds?

tonistiigi:
>  there is code in pkg/archive/archive.go using it - perhaps that's involved?

I can't see how this might be related. We don't create tarballs anymore for sending local files.

crazy-max:
@aaronlehmann 

> I think this revert was meant for #883 (but maybe I misunderstood).

Yes indeed, I have tested your PR(s):

* https://github.com/tonistiigi/fsutil/pull/121
* https://github.com/moby/moby/pull/43095

And seems to work fine for #850 issue: https://github.com/crazy-max/buildx/commit/57678ea1d06129b045064ad70bc85a6b9680221a

Repro: https://github.com/crazy-max/buildx-dockerignore-issue/runs/4609519849?check_suite_focus=true#step:4:136

Do you have a simple repro for #883 I could use in https://github.com/crazy-max/buildx-dockerignore-issue?

aaronlehmann:
> Then we would need to also revert tonistiigi/fsutil#107 ? Is there a simpler fix that just doesn't call the new methods that it adds?

I wasn't sure if the changes to `builder/remotecontext/detect.go` and `pkg/archive/archive.go` were relevant in this context. If they are not, then just reverting the changes in the walker code from tonistiigi/fsutil#107 should be enough. We should either keep the `copy` code as is, or also revert moby/buildkit#2319 (contenthash and copy need to stay consistent).

> Yes indeed, I have tested your PR(s):
> tonistiigi/fsutil#121
> moby/moby#43095
> And seems to work fine for #850 issue: crazy-max@57678ea

#850 has been overloaded to include multiple issues. Are you testing the original issue, or the one about file permissions? The fsutil PR you referenced is about the permissions thing (and the moby PR is an optimization that should not have any visible impact).

tonistiigi:
>  Is there a simpler fix that just doesn't call the new methods that it adds?

Another approach might be to try to rework https://github.com/tonistiigi/fsutil/commit/da5201e0b3af118e109157de2b76c0de57a6fff2 without https://github.com/tonistiigi/fsutil/commit/9d0ab16e32163800c6b463804cc217946e78a817 . Atm it conflicts a lot so I needed to revert both but not sure if the logic of the fix actually depends on the previous commit.

> Are you testing the original issue, or the one about file permissions?

Both. Current issue seems to be with the original report. The goal would be to have no regressions between v0.6 and v0.7. For this one I believe we fixed it in v0.7.1 but these reverts are now reverting the v0.7.1 fix as well.

aaronlehmann:
> Another approach might be to try to rework tonistiigi/fsutil@da5201e without tonistiigi/fsutil@9d0ab16 

I suppose, but that feels like relatively high effort/high risk. We would end up with different code in the release branch. I prefer to either revert back to a known good fsutil version (not sure if I'm missing something about why we can't do this), or fix forward with moby/moby#43037 + moby/moby#43047 + tonistiigi/fsutil#114 + tonistiigi/fsutil#120 + moby/buildkit#2486.

tonistiigi:
> known good fsutil version (not sure if I'm missing something about why we can't do this),

What is this version? Based on the previous comment it seems that we would need to split tonistiigi/fsutil#107

aaronlehmann:
Is the issue that you don't want to regress #41433? This was a longstanding issue and affected buildx < 0.7 AFAIK, but I can see that if it was fixed in 0.7, it would be unfortunate to reintroduce the bug.

You could revert tonistiigi/fsutil#107 and take only the first commit from moby/moby#42676. That first commit is a very simple fix for #41433, and the other commits are an optimization you asked for to create new functions that do the matching incrementally. The optimization is the cause of #850.

crazy-max (approved):
LGTM

tonistiigi:
ping @aaronlehmann

aaronlehmann:
This will regress moby/moby#41433, right? `(*PatternMatcher).Matches` [was not updated](https://github.com/moby/moby/pull/42676/files#r688179669) with the fix for moby/moby#41433 in the final version of the PR. If you want to keep the fix for moby/moby#41433, we need to take the first commit from https://github.com/moby/moby/pull/42676, which **does** update `Matches`.

If you don't care about removing the fix for moby/moby#41433, this looks okay to me.

crazy-max:
Cannot repro #850 and #883 since #889. I think we can close this one.

PR #909, opened 2022-01-09
feat: add env driver

### LINKED ISSUE #23: add "env" driver that connects to $BUILDKIT_HOST
This driver allows creating a buildx instance directly pointing to the existing buildkit endpoint.

`docker buildx create --driver=env unix:///var/lib/buildkitd.sock`
`docker buildx create --driver=env` uses `BUILDKIT_HOST` env value with UNIX sock default.

I guess TLS info is just passed with a custom driver-opt:

```
docker buildx create --driver=env --driver-opt ca=mycafile.pem https://foobar
```

`rm/stop` commands are no-op in this driver.

A better name could be "remote".

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Fixes #23 

cc: @AkihiroSuda @tonistiigi

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"strings"
 
+	buildkitclient "github.com/moby/buildkit/client"
+
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/driver"
 	ctxkube "github.com/docker/buildx/driver/kubernetes/context"
@@ -80,6 +82,13 @@ func driversForNodeGroup(ctx context.Context, dockerCli command.Cli, ng *store.N
 				defer func() {
 					dis[i] = di
 				}()
+
+				buildkitAPI, err := buildkitclient.New(ctx, n.Endpoint)
+				if err != nil {
+					di.Err = err
+					return nil
+				}
+
 				dockerapi, err := clientForEndpoint(dockerCli, n.Endpoint)
 				if err != nil {
 					di.Err = err
@@ -118,7 +127,7 @@ func driversForNodeGroup(ctx context.Context, dockerCli command.Cli, ng *store.N
 					}
 				}
 
-				d, err := driver.GetDriver(ctx, "buildx_buildkit_"+n.Name, f, dockerapi, imageopt.Auth, kcc, n.Flags, n.Files, n.DriverOpts, n.Platforms, contextPathHash)
+				d, err := driver.GetDriver(ctx, "buildx_buildkit_"+n.Name, f, dockerapi, n.Endpoint, buildkitAPI, imageopt.Auth, kcc, n.Flags, n.Files, n.DriverOpts, n.Platforms, contextPathHash, ng.Driver)
 				if err != nil {
 					di.Err = err
 					return nil
@@ -259,7 +268,7 @@ func getDefaultDrivers(ctx context.Context, dockerCli command.Cli, defaultOnly b
 		return nil, err
 	}
 
-	d, err := driver.GetDriver(ctx, "buildx_buildkit_default", nil, dockerCli.Client(), imageopt.Auth, nil, nil, nil, nil, nil, contextPathHash)
+	d, err := driver.GetDriver(ctx, "buildx_buildkit_default", nil, dockerCli.Client(), "", nil, imageopt.Auth, nil, nil, nil, nil, nil, contextPathHash, "")
 	if err != nil {
 		return nil, err
 	}
--- a/driver/env/driver.go
+++ b/driver/env/driver.go
@@ -1,0 +1,70 @@
+package env
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/docker/buildx/driver"
+	"github.com/docker/buildx/util/progress"
+	"github.com/moby/buildkit/client"
+	buildkitclient "github.com/moby/buildkit/client"
+	"github.com/pkg/errors"
+)
+
+type Driver struct {
+	factory driver.Factory
+	driver.InitConfig
+	BuldkitdAddr string
+	BuildkitAPI  *buildkitclient.Client
+}
+
+func (d *Driver) Bootstrap(ctx context.Context, l progress.Logger) error {
+	return nil
+}
+
+func (d *Driver) Info(ctx context.Context) (*driver.Info, error) {
+	if d.BuldkitdAddr == "" && d.Driver == "env" {
+		return nil, errors.Errorf("buldkitd addr must not be empty")
+	}
+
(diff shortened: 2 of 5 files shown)


commands/util.go
```diff
@@ -117,7 +118,13 @@ func driversForNodeGroup(ctx context.Context, dockerCli command.Cli, ng *store.N
...
 				}
 
-				d, err := driver.GetDriver(ctx, "buildx_buildkit_"+n.Name, f, dockerapi, imageopt.Auth, kcc, n.Flags, n.Files, n.DriverOpts, n.Platforms, contextPathHash)
+				buildkitAPI, err := buildkitclient.New(ctx, endpoint)
+				if err != nil {
+					di.Err = err
+					return nil
+				}
+
+				d, err := driver.GetDriver(ctx, "buildx_buildkit_"+n.Name, f, dockerapi, buildkitAPI, imageopt.Auth, kcc, n.Flags, n.Files, n.DriverOpts, n.Platforms, contextPathHash, endpoint)
```
errordeveloper, line 127: I think this functions needs some refactoring, the list of args is becoming very hard to read...

driver/manager.go
```diff
@@ -51,6 +52,8 @@ type InitConfig struct {
 	// This object needs updates to be generic for different drivers
 	Name             string
 	DockerAPI        dockerclient.APIClient
+	BuildkitAPI      *buildkitclient.Client
+	BuldkitdAddr     string
```
errordeveloper, line 56: Is there a way to get the address from the client?

commands/create.go
```diff
@@ -137,7 +137,8 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
 		if len(args) > 0 {
 			ep, err = validateEndpoint(dockerCli, args[0])
 			if err != nil {
-				return err
+				logrus.Warnf("No valid Docker endpoint found, will continue with buildkitd endpoint.")
```
errordeveloper, line 140: I'm not sure about this one, but I'm not yet familiar with the argument handling logic, perhaps something to look into.

commands/util.go
```diff
@@ -231,7 +238,7 @@ func getInstanceByName(ctx context.Context, dockerCli command.Cli, instance, con
 	if err != nil {
 		return nil, err
 	}
-	return driversForNodeGroup(ctx, dockerCli, ng, contextPathHash)
+	return driversForNodeGroup(ctx, dockerCli, ng, contextPathHash, ng.Nodes[0].Endpoint)
```
errordeveloper, line 241: This probably needs extra care, does `driversForNodeGroup` really need to take a concrete endpoint?

driver/manager.go
```diff
@@ -51,6 +52,8 @@ type InitConfig struct {
 	// This object needs updates to be generic for different drivers
 	Name             string
 	DockerAPI        dockerclient.APIClient
+	BuildkitAPI      *buildkitclient.Client
```
tonistiigi, line 55: This is very driver-specific and shouldn't be here. Only the remote driver itself should return the client. For the address, not exactly sure why it is here. Maybe we need a generic "endpoint" field in here that is buildkit address for remote driver and docker address for container one.

commands/create.go
```diff
@@ -137,7 +137,8 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
 		if len(args) > 0 {
 			ep, err = validateEndpoint(dockerCli, args[0])
 			if err != nil {
```
tonistiigi, line 139: If this is different things now for different drivers we should just move the validation to the driver level.

driver/remote/driver.go
```diff
@@ -0,0 +1,61 @@
...
+func (d *Driver) Features() map[driver.Feature]bool {
+	return map[driver.Feature]bool{}
+}
+
+func (d *Driver) Factory() driver.Factory {
+	return d.factory
+}
+
+func (d *Driver) IsMobyDriver() bool {
+	return true
```
tonistiigi, line 56: `false`

commands/util.go
```diff
@@ -80,6 +81,13 @@ func driversForNodeGroup(ctx context.Context, dockerCli command.Cli, ng *store.N
 				defer func() {
 					dis[i] = di
 				}()
+
+				buildkitAPI, err := buildkitclient.New(ctx, n.Endpoint)
```
Dentrax, line 85: i think we're initializing both `buildkitAPI` and `dockerapi` here if we set `--driver=env`. So `dockerapi` initializing unnecessarily. But couldn't find a better way.

developer-guy:
cc: @crazy-max

driver/env/driver.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (d *Driver) Info(ctx context.Context) (*driver.Info, error) {
+	if d.BuldkitdAddr == "" && d.Driver == "env" {
+		return nil, errors.Errorf("buldkitd addr must not be empty")
+	}
+
+	c, err := client.New(ctx, d.BuldkitdAddr)
+	if err != nil {
+		return nil, errors.Wrapf(driver.ErrNotConnecting, err.Error())
```
AkihiroSuda, line 32: `fmt.Errorf` with `%w`

driver/env/driver.go
```diff
@@ -0,0 +1,70 @@
...
+func (d *Driver) Rm(ctx context.Context, force, rmVolume, rmDaemon bool) error {
+	return fmt.Errorf("rm command is not implemented for this driver")
+}
+
+func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
+	return client.New(ctx, d.BuldkitdAddr, client.WithSessionDialer(d.BuildkitAPI.Dialer()))
+}
+
+func (d *Driver) Features() map[driver.Feature]bool {
+	return map[driver.Feature]bool{}
```
AkihiroSuda, line 57: Why empty?

driver/env/driver.go
tonistiigi: `fmt.Errorf` is not allowed because it doesn't store stacktrace

commands/util.go
tonistiigi: This is definitely not allowed. Only the `driver.Client()` is allowed to call this. `dockerapi` initialization is for the features like `--load` that move things to Docker.

 Another difference is that `clientForEndpoint` is just a struct setup, while `buildkitclient.New()` actually dials out to the endpoint.

jedevc:
@developer-guy have you seen #1078? I've tried to finish off the work you started, do let me know what you think :smile:

tonistiigi:
Merged in #1078

PR #913, opened 2022-01-11
Add simple function to setup default driver

errordeveloper:
This would help with https://github.com/docker/compose/pull/9099

### DIFF
--- a/driver/manager.go
+++ b/driver/manager.go
@@ -2,6 +2,7 @@ package driver
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


tonistiigi (commented):
I don't think this is correct. "default" is not something defined by `driver` package but handled by the store resolution in command level. Atm you are finding the default factory(nothing to do with builder with the name "default") based on current environment(eg. in some cases it could be container/k8s factory) and then try to get a named builder "default" from it. Such builder will not exist as "default" is a reserved name https://github.com/docker/buildx/blob/master/commands/create.go#L44-L46 and builder with that name is not allowed.

If I understand the intent of this code then you want to always get the "docker" factory and then initialize its driver. Docker driver does not use the name property. `ContextPathHash` is likely also unused(I think it is only for k8s).

errordeveloper:
thanks @tonistiigi! I will have a better look.

errordeveloper:
@tonistiigi I reworked this, please take another look :)

driver/manager.go
```diff
@@ -140,6 +141,18 @@ func GetFactories() []Factory {
 	return ds
 }
 
+func GetDefaultDockerDriver(ctx context.Context, api dockerclient.APIClient, auth Auth, contextPathHash string) (Driver, error) {
```
tonistiigi, line 144: `contextPathHash` is unused

driver/manager.go
```diff
@@ -140,6 +141,18 @@ func GetFactories() []Factory {
...
+	f := GetFactory("docker", false)
+	if f == nil {
+		return nil, fmt.Errorf("unable to get default builder")
```
tonistiigi, line 147: `s/default/docker/`

tonistiigi, line 147: use `errors.Errorf()` errors for stack capture

tonistiigi (commented):
I'm not sure how much value this function provides if it just calls 2 public functions internally and this function is itself is not called anywhere in buildx itself. Maybe it should at least be in the `drivers/docker` package? Remove "Docker" from the function name then.

PR #914, opened 2022-01-11
bake: keep target inheritance

### LINKED ISSUE #912: nested `inherits` only affects first target
it appears that `inherits` cannot be nested, or, (for me: worse), it only affects the *first* target.

For example `docker buildx bake --print --file FILE` on:

```hcl
target "grandparent" {
  output = ["type=docker"]
}

target "parent" {
  inherits = ["grandparent"]
  args = {
    FOO = "bar"
  }
}

target "child 1" {
  inherits = ["parent"]
}

target "child 2" {
  inherits = ["parent"]
}

group "default" {
  targets = [
    "child 1",
    "child 2"
  ]
} 
```

```
{
  "group": {
    "default": {
      "targets": [
        "child 1",
        "child 2"
      ]
    }
  },
  "target": {
    "child 1": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "FOO": "bar"
      },
      "output": [
        "type=docker"
      ]
    },
    "child 2": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "FOO": "bar"
      }
    }
  }
}
```

My expectation would have been that either:

- **both** child 1 and 2 inherit `output`
- **neither** inherit `output` and the parser tells me that I can't nest `inherits`

Took me a while to figure this out 🙃

(Amazing work on bake -- love it ❤️)

crazy-max:
fixes #912 

After first pass inheritance is lost for the next targets in a group.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -389,10 +389,11 @@ func (c Config) group(name string, visited map[string]struct{}) []string {
 }
 
 func (c Config) ResolveTarget(name string, overrides map[string]map[string]Override) (*Target, error) {
-	t, err := c.target(name, map[string]struct{}{}, overrides)
+	t, err := c.target(name, map[string]*Target{}, overrides)
 	if err != nil {
 		return nil, err
 	}
+	t.Inherits = nil
 	if t.Context == nil {
 		s := "."
 		t.Context = &s
@@ -404,11 +405,11 @@ func (c Config) ResolveTarget(name string, overrides map[string]map[string]Overr
 	return t, nil
 }
 
-func (c Config) target(name string, visited map[string]struct{}, overrides map[string]map[string]Override) (*Target, error) {
-	if _, ok := visited[name]; ok {
-		return nil, nil
+func (c Config) target(name string, visited map[string]*Target, overrides map[string]map[string]Override) (*Target, error) {
+	if t, ok := visited[name]; ok {
+		return t, nil
 	}
-	visited[name] = struct{}{}
+	visited[name] = nil
 	var t *Target
 	for _, target := range c.Targets {
 		if target.Name == name {
@@ -429,7 +430,6 @@ func (c Config) target(name string, visited map[string]struct{}, overrides map[s
 			tt.Merge(t)
 		}
 	}
-	t.Inherits = nil
 	m := defaultTarget()
 	m.Merge(tt)
 	m.Merge(t)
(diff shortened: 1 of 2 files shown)


bake/bake.go
```diff
@@ -421,15 +422,18 @@ func (c Config) target(name string, visited map[string]struct{}, overrides map[s
 	}
 	tt := &Target{}
 	for _, name := range t.Inherits {
-		t, err := c.target(name, visited, overrides)
+		if _, ok := inherited[name]; ok {
+			continue
+		}
+		inherited[name] = struct{}{}
+		t, err := c.target(name, visited, inherited, overrides)
```
tonistiigi, line 429: what case is it where passing `inherited` makes sense in here?

Wondering something like:
```
a: foo=123, bar=234
b: inherits:a, bar=567
c: inherits:a, baz=890
d: inherits:b,c
```

So what is the expected value for `bar`. I think it should be `234` as `c` is inherited after `b` and should overwrite its properties.

crazy-max: this was in my initial commit and forgot to remove this logic it afterwards. should be ok now.

crazy-max: > So what is the expected value for `d->bar`. I think it should be `234` as `c` is inherited after `b` and should overwrite its properties.

And yes should be `234`. I will add another test for this one.

bake/bake.go
```diff
@@ -404,11 +405,13 @@ func (c Config) ResolveTarget(name string, overrides map[string]map[string]Overr
 	return t, nil
 }
 
-func (c Config) target(name string, visited map[string]struct{}, overrides map[string]map[string]Override) (*Target, error) {
+func (c Config) target(name string, visited map[string]struct{}, inherit bool, overrides map[string]map[string]Override) (*Target, error) {
```
errordeveloper, line 408:
```suggestion
func (c Config) target(name string, visited map[string]struct{}, inherited bool, overrides map[string]map[string]Override) (*Target, error) {
```

errordeveloper (approved):
just one minor nit, I think it'd help to improve readability :)

bake/bake.go
```diff
@@ -404,11 +405,10 @@ func (c Config) ResolveTarget(name string, overrides map[string]map[string]Overr
 	return t, nil
 }
 
-func (c Config) target(name string, visited map[string]struct{}, overrides map[string]map[string]Override) (*Target, error) {
-	if _, ok := visited[name]; ok {
-		return nil, nil
+func (c Config) target(name string, visited map[string]*Target, overrides map[string]map[string]Override) (*Target, error) {
+	if t, ok := visited[name]; ok {
+		return t, nil
 	}
```
tonistiigi, line 411: Need to block the key with `nil` value here. Otherwise, it goes to an infinite loop.

crazy-max: Good catch yes

tonistiigi (approved):

PR #915, opened 2022-01-11
ci: build with stable buildkit image

crazy-max:
follow-up https://github.com/moby/buildkit/pull/2491#issuecomment-1010123172

Our actual pipeline is using a bleeding edge release of the BuildKit image which might introduce bugs unrelated to buildx like https://github.com/docker/buildx/runs/4777886890?check_suite_focus=true#step:7:135:

```
error: failed to solve: process "/bin/sh -c set -x; xx-go build -ldflags \"$(cat /tmp/.ldflags) ${LDFLAGS}\" -o /usr/bin/buildx ./cmd/buildx &&   xx-verify --static /usr/bin/buildx" did not complete successfully: failed to mount /tmp/buildkit-mount2802602399: [{Type:overlay Source:overlay Options:[index=off workdir=/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/24/work upperdir=/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/24/fs lowerdir=/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/15/fs:/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/14/fs:/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/13/fs:/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/12/fs:/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/11/fs:/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/10/fs:/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/9/fs:/var/lib/buildkit/runc-overlayfs/snapshots/snapshots/6/fs redirect_dir=off]}]: no such file or directory
```

This PR removes it for the main build job but also adds a small job to keep using it to detect this kind of behavior in the future.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,6 @@ on:
 
 env:
   REPO_SLUG: "docker/buildx-bin"
-  REPO_SLUG_ORIGIN: "moby/buildkit:master"
   RELEASE_OUT: "./release-out"
 
 jobs:
@@ -31,8 +30,6 @@ jobs:
       -
         name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
-        with:
-          driver-opts: image=${{ env.REPO_SLUG_ORIGIN }}
       -
         name: Test
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -90,3 +87,24 @@ jobs:
...
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          driver-opts: image=moby/buildkit:master
+          buildkitd-flags: --debug
+      -
+        name: Build
+        run: |
+          make release
```
tonistiigi, line 110: This doesn't conflict with the artifact creation etc?

crazy-max: No it just runs the build without any release

tonistiigi: Ok, so it builds all the release artifacts but it doesn't upload them as there is no extra step in this workflow. Maybe add a comment for it as it is a bit confusing. Otherwise, we could just run some equivalent bake command directly as well.

crazy-max: Done

tonistiigi (approved):

PR #924, opened 2022-01-21
root: filter out useless commandConn.CloseWrite warning message

crazy-max:
relates to https://github.com/docker/buildx/issues/764#issuecomment-961198793
closes #813 

also while doing further tests for https://github.com/docker/cli/pull/3314 I encounter a case where some logs don't match the output format from docker/cli:

```shell
$ docker build ...
WARNING: Error loading config file: ...
WARN[0000] No output specified for docker-container driver. Build result will only remain in ...
```

now using cli writers for logrus and simplify output format for parity:

```shell
$ docker build ...
WARNING: Error loading config file: ...
WARNING: No output specified for docker-container driver. Build result will only remain in ...
```

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -28,10 +28,22 @@ func NewRootCmd(name string, isPlugin bool, dockerCli command.Cli) *cobra.Comman
 		}
 	}
 
-	logrus.AddHook(logutil.NewFilter(
+	logrus.AddHook(logutil.NewFilter([]logrus.Level{
+		logrus.DebugLevel,
+	},
 		"serving grpc connection",
 		"stopping session",
 		"using default config store",
+	))
+
+	// filter out useless commandConn.CloseWrite warning message that can occur
+	// when listing builder instances with "buildx ls" for those that are
+	// unreachable: "commandConn.CloseWrite: commandconn: failed to wait: signal: killed"
+	// https://github.com/docker/cli/blob/3fb4fb83dfb5db0c0753a8316f21aea54dab32c5/cli/connhelper/commandconn/commandconn.go#L203-L214
+	logrus.AddHook(logutil.NewFilter([]logrus.Level{
+		logrus.WarnLevel,
+	},
+		"commandConn.CloseWrite:",
 	))
 
 	addCommands(cmd, dockerCli)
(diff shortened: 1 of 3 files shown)


commands/root.go
```diff
@@ -28,12 +29,45 @@ func NewRootCmd(name string, isPlugin bool, dockerCli command.Cli) *cobra.Comman
 		}
 	}
 
-	logrus.AddHook(logutil.NewFilter(
+	logrus.SetFormatter(&logutil.Formatter{})
```
tonistiigi, line 32: Why are the lines 32-51 needed?

crazy-max:
```go
logrus.SetFormatter(&logutil.Formatter{})
```

Logs don't match the output format from docker/cli, it feels weird:

```shell
$ docker build ...
WARNING: Error loading config file: ...
WARN[0000] No output specified for docker-container driver. Build result will only remain in ...
```

So use a simple formatter for logrus for parity:

```shell
$ docker build ...
WARNING: Error loading config file: ...
WARNING: No output specified for docker-container driver. Build result will only remain in ...
```

Ofc we could remove all `logrus.Warn|Debug|...` in our code and replace them with a simple printf but it's easier to use a formatter.

For lines 34-55 we are now using the docker/cli writers for logrus.

tonistiigi: > For lines 34-55 we are now using the docker/cli writers for logrus.

What's the benefit of that?

tonistiigi: For the warnings, there is also a case of the build warnings that have their own layout. Not strictly against the formatting change.

crazy-max: > What's the benefit of that?

Just to be aligned with the cli. If you feel it's not worth it I can remove this commit.

> For the warnings, there is also a case of the build warnings that have their own layout. Not strictly against the formatting change.

I think it's ok as you're not using the logrus formatter right?: https://github.com/docker/buildx/blob/11b771c789daa8d6c9bc3cf7ba7c86240f382bef/commands/build.go#L268-L306

tonistiigi: > I can remove this commit.

I think so if it doesn't change anything.

> I think it's ok as you're not using the logrus formatter right?:

Yes. Just I can see it being confusing for the user.

crazy-max: Done

tonistiigi (approved):

PR #925, opened 2022-01-23
Prevent crash when Exporter is not TracerDelegate

abergmeier:
There is no enforcing Exporter + TracerDelegate via type system nor by rigorous runtime checking.
Thus the code previously crashed since `withTracerDelegate` (nor any callers) check whether embedded `TracerDelegate` is `nil`.
Now only passes `withTracerDelegate` if it can actually be called.

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -341,11 +341,17 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
 		return nil, err
 	}
 
-	td, _ := exp.(client.TracerDelegate)
+	clientOpts := []client.ClientOpt{
+		client.WithContextDialer(func(context.Context, string) (net.Conn, error) {
+			return conn, nil
+		}),
+	}
 
-	return client.New(ctx, "", client.WithContextDialer(func(context.Context, string) (net.Conn, error) {
-		return conn, nil
-	}), client.WithTracerDelegate(td))
+	if td, _ := exp.(client.TracerDelegate); td != nil {
+		clientOpts = append(clientOpts, client.WithTracerDelegate(td))
+	}
+
+	return client.New(ctx, "", clientOpts...)
 }
 
 func (d *Driver) Factory() driver.Factory {
--- a/driver/kubernetes/driver.go
+++ b/driver/kubernetes/driver.go
@@ -211,11 +211,17 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
 		return nil, err
 	}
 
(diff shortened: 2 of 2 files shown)


driver/docker-container/driver.go
```diff
@@ -341,11 +341,18 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
...
 	}
 
+	clientOpts := []client.ClientOpt{
+		client.WithContextDialer(func(context.Context, string) (net.Conn, error) {
+			return conn, nil
+		}),
+	}
+
 	td, _ := exp.(client.TracerDelegate)
+	if td != nil {
```
tonistiigi, line 351: This should use `if td, ok := ...; ok {` pattern

abergmeier: I assume you mean `if td, _ := ...; td {`?

tonistiigi: No, I think `ok` is correct here and preferred to check for type matches. It would help if you would show where the panic case is happening for you to be sure.

abergmeier: Maybe you can share your thought process which went into writing the previous code.

That being said the crash happens in autogenerated code.
So first it extracts the option here: https://github.com/docker/buildx/blob/9a1267cd021083874b02c228374f42da83151c54/vendor/github.com/moby/buildkit/client/client.go#L77

It then passes the reference down and crashes here: https://github.com/docker/buildx/blob/9a1267cd021083874b02c228374f42da83151c54/vendor/github.com/moby/buildkit/client/client.go#L167

since `TracerDelegate` field of `withTracerDelegate` is still set to `nil`.
So IMO constructing `withTracerDelegate` without a valid reference makes the code more opaque (and fails at runtime).

Leaving `ok` out of the current code seems the sensible thing since the path to avoid is the one with having `TracerDelegate` as `nil` and thus checking for `nil` is the most transparent.

tonistiigi: If in your case the `exp` is nil then just check that directly and skip the option. If for some reason you want to support `WithTracerDelegate(nil)` then it can have a hander inside the function (but I don't think that is actually needed).

Do you have a case where `exp` is not nil and type check will return `nil` with `ok`? I'm not sure how that would be possible.

abergmeier: It seems to me like we have a communication problem.
I don't seem to be bringing across my thought process and I don't really understand your response to my posts.
So for now I think I will fork buildx and maintain the patches myself.

abergmeier:
Postponing to a later time perhaps.

PR #928, opened 2022-01-25
bake: add named contexts keys

### LINKED ISSUE #926: bake: enable named contexts
#904 only enabled named contexts for build. Same should happen for bake. Also you should be able to use one target as context for another. https://github.com/docker/buildx/issues/447#issuecomment-732537029

tonistiigi:
fixes #926

Enable using named contexts in bake targets. The naming is consistent with build args, and `--build-context` and `--build-arg` are similar and both take key-value pairts.

~Still working on creating contexts from other targets that is a bit complex. That can be a follow-up PR.~ Added to same PR. Targets can be linked with `target:` prefix.

### DIFF
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -339,7 +339,7 @@ target "db" {
 
 Complete list of valid target fields:
 
-`args`, `cache-from`, `cache-to`, `context`, `dockerfile`, `inherits`, `labels`,
+`args`, `cache-from`, `cache-to`, `context`, `contexts`, `dockerfile`, `inherits`, `labels`,
 `no-cache`, `output`, `platform`, `pull`, `secrets`, `ssh`, `tags`, `target`
 
 ### Global scope attributes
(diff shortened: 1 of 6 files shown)


docs/reference/buildx_bake.md
```diff
@@ -339,7 +339,7 @@ target "db" {
 
 Complete list of valid target fields:
 
-`args`, `cache-from`, `cache-to`, `context`, `dockerfile`, `inherits`, `labels`,
```
crazy-max, line 342: Could we have an example with named `contexts` on this page? Also clarify difference with `context` would be sweet.

tonistiigi: I need examples for the follow-up target linking anyway. Maybe better to do them together. WIP in https://github.com/docker/buildx/compare/master...tonistiigi:bake-named-contexts-2?expand=1

crazy-max (approved):
Ok I played with your PR yesterday and I must say that's a nice way to handle target dependencies. Will be really useful.

Docs is fine too, thanks for this. Also our reference docs like the [bake page](https://github.com/tonistiigi/buildx/blob/bake-named-contexts/docs/reference/buildx_bake.md) starts to be huge. I will create a follow-up to see how we could improve that. We might need something like a cookbook to showcase real-world examples. Will see.

PR #929, opened 2022-01-25
bake: restrict target name

### LINKED ISSUE #888: Bake : Allowed "/" in target names collides with bake --push option
The docker-bake.hcl file allows target names to contain "/" (in my case the target name was the same as the tag I planned to use).
However, this breaks the --push option because it is a shorthand for ```--set=*.output=type=registry```, and * does not match slashes wich results on this error, which took me time to understand since I did not use the * syntax myself :

```
error: could not find any target matching '*'
````

A simple workaround for me was to prevent using slashes in my target name. 

However I think it would be more consistent to prevent using slashes in the target name right from the beginning, or to find a way to make it compatible with the push option.

### PR DESCRIPTION
**Pull request by crazy-max**
fixes #888

This fix adds a restriction `[a-zA-Z0-9_-]+`
for target name. This is pretty much the same as the
container name restriction in moby.

___

```yaml
# docker-compose.yml
services:
  a.bc:
    build:
      context: .
```

```shell
$ docker buildx bake --print
error: invalid service name "a.bc": only "[a-zA-Z0-9_-]+" are allowed
```

___

```hcl
// docker-bake.hcl
group "default" {
  targets = ["a.bc"]
}

target "a.bc" {
  context = "."
}
```

```shell
$ docker buildx bake --print
docker-bake.hcl:5
--------------------
   3 |     }
   4 |
   5 | >>> target "a.bc" {
   6 |       context = "."
   7 |     }
--------------------
error: docker-bake.hcl:5,8-14: Invalid name; only "[a-zA-Z0-9_-]+" are allowed
```

### DIFF
--- a/bake/hclparser/hclparser.go
+++ b/bake/hclparser/hclparser.go
@@ -16,8 +16,9 @@ import (
 )
 
 type Opt struct {
-	LookupVar func(string) (string, bool)
-	Vars      map[string]string
+	LookupVar     func(string) (string, bool)
+	Vars          map[string]string
+	ValidateLabel func(string) error
 }
 
 type variable struct {
@@ -262,6 +263,12 @@ func Parse(b hcl.Body, opt Opt, val interface{}) hcl.Diagnostics {
 		}
 	}
 
+	if opt.ValidateLabel == nil {
+		opt.ValidateLabel = func(string) error {
+			return nil
+		}
+	}
+
 	p := &parser{
 		opt: opt,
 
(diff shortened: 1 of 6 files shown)


bake/hclparser/hclparser.go
```diff
@@ -483,6 +500,13 @@ func Parse(b hcl.Body, opt Opt, val interface{}) hcl.Diagnostics {
 	return nil
 }
 
+func ValidateLabel(lbl string) error {
```
tonistiigi, line 503: This should not be public

tonistiigi (commented):
Does compose side do something similar already?

bake/hclparser/hclparser.go
crazy-max: I checked on compose-spec and compose-go repos and have not seen any kind of validators for a service name. It's public so I can use it for compose in `bake` pkg. Do you prefer a dedicated pkg for this?

tonistiigi: Ah, that is for compose. It's a bit weird as `hclparser` package is supposed to be generic and not even know about the current schema. But I guess you want to detect it here so that you can send the `hcl.Diagnostics` error. Maybe the validator should be passed with parse opts?

crazy-max: > But I guess you want to detect it here so that you can send the `hcl.Diagnostics` error.

Yes that's it

> Maybe the validator should be passed with parse opts?

Yeah sounds better like this. Let me know if changes lgty.

PR #937, opened 2022-02-05
Relax to io.Writer requirement for NewPrinter

hinshun:
- Wanted to just provide `io.Writer` in the plain log output case

### DIFF
--- a/util/progress/printer.go
+++ b/util/progress/printer.go
@@ -41,7 +41,7 @@ func (p *Printer) Warnings() []client.VertexWarning {
 	return p.warnings
 }
 
-func NewPrinter(ctx context.Context, out console.File, mode string) *Printer {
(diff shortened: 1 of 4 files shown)


util/progress/printer.go
```diff
@@ -41,7 +42,7 @@ func (p *Printer) Warnings() []client.VertexWarning {
 	return p.warnings
 }
 
-func NewPrinter(ctx context.Context, out console.File, mode string) *Printer {
+func NewPrinter(ctx context.Context, w io.Writer, mode string) (*Printer, error) {
```
tonistiigi, line 45: Would passing console as a second parameter be suitable here? I don't want the tty mode to silently get disabled when some change starts to pass wrong parameter.

hinshun: Yeah that works for me.

tonistiigi (approved):

PR #946, opened 2022-02-09
build: enhance metadata json output

crazy-max:
follow-up https://github.com/moby/buildkit/pull/2610

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -251,11 +252,7 @@ func buildTargets(ctx context.Context, dockerCli command.Cli, opts map[string]bu
 	}
 
 	if len(metadataFile) > 0 && resp != nil {
-		mdatab, err := json.MarshalIndent(resp[defaultTargetName].ExporterResponse, "", "  ")
-		if err != nil {
-			return "", err
-		}
-		if err := ioutils.AtomicWriteFile(metadataFile, mdatab, 0644); err != nil {
+		if err := writeMetadataFile(metadataFile, decodeExporterResponse(resp[defaultTargetName].ExporterResponse)); err != nil {
 			return "", err
 		}
 	}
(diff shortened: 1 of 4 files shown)


commands/build.go
```diff
@@ -496,6 +493,32 @@ func parseContextNames(values []string) (map[string]string, error) {
...
 
+func writeMetadataFile(filename string, dt interface{}) error {
+	b, err := json.Marshal(dt)
+	if err != nil {
+		return err
+	}
+	return ioutils.AtomicWriteFile(filename, b, 0644)
+}
+
+func decExporterResponse(exporterResponse map[string]string) map[string]interface{} {
```
tonistiigi, line 504: nit: `decodeExporterResponse` .

hack/test-driver
```diff
@@ -44,58 +51,111 @@ if [ "$DRIVER" != "docker" ]; then
...
 
 # build
-${BUILDX_CMD} build ${buildPlatformFlag} \
-  --output="type=cacheonly" \
+buildxCmd build ${buildPlatformFlag} \
+  --output="$(buildOutput buildx-test-build)" \
   --builder="${builderName}" \
+  --metadata-file="${context}/metadata-build.json" \
   "${context}"
+jq '.' "${context}/metadata-build.json"
```
tonistiigi, line 125: Shouldn't this try some actual key?

crazy-max: I'm using `.` filter just to pretty print: https://github.com/docker/buildx/runs/5139711248?check_suite_focus=true#step:9:143
No validation there. Maybe we could add some though.

commands/build.go
```diff
@@ -496,6 +493,32 @@ func parseContextNames(values []string) (map[string]string, error) {
 	return result, nil
 }
 
+func writeMetadataFile(filename string, dt interface{}) error {
+	b, err := json.Marshal(dt)
```
tonistiigi, line 497: I guess removing `Marshalndent` was not intentional?

crazy-max: Yes it was. I should have commented about that sorry. I think we should have the same behavior as `buildctl` but can revert that change if you want to.

tonistiigi (approved):

PR #775, opened 2021-09-21
vendor: update buildkit

crazy-max:
https://github.com/moby/buildkit/compare/1a7543a105272362da388625c9192f79acc96cef...44891f4cb975ac70d62d40139ba819fe855da59c

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #948, opened 2022-02-09
vendor: update buildkit to 2f99651

crazy-max:
https://github.com/moby/buildkit/compare/539be1708964...2f996517bcbd

Was testing if everything was ok while vendoring buildkit, specially otel upgrade and named contexts.

For k8s dependencies, I use a replace directive now for them otherwise we always use the deps from containerd module which is not what we really want when vendoring buildkit as it's specific for the k8s driver (https://github.com/docker/buildx/pull/565, https://github.com/docker/buildx/pull/775, ...). So should be done in follow-up if we want to. cc @AkihiroSuda @morlay.


tonistiigi (approved):

AkihiroSuda (approved):

go.mod
```diff
@@ -44,21 +44,22 @@ require (
...
 )
 
 replace (
 	github.com/docker/cli => github.com/docker/cli v20.10.3-0.20210702143511-f782d1355eff+incompatible
 	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20211216190657-088afc99e4bf+incompatible
-	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace => github.com/tonistiigi/opentelemetry-go-contrib/instrumentation/net/http/httptrace/otelhttptrace v0.0.0-20211026174723-2f82a1e0c997
-	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => github.com/tonistiigi/opentelemetry-go-contrib/instrumentation/net/http/otelhttp v0.0.0-20211026174723-2f82a1e0c997
+	k8s.io/api => k8s.io/api v0.22.4
+	k8s.io/apimachinery => k8s.io/apimachinery v0.22.4
+	k8s.io/client-go => k8s.io/client-go v0.22.4
```
AkihiroSuda, line 64: Do we still need these replacements?

crazy-max: I'm fine removing them if you want to :+1:

PR #958, opened 2022-02-17
build: send buildinfo dependencies

crazy-max:
follow-up https://github.com/moby/buildkit/pull/2654


tonistiigi (approved):

build/build.go
```diff
@@ -1218,15 +1219,20 @@ func waitContextDeps(ctx context.Context, index int, results *waitmap.Map, so *c
...
-					return err
+				if dt, ok := rr.Metadata[exptypes.ExporterBuildInfo+"/"+platform]; ok {
+					metadata[exptypes.ExporterBuildInfo] = dt
+				}
+				if len(metadata) > 0 {
+					dt, err := json.Marshal(metadata)
+					if err != nil {
+						return err
+					}
+					so.FrontendAttrs["input-metadata:"+k+"::"+platform] = string(dt)
```
tonistiigi, line 1234: We should think about how to clean up this `k` so it isn't the `resultKey()` that is a bake artifact. If it ends up in buildinfo as "deps" key it should be a clean value.

PR #959, opened 2022-02-18
set build-args from docker proxy configuration

### LINKED ISSUE #947: Client config.json proxy settings ignored
Hi,

I'm currently trying to get Buildx with a docker-container driver up and running within our CI infrastructure. Our build server is behind a proxy and therefore image builds require the correct proxy settings to be configured. Currently, we directly invoke the docker build command and have configured the proxy settings within `~/.docker/config.json` as described in the Docker [documentation](https://docs.docker.com/network/proxy/#configure-the-docker-client). What this does it that it automatically populate the correct env variables within the container during build time, e.g. http_proxy, https_proxy, etc. This does not influence the Docker daemon behavior which uses the operating system's configuration.

I was able to get the docker daemon to work by passing through the env variables via  `--driver-opt env.http_proxy=$http_proxy` and additionally using a custom buildkit image with the company root-CAs included so that the images were able to be pulled. But what I'm still missing is the exact same behavior of that the correct env variables are configured within the to-be-built build container.

E.g. 

```
FROM debian:stable-slim

RUN echo $http_proxy
```

Running `docker build .` will print whatever is configured within `~/.docker/config.json`. However, when invoking `docker buildx build .` will print an empty string.

Is there a way to achieve that behavior?

tonistiigi:
fixes #947

For backward compatibility with docker build.

### DIFF
--- a/store/storeutil/storeutil.go
+++ b/store/storeutil/storeutil.go
@@ -37,6 +37,32 @@ func GetCurrentEndpoint(dockerCli command.Cli) (string, error) {
 	return de, nil
 }
 
+func GetProxyConfig(dockerCli command.Cli) map[string]string {
+	cfg := dockerCli.ConfigFile()
+	host := dockerCli.Client().DaemonHost()
+
+	proxy, ok := cfg.Proxies[host]
+	if !ok {
+		proxy = cfg.Proxies["default"]
+	}
+
+	m := map[string]string{}
+
+	if v := proxy.HTTPProxy; v != "" {
+		m["HTTP_PROXY"] = v
+	}
+	if v := proxy.HTTPSProxy; v != "" {
+		m["HTTPS_PROXY"] = v
+	}
+	if v := proxy.NoProxy; v != "" {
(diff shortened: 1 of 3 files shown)


crazy-max (approved):

store/storeutil/storeutil.go
```diff
@@ -37,6 +37,32 @@ func GetCurrentEndpoint(dockerCli command.Cli) (string, error) {
...
+	}
+	if v := proxy.HTTPSProxy; v != "" {
+		m["HTTPS_PROXY"] = v
+	}
+	if v := proxy.NoProxy; v != "" {
+		m["NO_PROXY"] = v
+	}
+	if v := proxy.FTPProxy; v != "" {
+		m["FTP_PROXY"] = v
+	}
```
Ka0o0, line 62: Is this now setting both, lower and upper case variants of the variables?

tonistiigi: Yes, buildkit does that automatically when one of the variants is set.

PR #965, opened 2022-02-24
bake: additional support for named context on remote inputs

tonistiigi:
- Update named context paths to be relative of remote inputs
- Add basic validation that named paths don't escape the current working directory when loaded from remote sources.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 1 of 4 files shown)


bake/bake.go
```diff
@@ -729,6 +745,59 @@ func updateContext(t *build.Inputs, inp *Input) {
...
 }
 
+// validateContextsEntitlements is a basic check to ensure contexts do not
+// escape local directories when loaded from remote sources. This is to be
+// replaced with proper entitlements support in the future.
+func validateContextsEntitlements(t build.Inputs, inp *Input) error {
+	if inp == nil || inp.State == nil {
+		return nil
+	}
+	if v, ok := os.LookupEnv("BAKE_ALLOW_REMOTE_FS_ACCESS"); ok {
```
tonistiigi, line 755: @crazy-max I think in our github actions we can define this by default as the whole system is in a vm anyway.

crazy-max: On GHA they could include `$HOME/.docker` which contains auth token in `config.json` if login step has been done in a previous step.

crazy-max (commented):
Can you add `BAKE_ALLOW_REMOTE_FS_ACCESS` to the list of built-in variables: https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md#built-in-variables

tonistiigi:
> Can you add BAKE_ALLOW_REMOTE_FS_ACCESS to the list of built-in variables:

I don't really want to make this very official. Hopefully, it would be removed in the next release if we have proper entitlements support.

crazy-max (approved):

PR #972, opened 2022-02-26
imagetools inspect: keep platform order

crazy-max:
follow-up #854

keep order from the manifest and also print formatted image config.

### DIFF
--- a/util/imagetools/printers.go
+++ b/util/imagetools/printers.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sort"
 	"strings"
 	"text/tabwriter"
 	"text/template"
@@ -17,6 +18,7 @@ import (
 	binfotypes "github.com/moby/buildkit/util/buildinfo/types"
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 )
 
 const defaultPfx = "  "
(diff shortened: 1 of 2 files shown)


tonistiigi (commented):
As discussed offline, the added lines are not worth it. Let's just keep it deterministic. JSON is already sorted, just update the `printBuildInfos` to also do the alphabetical sort.

util/imagetools/printers.go
```diff
@@ -33,6 +38,7 @@ type Printer struct {
 	manifest  ocispecs.Descriptor
 	index     ocispecs.Index
 	platforms []ocispecs.Platform
+	pkeys     []string
```
tonistiigi, line 41: Why does this need to be in a struct and not just local variable in `printBuildInfos` ?

util/imagetools/printers.go
```diff
@@ -229,15 +270,99 @@ func (p *Printer) printManifestList(out io.Writer) error {
 	return w.Flush()
 }
 
+func (p *Printer) printImageConfigs(ics map[string]*ocispecs.Image, out io.Writer) error {
```
tonistiigi, line 273: Not sure about this formatting. We don't even have anything like this for `docker inspect`. Maybe leave this out of release and discuss later.

crazy-max: Yes was just because it was "missing" following the fields we currently expose.

util/imagetools/printers.go
```diff
@@ -107,20 +110,36 @@ func (p *Printer) Print(raw bool, out io.Writer) error {
 		return err
 	}
 
+	var pkeys []string
```
tonistiigi, line 113: define this in `printBuildInfos` if it is only used there.

tonistiigi (approved):

PR #975, opened 2022-02-27
build: fix iidfile content

### LINKED ISSUE #420: --iidfile contains mismatching id
The documentation for `build` argument `--iidfile` states: `--iidfile, Write the image ID to the file`.
The image imported to docker does not correspond to the iidfile contents.

When using:
```
docker buildx build --tag kalamaja --load --iidfile id --progress plain .
```
On dockerfile:

```
FROM alpine
RUN echo a
```

```
...
#6 exporting manifest sha256:305dbeb87452bc1457231f97e698d19db38c61def97a7a2982e6eae83256f7b6 0.0s done
#6 exporting config sha256:e384d63460089a5a38d739dc1fd82092caaaff3c158c324faae54ffda974a455 0.0s done
#6 sending tarball
#6 sending tarball 0.4s done
#6 DONE 0.8s

#7 importing to docker
#7 DONE 0.2s
```

The result is:
```
# cat id; echo
sha256:305dbeb87452bc1457231f97e698d19db38c61def97a7a2982e6eae83256f7b6
```
```
# docker image ls | head -2
kalamaja            latest              e384d6346008        3 minutes ago       5.57MB
```
```
# docker image inspect e384d6346008 | jq .[0].Id
"sha256:e384d63460089a5a38d739dc1fd82092caaaff3c158c324faae54ffda974a455"
```

crazy-max:
fixes #420
fixes docker/build-push-action#461

output the image ID or digest with `--iidfile` flag and `containerimage.digest` response depending on the driver and output being used:

* `docker` driver : writes the image ID.
* `docker` driver with `--push`: writes the image digest.
* `docker-container` driver with `--load`: writes the image ID.
* `docker-container` driver with `--push`: writes the image digest.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -833,7 +833,7 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 							respMu.Lock()
 							resp[k] = &client.SolveResponse{
 								ExporterResponse: map[string]string{
-									"containerimage.digest": desc.Digest.String(),
+									exptypes.ExporterImageDigestKey: desc.Digest.String(),
 								},
 							}
 							respMu.Unlock()
@@ -918,7 +918,6 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 						if err != nil {
 							return err
 						}
-						res[i] = rr
 
 						d := drivers[dp.driverIndex].Driver
 						if d.IsMobyDriver() {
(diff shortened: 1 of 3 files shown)


tonistiigi:
> docker driver with --push: writes the image digest.

Is this the behavior that we want?

crazy-max:
> > docker driver with --push: writes the image digest.
> 
> Is this the behavior that we want?

I see what you mean as the image is in docker store but also push to the registry. I agree for feature parity it might be better to keep the image ID for this case. Wonder if we could somehow provide the image digest in the exporter response (other key?) or another flag?

crazy-max:
#980 will fix the imageID issue with the container driver. About image digest I think it's a misleading issue in our github action with the [`digest` output](https://github.com/docker/build-push-action/#outputs). Should be called `imageid` instead and we should change the `digest` output behavior to only add image digest if image is effectively pushed. WDYT?

tonistiigi:
> Should be called imageid instead and we should change the digest output behavior to only add image digest if image is effectively pushed. WDYT?

Yes, separate outputs for the action. The digest is filled whenever it is available, not only with push.

PR #979, opened 2022-03-01
ci: use kind for e2e tests

crazy-max:

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -74,22 +74,19 @@ jobs:
             echo "MULTI_NODE=0" >> $GITHUB_ENV
           fi
       -
-        name: Install k3s
+        name: Set up kind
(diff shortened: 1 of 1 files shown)


crazy-max:
Found out were using kind while implementing the kubernetes driver in #167 so I wanted to check its behavior in our CI. pros are we could use kind in a container in a follow-up so we don't need to rely on the "kind action" on GHA. WDYT?

AkihiroSuda:
I feel k3s is closer to production clusters, but no strong opinion.

@tonistiigi @ktock WDYT?

ktock:
Sorry for the late reply.

> I feel k3s is closer to production clusters, but no strong opinion.

I feel so too but not a strong opinion.

>  kind in a container in a follow-up 

Maybe "k3s in Docker" can also be used? : https://github.com/k3d-io/k3d

PR #980, opened 2022-03-01
build: return imageID when loading without docker driver

### LINKED ISSUE #420: --iidfile contains mismatching id

crazy-max:
fixes #420 

follow-up https://github.com/docker/buildx/pull/975#issuecomment-1053957423

repro https://github.com/crazy-max/buildx-buildkit-tests/runs/5374641648?check_suite_focus=true#step:7:2

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -762,9 +762,12 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 				resp[k] = res[0]
 				respMu.Unlock()
 				if len(res) == 1 {
-					digest := res[0].ExporterResponse["containerimage.digest"]
+					dgst := res[0].ExporterResponse[exptypes.ExporterImageDigestKey]
+					if v, ok := res[0].ExporterResponse[exptypes.ExporterImageConfigDigestKey]; ok {
+						dgst = v
+					}
 					if opt.ImageIDFile != "" {
-						return ioutil.WriteFile(opt.ImageIDFile, []byte(digest), 0644)
+						return ioutil.WriteFile(opt.ImageIDFile, []byte(dgst), 0644)
 					}
 					return nil
 				}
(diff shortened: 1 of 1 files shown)


tonistiigi (commented):
We should just use `containerimage.config.digest` always and if it is not set (eg. multi-platform) then the full digest as before.

No need to do anything special for `--load` as multi-platform don't load anyway.

Also need to make sure this doesn't break the github action and update if it may.

crazy-max:
> Also need to make sure this doesn't break the github action and update if it may.

It's fine for the GitHub Action with this behavior but as discussed we might create a new output named `imageid` that will contain the actual imageid and set `digest` output accordingly.

tonistiigi (approved):

tonistiigi:
> It's fine for the GitHub Action with this behavior but as discussed we might create a new output named imageid that will contain the actual imageid and set digest output accordingly.

But what field does the action use atm? If it uses `iidfile` then it needs to be updated before the release.

crazy-max:
Hum yeah atm `digest` output returns content of `iidfile`, so need to make changes for the action. Will do that.

tonistiigi (approved):

PR #989, opened 2022-03-06
build: set remote digest when pushed with docker driver

crazy-max:
closes #975

Set `containerimage.digest` with correct digest when image is pushed with moby. If `containerimage.digest` already exists, its value will be saved to `containerimage.config.digest` for backward compatibility with #980.

repro: https://github.com/crazy-max/buildx-buildkit-tests/runs/5440498633?check_suite_focus=true#step:7:153

In follow-up create the `imageid` output in our GitHub Action and also fix `digest` output. See https://github.com/docker/build-push-action/pull/569

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -937,12 +937,26 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 											return errors.Errorf("tag is needed when pushing to registry")
 										}
 										pw := progress.ResetTime(pw)
-										for _, name := range strings.Split(pushNames, ",") {
+										pushList := strings.Split(pushNames, ",")
+										for _, name := range pushList {
 											if err := progress.Wrap(fmt.Sprintf("pushing %s with docker", name), pw.Write, func(l progress.SubLogger) error {
 												return pushWithMoby(ctx, d, name, l)
 											}); err != nil {
 												return err
 											}
+										}
+										remoteDigest, err := remoteDigestWithMoby(ctx, d, pushList[0])
+										if err == nil && remoteDigest != "" {
+											// old daemons might not have containerimage.config.digest set
+											// in response so use containerimage.digest value for it if available
+											if _, ok := rr.ExporterResponse[exptypes.ExporterImageConfigDigestKey]; !ok {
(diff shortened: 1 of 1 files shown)


build/build.go
```diff
@@ -937,13 +937,23 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
...
 											if err := progress.Wrap(fmt.Sprintf("pushing %s with docker", name), pw.Write, func(l progress.SubLogger) error {
 												return pushWithMoby(ctx, d, name, l)
 											}); err != nil {
 												return err
 											}
 										}
+										remoteDigest, err := remoteDigestWithMoby(ctx, d, pushList[0])
+										if err == nil && remoteDigest != "" {
+											if v, ok := rr.ExporterResponse[exptypes.ExporterImageDigestKey]; ok {
+												rr.ExporterResponse[exptypes.ExporterImageConfigDigestKey] = v
```
tonistiigi, line 951: why this? Is this backward compatibility or smth? Why wouldn't `ExporterImageConfigDigestKey` be set already?

crazy-max: > Why wouldn't `ExporterImageConfigDigestKey` be set already?

As you can see in https://github.com/crazy-max/buildx-buildkit-tests/runs/5458103650?check_suite_focus=true#step:7:152 (using latest stable buildx version), metadata only returns then following content when image is pushed with moby:

```json
{
  "containerimage.digest": "sha256:1fd302965e74b804eccf7e6ab6ec1edcbd3dfa8b8371ebc7244492832c39dd81"
}
```

I think this is linked to the current implementation on moby: https://github.com/moby/moby/blob/327699c313f51f8062cf3affcb3746f56f3689c6/builder/builder-next/exporter/export.go#L176

tonistiigi: Ok, we should try to fix this in moby. To make sure we only do this for the old daemons lets only set `ExporterImageConfigDigestKey` when no such value exists and add a comment that this is for handling older daemons that don't return `ExporterImageConfigDigestKey`.

tonistiigi (approved):

PR #995, opened 2022-03-10
Remove uses of deprecated io/ioutil

thaJeztah:
The package has been deprecated since Go 1.16: https://go.dev/doc/go1.16#ioutil

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,6 +12,7 @@ linters:
     - gofmt
     - govet
     - deadcode
+    - depguard
     - goimports
     - ineffassign
     - misspell
@@ -22,6 +23,15 @@ linters:
     - typecheck
     - structcheck
   disable-all: true
+
+linters-settings:
+  depguard:
+    list-type: blacklist
+    include-go-root: true
+    packages:
+      # The io/ioutil package has been deprecated.
+      # https://go.dev/doc/go1.16#ioutil
+      - io/ioutil
 
 issues:
   exclude-rules:
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"encoding/csv"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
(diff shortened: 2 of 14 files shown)


crazy-max (commented):
LGTM, thanks!

We should also add a depguard rule for https://golang.org/pkg/syscall maybe:

> Deprecated: this package is locked down. Callers should use the corresponding package in the golang.org/x/sys repository instead. That is also where updates required by new systems or versions should be applied. See https://golang.org/s/go1.4-syscall for more information.

crazy-max (approved):

thaJeztah:
> We should also add a depguard rule for https://golang.org/pkg/syscall maybe:

That one is more tricky; the syscall packages has some "platform agnostic" implementations that do not live elsewhere (unless you create separate win/mac/linux variants etc.

It's ok to use those alternatives for most cases, but I've seen a couple where it was either complicated, or where you didn't want to use the extra dependency of golang.org/x/sys.

thaJeztah:
We should look at using `t.TempDir()` in tests though (also in buildkit). Not urgent (but it can be nice and clean in some cases) similar to https://github.com/docker/cli/pull/3443

tonistiigi (approved):

.golangci.yml
```diff
@@ -12,6 +12,7 @@ linters:
     - gofmt
     - govet
     - deadcode
+    - depguard
```
tonistiigi, line 15: why isn't there a linter that just checks for deprecated functions in stdlib?

thaJeztah: Because ...  _"do as we say, not as we do"_ (?) they didn't deprecated it in code; https://github.com/golang/go/blob/go1.16.15/src/io/ioutil/ioutil.go#L25-L26

Otherwise `staticcheck` would've picked it up (e.g. like for `strings.Title()` (which IMO shouldn't have been marked deprecated, just had its limitations better documented); see https://github.com/moby/moby/pull/43201

PR #1000, opened 2022-03-15
docs: make sure syntax with latest stable dockerfile image is defined

tonistiigi:

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,8 @@ On Windows:
 Here is how to install and use Buildx inside a Dockerfile through the
 [`docker/buildx-bin`](https://hub.docker.com/r/docker/buildx-bin) image:
 
-```Dockerfile
+```dockerfile
+# syntax=docker/dockerfile:1
 FROM docker
 COPY --from=docker/buildx-bin /buildx /usr/libexec/docker/cli-plugins/docker-buildx
 RUN docker buildx version
@@ -298,6 +299,7 @@ inside your Dockerfile and can be leveraged by the processes running as part
 of your build.
 
 ```dockerfile
+# syntax=docker/dockerfile:1
 FROM --platform=$BUILDPLATFORM golang:alpine AS build
 ARG TARGETPLATFORM
 ARG BUILDPLATFORM
--- a/docs/reference/buildx_build.md
+++ b/docs/reference/buildx_build.md
@@ -128,8 +128,8 @@ $ docker buildx build --build-context project=path/to/project/source .
 # docker buildx build --build-context project=https://github.com/myuser/project.git .
 ```
 
-```Dockerfile
-#syntax=docker/dockerfile:1.4
+```dockerfile
+# syntax=docker/dockerfile:1
 FROM alpine
 COPY --from=project myfile /
 ```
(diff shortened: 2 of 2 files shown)


docs/reference/buildx_bake.md
```diff
@@ -804,14 +804,17 @@ The value can be a local source directory, container image (with docker-image://
 
 #### Pinning alpine image
 
+_Dockerfile_
+
 ```Dockerfile
-# Dockerfile
+#syntax=docker/dockerfile:1.14
```
crazy-max, line 810:
```suggestion
#syntax=docker/dockerfile:1.4
```

crazy-max (commented):
I think it would be better to be aligned with [our docs](https://docs.docker.com/engine/reference/builder/#official-releases) and so use `docker/dockerfile:1` instead. cc @thaJeztah 

Also not a strong preference on this but I tend to add a space: `# syntax`. We discussed about this in https://github.com/google/addlicense/pull/88#pullrequestreview-722329613.

thaJeztah:
Yes, I think it'd be good to follow our own "best practice" / "recommended" approach (with the exception of features that require `labs` front-ends of course).

> We recommend using `docker/dockerfile:1`, which always points to the latest stable release of the version 1 syntax, and receives both “minor” and “patch” updates for the version 1 release cycle. BuildKit automatically checks for updates of the syntax when performing a build, making sure you are using the most current version.

Also agreed on trying to keep the formatting consistent to not confuse users (`# syntax=...`)

thaJeztah:
It's a bit of a tricky one; the problem I often see is that these things get "stuck". I've seen this happening with examples we have for Docker Compose, where a specific feature neeeded (e.g.) version `3.2` of the compose schema. What happened is that people took those examples and wrote their compose-file using it. Now a new feature was added after `3.2`, and when they tried to use it, compose failed (invalid option). I can imagine the same thing will happen for Dockerfiles; someone reads about a new feature in (e.g.) `1.3`, so updates their Dockerfile to use `:1.3`, now they copy an example to use a feature that's new in `1.4`, and it doesn't work (effectively, specifying `1.3` or `1.4` becomes: "don't use features _newer than_ the version I specified").

For compose, the version was actually needed, because older versions wouldn't automatically upgrade to a newer spec, so the version could be used to produce an error that the user should update their version of Compose (or Docker) but for BuildKit (`Dockerfile`), we don't have that problem; specifying `:1` will always pull the latest version, so there's no real need to specify a version (not even to "indicate" a specific version is needed, which should be a bit of an implementation detail).

From a documentation perspective, I can see _some_ value in having documented "when" it was added, but perhaps we should consider having (e.g.) an annotation / badge for this. That way we could show things as "new" in the docs (without having the problem of older versions of the front-end ending up in Dockerfiles). That said; even for that case, we should ask ourselves what the added value is, given that the syntax (front-end) automatically updates; would it be just for posterity? If users follow the best practices (specify `docker/dockerfile:1`), then they can expect that _"if it's documented, it works"_ (of course, with the exception of `:labs` features).

crazy-max (approved):

PR #1012, opened 2022-03-17
[v0.8] bake: allow dot in target name

crazy-max:
#1011 

(cherry picked from commit 8f61bfbbaa548eecce2f0b32c1e2ce37b8e87a0c)


tonistiigi:
I don't really like this. This is ambiguous to the `--set target.args.foo` and I also want to allow such selectors in HCL variable in the future. E.g. `dockerfile = othertarget.dockerfile`.

Maybe for compose we could always replace `.` with `_` after parsing and then also do that in bake arguments. Would that be an option?

PR #1018, opened 2022-03-21
[v0.8] bake: allow dot in target names for compose

tonistiigi:
This is a hotfix for v0.8 to unblock release and
restore backward compatibility. More proper fix
coming later.

replaces #1012


tonistiigi:
Self-merge to unblock release

PR #1021, opened 2022-03-23
bake: fix skipped group when already visited by another one

### LINKED ISSUE #1020: Resolve group failed with v0.8.0
Hi guys, we hit some target resolving issue after upgrading to v0.8.0 from v0.7.1

version: **v0.8.0**

```
cat buildtest.hcl
group "a" {
  targets = ["b", "c"]
}
group "b" {
  targets = ["d"]
}
group "c" {
  targets = ["b"]
}
target "d" {
  context = "."
  dockerfile = "./testdockerfile"
}

docker buildx bake a -f buildtest.hcl
[+] Building 0.0s (0/0)
error: failed to find target b
```

crazy-max:
fixes #1020 

when a group has already been visited, its resolution will be skipped on next iteration and therefore would fail to find the relevant target when nested groups calling the same target is being used in the definition.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -412,12 +412,12 @@ func (c Config) newOverrides(v []string) (map[string]map[string]Override, error)
 }
 
 func (c Config) ResolveGroup(name string) []string {
-	return c.group(name, map[string]struct{}{})
+	return dedupString(c.group(name, map[string][]string{}))
 }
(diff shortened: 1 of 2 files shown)


yifan-gu-anchorage:
@crazy-max Thank you for the fix!

PR #1024, opened 2022-03-24
vendor: github.com/containerd/containerd v1.6.2

thaJeztah:
includes a fix for CVE-2022-24769 (but doesn't affect the code we use)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/cloudflare/cfssl v0.0.0-20181213083726-b94e044bb51e // indirect
 	github.com/compose-spec/compose-go v1.0.8
 	github.com/containerd/console v1.0.3
-	github.com/containerd/containerd v1.6.1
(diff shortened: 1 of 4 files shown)


AkihiroSuda (approved):

tonistiigi:
These updates are not relevant to buildx. Let's update containerd when we update buildkit as this is where the dependency is coming from.

thaJeztah:
scanners may flag the binaries as vulnerable (and I think we have an obligation now to update such dependencies)

thaJeztah:
(and yes, that's silly for things like these that don't affect the code)

tonistiigi:
> scanners may flag the binaries as vulnerable

This has happened many times before, especially with CLI/Moby CVE-s. We can't let some scanner showing wrong data force us to merge in unrelated code to our releases.

thaJeztah:
not the scanner... contracts

PR #1033, opened 2022-03-29
update github.com/compose-spec/compose-go to v1.2.1

### LINKED ISSUE #1032: Bake not accepting string for published port in Yaml
`docker compose config` will normalise a port mapping like:

```yaml
    ports:
      - 3306:3306
```

to this:

```yaml
    ports:
    - mode: ingress
      target: 3306
      published: "3306"
      protocol: tcp
```

Trying to pipe this into `buildx bake` 0.8.0 hits an error:

```                        
error: failed to parse -: parsing yaml: services.mysql.ports.0.published must be a integer, parsing hcl: -.hcl:17,11-12: Invalid character; Single quotes are not valid. Use double quotes (") to enclose strings., and 13 other diagnostic(s)
```

The compose-spec [allows either int or string here](https://github.com/compose-spec/compose-spec/blob/master/schema/compose-spec.json#L328), because it's supposed to be able to support a range like `"3306-3310"` - I guess `compose` is choosing to always output string

crazy-max:
fixes #1032 


tonistiigi (commented):
Could we add a test for the case described in the issue.

tonistiigi (approved):

AkihiroSuda (approved):

PR #1037, opened 2022-04-01
chore: github issue template

crazy-max:
adds a bug report [issue forms template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms).

will look like this:

image

___

image

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,0 +1,137 @@
+# https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema
+name: Bug Report
+description: Report a bug
+labels:
+  - kind/bug
+  - status/triage
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for taking the time to report a bug!
+        If this is a security issue please report it to the [Docker Security team](mailto:security@docker.com).
+
+  - type: checkboxes
+    attributes:
+      label: Contributing guidelines
+      description: |
+        Please read the contributing guidelines before proceeding.
+      options:
+        - label: I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree
+          required: true
+
+  - type: checkboxes
+    attributes:
+      label: I've found a bug and checked that ...
+      description: |
+        Make sure that your request fulfills all of the following requirements.
+        If one requirement cannot be satisfied, explain in detail why.
+      options:
+        - label: ... the documentation does not mention anything about my problem
+        - label: ... there are no open or closed issues that are related to my problem
+
+  - type: textarea
+    attributes:
+      label: Description
+      description: |
+        Please provide a brief description of the bug in 1-2 sentences.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Expected behaviour
+      description: |
+        Please describe precisely what you'd expect to happen.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Actual behaviour
+      description: |
+        Please describe precisely what is actually happening.
+    validations:
+      required: true
+
+  - type: input
+    attributes:
+      label: Buildx version
+      description: |
+        Can be found using the `docker buildx version` command.
+        Example: `github.com/docker/buildx v0.8.1 5fac64c2c49dae1320f2b51f1a899ca451935554`
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Docker info
+      description: |
+        Output of `docker info`.
+      placeholder: |
+        ```
+        
+        ```
+      render: markdown
+
+  - type: textarea
+    attributes:
+      label: Builders list
+      description: |
+        Output of `docker buildx ls`.
+      placeholder: |
+        ```
+        
+        ```
+      render: markdown
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Configuration
+      description: |
+        Please provide a minimal Dockerfile, bake definition (if applicable) and invoked command to reproduce the bug.
+      placeholder: |
+        ```dockerfile
+        FROM alpine
+        echo hello
+        ```
+        
+        ```hcl
+        group "default" {
+          targets = ["app"]
+        }
+        target "app" {
+          dockerfile = "Dockerfile"
+          target = "build"
+        }
+        ```
+        
+        ```console
+        $ docker buildx build .
+        $ docker buildx bake
+        ```
+      render: markdown
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Logs
+      description: |
+        Please provide Buildx logs (also BuildKit logs if applicable).
+      placeholder: |
+        ```
+        
+        ```
+      render: markdown
+    validations:
+      required: false
+
+  - type: textarea
+    attributes:
+      label: Additional info
+      description: |
+        Please provide any additional information that could be useful.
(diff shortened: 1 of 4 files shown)


hack/dockerfiles/lint.Dockerfile
```diff
@@ -3,10 +3,8 @@
...
 
 FROM golang:${GO_VERSION}-alpine
-RUN apk add --no-cache gcc musl-dev yamllint
+RUN apk add --no-cache gcc musl-dev
 RUN wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.36.0
 WORKDIR /go/src/github.com/docker/buildx
 RUN --mount=target=/go/src/github.com/docker/buildx --mount=target=/root/.cache,type=cache \
   golangci-lint run
-RUN --mount=target=/go/src/github.com/docker/buildx --mount=target=/root/.cache,type=cache \
-  yamllint -c .yamllint.yml --strict .
```
crazy-max, line 12: looks like yamllint is not happy even if it's valid: https://github.com/docker/buildx/runs/5785153351?check_suite_focus=true#step:3:135

```
#14 [stage-0 6/6] RUN --mount=target=/go/src/github.com/docker/buildx --mount=target=/root/.cache,type=cache   yamllint -c .yamllint.yml --strict .
#14 0.744 ./.github/ISSUE_TEMPLATE/bug.yml
#14 0.744   95:1      error    trailing spaces  (trailing-spaces)
#14 0.744   116:1     error    trailing spaces  (trailing-spaces)
#14 0.744 
#14 ERROR: executor failed running [/bin/sh -c yamllint -c .yamllint.yml --strict .]: exit code: 1
```

I think we can remove this linter. It that was added in https://github.com/docker/buildx/pull/451 for GitHub Actions workflow but know GitHub reports YAML syntax issues. Maybe I could instead add an exception instead for the issue template folder. Let me know if it sgty.

thaJeztah: Curious if there were actually trailing spaces though, or was it because of the empty "code" block?

crazy-max: Yes that's because the empty code block :disappointed:

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,144 @@
...
+        Can be found using the `docker buildx version` command.
+        Example: `github.com/docker/buildx v0.8.1 5fac64c2c49dae1320f2b51f1a899ca451935554`
+    validations:
+      required: true
+
+  - type: input
+    attributes:
+      label: BuildKit version
+      description: |
+        Please provide the BuildKit version being used by your builder.
```
jedevc, line 75: We should provide info to how to get this information as well - I know the unreleased `buildx ls` has this, maybe we can call that out here? Or maybe just remove this section, since we ask for that later anyways?

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,144 @@
...
+        ```
+      render: markdown
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Configuration
+      description: |
+        Please provide a minimal Dockerfile, bake definition (if applicable) to reproduce the bug.
```
jedevc, line 107: Maybe include a `command line invocation` here as well?

jedevc:
Let's definitely get this in asap :tada: - should save a lot of time going back and forth on getting reproduction details!

crazy-max:
Also added "Feature request" and custom template for "Security and Vulnerabilities report" and "Questions and Discussions": https://github.com/crazy-max/buildx/issues/new/choose

image

.github/ISSUE_TEMPLATE/bug.yml
crazy-max: I think we can remove this now that buildx 0.9 with `buildx ls` displays BuildKit version.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,140 @@
...
+
+  - type: checkboxes
+    attributes:
+      label: I've found a bug and checked that ...
+      description: |
+        Make sure that your request fulfills all of the following requirements.
+        If one requirement cannot be satisfied, explain in detail why.
+      options:
+        - label: ... the documentation does not mention anything about my problem
+        - label: ... there are no open or closed issues that are related to my problem
```
jedevc, line 23: This reads a little weird to me with the little paragraph in between:
image

Not a blocker, but also not sure how to rephrase this to flow a little easier.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,140 @@
...
+    attributes:
+      label: Steps to reproduce
+      description: |
+        Please describe the steps to reproduce the bug.
+      placeholder: |
+        1. ...
+        2. ...
+        3. ...
+    validations:
+      required: true
```
jedevc, line 59: I think we could remove this section and maybe merge with the configuration section? It seems like there would be some decent overlap.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,140 @@
...
+      label: Logs
+      description: |
+        Please provide Buildx logs (also BuildKit logs if applicable).
+      placeholder: |
+        ```
+        
+        ```
+      render: markdown
+    validations:
+      required: true
```
jedevc, line 134: I suggest we can have this as optional, in many cases the reproduction information should be enough.

We can always ask for logs later if necessary / come back and require this field in a later PR if we need it.

.github/ISSUE_TEMPLATE/bug.yml
thaJeztah: Would it be more useful to have the information of the currently active builder (instead of `buildx ls` with all builders?)

jedevc: `buildx inspect` unfortunately doesn't print the buildkit version! That seems like an oversight, I'll take a look at that. I think atm `buildx ls` is the only way to get that info.

There's also a number of cases where having all the builders listed is super helpful for debugging, especially when doing things around builder management/etc.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,9 @@
+# https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
```
jedevc, line 1: So had another thought here :smile:

I wonder if we should have a little section here about `Dockerfile`-specific issues, where we link to the buildkit new issues page :thinking:

.github/ISSUE_TEMPLATE/bug.yml
crazy-max: Yeah `buildx ls` is more useful also `*` and `--builder` indicated which builder is being used from the repro.

.github/ISSUE_TEMPLATE/config.yml
crazy-max: Hum not sure how to phrase this, let's see in a follow-up but that sounds like a good idea.

jedevc (approved):

.github/SECURITY.md
```diff
@@ -0,0 +1,12 @@
+# Reporting security issues
+
+The project maintainers take security seriously. If you discover a security
+issue, please bring it to their attention right away!
+
+**Please _DO NOT_ file a public issue**, instead send your report privately to
+[security@docker.com](mailto:security@docker.com).
+
+Security reports are greatly appreciated, and we will publicly thank you for it.
```
thaJeztah, line 9: Not a blocker (we need to update in moby as well), but we should mention that we publicly thank them _if they want to_ (some people may not want to, or their employer may not allow them to be mentioned)

jedevc (approved):

PR #1025, opened 2022-03-24
bake: merge targets and vars from multiple JSON files

crazy-max:
fixes #756 
fixes docker/bake-action#61

while checking the issue reported in docker/bake-action#61, I found out that the HCL diagnostic being reported was wrong:

```
  /usr/bin/docker buildx bake --file ./docker-bake.hcl --file /tmp/docker-metadata-action-GGKyIw/docker-metadata-action-bake.json --file /tmp/docker-metadata-action-fZ5d0s/docker-metadata-action-bake.json --file /tmp/docker-metadata-action-toPOSl/docker-metadata-action-bake.json --file /tmp/docker-metadata-action-Cqo0nf/docker-metadata-action-bake.json --metadata-file /tmp/docker-build-push-Rpx0C2/metadata-file binary dind proxy tools --print
  ./docker-bake.hcl:2
  --------------------
     1 |     // filled by GitHub Actions
     2 | >>> target "docker-metadata-k3d" {}
     3 |     target "docker-metadata-k3d-dind" {}
     4 |     target "docker-metadata-k3d-proxy" {}
  --------------------
  error: ./docker-bake.hcl:2,1-7: Unexpected "target" block; Blocks are not allowed here., and 3 other diagnostic(s)
  Error: The process '/usr/bin/docker' failed with exit code 1
```

`Blocks are not allowed here.` should be filtered out as it's ignored while checking HCL attributes: https://github.com/docker/buildx/blob/1ca30a58c2f25a7f72a2fdb5d62e82bf1e2fa37c/bake/hclparser/hclparser.go#L306

Now that the diag is filtered out we have the right diagnostic being reported: https://github.com/crazy-max/buildx-buildkit-tests/tree/main/bake-action-61

```console
$ docker buildx bake -f docker-bake.hcl -f json-target/binary.json -f json-target/dind.json -f json-target/proxy.json binary --print
json-target/dind.json:2
--------------------
   1 |     {
   2 | >>>   "target": {
   3 |         "docker-metadata-k3d-dind": {
   4 |           "tags": [
--------------------
error: json-target/dind.json:2,3-11: Duplicate argument; Argument "target" was already set at json-target/binary.json:2,3-11, and 1 other diagnostic(s)
```

This PR also filters out the diagnostic `Argument "x" was already set at` only for reserved names so we can merge targets from multiple JSON files: https://github.com/crazy-max/buildx-buildkit-tests/tree/main/buildx-756

```console
[...]

PR #1062, opened 2022-04-11
bake: set attribute even if diagnosed as duplicated

crazy-max:
Carry changes from #1025 related to an issue with hcl attributes not set if diagnosed as duplicated: https://github.com/docker/buildx/pull/1025#discussion_r843336438.

~Needs to fork the [merge logic from hcl repo](https://github.com/hashicorp/hcl/blob/main/merged.go). It might be better if hcl exposed its mergedbodies interface. Don't think we should fork the repo and vendor it here with this small change though. Maybe there is a better way?~

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -247,7 +247,7 @@ func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error)
 	}
 
 	if len(hclFiles) > 0 {
-		renamed, err := hclparser.Parse(hcl.MergeFiles(hclFiles), hclparser.Opt{
+		renamed, err := hclparser.Parse(hclparser.MergeFiles(hclFiles), hclparser.Opt{
 			LookupVar:     os.LookupEnv,
 			Vars:          defaults,
 			ValidateLabel: validateTargetName,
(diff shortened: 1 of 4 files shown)


tonistiigi (commented):
Did we try to upstream this?

Could we avoid copying this bunch of code by just providing our own `JustAttributes()` function that ranges over array and calls internal `JustAttributes()` for each item.

Add more context to this issue about the actual bake case it addresses and bake tests.

jedevc:
@crazy-max do you have an test case here? (sorry, I know this is an older issue)

I think there might be an easier way to resolve this using some of the bake refactors recently, but I can't seem to reproduce the issue being discussed - the test from https://github.com/docker/buildx/pull/1025#discussion_r843336438 fails for different reasons.

crazy-max:
> Did we try to upstream this?
> 
> Could we avoid copying this bunch of code by just providing our own `JustAttributes()` function that ranges over array and calls internal `JustAttributes()` for each item.
> 
> Add more context to this issue about the actual bake case it addresses and bake tests.

Opened an issue on HCL repo and got some feedback https://github.com/hashicorp/hcl/issues/636#issuecomment-1773214417. Creating our own implementation like Terraform does looks to be the right way.

tonistiigi (approved):

PR #1069, opened 2022-04-13
bake: support compose build secrets

### LINKED ISSUE #1060: Support `build.secrets` in yaml
Compose has added a `<service>.build.secrets` key and `...ssh` is on the way

https://github.com/compose-spec/compose-spec/pull/238

It looks to me to be broadly similar to `build.x-bake.secrets` so hopefully can be supported simply, the only issue might be merging? 

I would suggest that if both are present they're merged, de-duplicating by `id`, but it might be simpler to have a hard error when both are used

crazy-max:
fixes #1060 

Adds support for compose build secrets with bake. Needs to vendor compose-go https://github.com/compose-spec/compose-go/compare/v1.2.1...v1.2.4. See spec https://github.com/compose-spec/compose-spec/blob/master/build.md#secrets.

We can keep the [`x-bake` secret field](https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md#extension-field-with-compose) for now imo.

Also [env secret type](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#env) is supported with the following format:

```yaml
services:
  webapp:
    build:
      context: .
      secrets:
        - ENV_TOKEN

secrets:
  ENV_TOKEN: {}
```

I will review our documentation about bake and create dedicated guides for compose and hcl formats for this kind of use cases in a follow-up.

cc @glours @ciaranmcnulty

### DIFF
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -74,6 +74,16 @@ func ParseCompose(dt []byte) (*Config, error) {
 				dockerfilePath := s.Build.Dockerfile
 				dockerfilePathP = &dockerfilePath
 			}
+
+			var secrets []string
+			for _, bs := range s.Build.Secrets {
+				secret, err := composeToBuildkitSecret(bs, cfg.Secrets[bs.Source])
+				if err != nil {
+					return nil, err
+				}
+				secrets = append(secrets, secret)
+			}
+
 			g.Targets = append(g.Targets, s.Name)
 			t := &Target{
 				Name:       s.Name,
@@ -89,6 +99,7 @@ func ParseCompose(dt []byte) (*Config, error) {
 				})),
 				CacheFrom:   s.Build.CacheFrom,
 				NetworkMode: &s.Build.Network,
+				Secrets:     secrets,
 			}
 			if err = t.composeExtTarget(s.Build.Extensions); err != nil {
 				return nil, err
@@ -208,4 +219,22 @@ func (t *Target) composeExtTarget(exts map[string]interface{}) error {
 		}
 	}
 	return nil
+}
+
+// composeToBuildkitSecret converts secret from compose format to buildkit's
+// csv format.
+func composeToBuildkitSecret(inp compose.ServiceSecretConfig, psecret compose.SecretConfig) (string, error) {
+	if psecret.External.External {
+		return "", errors.Errorf("unsupported external secret %s", psecret.Name)
+	}
+
+	var bkattrs []string
+	if inp.Source != "" {
+		bkattrs = append(bkattrs, "id="+inp.Source)
+	}
+	if psecret.File != "" {
+		bkattrs = append(bkattrs, "src="+psecret.File)
+	}
+
+	return strings.Join(bkattrs, ","), nil
 }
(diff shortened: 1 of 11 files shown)


glours (approved):
LGTM

bake/compose.go
```diff
@@ -209,3 +220,34 @@ func (t *Target) composeExtTarget(exts map[string]interface{}) error {
 	}
 	return nil
 }
+
+// composeToBuildkitSecret converts secret from compose format to buildkit's
+// csv format.
+func composeToBuildkitSecret(inp compose.ServiceSecretConfig, projectSecrets compose.Secrets) (string, error) {
+	psecret := projectSecrets[inp.Source]
+	if psecret.External.External {
+		return "", fmt.Errorf("unsupported external secret %s", psecret.Name)
```
tonistiigi, line 229: `errors.Errorf`

bake/compose.go
```diff
@@ -209,3 +220,34 @@ func (t *Target) composeExtTarget(exts map[string]interface{}) error {
 	}
 	return nil
 }
+
+// composeToBuildkitSecret converts secret from compose format to buildkit's
+// csv format.
+func composeToBuildkitSecret(inp compose.ServiceSecretConfig, projectSecrets compose.Secrets) (string, error) {
```
tonistiigi, line 226: Why doesn't this just take `psecret` as parameter?

tonistiigi (commented):
I guess in the future we can discuss how this could be changed to a typed structure. But hacky to encode csv so it can be parsed again.

bake/compose.go
```diff
@@ -209,3 +220,33 @@ func (t *Target) composeExtTarget(exts map[string]interface{}) error {
...
+	}
+
+	var bkattrs []string
+	if inp.Source != "" {
+		bkattrs = append(bkattrs, "id="+inp.Source)
+	}
+	if psecret.File != "" {
+		bkattrs = append(bkattrs, "src="+psecret.File)
+	}
+	if inp.Target != "" {
```
tonistiigi, line 238: What are these things: targets, uid, gid, mode? These can't be set by the flags but are determined by the definition in Dockerfile.

crazy-max:
> I guess in the future we can discuss how this could be changed to a typed structure. But hacky to encode csv so it can be parsed again. (There is also a plan to allow defining the secret keys in json/hcl directly instead of csv).

Yes typed structure would be nice and also expose a contract in BuildKit so we don't have hardcoded attributes that could potentially drift.

bake/compose.go
crazy-max: Oh my bad I confused the secret mount in the [Dockerfile syntax](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypesecret) with the secret flag!

crazy-max:
> What are these things: targets, uid, gid, mode? These can't be set by the flags but are determined by the definition in Dockerfile.

@glours Looking at this, adding these fields in the compose-spec might have been out of scope for buildx.

tonistiigi (approved):

PR #1078, opened 2022-04-25
Add remote driver

### LINKED ISSUE #23: add "env" driver that connects to $BUILDKIT_HOST

jedevc:
As a follow-up to #909, this PR introduces support for the new "remote" driver. This new PR is because the original contributor hasn't responded, so I've done my best at tidying up the review comments and completing the desired feature set.

Fixes #23.

I've also started work on some docs in https://github.com/jedevc/buildx/commit/97e79d1253954b89329a3353bfb52c41291a93fa (not part of this PR), and will do a follow-up once the feature is merged.

CC: @developer-guy @Dentrax


driver/remote/driver.go
```diff
@@ -0,0 +1,86 @@
...
+		Status: driver.Running,
+	}, nil
+}
+
+func (d *Driver) Stop(ctx context.Context, force bool) error {
+	return fmt.Errorf("stop command is not implemented for this driver")
+}
+
+func (d *Driver) Rm(ctx context.Context, force, rmVolume, rmDaemon bool) error {
+	return fmt.Errorf("rm command is not implemented for this driver")
```
tonistiigi, line 55: You still need to be able to remove the remote builder instance. The buildkitd itself remains running.

driver/remote/driver.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, errors.Wrapf(driver.ErrNotConnecting, err.Error())
+	}
+
+	return &driver.Info{
+		Status: driver.Running,
+	}, nil
+}
+
+func (d *Driver) Stop(ctx context.Context, force bool) error {
+	return fmt.Errorf("stop command is not implemented for this driver")
```
tonistiigi, line 51: Use `errors.Errorf`. Could imprive this error. eg. "this builder user remote driver that can't be stopped by buildx. Stop buildkitd process directly if needed."

tonistiigi, line 51: Use `errors.Errorf`. Could imprive this error. eg. "this builder user remote driver that can't be stopped by buildx. Stop buildkitd process directly if needed."

driver/remote/factory.go
```diff
@@ -0,0 +1,111 @@
...
+func (*factory) Name() string {
+	return "remote"
+}
+
+func (*factory) Usage() string {
+	return "remote"
+}
+
+func (*factory) Priority(ctx context.Context, api dockerclient.APIClient) int {
+	if api == nil {
```
tonistiigi, line 33: Priority of remote driver should not depend on Docker API.

tonistiigi, line 33: Priority of remote driver should not depend on Docker API.

driver/remote/factory.go
```diff
@@ -0,0 +1,111 @@
...
+		return priorityUnsupported
+	}
+
+	return prioritySupported
+}
+
+func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver, error) {
+	if len(cfg.Files) > 0 {
```
tonistiigi, line 41: I guess same would be needed for `--buildkitd-flags`

tonistiigi, line 41: I guess same would be needed for `--buildkitd-flags`

tonistiigi (commented):
Could we add some CI testing for this. Could just start buildkitd manually in a container and then use remote driver to dial into it via port.

driver/remote/driver.go
jedevc: So it looks like `Stop` is called on remove - returning an error here means that the remove command errors with the message we put here.
I'm tempted to do the same as the docker driver, and just `return nil`?

driver/remote/factory.go
jedevc: Makes sense, it should just return a fixed priority.

Is there a reason that the kubernetes driver priority depends on the Docker API?

tonistiigi:
> So it looks like Stop is called on remove - returning an error here means that the remove command errors with the message we put here.

Ok, leave it nil atm. As a follow-up would be good to return a typed error in here so we can show an error message for `stop` in these cases but let `rm` pass cleanly.

driver/remote/factory.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, errors.Errorf("setting config file is not supported for remote driver")
+	}
+	if len(cfg.BuildkitFlags) > 0 {
+		return nil, errors.Errorf("setting buildkit flags is not supported for remote driver")
+	}
+
+	d := &Driver{
+		factory:      f,
+		InitConfig:   cfg,
+		buildkitAddr: cfg.BuildkitAddr,
```
tonistiigi, line 46: Why the extra field if we can already access this from `InitConfig` ?

driver/manager.go
```diff
@@ -103,11 +104,12 @@ func GetFactory(name string, instanceRequired bool) Factory {
 	return nil
 }
 
-func GetDriver(ctx context.Context, name string, f Factory, api dockerclient.APIClient, auth Auth, kcc KubeClientConfig, flags []string, files map[string][]byte, do map[string]string, platforms []specs.Platform, contextPathHash string) (Driver, error) {
+func GetDriver(ctx context.Context, name string, f Factory, api dockerclient.APIClient, buildkitAddr string, auth Auth, kcc KubeClientConfig, flags []string, files map[string][]byte, do map[string]string, platforms []specs.Platform, contextPathHash string) (Driver, error) {
 	ic := InitConfig{
 		DockerAPI:        api,
 		KubeClientConfig: kcc,
 		Name:             name,
+		BuildkitAddr:     buildkitAddr,
```
tonistiigi, line 112: I guess a better name in here would be `EndpointAddr` so it is not so specific to remote driver..

driver/remote/driver.go
```diff
@@ -0,0 +1,85 @@
...
+	cert       string
+	key        string
+}
+
+func (d *Driver) Bootstrap(ctx context.Context, l progress.Logger) error {
+	return nil
+}
+
+func (d *Driver) Info(ctx context.Context) (*driver.Info, error) {
+	if d.buildkitAddr == "" {
```
tonistiigi, line 31: Should we default to `BUILDKIT_HOST` in here.

driver/remote/factory.go
```diff
@@ -0,0 +1,109 @@
...
+func (*factory) Name() string {
+	return "remote"
+}
+
+func (*factory) Usage() string {
+	return "remote"
+}
+
+func (*factory) Priority(ctx context.Context, api dockerclient.APIClient) int {
+	return priority
```
tonistiigi, line 32: I think this could use a bit of logic change. You could update the signature of the `Priority()` method to take in the endpoint/args value that you can pass to the `GetDefaultFactory()` . If it is set to a tcp/tls address (or `BUILDKIT_HOST` set) then this priority in here would be higher(below 30).

Ideally, we could also detect that `docker-container://` and `kube-pod://` addresses go to the correct driver but that can be a follow-up.

jedevc:
So, I've changed a few more things (as another commit on top, will squash when code is good).
- Validate the endpoint argument at point of creation, requiring that it exists, and is a valid url with a valid scheme.
- Use the value of `BUILDKIT_HOST` as fallback if the endpoint arg is not provided
- Fix priorities for the remote (so if the url matches a given pattern, we prioritize it more)
  - For now, I've done unix:// (which might not be the right call) and tcp:// - no tls://, since those args are handled at a different layer, and I think conceptually you could also do TLS over unix sockets?
- Refactoring around the `runCreate` function, to get it to play nicely with the remote driver.

tonistiigi (approved):

jedevc:
Squashed, should be good to go :tada:

tonistiigi (approved):

jaihwan104:
I want to use the remote driver.
When will this feature be released? 👀

PR #1093, opened 2022-05-05
Add remote driver guide

jedevc:
Added basic scenarios for:
- Connecting over a basic unix socket
- Connecting to a docker buildkit container
- Connecting to a kubernetes buildkit deployment

### DIFF
--- a/docs/guides/remote-builder.md
+++ b/docs/guides/remote-builder.md
@@ -1,0 +1,156 @@
+---
+title: "Remote builder"
+description: "Connect buildx to an external buildkitd instance"
+keywords: build, buildx, buildkit
+---
+
+The buildx remote driver allows for more complex custom build workloads that
+allow users to connect to external buildkit instances. This is useful for
+scenarios that require manual management of the buildkit daemon, or where a
+buildkit daemon is exposed from another source.
+
+To connect to a running buildkitd instance:
+
+```console
+$ docker buildx create \
+  --name remote \
+  --driver remote \
+  tcp://localhost:1234
+```
+
+## Remote Buildkit over Unix sockets
+
+In this scenario, we'll create a setup with buildkitd listening on a unix
+socket, and have buildx connect through it.
+
+Firstly, ensure that [buildkit](https://github.com/moby/buildkit) is installed.
+For example, you can launch an instance of buildkitd with:
+
+```console
+$ sudo ./buildkitd --group $(id -gn) --addr unix://$HOME/buildkitd.sock
+```
+
+Alternatively, [see here](https://github.com/moby/buildkit/blob/master/docs/rootless.md)
+for running buildkitd in rootless mode or [here](https://github.com/moby/buildkit/tree/master/examples/systemd)
+for examples of running it as a systemd service.
+
+You should now have a unix socket accessible to your user, that is available to
+connect to:
+
+```console
+$ ls -lh /home/user/buildkitd.sock
+srw-rw---- 1 root user 0 May  5 11:04 /home/user/buildkitd.sock
+```
+
+You can then connect buildx to it with the remote driver:
+
+```console
+$ buildx create \
+  --name remote-unix \
+  --driver remote \
+  unix://$HOME/buildkitd.sock
+```
+        
+If you list available builders, you should then see `remote-unix` among them:
+
+```console
+$ buildx ls
+NAME/NODE           DRIVER/ENDPOINT                        STATUS  PLATFORMS
+test                remote
+  test0             unix:///home/.../buildkitd.sock        running linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/386
+default *           docker
+  default           default                                running linux/amd64, linux/386
+```
+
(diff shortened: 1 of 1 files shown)


jedevc:
@crazy-max have updated to follow the guide to follow the new-style docs (and tested to make sure they'll look alright on docs.docker.com, once https://github.com/docker/docker.github.io/pull/14644 is merged).

docs/guides/remote-driver.md
```diff
@@ -0,0 +1,131 @@
+---
+title: "Remote driver"
```
crazy-max, line 2: Maybe `Remote builder` would be better

crazy-max (approved):
LGTM, ptal @tonistiigi @thaJeztah

docs/guides/remote-builder.md
```diff
@@ -0,0 +1,151 @@
...
+
+We can now connect to this running container using buildx:
+
+```console
+$ docker buildx create \
+  --name remote-kubernetes \
+  --driver remote \
+  --driver-opt cacert=.certs/ca.pem,cert=.certs/client-cert.pem,key=.certs/client-key.pem,servername=... \
+  tcp://buildkitd.default.svc:1234
+```
```
AkihiroSuda, line 111: This seems about Kubernetes, not Docker

docs/guides/remote-builder.md
```diff
@@ -0,0 +1,151 @@
...
+
+In this scenario, we'll create a setup with buildkitd listening on a unix
+socket, and have buildx connect through it.
+
+Firstly, ensure that [buildkit](https://github.com/moby/buildkit) is installed.
+Then launch an instance of buildkitd:
+
+```console
+$ sudo ./buildkitd --group $(id -gn) --addr unix://$HOME/buildkitd.sock
+```
```
AkihiroSuda, line 31: I’d rather recommend using rootless mode, but that can be another PR

tonistiigi: Lets keep it simple as it is and add a link to https://github.com/moby/buildkit/blob/master/docs/rootless.md for rootless. Could even leave the default socket path. Can also point to https://github.com/moby/buildkit/tree/master/examples/systemd for running buildkitd with systemd.

docs/guides/remote-builder.md
```diff
@@ -0,0 +1,151 @@
...
+
+(remember that `--load` is necessary when not using the default `docker`
+driver, to load the build result into the docker daemon)
+
+## Remote Buildkit in Docker container
+
+In this scenario, we'll create a similar setup to the `docker-container`
+driver, by manually booting a buildkit docker container and connecting to it
+using the buildx remote driver. In most cases you'd probably just use the
+`docker-container` driver directly, but imagine that for some reason you want
```
tonistiigi, line 76: > but imagine

Smth like

In most cases, you'd probably just use the `docker-container` driver that connects to buildkit through Docker API, but in this case we are manually creating a container and accessing it with via an exposed port.

docs/guides/remote-builder.md
```diff
@@ -0,0 +1,151 @@
...
+to manually manage the container instead of having buildx manage it for you.
+
+First, we need to generate certificates for buildkit - you can use the 
+[create-certs.sh](https://github.com/moby/buildkit/blob/master/examples/kubernetes/create-certs.sh)
```
tonistiigi, line 80: `/blob/master` -> lets use a tagged release link in case the file paths should change.

docs/guides/remote-builder.md
```diff
@@ -0,0 +1,151 @@
...
+allow arbitrary access to buildkit without credentials.
+
+With our certificates generated in `.certs/`, we startup the container:
+
+```console
+$ docker run -d --rm \
+  --name=remote-buildkitd \
+  --privileged \
+  -p 1234:1234 \
+  -v $PWD/.certs:/certs \
```
tonistiigi, line 92: Mounting to `/etc/buildkit/certs` might be cleaner

docs/guides/remote-builder.md
jedevc: Have setup those links right below this to point users in the right direction - I think we should keep the specified socket path, otherwise we'll also need to tell users later on to use sudo buildx (since the `/var/run/buildkit/` directory is owned by `root:root` and inaccessible, even with `--group`, though maybe that's a bug?).

docs/guides/remote-builder.md
```diff
@@ -0,0 +1,157 @@
...
+socket, and have buildx connect through it.
+
+Firstly, ensure that [buildkit](https://github.com/moby/buildkit) is installed.
+Then launch an instance of buildkitd:
+
+```console
+$ sudo ./buildkitd --group $(id -gn) --addr unix://$HOME/buildkitd.sock
+```
+
+In this example, we run buildkitd as the root user, however, in the real world,
```
tonistiigi, line 33: "For example, launch an instance of buildkitd with:

..

Alternatively, [look here] for running buildkitd in rootless mode or [here] for examples of running it as a systems service.

tonistiigi (approved):

PR #1096, opened 2022-05-05
ci: enable cache

crazy-max:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,9 +42,15 @@ jobs:
         with:
           file: ./coverage/coverage.txt
       -
+        name: Expose GitHub Runtime
+        uses: crazy-max/ghaction-github-runtime@v1
+      -
         name: Build binaries
         run: |
           make release
+        env:
+          CACHE_FROM: type=gha,scope=release
+          CACHE_TO: type=gha,scope=release
       -
         name: Upload artifacts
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -41,10 +41,16 @@ jobs:
...
+      -
+        name: Expose GitHub Runtime
+        uses: crazy-max/ghaction-github-runtime@v1
       -
         name: Build binaries
         run: |
           make release
+        env:
+          CACHE_FROM: type=gha,scope=release
+          CACHE_TO: type=gha,scope=release,mode=max
```
tonistiigi, line 53: Is the `mode=max` beneficial for our Dockerfile. Maybe better results just upgrading to `COPY --link`?

crazy-max: Ah indeed `COPY --link` might be better

crazy-max: By mounting `.git` folder, `buildx-version` stage is not cached: https://github.com/crazy-max/buildx/runs/6336356712?check_suite_focus=true#step:8:72. I think on CI we should not mount `.git` but relying on [`GITHUB_REF` env var](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) so we can cache effectively. WDYT?

tonistiigi (approved):

PR #1097, opened 2022-05-06
dockerfile: enhance test stage

crazy-max:
optimize test stage to use gobase as based stage instead of build one. also move test step in our ci workflow to a dedicated job.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ FROM golatest AS gobase
 COPY --from=xx / /
 RUN apk add --no-cache file git
 ENV GOFLAGS=-mod=vendor
+ENV CGO_ENABLED=0
 WORKDIR /src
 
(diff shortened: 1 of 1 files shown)


jedevc (approved):
LGTM. Curious if we'd want to add a dependency between the jobs so that the build job runs after the test job, to avoid making a release/pushing to dockerhub if the tests were to fail?

crazy-max:
> Curious if we'd want to add a dependency between the jobs so that the build job runs after the test job, to avoid making a release/pushing to dockerhub if the tests were to fail?

Yes mainly to make clear what job fails and avoid mixing steps in our pipeline.

tonistiigi:
Do I get it correctly that these build do not use remote cache at all? Normally we would want to cache the first build so both of these targets can reuse it.

tonistiigi (approved):

PR #1103, opened 2022-05-09
Use --bootstrap to wait for remote to become active

jedevc:
Previously, the `--bootstrap` flag didn't do anything with the remote driver - now it should wait until the remote becomes active, with a small linear backoff.

This should help improve scenarios where the remote may not come up instantly if it has been started just before as was the case in the issue that caused #1094.


crazy-max:
@jedevc Can we remove this step now since these changes: https://github.com/docker/buildx/blob/43968ffa6899a3cf145cc1290c0744e5017ede48/.github/workflows/e2e.yml#L150-L163?

driver/remote/driver.go
```diff
@@ -23,17 +23,39 @@ type tlsOpts struct {
...
+			time.Sleep(time.Duration(i) * time.Second)
+		}
+	}
 }
 
 func (d *Driver) Info(ctx context.Context) (*driver.Info, error) {
 	c, err := d.Client(ctx)
 	if err != nil {
-		return nil, errors.Wrapf(driver.ErrNotConnecting, err.Error())
+		return &driver.Info{
```
tonistiigi, line 50: I think this should still return error and we should check for "network error" on the caller side.

jedevc: I agree about preserving the error to be able to display - I'm not sure about how much to change the caller side. `Info` gets called before `Boot`, which requires some fiddly logic changes in generic code to ignore network errors in that case.

I think there's two scenarios for errors here:
- Failure to connect to the endpoint providing the driver (can't connect to kubernetes, can't connect to docker daemon, etc) - this kind of failure means it's not worth trying to bootstrap, something is wrong.
- Failure to connect to the individual driver node (as in this case) - bootstrapping here is fine

Could we maybe rework status to be a struct and add an `Err` field to it? Then we could express a status of a node that's `inactive` because of an `err`. It currently feels like the error returned from here is quite tied to it's status, and is a little inconsistent across the drivers.

PR #1111, opened 2022-05-12
create: load default buildkit config if none specified

crazy-max:
fixes https://github.com/docker/setup-buildx-action/issues/127

if configuration file is not specified, will look for one by default in:
* `$BUILDX_CONFIG/buildkitd.default.toml`
* `$DOCKER_CONFIG/buildx/buildkitd.default.toml`
* `~/.docker/buildx/buildkitd.default.toml`

this can be useful when a user session should inherit from the same configuration when a builder is created without bothering specifying the config path.

### DIFF
--- a/docs/reference/buildx_create.md
+++ b/docs/reference/buildx_create.md
@@ -79,6 +79,11 @@ Specifies the configuration file for the buildkitd daemon to use. The configurat
 can be overridden by [`--buildkitd-flags`](#buildkitd-flags).
 See an [example buildkitd configuration file](https://github.com/moby/buildkit/blob/master/docs/buildkitd.toml.md).
 
+If the configuration file is not specified, will look for one by default in:
+* `$BUILDX_CONFIG/buildkitd.default.toml`
+* `$DOCKER_CONFIG/buildx/buildkitd.default.toml`
+* `~/.docker/buildx/buildkitd.default.toml`
+
 Note that if you create a `docker-container` builder and have specified
 certificates for registries in the `buildkitd.toml` configuration, the files
 will be copied into the container under `/etc/buildkit/certs` and configuration
--- a/commands/create.go
+++ b/commands/create.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/store/storeutil"
 	"github.com/docker/buildx/util/cobrautil"
+	"github.com/docker/buildx/util/confutil"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/google/shlex"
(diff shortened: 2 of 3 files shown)


docs/reference/buildx_create.md
```diff
@@ -79,6 +79,11 @@ Specifies the configuration file for the buildkitd daemon to use. The configurat
 can be overridden by [`--buildkitd-flags`](#buildkitd-flags).
 See an [example buildkitd configuration file](https://github.com/moby/buildkit/blob/master/docs/buildkitd.toml.md).
 
+If the configuration file is not specified, will look for one by default in:
+* `$BUILDX_CONFIG/buildkitd.default.toml`
+* `$DOCKER_CONFIG/buildx/buildkitd.default.toml`
+* `~/.docker/buildx/buildkitd.default.toml`
```
jedevc, line 85: Is there a way to run without a configuration file at all, if the default one exists? I suppose you'd just point the config file to `/dev/null`?

jedevc (approved):
Aha, I will very much end up using this a lot :tada:

docs/reference/buildx_create.md
jedevc: I think somehow calling out that we have loaded a default would be useful - I can definitely imagine it being frustrating to create a default, forget about it, and then be confused about why new builders are acting a certain way.
Could we attach messages like this to stderr, which won't affect command substitutions like that?

thaJeztah: We need to look at adopting XDG in general (where possible). ISTR there's still some "gaps" in the specification, and if we adopt, we must look at (e.g.) `macOS` (what's the most logical location? "mac native directories" (`~/Library/xxx`) or "Linux(ish)" directories (XDG / `~/.docker`)


Oh! Found it; https://github.com/moby/moby/issues/20693


Question; should this config be "rootless" aware? I recall the `pkg/homedir` package in Moby takes that into account; https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/pkg/homedir/homedir_linux.go#L80-L93

crazy-max: > We need to look at adopting XDG in general

Yeah would like too but atm we rely on docker cli conf path if `BUILDX_CONFIG` is not specified: https://github.com/docker/buildx/blob/98439f7f08eae1b619663ea5e527a5f84894feb7/util/confutil/config.go#L22

So I would say it's already supported in https://github.com/docker/cli/blob/master/cli/config/config.go which relies on https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/pkg/homedir/homedir_linux.go#L80-L93.

crazy-max: Also instead of using

https://github.com/docker/buildx/blob/98439f7f08eae1b619663ea5e527a5f84894feb7/util/confutil/config.go#L22

we could directly use https://github.com/docker/cli/blob/1b922a47c1bb0dda1239a20616e533dfd16d8ac8/cli/config/config.go#L62-L66

tonistiigi (approved):

docs/reference/buildx_create.md
crazy-max: Added a warning log msg if default config is being used which is streamed to stderr so it will work if called out with another command.

jedevc (approved):

PR #1123, opened 2022-05-12
ci: update gha bundle

crazy-max:
follow-up #1098 

[many PRs](https://github.com/docker/buildx/pulls?q=is%3Apr+is%3Aopen+label%3Abot) have been opened after merging #1098. this PR updates at once our docker actions to latest major, as well as the official actions. also remove the godev workflow which is not necessary anymore and also pin external/untrusted actions being used (dependabot still track sha linked to semver releases so it's still usable this way).

this will automatically close the dependabot PRs currently opened when merged. it's just to avoid approving and merging them one by one to begin with that will fill up our pipeline jobs.


.github/workflows/build.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           file: ./coverage/coverage.txt
       -
         name: Expose GitHub Runtime
-        uses: crazy-max/ghaction-github-runtime@v1
+        uses: crazy-max/ghaction-github-runtime@94085cef04e835448ac610090128515e781b4939
```
jedevc, line 46: Why is this changed from a tag -> commit hash? It looks like the same commit https://github.com/crazy-max/ghaction-github-runtime/releases/tag/v1 

(and also a couple of other places below)

crazy-max: Yes as I said:

> also pin external/untrusted actions being used (dependabot still track sha linked to semver releases so it's still usable this way).

Because Git tags are mutable we don't want to take the risk to fetch unchecked changes.

jedevc (approved):

PR #1128, opened 2022-05-13
rm: display removed builder and error if removing default builder

crazy-max:
was looking at allowing removing multiple builders with a single invocation of `rm` command and also fixing removing timed out runners not in running state with `--all-inactive` but before that small enhancements:

* display error message if trying to remove default builder

atm if "default" is used and try to remove it, command exits without saying it cannot be removed. this PR fixes it and now error if we try to do so:

```shell
$ docker buildx use default
$ docker buildx rm # or docker buildx rm default
ERROR: default builder cannot be removed
```

* display name of removed builder

we don't log anything if the builder has been effectively removed. It's useful in case we don't specify the builder so we know what has been removed:

```shell
$ docker buildx create --use
nostalgic_gould
$ docker buildx rm # or docker buildx rm nostalgic_gould
nostalgic_gould removed
```

also adds another commit to uppercase error level to match the one of logrus.


commands/rm.go
```diff
@@ -60,6 +60,9 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
 		return err
 	}
 	if ng != nil {
+		if ng.Name == "default" {
```
tonistiigi, line 64: Is this "default" case or builder with driver "docker"

commands/rm.go
```diff
@@ -67,7 +68,11 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
 		if err := txn.Remove(ng.Name); err != nil {
 			return err
 		}
-		return err1
+		if err1 != nil {
+			return err1
+		}
+		fmt.Printf("%s removed\n", ng.Name)
```
tonistiigi, line 74: stderr?

commands/rm.go
crazy-max: Hum yes so in my mind it is the default builder, whatever driver we use because looking at standalone mode without docker, "default" builder is created after using `ls`:

```dockerfile
FROM alpine
COPY --from=docker/buildx-bin / /usr/bin/
```

```
$ docker buildx build --tag buildx --load .
$ docker run --rm -it buildx sh
/ # apk add tree

/ # tree -a /root
/root
└── .ash_history

0 directories, 1 file

/ # buildx ls
NAME/NODE DRIVER/ENDPOINT  STATUS PLATFORMS
default * docker-container
  default default          Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

/ # tree -a /root
/root
├── .ash_history
└── .docker
    └── buildx
        ├── .lock
        ├── current
        ├── defaults
        └── instances

4 directories, 3 files

/ # buildx rm default

/ # buildx ls
NAME/NODE DRIVER/ENDPOINT  STATUS PLATFORMS
default * docker-container
  default default          Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

I guess this is linked to the store pkg?

crazy-max: > builder with driver "docker"

Added another condition to disallow removing context builders which is not working atm anyway and should be done with `docker context rm`.

jedevc (approved):

commands/rm.go
```diff
@@ -43,30 +44,41 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
...
+	ctxbuilders, err := dockerCli.ContextStore().List()
 	if err != nil {
 		return err
 	}
-	if ng != nil {
-		err1 := rm(ctx, dockerCli, in, ng)
-		if err := txn.Remove(ng.Name); err != nil {
-			return err
+	for _, cb := range ctxbuilders {
+		if ng.Name == "default" && len(ng.Nodes) == 1 && ng.Nodes[0].Endpoint == cb.Name {
```
tonistiigi, line 68: Not sure why there is a "default" name check in here. Shouldn't it check that `ng.Driver == "docker"` ? What does the "default" mean in here?

Didn't we already have an error suggesting to use `docker context` somewhere or am I mixing it up with some other case.

tonistiigi: `docker-container` is not the default ever, it always requires an instance to be created and therefore can always be removed.

crazy-max: as discussed I switched to the cond on the driver name.

tonistiigi (approved):

PR #1134, opened 2022-05-17
Remote driver connhelpers

jedevc:
This patch pulls in the `ssh` and `docker-container` connhelpers from buildkit to allow connections over `ssh` and to managed `docker-containers`.

I was thinking we pull in at least the `docker-container` and `ssh` helpers, and maybe the `kubepod` one?

I've also attempted to update the buildkit dependency to pull in https://github.com/moby/buildkit/pull/2843 - though in doing so that's also bumped some other things, so not entirely sure whether that's the right way to go.


tonistiigi:
> I was thinking we pull in at least the docker-container and ssh helpers, and maybe the kubepod one?

Yes, include the k8s one too.

go.mod
```diff
@@ -5,8 +5,8 @@ go 1.17
 require (
 	github.com/compose-spec/compose-go v1.2.4
 	github.com/containerd/console v1.0.3
-	github.com/containerd/containerd v1.6.3-0.20220401172941-5ff8fce1fcc6
```
AkihiroSuda, line 8: not sure this change is needed in this PR

jedevc: buildkit master has updated to containerd v1.6.4, I think that's why this gets bumped? https://github.com/moby/buildkit/blob/master/go.mod#L11

cmd/buildx/main.go
```diff
@@ -21,6 +21,10 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/openstack"
 
+	_ "github.com/moby/buildkit/client/connhelper/dockercontainer"
```
tonistiigi, line 24: These should probably go to remote driver pkg. The driver expects them to be always loaded.

docs/guides/remote-builder.md
```diff
@@ -115,6 +115,16 @@ $ docker buildx create \
   tcp://buildkitd.default.svc:1234
 ```
 
+Alternatively, we could use the `docker-container://` URL scheme to connect
+through the underlying docker daemon:
```
tonistiigi, line 119: > through the underlying docker daemon

"to the buildkit container without specifying a port."

tonistiigi (approved):

PR #1137, opened 2022-05-18
Imagetools multiple repositories

### LINKED ISSUE #486: imagetools create should support copying images between repositories
It's [currently not possible to create a new image based on source images with multiple repositories](https://github.com/docker/buildx/blob/d1a46faf8475564fbb27c9228b79a806e59b8f65/commands/imagetools/create.go#L79):

```
multiple repositories currently not supported, found map[docker.io/crazymax/diun:{} localhost:5000/name/app:{}]
```

jedevc:
Fixes #486

This PR introduces support for multiple repositories: this allows creating manifest lists with manifests pulled from multiple other locations.

This is split into two commits:
- The first refactors surrounding code to ensure that multiple repositories are recognized - this has changed some of the API surface, making the private `src` struct public to allow keeping the oci descriptor next to where it can be fetched from.
- The second to actually copy the manifests around to the new repository, and enable the multiple repositories feature

### DIFF
--- a/commands/imagetools/create.go
+++ b/commands/imagetools/create.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -9,6 +10,7 @@ import (
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/store/storeutil"
 	"github.com/docker/buildx/util/imagetools"
+	"github.com/docker/buildx/util/progress"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/distribution/reference"
 	"github.com/moby/buildkit/util/appcontext"
@@ -25,6 +27,7 @@ type createOptions struct {
 	tags         []string
 	dryrun       bool
 	actionAppend bool
+	progress     string
 }
 
 func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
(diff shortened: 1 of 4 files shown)


tonistiigi:
Could we add some progress logs so the user can see what is going on when layer transfers happen? Ideally, we could just reuse the build progressbar and show that transfers the same way we see on image pulls during build. For now, just printing logs like "copying blob sha256: from .. to .. & done` would do as well.

I don't this currently does cross-repo mounting but it should be quite easy to add support for it. Supposedly the label for the origin of the blob gets missing somewhere. https://github.com/containerd/containerd/blob/main/remotes/handlers.go#L330

jedevc:
Have got the cross-repo mounting working now in a `wip` commit on top - I think the cleanest solution is to add an `annotateDistributionSourceHandler` to `contentutil.CopyChain` in buildkit, by having it copy the annotation attached to the top level descriptor to the lower level descriptors. Then, the resolver's `Copy` function should just be able to attach it to the manifest descriptor.

Let me know, and I can actually open a PR on buildkit, instead of hacking around with the `vendor/` directory :see_no_evil:

tonistiigi:
Yes, go ahead with the PR to buildkit.

vendor/github.com/moby/buildkit/util/contentutil/copy.go
```diff
@@ -75,7 +76,7 @@ func CopyChain(ctx context.Context, ingester content.Ingester, provider content.
 		}
 	})
 	handlers := []images.Handler{
-		images.ChildrenHandler(provider),
+		annotateDistributionSourceHandler(images.ChildrenHandler(provider), desc.Annotations),
```
tonistiigi, line 79: Maybe the annotated descriptor should be already provided by the `content.Provider` implementation?

crazy-max:
We should review this example in build-push-action repo when 0.9 is out with this PR: https://github.com/docker/build-push-action/blob/master/docs/advanced/copy-between-registries.md

jedevc:
Have rebased onto master, this should be good to merge.

crazy-max (approved):
LGTM

commands/imagetools/create.go
```diff
@@ -180,23 +180,46 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
...
 	for _, t := range tags {
-		if err := r.Push(ctx, t, desc, dt); err != nil {
-			return err
-		}
-		fmt.Println(t.String())
+		t := t
+		eg.Go(func() error {
+			return progress.Wrap(fmt.Sprintf("pushing %s", t.String()), pw.Write, func(sub progress.SubLogger) error {
+				eg2, _ := errgroup.WithContext(ctx)
+				for _, s := range srcs {
```
tonistiigi, line 195: Shouldn't this only happen when src and destination repo are actually different?

commands/imagetools/create.go
```diff
@@ -78,18 +81,21 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
...
 		return errors.Errorf("no repositories specified, please set a reference in tag or source")
 	}
-	if len(repos) > 1 {
-		return errors.Errorf("multiple repositories currently not supported, found %v", repos)
-	}
 
-	var repo string
-	for r := range repos {
-		repo = r
+	var defaultRepo *string
```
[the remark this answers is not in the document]
jedevc: `repos` is only used to determine the `defaultRepo`. I think some sort of collection of all the repo names is required, since we only derive a default if all the other repo names present are the same.

We could potentially have a `bool` to track if we found a different repo that doesn't match the previously selected default during iteration, but I think that's potentially more complex than just tracking all the found repo names and then having a `len()`.

tonistiigi (approved):

PR #1140, opened 2022-05-22
docs: bake guides and refactor reference

crazy-max:
follow-up https://github.com/docker/buildx/pull/928#pullrequestreview-866904260

creates guides for bake as our reference docs for this command is quite huge and I think bake deserves its own section.

added a commit using my fork in https://github.com/docker/docker.github.io/pull/14644/commits/b44504072b1540118205d2ceafd7761574bf2ea5 to see how it looks. Here is a preview link: https://deploy-preview-14644--docsdocker.netlify.app/build/bake/file-definition/

I plan to add more concrete examples and use cases. let me now if the structure sounds good to you. cc @tonistiigi @dockertopia 


docs/guides/bake/file-definition.md
```diff
@@ -0,0 +1,363 @@
...
+directory in the following order:
+
+* `docker-compose.yml`
+* `docker-compose.yaml`
+* `docker-bake.json`
+* `docker-bake.override.json`
+* `docker-bake.hcl`
+* `docker-bake.override.hcl`
+
+A target reflects a single docker build invocation with the same options that
```
tonistiigi, line 17: This should start with an intro like "Inside a bake file you can define group, target and variable blocks". 

Also mention that "when you invoke `bake` you specify what targets/groups you want to build. If no arguments is specified group/target named "default" will be built"

docs/guides/bake/file-definition.md
```diff
@@ -0,0 +1,363 @@
...
+> `inherits` service field is also not supported. Use [YAML anchors](https://docs.docker.com/compose/compose-file/#fragments)
+> to reference other services.
+>
+> Specifying variables or global scope attributes is not yet supported for
+> compose files.
+{: .warning }
+
+## Remote definition
+
+You can also use a remote `git` bake definition:
```
tonistiigi, line 167: You can also build bake files directly from a remote Git repository or HTTPS URL.

docs/guides/bake/file-definition.md
```diff
@@ -0,0 +1,363 @@
...
+## Global scope attributes
+
+You can define global scope attributes in HCL/JSON and use them for code reuse
+and setting values for variables. This means you can do a "data-only" HCL file
+with the values you want to set/override and use it in the list of regular
+output files.
+
+```hcl
+# docker-bake.hcl
+variable "FOO" {
```
tonistiigi, line 294: Have not explained variables yet.

docs/guides/bake/file-definition.md
```diff
@@ -0,0 +1,363 @@
...
+#8 0.136 -rw-rw-rw-    1 root     root          1893 Jul 27 18:31 poule.yml
+#8 0.136 drwxrwxrwx    7 root     root          4096 Jul 27 18:31 scripts
+#8 0.136 drwxrwxrwx    3 root     root          4096 Jul 27 18:31 service
+#8 0.136 drwxrwxrwx    2 root     root          4096 Jul 27 18:31 templates
+#8 0.136 drwxrwxrwx   10 root     root          4096 Jul 27 18:31 vendor
+#8 0.136 -rwxrwxrwx    1 root     root          9620 Jul 27 18:31 vendor.conf
+#8 0.136 /bin/sh: stop: not found
+```
+
+## Global scope attributes
```
tonistiigi, line 285: I maybe makes sense to explain this together with `--set`. Section like "Configuring Bake builds" that covers vars/attrs and `--set`?

dockertopia (approved):
LGTM.

docs/guides/bake/configuring-build.md
```diff
@@ -0,0 +1,150 @@
+---
+title: "Configuring builds"
+keywords: build, buildx, bake, buildkit, hcl, json
+---
+
```
tonistiigi, line 5: This should start with an intro, e.g. "Bake supports loading build definition from files. But sometimes you need even more flexibility to configure this definition. For this use case, you can define variables inside the bake files that can be set by the user with environment variables or by attribute definitions in other bake files. If you wish to change a specific value for a single invocation you can use a `--set` flag."

docs/guides/bake/compose-xbake.md
```diff
@@ -0,0 +1,125 @@
+---
+title: "Extension field with Compose"
```
tonistiigi, line 2: Let's just call it "Building Compose files" and add an extra intro block explaining that you can just build compose files directly and how each service definition inside the compose files maps to a target (could have a simple example for this as well).

docs/guides/bake/file-definition.md
tonistiigi: ^

tonistiigi (commented):
If you want, you can merge as-is and follow up on the next PRs.

docs/guides/bake/index.md
```diff
@@ -0,0 +1,39 @@
...
+request. If multiple files are specified they are all read and configurations are
+combined.
+
+We recommend using HCL files as its experience is more aligned with buildx UX
+and also allows better code reuse, different target groups and extended features.
+
+## Next steps
+
+* [File definition](file-definition.md)
+* [Configuring builds](configuring-build.md)
```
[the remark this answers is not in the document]
crazy-max: We define variables in "File definition" so I think it's fine? But the "HCL variables and functions" doesn't look quite right and should only be about HCL user-defined functions as you said. I made some changes about it.

docs/guides/bake/configuring-build.md
crazy-max: Added an intro and also moved some bits from HCL user-defined funcs specific to build configuration in this page.

docs/guides/bake/compose-xbake.md
crazy-max: There is already a basic example in "File definition" for compose files. Also in "Target" specification we already have a note saying "In the case of compose files, each service corresponds to a target.".

This "Extension field with Compose" is very specific to the case where some fields are not supported in the compose spec.

Do you think we should move the [Compose file](https://deploy-preview-14644--docsdocker.netlify.app/build/bake/file-definition/#compose-file) section to this page so we don't have many things around about compose but on the same page?

tonistiigi: I don't think this provides enough context for the new reader. They already need to know that they are looking for `x-bake` to find this useful. You can leave the example in the file definition but there is much more to explain in here.

docs/guides/bake/index.md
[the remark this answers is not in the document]
tonistiigi: grouping variables with "configuring" and functions with the rest of the function section makes more sense to me. Yeah, the title should be just `HCL functions` as there are stdlib functions as well(should we list them?).

docs/guides/bake/compose-xbake.md
crazy-max: Added some context for building compose files. In a follow-up we could add new use cases to show how to use compose build fields from the specification and how they are translated to target with bake (secrets, cache).

tonistiigi (approved):
👍 Let's get this in and iterate from here.

PR #1153, opened 2022-06-08
bake: support compose build tags and secrets env

crazy-max:
Adds support for compose build `tags` field: https://github.com/compose-spec/compose-spec/blob/master/build.md#tags

Also takes into account the new `environment` field added in compose secrets: https://github.com/compose-spec/compose-spec/blob/master/spec.md#secrets-top-level-element

Needs to vendor compose-go https://github.com/compose-spec/compose-go/compare/v1.2.4...v1.2.7


bake/compose.go
```diff
@@ -229,7 +230,9 @@ func composeToBuildkitSecret(inp compose.ServiceSecretConfig, psecret compose.Se
 	}
 
 	var bkattrs []string
-	if inp.Source != "" {
+	if psecret.Environment != "" {
+		bkattrs = append(bkattrs, "id="+psecret.Environment)
```
tonistiigi, line 234: Looks like this should use `env=` (and leave id as-is?)

crazy-max: Can be [one-of `id` or `env` right](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#env)? I can change to `env` ofc if it makes more sense.

PR #1155, opened 2022-06-09
bake: support compose build cache_to

### LINKED ISSUE #1154: Bake does not support `build.cache_to`
[Compose-spec defines this field](https://github.com/compose-spec/compose-spec/blob/master/build.md#cache_to) but bake doesn't seem to see it

```yaml
services:
  web:
    build:
      context: .
      cache_to:
        - type=local,dest=path/to/cache
      cache_from:
        - type=local,dest=path/to/cache
```

is read as:

```json
{
  "group": {
    "default": {
      "targets": [
        "web"
      ]
    }
  },
  "target": {
    "web": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "cache-from": [
        "type=local,dest=path/to/cache"
      ]
    }
  }
```

This is maybe because `compose` doesn't really support it and bake uses the parser? It should really just be an alias or replacement for `build.x-bake.cache-to` I think?

crazy-max:
fixes #1154 

### DIFF
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -100,6 +100,7 @@ func ParseCompose(dt []byte) (*Config, error) {
 					return val, ok
 				})),
 				CacheFrom:   s.Build.CacheFrom,
+				CacheTo:     s.Build.CacheTo,
 				NetworkMode: &s.Build.Network,
 				Secrets:     secrets,
 			}
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -187,25 +187,25 @@ func (t *Target) composeExtTarget(exts map[string]interface{}) error {
 	}
 
 	if len(xb.Tags) > 0 {
-		t.Tags = append(t.Tags, xb.Tags...)
+		t.Tags = dedupString(append(t.Tags, xb.Tags...))
 	}
 	if len(xb.CacheFrom) > 0 {
-		t.CacheFrom = append(t.CacheFrom, xb.CacheFrom...)
+		t.CacheFrom = dedupString(append(t.CacheFrom, xb.CacheFrom...))
 	}
 	if len(xb.CacheTo) > 0 {
-		t.CacheTo = append(t.CacheTo, xb.CacheTo...)
+		t.CacheTo = dedupString(append(t.CacheTo, xb.CacheTo...))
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
I'd maybe add a test where it  + `x-bake.cache-to` are both defined

ciaranmcnulty:
Can you remember why this was necessary? https://github.com/docker/buildx/blame/59e2e24847e8b659b5cd944958afc49ba115d1f8/bake/compose.go#L153

Is the same necessary for cache_to?

crazy-max:
> Can you remember why this was necessary? https://github.com/docker/buildx/blame/59e2e24847e8b659b5cd944958afc49ba115d1f8/bake/compose.go#L153
> 
> Is the same necessary for cache_to?

I recall we didn't want to mix-up fields from compose and `x-bake`. I think we should merge them now that cache_to is supported by compose. I will add another commit for this case.

> I'd maybe add a test where it + `x-bake.cache-to` are both defined

Yes indeed.

bake/compose.go
```diff
@@ -149,7 +150,6 @@ func (t *Target) composeExtTarget(exts map[string]interface{}) error {
 					}
 				}
 			case "cache-from":
-				t.CacheFrom = []string{} // Needed to override the main field
```
tonistiigi, line 152: unrelated but what is this `val.([]interface{})` `res.(string)` construction. Looks like asking for panic.

tonistiigi: > Needed to override the main field

Can you explain why we are reverting this?

tonistiigi (commented):
Needs rebase

bake/compose.go
crazy-max: I see where it was discussed to replace the main field if it was defined for `cache-from`: https://github.com/docker/buildx/pull/721#discussion_r687950699. I think it makes sense to merge both fields. Do you have some concerns about it (expect the breaking change)?

> unrelated but what is this `val.([]interface{})` `res.(string)` construction. Looks like asking for panic.

https://github.com/docker/buildx/pull/1167

tonistiigi (approved):

PR #1165, opened 2022-06-14
Bake function crash

### LINKED ISSUE #1160: Bake segfaults on missing function params
Reproducer:

```hcl
function "foo" {
  result = "bar"
}
```

Output:
```
[+] Building 0.0s (0/0)                                                                                                                                                     
[signal SIGSEGV: segmentation violation code=0x2 addr=0x10 pc=0x10454a0a8]

goroutine 1 [running]:
github.com/docker/buildx/bake/hclparser.(*parser).resolveFunction(0x140000aad70, {0x140000553e6, 0x3})
        /src/bake/hclparser/hclparser.go:114 +0x108
github.com/docker/buildx/bake/hclparser.Parse({0x105e89a60, 0x14000113110}, {0x105e36020, 0x140005ef6e0, 0x105e32310}, {0x105cad080, 0x140005ef710})
        /src/bake/hclparser/hclparser.go:360 +0xcb4
```

Docker info:
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2)
  compose: Docker Compose (Docker Inc., v2.5.1)
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc., 0.6.0)
  scan: Docker Scan (Docker Inc., v0.17.0)
```

jedevc:
Fix #1160 for functions missing `params` attribute and similar issue with missing `result` attribute.


tonistiigi (commented):
Do I get it correct that in https://github.com/hashicorp/hcl/blob/main/ext/userfunc/README.md params is always required? If that is the case, then I think we should keep it the same and show error about "params not being set". We can propose upstream if they want to support this case and if they do then update both later as a feature update.

crazy-max:
Yes it's required atm and I agree that we can return an error for now to fix the segfault but looking at https://github.com/hashicorp/hcl/pull/455, I'm afraid a proposal to make it optional will take quite some time to be reviewed :sweat_smile:

jedevc:
So weirdly - it seems as if the current state of code works fine, and you can invoke functions without parameters.

We could just forbid it entirely, but I think it actually works now.

jedevc:
Upstream HCL definitely forbids non-present results and params - which are checked using a schema [here](https://github.com/hashicorp/hcl/blob/main/ext/userfunc/decode.go#L9-L24).

Because we don't use that schema, we need to provide other ways of checking, which are now done :tada:

crazy-max (approved):
LGTM

Issue #1104, opened 2022-05-10
Proposal: build debugging in buildx (interactive sessions)


tonistiigi:
This is a follow-up to BuildKit debugging issue https://github.com/moby/buildkit/issues/1472  so we can discuss the UX and next steps for features like https://github.com/moby/buildkit/pull/2813  . https://github.com/moby/buildkit/issues/1472  is mostly implemented in BuildKit v0.9 with additional signaling patches in v0.10 so unless we missed something no more (breaking) BuildKit changes should be needed.

All of this work does not need to end up in buildx repository, and some may be moved out later. I don't want opinionated dev features in buildctl that should be vendor-agnostic test tool, and I also don't want to maintain two similar, but different debugging stacks. Aside of that, code reuse is encouraged.

BuildKit issue concentrated on internal building blocks for this feature and you should read it first. In here, I'm proposing steps for incremental PRs to end up with a user-friendly debugging feature.

All flag names up to later discussion. Naming is hard.

PRs may be combined where it makes sense for review, but all the described steps should be quite independent.

### PR1

Add possibility to interactively run a process using the `NewContainer` API after the build has completed.

```
docker buildx build --invoke bash .
docker buildx build --invoke entrypoint=/usr/bin/cat,args=/etc/passwd,env=FOO=bar . # optional longform (not a requirement)
```

The build will run with the progressbar until completion. Progressbar will finish and container will be launched. Container redirects all stdio and signals. TTY is enabled if user enabled TTY for main process.

### PR2

Add "monitor mode" to the interactive process. When running an interactive process, the user can switch between to process io and monitor process io. Similar to QEMU monitor mode. In monitor mode they can issue additional commands.

In the very first PR only supported command may be "exit".

```
(buildx) exit
```

### PR3

Add "rollback" command to monitor mode. With this command the user can make modifications in the interactive container and when they issue "rollback" command they are brought back to the initial state.

### PR4

Refactor build command to invoke builds in a background process. This is important as we don't want the lifecycle of "debugging session" to be locked into a single process. A socket should be created under `~/.buildx` and even if the current process (unexpectedly) dies its state can be accessed again via the socket.

### PR5

Add `list` and `attach` commands to the monitor mode. List would show all the current active sessions (via the socket described in the previous section). Attach would make that session active in current process. If the session is already active in another process these processes would detach and got to the monitor mode.

### PR6

Add `exec` command to execute new processes in the same debug container. All processes are attachable as described in previous section.

### PR7

[...]

PR #1168, opened 2022-06-15
Add `--invoke` option to launch a container from the build result

ktock:
https://github.com/docker/buildx/issues/1104

This is a PoC of `--invoke` option to launch a container based on the build result.
Please give me feedbacks on the design of the monitor and CLI design.

This covers PR1 to PR3 of https://github.com/docker/buildx/issues/1104.

> PR1
> Add possibility to interactively run a process using the NewContainer API after the build has completed. The build will run with the progressbar until completion. Progressbar will finish and container will be launched. Container redirects all stdio and signals. TTY is enabled if user enabled TTY for main process.

> PR2
> Add "monitor mode" to the interactive process. When running an interactive process, the user can switch between to process io and monitor process io. Similar to QEMU monitor mode. In monitor mode they can issue additional commands. In the very first PR only supported command may be "exit".

> PR3
> Add "rollback" command to monitor mode. With this command the user can make modifications in the interactive container and when they issue "rollback" command they are brought back to the initial state. Add "reload" command to monitor mode. This will run the build again(now with possibly updated source) and invoke the shell again.

## Example:

```dockerfile
FROM ghcr.io/stargz-containers/busybox:1.32.0-org
RUN echo hello > /hello
```

### rollback

`Ctrl-a-c` switches IO between monitor and container with printing `Switched IO` message.
`rollback` rollbacks the container by restarting it.

```console
# buildx build --invoke sh /tmp/ctx
WARNING: No output specified with remote driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
[+] Building 0.5s (5/5) FINISHED                                                  
 => [internal] load build definition from Dockerfile                         0.1s
 => => transferring dockerfile: 111B                                         0.0s
 => [internal] load .dockerignore                                            0.0s
 => => transferring context: 2B                                              0.0s
 => [internal] load metadata for ghcr.io/stargz-containers/busybox:1.32.0-o  0.4s
 => [1/2] FROM ghcr.io/stargz-containers/busybox:1.32.0-org@sha256:bde48e17  0.0s
 => => resolve ghcr.io/stargz-containers/busybox:1.32.0-org@sha256:bde48e17  0.0s
 => CACHED [2/2] RUN echo hello > /hello                                     0.0s
/ # cat /hello
hello
/ # echo hi > /hi
/ # ls
bin    etc    hi     proc   sys    usr
dev    hello  home   root   tmp    var
/ # Switched IO

(buildx) rollback
(buildx) Switched IO

/ # ls
bin    dev    etc    hello  home   proc   root   sys    tmp    usr    var
```

### reload

`reload` reloads and builds the Dockerfile.

```dockerfile
FROM ghcr.io/stargz-containers/busybox:1.32.0-org
RUN echo hello2 > /hello2
```

```console
(buildx) reload
WARNING: No output specified with remote driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
[+] Building 0.4s (5/5) FINISHED                                                  
 => [internal] load build definition from Dockerfile                         0.0s
 => => transferring dockerfile: 113B                                         0.0s
 => [internal] load .dockerignore                                            0.1s
 => => transferring context: 2B                                              0.0s
 => [internal] load metadata for ghcr.io/stargz-containers/busybox:1.32.0-o  0.3s
 => [1/2] FROM ghcr.io/stargz-containers/busybox:1.32.0-org@sha256:bde48e17  0.1s
 => => resolve ghcr.io/stargz-containers/busybox:1.32.0-org@sha256:bde48e17  0.0s
 => CACHED [2/2] RUN echo hello2 > /hello2                                   0.0s
(buildx) Switched IO

/ # ls
bin     etc     home    root    tmp     var
dev     hello2  proc    sys     usr
/ # cat /hello2
hello2
```

### exit

`exit` exits the monitor.

```console
(buildx) exit
```

### long form

Container is configurable through the long form of `--invoke`.

```console
# buildx build --invoke entrypoint=sh,env=FOO=bar /tmp/ctx
WARNING: No output specified with remote driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
[+] Building 0.4s (5/5) FINISHED                                                  
 => [internal] load build definition from Dockerfile                         0.1s
 => => transferring dockerfile: 111B                                         0.0s
 => [internal] load .dockerignore                                            0.0s
 => => transferring context: 2B                                              0.0s
 => [internal] load metadata for ghcr.io/stargz-containers/busybox:1.32.0-o  0.3s
 => [1/2] FROM ghcr.io/stargz-containers/busybox:1.32.0-org@sha256:bde48e17  0.0s
 => => resolve ghcr.io/stargz-containers/busybox:1.32.0-org@sha256:bde48e17  0.0s
 => CACHED [2/2] RUN echo hello > /hello                                     0.0s
/ # env | grep FOO
FOO=bar
```

### TODOs

- Remaining concerns
  - Which sequence should be used to toggle IO? Currently C-a-c is used. Should it be configurable?
  - Should container IO be accessible and readable even after it exits?
  - Which driver and result should be used when multiple results are provided?

- TODOs
  - [x] Discuss and determine the design of monitor and CLI
  - [x] Add tests

Please tell me if I'm missing anything.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -615,7 +615,101 @@ func toSolveOpt(ctx context.Context, di DriverInfo, multiDriver bool, opt Option
 	return &so, releaseF, nil
 }
 
+// ContainerConfig is configuration for a container to run.
+type ContainerConfig struct {
+	ResultCtx *ResultContext
+	Args      []string
+	Env       []string
+	User      string
+	Cwd       string
+	Tty       bool
+	Stdin     io.ReadCloser
+	Stdout    io.WriteCloser
+	Stderr    io.WriteCloser
+}
+
+// ResultContext is a build result with the client that built it.
+type ResultContext struct {
+	Client *client.Client
+	Res    *gateway.Result
+}
+
+// Invoke invokes a build result as a container.
+func Invoke(ctx context.Context, cfg ContainerConfig) error {
+	if cfg.ResultCtx == nil {
+		return errors.Errorf("result must be provided")
+	}
+	c, res := cfg.ResultCtx.Client, cfg.ResultCtx.Res
+	_, err := c.Build(ctx, client.SolveOpt{}, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
+		if res.Ref == nil {
+			return nil, errors.Errorf("no reference is registered")
+		}
+		st, err := res.Ref.ToState()
+		if err != nil {
+			return nil, err
+		}
+		def, err := st.Marshal(ctx)
+		if err != nil {
+			return nil, err
+		}
+		imgRef, err := c.Solve(ctx, gateway.SolveRequest{
+			Definition: def.ToPB(),
+		})
+		if err != nil {
+			return nil, err
+		}
+		ctr, err := c.NewContainer(ctx, gateway.NewContainerRequest{
+			Mounts: []gateway.Mount{
+				{
+					Dest:      "/",
+					MountType: pb.MountType_BIND,
+					Ref:       imgRef.Ref,
+				},
+			},
+		})
+		if err != nil {
+			return nil, err
+		}
+		defer ctr.Release(ctx)
+		proc, err := ctr.Start(ctx, gateway.StartRequest{
+			Args:   cfg.Args,
+			Env:    cfg.Env,
+			User:   cfg.User,
+			Cwd:    cfg.Cwd,
+			Tty:    cfg.Tty,
+			Stdin:  cfg.Stdin,
+			Stdout: cfg.Stdout,
+			Stderr: cfg.Stderr,
+		})
+		if err != nil {
+			return nil, errors.Errorf("failed to start container: %v", err)
+		}
+		errCh := make(chan error)
+		doneCh := make(chan struct{})
+		go func() {
+			if err := proc.Wait(); err != nil {
+				errCh <- err
+				return
+			}
+			close(doneCh)
+		}()
+		select {
+		case <-doneCh:
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case err := <-errCh:
+			return nil, err
+		}
+		return nil, nil
+	}, nil)
+	return err
+}
+
 func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, docker DockerAPI, configDir string, w progress.Writer) (resp map[string]*client.SolveResponse, err error) {
+	return BuildWithResultHandler(ctx, drivers, opt, docker, configDir, w, nil)
+}
+
+func BuildWithResultHandler(ctx context.Context, drivers []DriverInfo, opt map[string]Options, docker DockerAPI, configDir string, w progress.Writer, resultHandleFunc func(driverIndex int, rCtx *ResultContext)) (resp map[string]*client.SolveResponse, err error) {
 	if len(drivers) == 0 {
 		return nil, errors.Errorf("driver required for build")
 	}
@@ -927,12 +1021,16 @@ func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, do
 						ch, done := progress.NewChannel(pw)
 						defer func() { <-done }()
 
+						cc := c
 						rr, err := c.Build(ctx, so, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
 							res, err := c.Solve(ctx, req)
 							if err != nil {
 								return nil, err
 							}
 							results.Set(resultKey(dp.driverIndex, k), res)
+							if resultHandleFunc != nil {
+								resultHandleFunc(dp.driverIndex, &ResultContext{cc, res})
+							}
 							return res, nil
 						}, ch)
 						if err != nil {
--- a/commands/build.go
+++ b/commands/build.go
@@ -4,14 +4,19 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/csv"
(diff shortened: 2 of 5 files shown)


ktock:
cc @tonistiigi

build/build.go
```diff
@@ -615,7 +615,102 @@ func toSolveOpt(ctx context.Context, di DriverInfo, multiDriver bool, opt Option
...
+type ResultContext struct {
+	Client   *client.Client
+	SolveOpt client.SolveOpt
+	Res      *gateway.Result
+}
+
+// Invoke invokes a build result as a container.
+func Invoke(ctx context.Context, cfg ContainerConfig) error {
+	if cfg.ResultCtx == nil {
+		return fmt.Errorf("result must be provided")
```
tonistiigi, line 641: use `pkg/errors` only in these repos.

build/build.go
```diff
@@ -615,7 +615,102 @@ func toSolveOpt(ctx context.Context, di DriverInfo, multiDriver bool, opt Option
...
+	Tty       bool
+	Stdin     io.ReadCloser
+	Stdout    io.WriteCloser
+	Stderr    io.WriteCloser
+}
+
+// ResultContext is a build result with the client that built it.
+type ResultContext struct {
+	Client   *client.Client
+	SolveOpt client.SolveOpt
```
tonistiigi, line 634: Do you even need `SolveOpt` in here? The whole build is based on `Res`. If `SolveOpt` contains output options it may be even wrong to reuse it here.

tonistiigi (commented):
This looks like a great start 👍 

> Which sequence should be used to toggle IO? Currently C-a-c is used. Should it be configurable?

I think it is pretty good atm. We should consider also including `ctrl-C`. Eg. when you are attached to container and press `Ctrl-C` it could bring you to monitor mode. Pressing it again could give a warning about how to attach back, and another one close the program.

C-a-c atm doesn't show prompt. Need to press enter to get to the prompt.

Maybe for something common like reload we could add a special shortcut in the future.

> Should container IO be accessible and readable even after it exits?

You mean stored in some file? I think we can think about that separately together with thinking if build logs need a similar feature.

> Which driver and result should be used when multiple results are provided?

Native arch of the first node for selected driver seems like the best option.

---

For reload, currently it triggers the whole build again. This could be faster by just running a new solve with the same gateway connection (same `Build()` callback). That would mean that, for example resolving a tag from registry for the base image is not needed between reloads. Or maybe we need two modes of reload.

Let's try to get this in quickly, mark the flags as experimental (maybe `BUILDX_EXPERIMENTAL` env?), and then start to interate on it instead of trying to get the first PR to be the perfect experience.

ktock:
> We should consider also including ctrl-C. Eg. when you are attached to container and press Ctrl-C it could bring you to monitor mode. Pressing it again could give a warning about how to attach back, and another one close the program.

Let's work on it in following-up patches. But it looks like it conflicts with the terminal keybind (for sending signals to the container)?

> C-a-c atm doesn't show prompt. Need to press enter to get to the prompt.

Fixed to show `(buildx)` prompt when the IO is swithced to the monitor.
Container doesn't still show the prompt just after `C-a-c`.
This is because `C-a-c` just toggle the IO stream but doesn't request the prompt to the attached container.
User need to explicitly send newline or something to get the prompt.

> You mean stored in some file? I think we can think about that separately together with thinking if build logs need a similar feature.

Sure.

> Native arch of the first node for selected driver seems like the best option.

Current implementation uses the lowest index of the selected drivers. Does this look good?

> `BUILDX_EXPERIMENTAL`

Added this.

build/build.go
ktock: Thank you for the comment. fixed this.

build/build.go
ktock: Thank you for the comment. fixed this.

commands/build.go
```diff
@@ -225,47 +231,135 @@ func runBuild(dockerCli command.Cli, in buildOptions) (err error) {
...
 
-	imageID, err := buildTargets(ctx, dockerCli, map[string]build.Options{defaultTargetName: opts}, in.progress, contextPathHash, in.builder, in.metadataFile)
+	imageID, res, err := buildTargets(ctx, dockerCli, map[string]build.Options{defaultTargetName: opts}, in.progress, contextPathHash, in.builder, in.metadataFile)
 	err = wrapBuildError(err, false)
 	if err != nil {
 		return err
 	}
 
+	if in.invoke != "" {
+		if os.Getenv("BUILDX_EXPERIMENTAL") != "1" {
```
tonistiigi, line 241: Better to just put the flag definition `flags.StringVar(&options.invoke` behind env.

tonistiigi (commented):
> But it looks like it conflicts with the terminal keybind (for sending signals to the container)?

Yes, indeed it seems hard to predict if the signal was for cancelling the process or shell. Maybe if the process dies it goes to the buildx prompt first. We can think about this more later.

> User need to explicitly send newline or something to get the prompt.

Could we just print enter for the user. I think we should either print enter or buffer the last line user saw and reprint it again. https://github.com/tonistiigi/vt100 could help if want to figure out what the last tty line is from a stream.

ktock:
> Could we just print enter for the user.

Fixed to automatically send a newline to the container on `C-a-c` as it seems the easiest solution as of now.
In the future we can consider richer UI like re-printing the last line.

> Better to just put the flag definition flags.StringVar(&options.invoke behind env.

Fixed this.

PR #1173, opened 2022-06-19
add support for oci-layout build-context

deitch:
Adds to buildx the oci-layout: build-context capabilities added to buildkit in [buildkit 2827](http://github.com/moby/buildkit/pull/2827).

Recommended by @tonistiigi . Remains incomplete, some things I wasn't sure about. Will comment inline.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
+	_ "crypto/sha256" // ensure digests can be computed
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -17,6 +18,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/content/local"
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/platforms"
 	"github.com/docker/buildx/driver"
@@ -1331,6 +1334,32 @@ func LoadInputs(ctx context.Context, d driver.Driver, inp Inputs, pw progress.Wr
 
 		if urlutil.IsGitURL(v.Path) || urlutil.IsURL(v.Path) || strings.HasPrefix(v.Path, "docker-image://") || strings.HasPrefix(v.Path, "target:") {
 			target.FrontendAttrs["context:"+k] = v.Path
+			continue
+		}
+
+		// handle OCI layout
+		if strings.HasPrefix(v.Path, "oci-layout://") {
+			pathAlone := strings.TrimPrefix(v.Path, "oci-layout://")
+			parts := strings.SplitN(pathAlone, "@", 2)
+			if len(parts) != 2 {
+				return nil, errors.Errorf("invalid oci-layout context %s, must be oci-layout:///path/to/layout@sha256:hash", v.Path)
+			}
+			localPath := parts[0]
+			dgst, err := digest.Parse(parts[1])
+			if err != nil {
+				return nil, errors.Wrapf(err, "invalid oci-layout context %s, does not have proper hash, must be oci-layout:///path/to/layout@sha256:hash", v.Path)
+			}
+			store, err := local.NewStore(localPath)
+			if err != nil {
+				return nil, errors.Wrapf(err, "invalid store at %s", localPath)
+			}
+			// now we can add it
+			if target.OCIStores == nil {
+				target.OCIStores = map[string]content.Store{}
+			}
+			target.OCIStores[k] = store
+
+			target.FrontendAttrs["context:"+k] = fmt.Sprintf("oci-layout:%s@%s", k, dgst.String())
 			continue
 		}
 		st, err := os.Stat(v.Path)
--- a/docs/reference/buildx_build.md
+++ b/docs/reference/buildx_build.md
@@ -106,7 +106,7 @@ More built-in build args can be found in [dockerfile frontend docs](https://gith
 Define additional build context with specified contents. In Dockerfile the context can be accessed when `FROM name` or `--from=name` is used.
 When Dockerfile defines a stage with the same name it is overwritten.
 
-The value can be a local source directory, container image (with docker-image:// prefix), Git or HTTP URL.
+The value can be a local source directory, [local OCI layout compliant directory](https://github.com/opencontainers/image-spec/blob/main/image-layout.md), container image (with docker-image:// prefix), Git or HTTP URL.
 
 Replace `alpine:latest` with a pinned one:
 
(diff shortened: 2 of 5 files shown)


deitch:
Pulling in the updated buildkit lib updated a whole bunch of other things. go mod worked its magic. It all worked and compiled, so I didn't complain.

build/build.go
```diff
@@ -1235,6 +1238,33 @@ func LoadInputs(ctx context.Context, d driver.Driver, inp Inputs, pw progress.Wr
...
+			localPath := fmt.Sprintf("/%s", reference.Path(named))
+			store, err := local.NewStore(localPath)
+			if err != nil {
+				return nil, errors.Wrapf(err, "invalid store at %s", localPath)
+			}
+			// now we can add it
+			target.OCIStores = map[string]content.Store{
+				k: store,
+			}
+			target.FrontendAttrs["context:"+k] = "oci-layout:" + k
```
deitch, line 1265: This didn't seem to make sense to me, but I wasn't sure what else we needed to put here, if anything? We need to have the name of the context, and a unique store ID. The name gets mapped to the store in `target.OCIStores` (2 lines above), and then the context to the unique store ID here. So should this be like I did here?

build/build.go
```diff
@@ -1235,6 +1238,33 @@ func LoadInputs(ctx context.Context, d driver.Driver, inp Inputs, pw progress.Wr
...
+		// handle OCI layout
+		if strings.HasPrefix(v.Path, "oci-layout://") {
+			pathAlone := strings.TrimPrefix(v.Path, "oci-layout://")
+			// need to add "dummy.io", because reference.Parse*() funcs require we do not start with a '/'
+			named, err := reference.ParseNormalizedNamed(fmt.Sprintf("dummy.io%s", pathAlone))
+			if err != nil {
+				return nil, errors.Wrapf(err, "invalid oci-layout context %s, must be oci-layout:///path/to/layout@sha256:hash", v.Path)
+			}
+			// check that we had a digest
+			digested, ok := named.(reference.Digested)
```
deitch, line 1251: I wasn't sure where this goes. In the client test in buildkit [here](https://github.com/moby/buildkit/blob/a6a114a1a476c99c2501aa34811159b849df4005/client/client_test.go#L1669-L1689), we pulled the digest (which I can get from `digested.Digest()`) and put it in `llb.OCILayout(storeID, digest)`, which is used in the `client.Solve(ctx, def, opts)` in `def`.

Where do we put the digest here in buildx so it knows that the build-context `foo` refers not only to the OCI store, but to a specific digest?

build/build.go
```diff
@@ -1235,6 +1238,33 @@ func LoadInputs(ctx context.Context, d driver.Driver, inp Inputs, pw progress.Wr
...
+				return nil, errors.Wrapf(err, "invalid oci-layout context %s, must be oci-layout:///path/to/layout@sha256:hash", v.Path)
+			}
+			// now ensure that we start with a `/`
+			localPath := fmt.Sprintf("/%s", reference.Path(named))
+			store, err := local.NewStore(localPath)
+			if err != nil {
+				return nil, errors.Wrapf(err, "invalid store at %s", localPath)
+			}
+			// now we can add it
+			target.OCIStores = map[string]content.Store{
```
tonistiigi, line 1262: This needs to add a store, not replace the whole map

build/build.go
```diff
@@ -1235,6 +1238,33 @@ func LoadInputs(ctx context.Context, d driver.Driver, inp Inputs, pw progress.Wr
 			target.FrontendAttrs["context:"+k] = v.Path
 			continue
 		}
+
+		// handle OCI layout
+		if strings.HasPrefix(v.Path, "oci-layout://") {
+			pathAlone := strings.TrimPrefix(v.Path, "oci-layout://")
+			// need to add "dummy.io", because reference.Parse*() funcs require we do not start with a '/'
+			named, err := reference.ParseNormalizedNamed(fmt.Sprintf("dummy.io%s", pathAlone))
```
tonistiigi, line 1246: Rather than this hack just split by `@` and then `digest.Parse` maybe.

build/build.go
tonistiigi, line 1265: The value in here nees to include digest that you are expecting in https://github.com/moby/buildkit/blob/master/frontend/dockerfile/builder/build.go#L892

docs/reference/buildx_build.md
```diff
@@ -126,6 +126,30 @@ FROM alpine
 COPY --from=project myfile /
 ```
 
+Source an image from a local [OCI layout compliant directory](https://github.com/opencontainers/image-spec/blob/main/image-layout.md):
+
+```console
+$ docker buildx build --build-context foo=oci-layout:///path/to/local/layout@sha256:abcd12345 .
```
tonistiigi, line 132: I think we should allow this input without the digest as well in here. If no digest is set it can look up `index.json` inside this dir and see if it has one descriptor and then set it as the expected digest. If there is ambiguity then it can produce an error.

tonistiigi (commented):
In addition

build/build.go
deitch: Good point. It should check if the map is nil, if so create it, and then add it either way.

build/build.go
deitch: One ugly hack vs another? I was _so_ disappointed than `reference.Parse` returns a thing that does not have component parts that can be looked up, like `url.Parse()` does.

build/build.go
deitch: Ah, that is it. I went through our buildkit PR, and could not recall which part did this. That helps.

docs/reference/buildx_build.md
deitch: I am hesitant about this, as it wades into a deep unresolved issue around what actually should be in `index.json`. I have seen those that _only_ include tags, others that include entire image refs, making the layout directory essentially into a full registry. I have implemented. both ways.

containerd does a bit of a hybrid: directory _is_ a full registry, but tag->manifest-hash is in the boltdb.

Let's get this working with the hash required, as is, and then let's deal with supporting hash-less - both in buildkit and in buildx - as a separate iteration.

build/build.go
deitch: Take a look, I _think_ I got it right.

deitch:
Most things resolved, one open to be sure that this is right, another subject to discussion.

deitch:
Fixed the go mod issues. Phew. 

I am seeing vet issues with:

```
# github.com/docker/buildx/bake
bake/bake.go:885:69: cannot use os.Stderr (variable of type *os.File) as type *configfile.ConfigFile in argument to authprovider.NewDockerAuthProvider
# github.com/docker/buildx/bake
vet: bake/bake.go:885:69: cannot use os.Stderr (variable of type *os.File) as *configfile.ConfigFile value in argument to authprovider.NewDockerAuthProvider
```

I didn't think I changed those, but perhaps downstream dependencies?

deitch:
Yeah, it looks like the definition of `github.com/moby/buildkit/session/auth/authprovider.NewDockerAuthProvider()` changed in buildkit from:

```go
func NewDockerAuthProvider(stderr io.Writer) session.Attachable {
	return &authProvider{
		config:      config.LoadDefaultConfigFile(stderr),
		seeds:       &tokenSeeds{dir: config.Dir()},
		loggerCache: map[string]struct{}{},
	}
}
```

to

```go
func NewDockerAuthProvider(cfg *configfile.ConfigFile) session.Attachable {
	return &authProvider{
		authConfigCache: map[string]*types.AuthConfig{},
		config:          cfg,
		seeds:           &tokenSeeds{dir: config.Dir()},
		loggerCache:     map[string]struct{}{},
	}
}
```

I got it.

build/build.go
```diff
@@ -1235,6 +1238,32 @@ func LoadInputs(ctx context.Context, d driver.Driver, inp Inputs, pw progress.Wr
...
+			if err != nil {
+				return nil, errors.Wrapf(err, "invalid store at %s", localPath)
+			}
+			// now we can add it
+			if target.OCIStores == nil {
+				target.OCIStores = map[string]content.Store{}
+			}
+			target.OCIStores[k] = store
+
+			target.FrontendAttrs["context:"+k] = fmt.Sprintf("oci-layout:%s@sha256:%s", k, hash.Encoded())
```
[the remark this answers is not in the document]
tonistiigi: Did you forget to push? Looks the same as before.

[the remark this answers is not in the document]
deitch: Doh!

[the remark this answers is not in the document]
deitch: Done

deitch:
Figured out the failing CI test, updated, should be good now.

deitch:
How done stand here @tonistiigi ?

deitch:
Been almost a month. I just rebased.

Is there anything else we need to get this merged in?

docs/reference/buildx_build.md
```diff
@@ -126,6 +126,30 @@ FROM alpine
 COPY --from=project myfile /
 ```
 
+Source an image from a local [OCI layout compliant directory](https://github.com/opencontainers/image-spec/blob/main/image-layout.md):
```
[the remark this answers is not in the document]
deitch: Done. Back to you @tonistiigi

tonistiigi (approved):

PR #1180, opened 2022-06-22
Update link to High-level build options with Bake

rdmolony:
Current link is invalid, so I replaced it with the High-level build options
from https://docs.docker.com/engine/reference/commandline/buildx_bake/


crazy-max:
Reference documentation is published on docs web site. Recent changes to our documentation on the master branch is not (yet) synced with docs website as there is on going work. This will be fixed with https://github.com/docker/docker.github.io/pull/14905. In the meantime you can directly read the current stable docs for bake at https://docs.docker.com/engine/reference/commandline/buildx_bake/

Issue #1186, opened 2022-06-26
"dial tcp: lookup" "no such host" for kubernetes node, when combining with docker-driver node (TLS DOCKER_HOST)


jobcespedes:
## Summary
Error about "dial tcp: lookup" "no such host" for kubernetes node, when combining with docker-driver node (TLS DOCKER_HOST):
`kubernetes:///multiarch-builder?deployment=&kubeconfig= error during connect: Get "http://%2Fmultiarch-builder%3Fdeployment=&kubeconfig=/v1.24/info": dial tcp: lookup /multiarch-builder?deployment=&kubeconfig=: no such host`

## Reproduce
```
# docker context create --docker host=tcp://public_dns:2376,ca="/tmp/ca.pem",cert="/tmp/cert.pem",key="/tmp/key.pem" arm64
docker buildx create --append --name=multiarch-builder --platform=linux/amd64 --driver=kubernetes amd64
## At this point, there is no error with amd64 node
docker buildx create --append --name multiarch-builder --driver docker-container --platform linux/arm64 arm64
## At this point, there is an error with amd64 node, no error with arm64 node
```
## Info
Buildx version: `github.com/docker/buildx v0.8.2 6224def4dd2c3d347eee19db595348c50d7cb491`
Docker info:
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc., v0.8.2)
  pushrm: Push Readme to container registry (Christian Korneck, 1.7.0)

Server:
 Containers: 3
  Running: 3
  Paused: 0
  Stopped: 0
 Images: 2315
 Server Version: 20.10.16
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: journald
 Cgroup Driver: systemd
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc io.containerd.runc.v2 io.containerd.runtime.v1.linux
 Default Runtime: runc
 Init Binary: /usr/libexec/docker/docker-init
 containerd version: 
 runc version: eae618f-dirty
 init version: 
 Security Options:
  seccomp
   Profile: default
  selinux
 Kernel Version: 5.17.12-100.fc34.x86_64
[...]

PR #1188, opened 2022-06-28
Introduce new errors for unsupported driver behaviors

jedevc:
Around discussion in https://github.com/docker/buildx/issues/1186, a few issues with buildx drivers came up. These scenarios are unsupported, but the user isn't warned, for example:
- The kubernetes endpoint cli arg will be overwritten internally, so the user's specification is ignored
- Appending a new node to a builder with a different driver than the original is unsupported (and atm will overwrite the previous builder's driver)
- ~~Creating builders with the same name as context builders gives a warning, but appending and deleting doesn't.~~

This PR should resolve the above issues, and provide a more helpful local debugging experience.

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -320,6 +320,9 @@ func createCmd(dockerCli command.Cli) *cobra.Command {
 }
 
 func csvToMap(in []string) (map[string]string, error) {
+	if len(in) == 0 {
+		return nil, nil
+	}
 	m := make(map[string]string, len(in))
 	for _, s := range in {
 		csvReader := csv.NewReader(strings.NewReader(s))
--- a/commands/create.go
+++ b/commands/create.go
@@ -61,32 +61,6 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
 		}
 	}
 
-	buildkitHost := os.Getenv("BUILDKIT_HOST")
-
-	driverName := in.driver
-	if driverName == "" {
-		if len(args) == 0 && buildkitHost != "" {
-			driverName = "remote"
-		} else {
-			var arg string
-			if len(args) > 0 {
-				arg = args[0]
-			}
-			f, err := driver.GetDefaultFactory(ctx, arg, dockerCli.Client(), true)
-			if err != nil {
-				return err
-			}
-			if f == nil {
-				return errors.Errorf("no valid drivers found")
-			}
-			driverName = f.Name()
-		}
-	}
-
-	if driver.GetFactory(driverName, true) == nil {
-		return errors.Errorf("failed to find driver %q", in.driver)
-	}
-
 	txn, release, err := storeutil.GetStore(dockerCli)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


commands/create.go
```diff
@@ -61,30 +61,15 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
...
+		return err
 	}
-
-	if driver.GetFactory(driverName, true) == nil {
-		return errors.Errorf("failed to find driver %q", in.driver)
+	for _, c := range contexts {
+		if c.Name == in.name {
+			logrus.Warnf("instance name %q already exists as context builder", in.name)
+			break
+		}
```
crazy-max, line 72: this should only be checked on creation. you can still leave or append a context builder to an existing one.

commands/create.go
```diff
@@ -101,49 +86,64 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
...
+			return errors.Errorf("existing instance for %q but no append mode, specify --node to make changes for existing instances", name)
+		}
+		if driverName != ng.Driver {
+			return errors.Errorf("existing instance for %q but has mismatched driver %q", name, ng.Driver)
 		}
 	}
 
+	if driver.GetFactory(driverName, true) == nil {
+		return errors.Errorf("failed to find driver %q", in.driver)
+	}
```
crazy-max, line 138: empty driver is a correct behavior afaik. default is `docker-container` if empty.

jedevc: Yup, at this point we have set `driverName` to `docker-container` if empty using `driver.GetDefaultFactory`
So `driverName` should always be a non-empty string here.

jedevc: oh wait i see, `in.driver` should be `driverName` here

commands/create.go
```diff
@@ -61,32 +61,19 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
...
-				return errors.Errorf("no valid drivers found")
+	if !in.actionLeave && !in.actionAppend {
+		contexts, err := dockerCli.ContextStore().List()
+		if err != nil {
+			return err
+		}
+		for _, c := range contexts {
+			if c.Name == in.name {
+				logrus.Warnf("instance name %q already exists as context builder", in.name)
+				break
```
crazy-max, line 72: Any reason to move that logic before https://github.com/docker/buildx/blob/3cf549a7f7ee015cabe99d4a08e4b9aabfc83ccc/commands/create.go#L96-L102

If `in.name` is empty it could still be the same name as an existing context when generated.

jedevc:
Have added another check, to error on attempting to edit nodes with `--node`, while including `--driver-opt` (which are otherwise silently discarded).

crazy-max (approved):
LGTM, PTAL @tonistiigi

tonistiigi (commented):
For the node update, if we are allowing flags then I don't see a reason why we would error on driveropt/config.

For the mixed drivers, looks like an oversight as well. Although this one is harder to fix. I don't see a specific reason why this should not be allowed. All the driver instances are passed separately to the build function. They do not need to be of the same type there.

jedevc:
> For the node update, if we are allowing flags then I don't see a reason why we would error on driveropt/config.

I think at the very least we should warn here - we allow the `--driver-opt` flag on a node update, but it will be silently discarded. The flag has to be accepted here since it's the same `create` subcommand that we use to create drivers in the first place. 

Will update the PR to do this.

> For the mixed drivers, looks like an oversight as well. Although this one is harder to fix. I don't see a specific reason why this should not be allowed. All the driver instances are passed separately to the build function. They do not need to be of the same type there.

True. However, the way we store the drivers in config makes this harder to do - the driver is attached to the `NodeGroup`, so to do this we'd need to push down the driver config into each individual `Node`, not quite sure what implications that would have, but seems like an interesting possibility :eyes:

store/nodegroup.go
```diff
@@ -70,6 +79,15 @@ func (ng *NodeGroup) Update(name, endpoint string, platforms []string, endpoints
...
 		}
+		if do != nil {
+			n.DriverOpts = do
+		}
+		if configFile != "" {
+			for k, v := range files {
+				n.Files[k] = v
+			}
+		}
+		logrus.Warn("new settings may not be used until builder is restarted")
```
tonistiigi, line 90: Only warn if there was a change in one of the fields that require restart (I believe all if conditions except the platform one).

tonistiigi (approved):

PR #1197, opened 2022-07-05
bake: add new --sync-output flag

### LINKED ISSUE #1089: Option to only tag images at end of bake
Currently `bake --push` will tag images in the remote registry as they are completed. This can be an issue if later images in the same high-level build error out.

Either the default behaviour could change, or a CLI option could be added that forces tagging to only happen at the end - I don't think pushing new layers would need to be held back.

This example results in a `foo` tag but not the `bar` one:

```Dockerfile
FROM alpine:3.15 as foo
FROM alpine:3.15 as bar
RUN sleep 30 && exit 1
```

```hcl
group "default" {
  targets = ["foo", "bar"]
}

target "foo" {
  target = "foo"
  tags = [
    "ciaranmcnulty/test:foo"
  ]
}

target "bar" {
  target = "bar"
  tags = [
    "ciaranmcnulty/test:bar"
  ]
}
```

jedevc:
Fixes #1089. Depends on https://github.com/moby/buildkit/pull/2947.

This patch introduces a new syncable output option, which ensures that all builds finish simultaneously in the solver, so that no outputs are completed independently of each other. This allows bake to easily express the notion that either all builds should succeed and output or none of them should.


build/build.go
```diff
@@ -737,11 +738,11 @@ func Invoke(ctx context.Context, cfg ContainerConfig) error {
...
 }
 
-func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, docker DockerAPI, configDir string, w progress.Writer) (resp map[string]*client.SolveResponse, err error) {
-	return BuildWithResultHandler(ctx, drivers, opt, docker, configDir, w, nil, false)
+func Build(ctx context.Context, drivers []DriverInfo, opt map[string]Options, docker DockerAPI, configDir string, w progress.Writer, syncOutputs bool) (resp map[string]*client.SolveResponse, err error) {
+	return BuildWithResultHandler(ctx, drivers, opt, docker, configDir, w, nil, syncOutputs, false)
 }
 
-func BuildWithResultHandler(ctx context.Context, drivers []DriverInfo, opt map[string]Options, docker DockerAPI, configDir string, w progress.Writer, resultHandleFunc func(driverIndex int, rCtx *ResultContext), allowNoOutput bool) (resp map[string]*client.SolveResponse, err error) {
+func BuildWithResultHandler(ctx context.Context, drivers []DriverInfo, opt map[string]Options, docker DockerAPI, configDir string, w progress.Writer, resultHandleFunc func(driverIndex int, rCtx *ResultContext), syncOutputs bool, allowNoOutput bool) (resp map[string]*client.SolveResponse, err error) {
```
crazy-max, line 745: We should look at refactoring these funcs in a follow-up

docs/reference/buildx_bake.md
```diff
@@ -25,6 +25,7 @@ Build from a file
 | [`--pull`](#pull) |  |  | Always attempt to pull all referenced images |
 | `--push` |  |  | Shorthand for `--set=*.output=type=registry` |
 | [`--set`](#set) | `stringArray` |  | Override target value (e.g., `targetpattern.key=value`) |
+| `--sync-output` |  |  | Ensure all builds complete before beginning output |
```
crazy-max, line 28: Maybe `--sync` would be enough? There are some cases where we might not need to output anything but keep build result synced (`type=cacheonly`)?

```suggestion
| `--sync` |  |  | Ensure all builds complete before returning result |
```

Let's also add a simple example here for this flag.

crazy-max (commented):
I pushed an extra commit as simple test case but we should look at implementing integration tests (pretty much like BuildKit with [testutil](https://github.com/moby/buildkit/tree/master/util/testutil/integration)).

docs/reference/buildx_bake.md
ciaranmcnulty: I think `--sync` isn't enough, you want to hint it is about synching the 'stuff at the end'

build/build.go
jedevc: Yup, have avoided making too many changes since #buildx server

jedevc:
Have updated with the new Evaluate API introduced in https://github.com/moby/buildkit/pull/3137.

Still not sure about the flag name, but the functionality should be correct now (have tested with buildkit before+after the evaluate API merged to check the StatFile fallback works correctly).

jedevc:
Adding to the v0.10 milestone, think this would be good to get in there.

Revisiting this after a while, I wonder if maybe we might want to make this the default behavior? Or some other way that doesn't involve needing to add lots more flags, IMO not needing a lot of extra flags on the cli is one of the selling points of bake :thinking:

jedevc:
Ping @tonistiigi @crazy-max any thoughts on the overall approach here? I'm a bit hesitant around adding new flags to the CLI though :thinking:

Possible approaches:
- Just add the new flag (clutters the command line)
- Make the synced outputs default (slows down quite a few builds)
- Maybe add a `sync` key or similar to each group in bake, all targets part of that group would be synced? We could allow setting it at the command line using `--set`.

Nithos:
Does the current action support `group` entries currently into the `targets` input? If so adding a `sync` to a group definition is a fairly elegant solution that avoids the flags. I am assuming that the group would get decomposed and automatically parallelize the process rather than having to manually deconstruct the list of targets present within the group and run through the GH matrix.

crazy-max:
Having a `sync` field for `group` is a good idea. Hence my suggestion to have a less opinionated flag named `--sync` instead of `--sync-output`.

I'm thinking about another use case now. Should we have a `fail-fast` strategy that cancels all in-progress targets in the group if any target fails? I think this is the current behavior.

jedevc:
Yeah, if we have a sync field, having the ability to sync at different points would be good:

- `fail-fast` is the current behavior
- `output` syncs at the output stage, like this patch
- `fail-slow` (naming is hard) is the opposite of `fail-fast`, all builds are run to completion (though we might *also* want to have variants of this that sync at output or after).

PhilippHomann:
Are there any plans on merging this one or implementing the previously mentioned `--sync` flag instead?
Think this is a potential fix of #1668 and also we would like to have our images only pushed to the registry if all targets succeed.

PR #1206, opened 2022-07-12
buildx: log errors in initializing builders

jedevc:
Previously, errors within the driver config would not be reported to the user until they tried to use the driver, even though they are easily accessible from the node group info.

For example:

```console
# before
$ buildx create --bootstrap --name=kube --driver=kubernetes --driver-opt=namespace=buildkit,replicas=4,loadbalance=test
kube

# after
$ buildx create --bootstrap --name=kube --driver=kubernetes --driver-opt=namespace=buildkit,replicas=4,loadbalance=test
ERROR: failed to initialize builder kube (kube0): invalid loadbalance "test"
kube
```

This is in a similar vein to #1188

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -134,6 +134,11 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
 		}
 	}
 
+	ngOriginal := ng
+	if ngOriginal != nil {
+		ngOriginal = ngOriginal.Copy()
+	}
+
 	if ng == nil {
 		ng = &store.NodeGroup{
 			Name: name,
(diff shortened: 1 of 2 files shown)


crazy-max (approved):
LGTM

jedevc:
Done, this should now properly error and attempt to rollback to the previous configuration if it failed.

tonistiigi (approved):
Letting this in for now but we should refactor this so that we can validate the driver inputs without creating the nodegroup instance and then deleting it. There is a race condition in the current implementation when the rollback runs.

Issue #1201, opened 2022-07-07
bake-hcl: add function to access current time / build time


tnaroska:
I was looking for a way to access the current time (or build time) in my docker-bake.hcl file but couldn't find an obvious way to do so.
There are some cty stdlib functions that work on timestamps (e.g. `timeAdd`, `formatDate`) but no obvious way to get the current time. I would suggest adding a wrapper function for `time.Now()` or maybe a builtin variable to access the build time.

Environment: 
```
buildx   v0.8.2
buildkit v0.10.3
```

PR #1214, opened 2022-07-18
bake: add timestamp function

jedevc:
Fix https://github.com/docker/buildx/issues/1201.

Terraform includes a `timestamp` function to get the current time. go-cty has imported a number of the timestamp functions to it's standard library in https://github.com/zclconf/go-cty/pull/37, however, this was one was not included:

> All but one of the "impure" functions fit into the cryptographic and filesystem categories I covered already anyway. The remaining one is timestamp, and I think it would be overkill to establish a new codebase just for that so maybe again it's sufficient to just copy-paste it into another application that needs it, as long as that application is prepared to deal with a function that doesn't return a stable result.

The result isn't fully stable, but user's should expect that, and can assign the result to a top-level variable and use that as a consistent timestamp throughout the file.

### DIFF
--- a/bake/hclparser/stdlib.go
+++ b/bake/hclparser/stdlib.go
@@ -1,12 +1,15 @@
 package hclparser
 
 import (
+	"time"
+
 	"github.com/hashicorp/go-cty-funcs/cidr"
 	"github.com/hashicorp/go-cty-funcs/crypto"
 	"github.com/hashicorp/go-cty-funcs/encoding"
 	"github.com/hashicorp/go-cty-funcs/uuid"
 	"github.com/hashicorp/hcl/v2/ext/tryfunc"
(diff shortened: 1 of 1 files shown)


crazy-max (approved):
SGTM

tonistiigi:
Shouldn't a "timestamp" be a number?

crazy-max:
> Shouldn't a "timestamp" be a number?

Strictly speaking I agree, but terraform returns a UTC timestamp string in RFC 3339 format: https://www.terraform.io/language/functions/timestamp

jedevc:
Personally I'd rather keep consistency with how terraform does this (since we also import a bunch of functions from their stdlib). But I do agree, I think as a variable would make more sense (even if it's not a number, but a RFC 3339 string).

tonistiigi (approved):

PR #1215, opened 2022-07-18
docs(bake): fix target fields and show type

### LINKED ISSUE #1192: Docs - bake and secrets
The only reference in the docs is that `secrets` is an option, but that's not correct it's `secret`

There are no concrete examples so you'll have to guess the syntax :/

crazy-max:
fixes #1192 

There is also some confusion around the list of valid target fields available and the overrides with the `--set` flag:

```hcl
target foo {
  secret = ["..."]
}
```

```console
$ docker buildx bake --set *.secrets=...
```

In the bake definition we have `secret` and in the override `secrets` for example.

I think we should align that in a follow-up if it makes sense:

* `no-cache-filter` > `no-cache-filters`
* `output` > `outputs`
* `secret` > `secrets`
* `ssh` > `sshs` (don't like this one :neutral_face:)

### DIFF
--- a/docs/guides/bake/file-definition.md
+++ b/docs/guides/bake/file-definition.md
@@ -43,23 +43,26 @@ $ docker buildx bake webapp-dev
 Complete list of valid target fields available for [HCL](#hcl-definition) and
 [JSON](#json-definition) definitions:
 
-* `args`
-* `cache-from`
-* `cache-to`
-* `context`
-* `contexts`
-* `dockerfile`
-* `inherits`
-* `labels`
-* `no-cache`
-* `no-cache-filter`
-* `output`
-* `platform`
-* `pull`
-* `secrets`
-* `ssh`
-* `tags`
-* `target`
+| Name                | Type   | Description                                                                                                                                     |
+|---------------------|--------|-------------------------------------------------------------------------------------------------------------------------------------------------|
+| `inherits`          | List   | [Inherit build options](#merging-and-inheritance) from other targets                                                                            |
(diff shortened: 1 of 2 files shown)


crazy-max:
> Is there a reason we can't have them all match identically? So `secret` in bake, would translate to `secret` at the cli `--set`. To me that makes more sense 🤔 The plurals might get confusing, especially since we already have distinct fields `context` and `contexts`.

Yes sgtm, we should do that in a follow-up

PTAL @tonistiigi

tonistiigi (approved):
Could we also add some description and maybe a link to the field documentation under build command?

crazy-max:
> Could we also add some description and maybe a link to the field documentation under build command?

Done, let me know if it LGTY.

docs/guides/bake/file-definition.md
```diff
@@ -437,3 +440,6 @@ $ docker buildx bake "https://github.com/tonistiigi/buildx.git#remote-test" "htt
 #8 0.136 -rwxrwxrwx    1 root     root          9620 Jul 27 18:31 vendor.conf
 #8 0.136 /bin/sh: stop: not found
 ```
+
+
+[^1]: See [`docker buildx build` command](https://docs.docker.com/engine/reference/commandline/buildx_build/) reference docs.
```
jedevc, line 445: image

That's a bit weird :confused:

An alternative would be to copy an exact link on each `--flag` to the exact ref in the buildx docs, but I don't think every flag has a section :cry:

jedevc (approved):

docs/guides/bake/file-definition.md
crazy-max: Yes does not look so great :persevere:

Will change that

jedevc (approved):

PR #1218, opened 2022-07-19
ci: upstream docs conformance validation

crazy-max:
~needs https://github.com/docker/docker.github.io/pull/14905 or https://github.com/docker/docker.github.io/pull/15194~

adds a job to check if links are still valid with upstream docs repository.

### DIFF
--- a/.github/workflows/docs-upstream.yml
+++ b/.github/workflows/docs-upstream.yml
@@ -1,0 +1,118 @@
+# this workflow runs the remote validate bake target from docker/docker.github.io
+# to check if yaml reference docs and markdown files used in this repo are still valid
+# https://github.com/docker/docker.github.io/blob/98c7c9535063ae4cd2cd0a31478a21d16d2f07a3/docker-bake.hcl#L34-L36
+name: docs-upstream
+
+concurrency:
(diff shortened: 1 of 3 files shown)


jedevc (approved):

tonistiigi (commented):
Does this tool actually pull down all the URLs (comments could help).

If it does could we filter this to only run on master or if `*.md` changes etc.

crazy-max:
> Does this tool actually pull down all the URLs (comments could help).

Yes it will run the validation for the whole docs repo as cross links could be broken outside our scope so we can take that into account and react accordingly (here or on docs repo).

> If it does could we filter this to only run on master or if `*.md` changes etc.

Done

___

Also added reference yaml docs that was not covered in the validation and we have a dead link: https://github.com/docker/buildx/runs/7622740288?check_suite_focus=true#step:8:667

```
#23 166.5 - ./_site/engine/reference/commandline/buildx_build/index.html
#23 166.5   *  internally linking to /build/guides/color-output/, which does not exist (line 1040)
#23 166.5      <a href="/build/guides/color-output/">Color output controls guide</a>
```

Fixed with last commit.

tonistiigi (approved):

Issue #1220, opened 2022-07-20
Reopen #59


larsqa:
Reopen issue #59 since:
- In 2019, [it's been mentioned that `--load` can't be used temporarily for multi-platform builds.](https://github.com/docker/buildx/issues/59#issuecomment-488902807)
- 2022 and the issue still exists, [as well the docs are misleading mentioning that it works](https://docs.docker.com/engine/reference/commandline/buildx_build/#output )
- It's been asked in the discussions to re-open the issue but it seems like they are ignored.

thaJeztah:
The docs mentions;

> Will automatically load the single-platform build result to docker images.

The detail is in "single-platform" here (as the docker daemon storage does not allow for multiple architectures for the same image to be stored).

Wondering if that section would be clearer if we added a not that it cannot be used if multiple platforms are specified

larsqa:
@jedevc Thanks for the quick reply! Glad to hear things are moving forward 😃

@thaJeztah Would agree - I had to read through many Error message descriptions and what they actually meant, before I figured out that there's an issue with the default `--load` flag.

thaJeztah:
@larsqa if you're interested in contributing; the source for those docs are in this repository; https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#load

(happy to give it a go myself as well if you want)

larsqa:
@thaJeztah I might not have enough knowledge to understand the full scope of the issue - but I provided my go, from a higher abstract level, from my understanding. Not sure if the majority of readers have a better understanding and are interested in deeper details?

PR: https://github.com/docker/buildx/pull/1221

PR #1221, opened 2022-07-20
Update --load with a note highlight

larsqa:
Proposal after discussion of #1220 

Single platform for the `--load` option should be more highlighted, with the help of a Note?

### DIFF
--- a/docs/reference/buildx_build.md
+++ b/docs/reference/buildx_build.md
@@ -230,6 +230,9 @@ More info about cache exporters and available attributes: https://github.com/mob
 Shorthand for [`--output=type=docker`](#docker). Will automatically load the
 single-platform build result to `docker images`.
 
+> **Note**
+> `--load` is the default option during a `docker buildx build` command. These builds store images locally. At the moment, multi-platform builds are only possible using remote registries. The `--push` command pushes the images directly to a registry. Local multi-platform build support is part [part of our current roadmap](https://github.com/docker/roadmap/issues/371) and should be available soon.
+
 ### <a name="metadata-file"></a> Write build result metadata to the file (--metadata-file)
 
 To output build metadata such as the image digest, pass the `--metadata-file` flag.


tonistiigi (commented):
I don't understand that the first part of this comment is trying to say. And push is documented separately. If there is a note, it should just say that `docker load` command currently does not support loading multi-platform images so only a single platform can be specified for such builds.

larsqa:
> I don't understand that the first part of this comment is trying to say. And push is documented separately. If there is a note, it should just say that `docker load` command currently does not support loading multi-platform images so only a single platform can be specified for such builds.

That's fair.

The part in question:
> `--load` is the default option during a `docker buildx build` command. These builds store images locally. At the moment, multi-platform builds are only possible using remote registries.

From my point of view and understanding, I assume that 
- `--load` is the default option for the `buildx` command. I wasn't able to figure out any other logic default option for this setting. 
- `--load` is the common-expected behaviour known in Docker by storing images locally.

I felt that mentioning that, before explaining the `--push` alternative which has multi-arch support, would be helpful for less-knowledgable devs like me.

tonistiigi:
`--load` is not default(why would you set it then). You might be thinking about Docker driver without `--output`. 

Looking at https://github.com/docker/buildx/blob/e12827910c278b76b9bb1050e7157354db3b1f42/docs/reference/buildx_build.md#docker the multi-platform case is already described there. And it is linked from `--load`. We can update the wording there to make it more clear that a single platform works.

larsqa:
> --load is not default(why would you set it then). You might be thinking about Docker driver without --output.

Ohhh alright!

> Looking at https://github.com/docker/buildx/blob/e12827910c278b76b9bb1050e7157354db3b1f42/docs/reference/buildx_build.md#docker the multi-platform case is already described there. And it is linked from --load. We can update the wording there to make it more clear that a single platform works.

I looked at your linked resource, which mentions:
> Currently, multi-platform images cannot be exported with the docker export type. The most common usecase for multi-platform images is to directly push to a [registry](https://github.com/docker/buildx/blob/e12827910c278b76b9bb1050e7157354db3b1f42/docs/reference/buildx_build.md#registry) (see registry).

The "registry" this part links to mentions:
> The registry exporter is a shortcut for type=image,push=true.

Looking a bit further, the [Push the build result to a registry (--push)](https://github.com/docker/buildx/blob/e12827910c278b76b9bb1050e7157354db3b1f42/docs/reference/buildx_build.md#-push-the-build-result-to-a-registry---push) mentions:
> Shorthand for [--output=type=registry](https://github.com/docker/buildx/blob/e12827910c278b76b9bb1050e7157354db3b1f42/docs/reference/buildx_build.md#registry). Will automatically push the build result to registry.

As a reader, I get a feeling that there are now 3 ways to push to the registry in order to use multi-platform:
- `-- type=image,push=true`
- `--output=type=registry`
- `--push` (shorthand of above)

While I believe they all use the underlying process, it still confuses me as a reader.

Here are my propositions, then:
- As you mention,  mention that `docker load` does **currently** not support loading multi-platform images so only a single platform can be specified for such builds. 
- Add to `docker load` that pushing to a registry with `--push` allows for mulit-platform support.
- Improve the push/registry to allow for one option to push to a registry and shorten it with push?

thaJeztah:
>> --load is not default(why would you set it then). You might be thinking about Docker driver without --output.
> Ohhh alright!

Some related discussion on that in  https://github.com/docker/cli/pull/3676#issuecomment-1157622147

tonistiigi:
> Here are my propositions, then:

I don't really understand these.

> As you mention, mention that docker load does currently not support loading multi-platform images so only a single platform can be specified for such builds.

This is already documented. If you think you can improve its readability, go ahead, and we will check.

> Add to docker load that pushing to a registry with --push allows for mulit-platform support.

Load should not document push options. Push should document push options.

> Improve the push/registry to allow for one option to push to a registry and shorten it with push?

Don't understand what you mean here.

jedevc:
Closing this due to inactivity - @larsqa, if you want to re-open, please go ahead! :tada:

PR #1230, opened 2022-07-23
docs(guide): color output controls

crazy-max:
follow-up #1226

behavior lgtm:

image

image

but looks like the [example in README of BuildKit repo](https://github.com/moby/buildkit#color-output-controls) is not correct:

```console
$ BUILDKIT_COLORS="run=blue;cancel=yellow;warn=orange;error=123,12,0" docker buildx bake
WARN[0000] Could not parse BUILDKIT_COLORS component: run=blue;cancel=yellow;warn=orange;error=123,12,0  error="A valid entry must have exactly two fields"
```

should be `run=blue:cancel=yellow:warn=orange:error=123,12,0` as correct separator but then:

```console
$ BUILDKIT_COLORS="run=blue:cancel=yellow:warn=orange:error=123,12,0" docker buildx bake
WARN[0000] Unknown color value found in BUILDKIT_COLORS: warn=orange  error="Colors must be a name from the pre-defined list or a valid 3-part RGB value"
```

`orange` does not exist in pre-defined colors.

cc @spkane

### DIFF
--- a/docs/guides/color-output.md
+++ b/docs/guides/color-output.md
@@ -1,0 +1,24 @@
+---
+title: "Color output controls"
+description: "Modifying colors of progress output"
+keywords: build, buildx, buildkit
+---
+
+Buildx has support for modifying the colors that are used to output information
+to the terminal. You can set the environment variable `BUILDKIT_COLORS` to
+something like `run=123,20,245:error=yellow:cancel=blue:warning=white` to set
+the colors that you would like to use:
+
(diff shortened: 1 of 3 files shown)


tonistiigi (approved):

spkane:
@crazy-max Thank you for catching that. It was a last-minute change to the PR that removed the named `orange` color since it is not a standard 4-bit color like the other named colors.

PR #1241, opened 2022-07-29
docs: add minimal docker driver docs

jedevc:
Mostly for completeness, this patch adds basic documentation for the docker driver :tada: After this, we'll have guides for all 4 of our drivers.

### DIFF
--- a/docs/guides/drivers/docker.md
+++ b/docs/guides/drivers/docker.md
@@ -1,0 +1,50 @@
+# Docker driver
+
+The buildx docker driver is the default builtin driver, that uses the BuildKit
+server components built directly into the docker engine.
+
+No setup should be required for the docker driver - it should already be
+configured for you:
+
+```console
+$ docker buildx ls
+NAME/NODE       DRIVER/ENDPOINT      STATUS   BUILDKIT PLATFORMS
+default         docker                                 
+  default       default              running  20.10.17 linux/amd64, linux/386
+```
+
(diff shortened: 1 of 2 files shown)


crazy-max (approved):
Nice LGTM, PTAL @tonistiigi

docs/guides/drivers/docker.md
```diff
@@ -0,0 +1,45 @@
...
+default         docker                           
+  default       default         running 20.10.17 linux/amd64, linux/386
+desktop-linux * docker                           
+  desktop-linux desktop-linux   running 20.10.17 linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
+```
+
+Unlike the [other drivers](../index.md), builders using the docker driver
+cannot be manually created, and are automatically constructed from the
+available [Docker Contexts](https://docs.docker.com/engine/context/working-with-contexts/).
+Additionally, they cannot be configured to a specific BuildKit version, and
```
tonistiigi, line 35: explain that the configuration/version is set by the Docker engine.

docs/guides/drivers/docker.md
```diff
@@ -0,0 +1,45 @@
...
+cannot take any extra parameters.
+
```
tonistiigi, line 37: Add a sentence explaining that in order to add new Docker instances here user needs to create a `docker context` pointing to that instance, and then they will show up in the builders list.

PR #1250, opened 2022-08-02
docs: add correct definition of prune --all flag

jedevc:
The previous definition was the same as the docker images prune command and referenced dangling images, which isn't what the command does. This commit brings the command description more inline with the buildctl definition.

Additionally, add some more description of what the various flags do in our reference pages.

### DIFF
--- a/docs/reference/buildx_prune.md
+++ b/docs/reference/buildx_prune.md
@@ -11,7 +11,7 @@ Remove build cache
 
 | Name | Type | Default | Description |
 | --- | --- | --- | --- |
-| `-a`, `--all` |  |  | Remove all unused images, not just dangling ones |
+| `-a`, `--all` |  |  | Include internal/frontend images |
 | [`--builder`](#builder) | `string` |  | Override the configured builder instance |
 | `--filter` | `filter` |  | Provide filter values (e.g., `until=24h`) |
 | `-f`, `--force` |  |  | Do not prompt for confirmation |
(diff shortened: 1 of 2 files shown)


docs/reference/buildx_prune.md
```diff
@@ -21,6 +21,18 @@ Remove build cache
...
+Clears the build cache of the selected builder.
+
+You can finely control what cache data is kept using:
+
+- The `--filter=until=<duration>` flag to keep images that have been used in
+  the last `<duration>` time.
+- The `--keep-storage=<space>` flag to keep `<space>` bytes of data in the cache.
+- The `--all` flag to allow clearing internal helper images and frontend images
+  set using the `#syntax=` directive or the `BUILDKIT_SYNTAX` build argument.
+
```
crazy-max, line 35: I think we should also carry some bits from https://github.com/docker/cli/pull/2864

crazy-max (approved):

docs/reference/buildx_prune.md
jedevc: Let's do this in a follow-up, maybe adding that info into a new guide around cache management?

Would be a good place to tie together `prune` and `du`, though not sure how we'd link in the GC policy, I think that's managed by the docker daemon?

docs/reference/buildx_prune.md
```diff
@@ -21,6 +21,18 @@ Remove build cache
...
 
+## Description
+
+Clears the build cache of the selected builder.
+
+You can finely control what cache data is kept using:
+
+- The `--filter=until=<duration>` flag to keep images that have been used in
+  the last `<duration>` time.
+- The `--keep-storage=<space>` flag to keep `<space>` bytes of data in the cache.
```
tonistiigi, line 32: We should explain more what this `<space>` means in here as we don't expect people to write bytes. Eg. Common example would be `--keep-storage 5GB`. Same for the duration(what is probably seconds if no unit?).

docs/reference/buildx_prune.md
```diff
@@ -21,6 +21,26 @@ Remove build cache
...
+
+You can finely control what cache data is kept using:
+
+- The `--filter=until=<duration>` flag to keep images that have been used in
+  the last `<duration>` time.
+
+  `<duration>` is a duration string, e.g. `24h` or `2h30m`, with allowable
+  units of `(h)ours`, `(m)inutes` and `(s)econds`.
+
+- The `--keep-storage=<space>` flag to keep `<space>` bytes of data in the cache.
```
thaJeztah, line 36: Just a quick thought; perhaps `<size>` would be a good alternative, as we already use "size" in various places

thaJeztah (approved):
LGTM

PR #1252, opened 2022-08-02
Fix prune `--filter until=<duration>` option

### LINKED ISSUE #400: --filter "until=24h" not working
I'm using buildkit in my GitLab CI pipelines to build my images with build caches. To accelerate subsequent runs, I export and save the build cache from buildx after each job and import it before another. This works fine but leads to ever growing cache sizes. To clean things up, I tried running `docker buildx prune --filter "until=24h"` as suggested [here](https://docs.docker.com/engine/reference/commandline/builder_prune/) and judging from the [source code](https://github.com/docker/buildx/blob/master/commands/prune.go#L153) I assumed this would work with buildx too. But when using this filter, nothing gets removed. Using no filter leads to all caches being removed, even those, I just created with the latest build (and I need those to build the next image from cache).

jedevc:
Fixes #400.

Previously, when specifying the filter option with the until value, no cache would be cleaned, preventing users from clearing by time. This bug arises from passing the until field through into buildkit, where, on filtering, a non-existent field returns false for a match.

The fix is simple, as we build up our list of filters to pass to buildkit, we skip over the until key, so create a valid list of filters for buildkit.

### DIFF
--- a/commands/prune.go
+++ b/commands/prune.go
@@ -176,6 +176,10 @@ func toBuildkitPruneInfo(f filters.Args) (*client.PruneInfo, error) {
 
 	bkFilter := make([]string, 0, f.Len())
 	for _, field := range f.Keys() {
+		if field == filterKey {
+			continue
+		}
+
 		values := f.Get(field)
 		switch len(values) {
 		case 0:
--- a/commands/prune.go
+++ b/commands/prune.go
@@ -155,9 +155,9 @@ func toBuildkitPruneInfo(f filters.Args) (*client.PruneInfo, error) {
 	if len(untilValues) > 0 && len(unusedForValues) > 0 {
 		return nil, errors.Errorf("conflicting filters %q and %q", "until", "unused-for")
 	}
-	filterKey := "until"
+	untilKey := "until"
 	if len(unusedForValues) > 0 {
-		filterKey = "unused-for"
+		untilKey = "unused-for"
 	}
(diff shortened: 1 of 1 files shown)


crazy-max (approved):

crazy-max:
Would also close https://github.com/moby/moby/issues/43734 (cc @thaJeztah)

commands/prune.go
```diff
@@ -176,6 +176,10 @@ func toBuildkitPruneInfo(f filters.Args) (*client.PruneInfo, error) {
 
 	bkFilter := make([]string, 0, f.Len())
 	for _, field := range f.Keys() {
+		if field == filterKey {
```
tonistiigi, line 179: This is quite confusing to read. We should rename so that this reads `filter == untilKey`

tonistiigi (approved):

commands/prune.go
tonistiigi: The current one LGTM, but noticed that I messed up and actually meant to write `field == untilKey` here that would have made a simpler patch. No need to change anything with current PR.

thaJeztah:
> Would also close [moby/moby#43734](https://github.com/moby/moby/issues/43734) (cc @thaJeztah)

thx! Thinking of that; do you recall if only `docker builder prune` was affected, or also `docker system prune`? (As this is a client-side change only)

crazy-max:
@thaJeztah Yes only `builder prune`

PR #1253, opened 2022-08-02
ci: enhanced build workflow

crazy-max:
Split binaries job so now each platform is built on its own runner.

Before: ~19m

image

Now: ~3m

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,9 +22,63 @@ env:
   RELEASE_OUT: "./release-out"
 
 jobs:
-  build:
+  test:
     runs-on: ubuntu-latest
     steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          version: latest
+      -
+        name: Test
+        uses: docker/bake-action@v2
+        with:
+          targets: test
+          set: |
+            *.cache-from=type=gha,scope=test
+            *.cache-to=type=gha,scope=test
+      -
+        name: Upload coverage
+        uses: codecov/codecov-action@v3
+        with:
+          file: ./coverage/coverage.txt
+
+  prepare:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -22,46 +22,105 @@ env:
...
       -
-        name: Build binaries
+        name: Create matrix
+        id: platforms
         run: |
-          make release
-        env:
-          CACHE_FROM: type=gha,scope=release
-          CACHE_TO: type=gha,scope=release
+          echo ::set-output name=matrix::$(docker buildx bake binaries-cross --print | jq -cr '.target."binaries-cross".platforms')
```
crazy-max, line 62: Matrix is dynamically created by looking at supported platforms in `docker-bake.hcl`

.github/workflows/build.yml
```diff
@@ -22,46 +22,105 @@ env:
...
           file: ./coverage/coverage.txt
+
+  prepare:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.platforms.outputs.matrix }}
+    steps:
       -
-        name: Expose GitHub Runtime
-        uses: crazy-max/ghaction-github-runtime@906832f62b7baa936e3fbef72b029308af505ee7
```
jedevc, line 50: Can't quite work out where we using this before :eyes:, is this just a cleanup from existing code?

crazy-max: Oh we were using it because the[ `make release` step](https://github.com/docker/buildx/blob/603595559f64233efa3052a1e825bdee500233e5/.github/workflows/build.yml#L53-L54) runs outside a GitHub Action (`run:`) and therefore requires to expose GitHub runtime to the workflow with https://github.com/crazy-max/ghaction-github-runtime

crazy-max:
@tonistiigi As discussed, `bin-image` job will be skipped on PR as it's basically calling the same target as `binaries`. And `buildkit-edge` will build the `binaries` target instead of `binaries-cross` as it seems enough for this case. Let me know if LGTY.

tonistiigi (approved):

PR #1256, opened 2022-08-05
bake: contexts support with x-bake

### LINKED ISSUE #1107: Support multiple build contexts in x-bake
I can't see any way to do this in the docs / compose-test.go

crazy-max:
fixes #1107

### DIFF
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -151,10 +151,12 @@ type xbake struct {
 	Pull          *bool       `yaml:"pull,omitempty"`
 	NoCache       *bool       `yaml:"no-cache,omitempty"`
 	NoCacheFilter stringArray `yaml:"no-cache-filter,omitempty"`
+	Contexts      stringMap   `yaml:"contexts,omitempty"`
 	// don't forget to update documentation if you add a new field:
 	// docs/guides/bake/compose-file.md#extension-field-with-x-bake
 }
 
+type stringMap map[string]string
 type stringArray []string
 
 func (sa *stringArray) UnmarshalYAML(unmarshal func(interface{}) error) error {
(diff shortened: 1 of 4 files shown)


bake/compose.go
```diff
@@ -141,16 +141,17 @@ func flatten(in compose.MappingWithEquals) compose.Mapping {
...
+	CacheFrom     stringArray       `yaml:"cache-from,omitempty"`
+	CacheTo       stringArray       `yaml:"cache-to,omitempty"`
+	Secrets       stringArray       `yaml:"secret,omitempty"`
+	SSH           stringArray       `yaml:"ssh,omitempty"`
+	Platforms     stringArray       `yaml:"platforms,omitempty"`
+	Outputs       stringArray       `yaml:"output,omitempty"`
+	Pull          *bool             `yaml:"pull,omitempty"`
+	NoCache       *bool             `yaml:"no-cache,omitempty"`
+	NoCacheFilter stringArray       `yaml:"no-cache-filter,omitempty"`
+	Contexts      map[string]string `yaml:"contexts,omitempty"`
```
jedevc, line 154: Not sure if we should call this `stringMap` for consistency? (even though the only reason we have `stringArray` is for the custom `UnmarshalYAML` case with CSV parsing).

bake/compose.go
```diff
@@ -263,3 +267,21 @@ func composeToBuildkitSecret(inp compose.ServiceSecretConfig, psecret compose.Se
 
 	return strings.Join(bkattrs, ","), nil
 }
+
+func dedupMap(ms ...map[string]string) map[string]string {
```
jedevc, line 271: We should put this next to `dedupString` - maybe in a separate file if they don't both fit in a single file neatly?

`dedupString` is a weird name as well for that now we'd have `dedupMap`, maybe we could change it to `dedupSlice` or equivalent?

bake/compose.go
crazy-max: Good point to be consistent, I'm fine with it.

jedevc (approved):

tonistiigi (approved):

PR #1257, opened 2022-08-08
build: fix issues with leaving invoke containers running

tonistiigi:
Currently, when running `--invoke`, there exist the following issues:

When invoke container is not shut down by gracefully killing the container process itself then the container is left running in buildkit. This means that it also holds onto resources and cache records in `buildx du` do not become releasable. To reproduce, switch IO to monitor mode and then type "exit" or `ctrl-D`/SIGINT.

When switching IO to monitor mode and issuing "rollback" command the buildx process goes to 100% CPU (200% on repeat). Looks like there is a busy loop.

The release problem should be fixed on buildkit side as well as buildkit should handle misbehaving clients. I plan to fix it but we need this for old daemons anyway.

Tbh, I never figured out where the 100% CPU was coming from. Profiling CPU/Blocking was quite useless showing lots of work in stdlib `runtime.selectgo` but no call path. I tried to debug the new `for-select` code in monitor, but I couldn't confirm that it started to loop. But after I fixed the release bug the 100% CPU was gone as well.

@ktock 

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -643,7 +643,16 @@ func Invoke(ctx context.Context, cfg ContainerConfig) error {
 		return errors.Errorf("result must be provided")
 	}
 	c, res := cfg.ResultCtx.Client, cfg.ResultCtx.Res
-	_, err := c.Build(ctx, client.SolveOpt{}, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
+
+	mainCtx := ctx
+
+	_, err := c.Build(context.TODO(), client.SolveOpt{}, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
+		ctx, cancel := context.WithCancel(ctx)
+		go func() {
+			<-mainCtx.Done()
+			cancel()
+		}()
+
 		if res.Ref == nil {
 			return nil, errors.Errorf("no reference is registered")
 		}
@@ -673,7 +682,8 @@ func Invoke(ctx context.Context, cfg ContainerConfig) error {
 		if err != nil {
 			return nil, err
 		}
-		defer ctr.Release(ctx)
+		defer ctr.Release(context.TODO())
+
 		proc, err := ctr.Start(ctx, gateway.StartRequest{
 			Args:   cfg.Args,
 			Env:    cfg.Env,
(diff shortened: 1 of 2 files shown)


build/build.go
```diff
@@ -643,7 +643,16 @@ func Invoke(ctx context.Context, cfg ContainerConfig) error {
 		return errors.Errorf("result must be provided")
 	}
 	c, res := cfg.ResultCtx.Client, cfg.ResultCtx.Res
-	_, err := c.Build(ctx, client.SolveOpt{}, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
+
+	mainCtx := ctx
```
tonistiigi, line 647: Worth looking into this case in buildkit client later on so hacks like this are not required. It looks like that if this context gets called in a certain timeframe the `/solve` grpc call is canceled before the callback returns, causing some calls inside callback to fail with "no such job". `Build()` should never return if the callback has not returned, even if the grpc already goes down.

build/build.go
```diff
@@ -673,7 +682,8 @@ func Invoke(ctx context.Context, cfg ContainerConfig) error {
 		if err != nil {
 			return nil, err
 		}
-		defer ctr.Release(ctx)
+		defer ctr.Release(context.TODO())
```
tonistiigi, line 685: Otherwise, if `ctx` is already canceled then this does nothing.

monitor/monitor.go
```diff
@@ -38,9 +38,7 @@ func RunMonitor(ctx context.Context, containerConfig build.ContainerConfig, relo
 	}
 
 	// Start container automatically
-	go func() {
-		m.rollback(ctx, containerConfig)
-	}()
+	m.rollback(ctx, containerConfig)
```
tonistiigi, line 41: Not really part of the fix but seemed useless goroutine.

monitor/monitor.go
```diff
@@ -52,6 +50,10 @@ func RunMonitor(ctx context.Context, containerConfig build.ContainerConfig, relo
 		go func() {
 			defer close(doneCh)
 			defer in.Close()
+			go func() {
+				<-ctx.Done()
```
tonistiigi, line 54: This is to handle `appcontext` canceling from `sigint`. Otherwise this goroutine will keep running although everything else shuts down. Worth thinking through this goroutine structure again to make the cancel path more clear and avoid parallel goroutines with no defined shutdown order. Atm there is a chance for a small race if this happens when the initial `Invoke()` call is also in the middle of processing.

ktock (approved):
Thank you.

crazy-max (approved):

PR #1260, opened 2022-08-09
detect moby worker supports multiplatform feature through containerd snapshotter

ndeloof:


driver/docker/driver.go
```diff
@@ -54,11 +54,22 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
 }
 
 func (d *Driver) Features() map[driver.Feature]bool {
+	var useContainerdSnapshotter bool
+	ctx := context.Background()
+	c, err := d.Client(ctx)
+	if err == nil {
+		workers, _ := c.ListWorkers(ctx)
+		for _, w := range workers {
+			if _, ok := w.Labels["org.mobyproject.buildkit.worker.snapshotter"]; ok {
```
crazy-max, line 63: Could you use LabelSnapshotter instead: https://github.com/moby/buildkit/blob/86c33b66e176a6fc74b88d6f46798d3ec18e2e73/worker/worker.go#L50

Otherwise LGTM

tonistiigi (approved):

driver/docker/driver.go
tonistiigi: I think this is in a package we don't currently import

crazy-max: Ah yes and this one import quite a lot.

ndeloof: I tried to, but adding buildkit/vendor import require me to run `make vendor` and this brings a crazy amount of changes

crazy-max: @ndeloof Yes let's keep it like this in the meantime

crazy-max (approved):

PR #1261, opened 2022-08-09
bake: load .env file from working dir for compose files

### LINKED ISSUE #282: docker buildx bake should apply env variable from .env

### LINKED ISSUE #1135: Allow `.env` file override (equivalent to docker-compose build's `--env-file` option) when using bake
As a follow up to [#282](https://github.com/docker/buildx/issues/282)

Depending on our build type (dev/ci/prod) we use different versions of a `.env` file when running our builds with `docker-compose build`. Would it be possible to add a [`--env-file` option](https://docs.docker.com/compose/environment-variables/#using-the---env-file--option) to bake so it can provide the same functionality?

crazy-max:
fixes #282
carry https://github.com/docker/buildx/pull/489

Loads `.env` file if available in the working directory for compose files. The file format is the same as the compose library: https://github.com/compose-spec/compose-go/tree/master/dotenv.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -179,6 +179,21 @@ func dedupMap(ms ...map[string]string) map[string]string {
 	return res
 }
 
+func sliceToMap(env []string) (res map[string]string) {
+	res = make(map[string]string)
+	for _, s := range env {
+		kv := strings.SplitN(s, "=", 2)
+		key := kv[0]
+		switch {
+		case len(kv) == 1:
+			res[key] = ""
+		default:
+			res[key] = kv[1]
+		}
+	}
+	return
+}
+
 func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error) {
 	defer func() {
 		err = formatHCLError(err, files)
@@ -242,15 +257,22 @@ func ParseFile(dt []byte, fn string) (*Config, error) {
 }
 
 func ParseComposeFile(dt []byte, fn string) (*Config, bool, error) {
+	envs := sliceToMap(os.Environ())
+	if wd, err := os.Getwd(); err == nil {
+		envs, err = loadDotEnv(envs, wd)
+		if err != nil {
+			return nil, true, err
+		}
+	}
 	fnl := strings.ToLower(fn)
 	if strings.HasSuffix(fnl, ".yml") || strings.HasSuffix(fnl, ".yaml") {
-		cfg, err := ParseCompose(dt)
+		cfg, err := ParseCompose(dt, envs)
 		return cfg, true, err
 	}
 	if strings.HasSuffix(fnl, ".json") || strings.HasSuffix(fnl, ".hcl") {
 		return nil, false, nil
 	}
-	cfg, err := ParseCompose(dt)
+	cfg, err := ParseCompose(dt, envs)
 	return cfg, err == nil, err
 }
 
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -3,8 +3,10 @@ package bake
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
+	"github.com/compose-spec/compose-go/dotenv"
 	"github.com/compose-spec/compose-go/loader"
 	compose "github.com/compose-spec/compose-go/types"
 	"github.com/pkg/errors"
@@ -14,34 +16,18 @@ import (
 // errComposeInvalid is returned when a compose file is invalid
 var errComposeInvalid = errors.New("invalid compose file")
 
-func parseCompose(dt []byte) (*compose.Project, error) {
-	return loader.Load(compose.ConfigDetails{
+func ParseCompose(dt []byte, envs map[string]string) (*Config, error) {
+	cfg, err := loader.Load(compose.ConfigDetails{
 		ConfigFiles: []compose.ConfigFile{
 			{
 				Content: dt,
 			},
 		},
-		Environment: envMap(os.Environ()),
+		Environment: envs,
 	}, func(options *loader.Options) {
 		options.SkipNormalization = true
 		options.SkipConsistencyCheck = true
 	})
-}
-
-func envMap(env []string) map[string]string {
-	result := make(map[string]string, len(env))
-	for _, s := range env {
-		kv := strings.SplitN(s, "=", 2)
-		if len(kv) != 2 {
-			continue
-		}
-		result[kv[0]] = kv[1]
-	}
-	return result
-}
-
-func ParseCompose(dt []byte) (*Config, error) {
-	cfg, err := parseCompose(dt)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


bake/bake.go
```diff
@@ -179,15 +181,30 @@ func dedupMap(ms ...map[string]string) map[string]string {
...
 
-func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error) {
+func sliceToMap(env []string) (res map[string]string) {
+	res = make(map[string]string)
+	for _, s := range env {
+		kv := strings.SplitN(s, "=", 2)
+		key := kv[0]
+		switch {
+		case len(kv) == 1:
+			res[key] = ""
```
jedevc, line 191:
```suggestion
		switch len(kv) {
		case 1:
			res[key] = ""
```

bake/env_test.go
```diff
@@ -0,0 +1,69 @@
...
+			dt:       "",
+			expected: map[string]string{"FOO": "bar"},
+		},
+	}
+	for _, tt := range cases {
+		tt := tt
+		t.Run("", func(t *testing.T) {
+			var filename string
+			if tt.dt != "" {
+				envf, err := os.CreateTemp(tmpdir, "test.env")
```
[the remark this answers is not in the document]
crazy-max: I'm using it at the beginning https://github.com/docker/buildx/pull/1261/files#diff-07dfe15c3516d8fcfaaf136e940a26079b71e9de33ee378939a29d2c64dcbebbR12

commands/bake.go
```diff
@@ -190,6 +196,7 @@ func bakeCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags.BoolVar(&options.printOnly, "print", false, "Print the options without building")
 	flags.BoolVar(&options.exportPush, "push", false, `Shorthand for "--set=*.output=type=registry"`)
 	flags.StringArrayVar(&options.overrides, "set", nil, `Override target value (e.g., "targetpattern.key=value")`)
+	flags.StringVar(&options.envFile, "env-file", "", "Specify an alternate environment file")
```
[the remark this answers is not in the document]
crazy-max: Good point

bake/env.go
```diff
@@ -0,0 +1,61 @@
...
+	if os.IsNotExist(err) {
+		return nil, errors.Errorf("couldn't find env file %q", ef)
+	}
+	if err != nil {
+		return res, err
+	}
+
+	if s.IsDir() {
+		return res, errors.Errorf("%q is a directory", ef)
+	}
```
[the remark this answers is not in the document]
crazy-max: Hum yeah indeed

bake/env_test.go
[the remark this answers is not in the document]
jedevc: ah of course :facepalm:

crazy-max:
As discussed with @tonistiigi I changed the current behavior (see updated PR description).

bake/bake.go
```diff
@@ -179,46 +180,82 @@ func dedupMap(ms ...map[string]string) map[string]string {
...
-					return nil, err
+			hclFiles = append(hclFiles, hf)
+		case ".hcl":
+			hf, diags := hclparse.NewParser().ParseHCL(f.Data, f.Name)
+			if diags.HasErrors() {
+				return nil, diags
+			}
+			hclFiles = append(hclFiles, hf)
+		default:
+			e, err := dotenv.UnmarshalBytes(f.Data)
```
crazy-max, line 224: Previously, files with an unknown extension were parsed using the HCL parser. Now they will be parsed as env files. @tonistiigi Do you think it could be an issue? Maybe instead we could check for `.env` extension and default to hcl?

bake/bake.go
```diff
@@ -179,46 +180,82 @@ func dedupMap(ms ...map[string]string) map[string]string {
...
+	var hclFiles []*hcl.File
+	envs := sliceToMap(os.Environ())
 	for _, f := range files {
-		cfg, isCompose, composeErr := ParseComposeFile(f.Data, f.Name)
-		if isCompose {
-			if composeErr != nil {
-				return nil, composeErr
+		switch filepath.Ext(strings.ToLower(f.Name)) {
+		case ".yml", ".yaml":
+			composeFiles = append(composeFiles, f)
```
crazy-max, line 210: Like hcl and json, compose files need to be parsed later because we need to collect envs from files first.

bake/bake.go
tonistiigi: > Previously, files with an unknown extension were parsed using the HCL parser. Now they will be parsed as env files.

What is the difference?

I would prefer to not change the `-f` behavior and only handle `.env` as a special case when parsing compose.

tonistiigi: Ah, I think I understand you intentions better now.

Still in the first PR maybe `.env` is enough. Just to have compatibility.

For the `-f` we can maybe first split the files between compose and HCL. Then parse HCL and store all the attributes in the process. Then parse compose(for the second time if needed) and pass the attributes and variables.

If we support `-f` this way it shouldn't be only for some special limited set of HCL files that don't have any blocks but ideally all the vars that would be replaced in HCL would be replaced in Compose yaml as well(and additionally only Compose does `.env` as well).

crazy-max: > For the `-f` we can maybe first split the files between compose and HCL. Then parse HCL and store all the attributes in the process. Then parse compose(for the second time if needed) and pass the attributes and variables.
> 
> If we support `-f` this way it shouldn't be only for some special limited set of HCL files that don't have any blocks but ideally all the vars that would be replaced in HCL would be replaced in Compose yaml as well(and additionally only Compose does `.env` as well).

Ah right we also need to store the attributes before passing them to HCL/JSON/Compose files when parsed. I missed that.

I will just add `.env` support for compose in this PR and open another one for `-f`.

bake/compose.go
```diff
@@ -14,34 +16,22 @@ import (
 // errComposeInvalid is returned when a compose file is invalid
 var errComposeInvalid = errors.New("invalid compose file")
 
-func parseCompose(dt []byte) (*compose.Project, error) {
-	return loader.Load(compose.ConfigDetails{
+func ParseCompose(dt []byte, envs map[string]string) (*Config, error) {
+	if wd, err := os.Getwd(); err == nil {
```
tonistiigi, line 20: Looks like this suits better in `ParseComposeFile`. `envs` is also passed to this function so shouldn't be extended. But `ParseComposeFile` already reads env from the host.

bake/compose.go
```diff
@@ -124,6 +114,40 @@ func ParseCompose(dt []byte) (*Config, error) {
...
+	ef, err := filepath.Abs(filepath.Join(workingDir, ".env"))
+	if err != nil {
+		return curenv
+	}
+
+	if _, err := os.Stat(ef); err != nil {
+		return curenv
+	}
+
+	dt, err := os.ReadFile(ef)
```
tonistiigi, line 131: I think these syscalls should still error unless the error is `notexist`

bake/compose.go
```diff
@@ -124,6 +114,40 @@ func ParseCompose(dt []byte) (*Config, error) {
...
+	if err != nil {
+		return curenv
+	}
+
+	envs, err := dotenv.UnmarshalBytes(dt)
+	if err != nil {
+		return curenv
```
tonistiigi, line 138: Is the behavior to ignore parse error the behavior in compose as well?

tonistiigi (approved):

bake/compose.go
crazy-max: I read the wrong implementation. The right one from compose-go indeed check for errors.

bake/bake.go
saurabhdes: @crazy-max Thankyou for this PR to fix  load .env file from working dir for compose files ! I was facing the exact issue  that's discussed here and it's resolved. 
Very much looking forward to add another -f file to load another .env data file such that the below command will run or please suggest another option to pass a  custom env variable file. Essentially same as https://github.com/docker/buildx/issues/1135

`docker buildx bake -f docker-compose.yml -f .ghcr.env`

PR #1266, opened 2022-08-11
correct the file location of plugin when installed on Windows OS

BlackFrog1:
Documentation update


crazy-max:
I think the current location is the right one according to the plugin manager on docker/cli repo: https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/cli-plugins/manager/manager_windows.go#L10 (cc @thaJeztah)

PR #1267, opened 2022-08-12
commands: use buildx env for experimental opt-in

tonistiigi:
I messed this up when doing the final review edits for #1100

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -701,7 +701,7 @@ func (w *wrapped) Unwrap() error {
 }
 
 func isExperimental() bool {
-	if v, ok := os.LookupEnv("BUILDKIT_EXPERIMENTAL"); ok {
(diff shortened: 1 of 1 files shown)


jedevc (approved):

crazy-max:
Missed https://github.com/docker/buildx/blob/da1f4b84969c56dd6960d1f270a365730ad8ef00/commands/build.go#L704

Also don't think we need to say `BUILDX_EXPERIMENTAL=1 is required.` for this flag as it's hidden behind the `isExperimental()` cond anyway.

jedevc:
Woops, we should definitely be consistent with keeping `BUILDX_EXPERIMENTAL=1 is required.` between `--invoke` and `--print`. I think some sort of `(experimental)` note after both would be nice though - just to indicate that it has been activated with the `isExperimental` condition.

crazy-max:
@jedevc Sgtm

PR #1268, opened 2022-08-12
hack: use single output dir

crazy-max:
reduce footprint in our gitignore by using a single output dir for build, cross and coverage. also remove unused `cross-out`. structure will look like this:

* `./bin/build`
* `./bin/coverage`
* `./bin/release`

can be overridden using the `DESTDIR` env var. name of this en var seems to be the more approriate in our case: https://www.gnu.org/prep/standards/html_node/DESTDIR.html.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,6 @@ else
 	$(error "Buildx is required: https://github.com/docker/buildx#installing")
 endif
 
-export BIN_OUT = ./bin
-export RELEASE_OUT = ./release-out
-
 shell:
 	./hack/shell
 
@@ -22,7 +19,7 @@ binaries-cross:
 
 install: binaries
 	mkdir -p ~/.docker/cli-plugins
-	install bin/buildx ~/.docker/cli-plugins/docker-buildx
+	install bin/build/buildx ~/.docker/cli-plugins/docker-buildx
 
 release:
 	./hack/release
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,1 @@
 bin/
-cross-out/
-release-out/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ on:
 
 env:
   REPO_SLUG: "docker/buildx-bin"
-  RELEASE_OUT: "./release-out"
+  DESTDIR: "./bin"
 
 jobs:
   test:
@@ -48,7 +48,7 @@ jobs:
         name: Upload coverage
         uses: codecov/codecov-action@v3
         with:
-          file: ./coverage/coverage.txt
+          directory: ${{ env.DESTDIR }}/coverage
 
   prepare:
     runs-on: ubuntu-latest
@@ -107,7 +107,7 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: buildx
-          path: ${{ env.RELEASE_OUT }}/*
+          path: ${{ env.DESTDIR }}/release/*
           if-no-files-found: error
 
   bin-image:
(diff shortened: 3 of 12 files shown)


Dockerfile
```diff
@@ -13,25 +13,23 @@ FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS golatest
...
 FROM gobase AS buildx-build
-ARG LDFLAGS="-w -s"
 ARG TARGETPLATFORM
+RUN xx-go --wrap
 RUN --mount=type=bind,target=. \
   --mount=type=cache,target=/root/.cache \
   --mount=type=cache,target=/go/pkg/mod \
   --mount=type=bind,source=/tmp/.ldflags,target=/tmp/.ldflags,from=buildx-version \
-  set -x; xx-go build -ldflags "$(cat /tmp/.ldflags) ${LDFLAGS}" -o /usr/bin/buildx ./cmd/buildx && \
+  make build CGO_ENABLED=$CGO_ENABLED LDFLAGS="$(cat /tmp/.ldflags)" DESTDIR=/usr/bin && \
```
thaJeztah, line 32: should the target be last there? so

```
make CGO_ENABLED=$CGO_ENABLED LDFLAGS="$(cat /tmp/.ldflags)" DESTDIR=/usr/bin build
```

Also curious; setting `LDFLAGS` here; does `/tmp/.ldflags` contain the `-X` options?

Makefile
```diff
@@ -1,59 +1,83 @@
...
-	export BUILDX_CMD = $(which buildx)
-else
-	$(error "Buildx is required: https://github.com/docker/buildx#installing")
-endif
+PKG := github.com/docker/buildx
+VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
+REVISION ?= $(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
+
+CGO_ENABLED ?= 0
+LDFLAGS ?= -s -w -X ${PKG}/version.Version=${VERSION} -X ${PKG}/version.Revision=${REVISION} -X ${PKG}/version.Package=${PKG}
```
thaJeztah, line 6: ... because here, it will skip those if they're not set (i.e., it doesn't _append_ these to the existing ones set, correct?

thaJeztah: ^^ ignore this one; I thought `/tmp/.ldflags` was a "generic" LDFLAGS for the platform, but it includes these

Dockerfile
```diff
@@ -13,25 +13,23 @@ FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS golatest
...
 FROM gobase AS buildx-build
-ARG LDFLAGS="-w -s"
 ARG TARGETPLATFORM
+RUN xx-go --wrap
 RUN --mount=type=bind,target=. \
   --mount=type=cache,target=/root/.cache \
   --mount=type=cache,target=/go/pkg/mod \
   --mount=type=bind,source=/tmp/.ldflags,target=/tmp/.ldflags,from=buildx-version \
-  set -x; xx-go build -ldflags "$(cat /tmp/.ldflags) ${LDFLAGS}" -o /usr/bin/buildx ./cmd/buildx && \
+  make build CGO_ENABLED=$CGO_ENABLED LDFLAGS="$(cat /tmp/.ldflags)" DESTDIR=/usr/bin BINNAME=buildx && \
```
tonistiigi, line 32: Why this change? How should the user know if they need a specific environment for (some of) the make targets. Seems quite messy to combine these. Same for the `buildx-version`

crazy-max: I changed the implementation so it does not require `make` in the Dockerfile. This commit is to avoid drifting with build opts when buildx is packaged in https://github.com/docker/docker-ce-packaging/blob/5683c6bdcabf000cec58677ebb5f411f5410512a/deb/common/rules#L22. Let me know if you prefer to move it in another PR.

hack/build
```diff
@@ -0,0 +1,19 @@
...
+: "${GO_PKG=github.com/docker/buildx}"
+: "${GO_LDFLAGS=-X ${GO_PKG}/version.Version=${VERSION} -X ${GO_PKG}/version.Revision=${REVISION} -X ${GO_PKG}/version.Package=${PACKAGE}}"
+: "${GO_STRIP=}"
+
+if [ -n "$GO_STRIP" ]; then
+  GO_LDFLAGS="$GO_LDFLAGS -s -w"
+fi
+
+set -x
+CGO_ENABLED=0 GO111MODULE=on go build -mod vendor -trimpath -ldflags "${GO_LDFLAGS}" -o "${DESTDIR}/docker-buildx" ./cmd/buildx
```
tonistiigi, line 19: where is the `CGO_ENABLED` coming from?

crazy-max:
Removed commits mutualizing local and Dockerfile build opts and hash-files script changes as well. Will do that in a follow-up.

thaJeztah (approved):
LGTM

PR #1271, opened 2022-08-14
docker-container: restart-policy opt

crazy-max:
Currently we can't set the restart policy for buildkit containers created with `docker-container` driver. In my use case I have restarted some machines but as the default restart policy is `no`, the containers were not running. Even if we ensure nodes are booted before building I think it's still useful to have this driver opt.

Current behavior (`unless-stopped`) is kept.

Needs docs update in follow-up (cc @dvdksn): https://docs.docker.com/build/drivers/docker-container/#synopsis

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -44,17 +44,18 @@ type Driver struct {
 
 	// if you add fields, remember to update docs:
 	// https://github.com/docker/docs/blob/main/content/build/drivers/docker-container.md
-	netMode      string
-	image        string
-	memory       opts.MemBytes
-	memorySwap   opts.MemSwapBytes
-	cpuQuota     int64
-	cpuPeriod    int64
-	cpuShares    int64
-	cpusetCpus   string
-	cpusetMems   string
-	cgroupParent string
-	env          []string
+	netMode       string
+	image         string
+	memory        opts.MemBytes
+	memorySwap    opts.MemSwapBytes
+	cpuQuota      int64
+	cpuPeriod     int64
+	cpuShares     int64
+	cpusetCpus    string
+	cpusetMems    string
+	cgroupParent  string
+	restartPolicy container.RestartPolicy
+	env           []string
 }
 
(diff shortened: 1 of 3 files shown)


jedevc:
LGTM.

Personally, not sure about defaulting to `always`, I quite like the default lazy behavior with `restart-policy=no`, since I can sometimes have a lot of builders, which might not all actually be running. But, also, this would then match the kubernetes driver behavior, where the replicas are always running. If we do switch the default, we should default to `unless-stopped` I think, since `buildx stop` will stop the container, which then shouldn't automatically restart.

jedevc (approved):

crazy-max:
> 

> If we do switch the default, we should default to `unless-stopped` I think, since `buildx stop` will stop the container, which then shouldn't automatically restart.

Good point for `unless-stopped`, tbf I never used `buildx stop` and didn't recall it ever exists :sweat_smile:

crazy-max:
PTAL @tonistiigi

AkihiroSuda:
Needs rebase

AkihiroSuda (approved):

tonistiigi:
What is the use case for this? In what case would the user need to change this?

crazy-max:
> What is the use case for this? In what case would the user need to change this? Running a builder takes resources and memory so if it is not being used it is better if it is not running.

Main use case is for the Build UI in Docker Desktop so I can see my build history without needing to start my builders.

tonistiigi (approved):
I think opt-in this is fine but I would be against making this default behavior.

driver/docker-container/driver.go
```diff
@@ -44,17 +44,18 @@ type Driver struct {
 
 	// if you add fields, remember to update docs:
 	// https://github.com/docker/docs/blob/main/content/build/drivers/docker-container.md
```
tianon, line 46: Looks like this PR might've missed updating these docs :eyes:

dvdksn: It's here :-) https://github.com/docker/docs/pull/19355

Issue #1269, opened 2022-08-14
"docker" driver is not available on Docker Desktop for Mac


rgov:
I don't know whether it's supposed to be available, but the usage text suggests so:

```
% docker buildx create --help

Usage:  docker buildx create [OPTIONS] [CONTEXT|ENDPOINT]

Options:
      --driver string            Driver to use (available: "docker", "docker-container", "kubernetes")
                                                            ^^^^^^

% docker buildx create --driver docker
error: failed to find driver "docker"


% docker buildx version
github.com/docker/buildx v0.8.2 6224def4dd2c3d347eee19db595348c50d7cb491
```

jedevc:
Heya :wave:

Yeah, you're right, this isn't supposed to be available, but the docs and help definitely suggest it is - I've pushed a PR to hopefully make the behavior a little more clear.

You can create new instances of the docker driver by creating a new docker context, which will automatically be added to the list of builders in this case - you can find more information in [docs/guides/drivers/docker.md](https://github.com/docker/buildx/blob/master/docs/guides/drivers/docker.md).

PR #1273, opened 2022-08-16
create: improve interface when attempting to create docker driver

jedevc:
Fix https://github.com/docker/buildx/issues/1269.

Previously, the help information for buildx indicated that users could create a new instance of the docker driver - which is explicitly
something we don't support, driver of this form are automatically derived from the available list of docker contexts.

This patch ensures that don't have AllowsInstance set will not appear in the help text, and additionally provide a new more specific error message instead of the generic "failed to find driver". This should help point users in the correct direction.

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -135,8 +135,8 @@ func runCreate(dockerCli command.Cli, in createOptions, args []string) error {
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


crazy-max (approved):
LGTM

tonistiigi (approved):

PR #1274, opened 2022-08-16
Consistent experimental help

jedevc:
Follow-up to https://github.com/docker/buildx/pull/1267

This patch ensures that we have consistent help messages between the two experimental options, and additionally adds the experimental options to the command reference along with information on how to enable them.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -479,7 +479,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags.StringArrayVar(&options.platforms, "platform", platformsDefault, "Set target platform for build")
 
 	if isExperimental() {
-		flags.StringVar(&options.printFunc, "print", "", "Print result of information request (outline, targets)")
+		flags.StringVar(&options.printFunc, "print", "", "Print result of information request (e.g., outline, targets) [experimental]")
 	}
 
 	flags.BoolVar(&options.exportPush, "push", false, `Shorthand for "--output=type=registry"`)
@@ -501,7 +501,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags.Var(options.ulimits, "ulimit", "Ulimit options")
 
 	if isExperimental() {
-		flags.StringVar(&options.invoke, "invoke", "", "Invoke a command after the build. BUILDX_EXPERIMENTAL=1 is required.")
+		flags.StringVar(&options.invoke, "invoke", "", "Invoke a command after the build [experimental]")
 	}
 
 	// hidden flags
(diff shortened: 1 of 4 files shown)


hack/dockerfiles/docs.Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN apk add --no-cache rsync git
 WORKDIR /src
 COPY --from=docsgen /out/docsgen /usr/bin
 ARG FORMATS
+ARG BUILDX_EXPERIMENTAL=0
```
crazy-max, line 17: Is it to not generate experimental flags for docs? Not sure how docs validation is ok with this as it should not set `--invoke` and `--print` flags in markdown table. Maybe I miss smth.

jedevc: I've overwritten the flag in `docker-bake.hcl`, I think that should be source of truth for the config? I guess it would be a good idea to remove the default here then, I think I just wanted to set it to a "definitely-falsey" value.

crazy-max: Oh I see, then:

```suggestion
ARG BUILDX_EXPERIMENTAL
```

should be enough

commands/build.go
```diff
@@ -501,7 +501,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags.Var(options.ulimits, "ulimit", "Ulimit options")
 
 	if isExperimental() {
-		flags.StringVar(&options.invoke, "invoke", "", "Invoke a command after the build. BUILDX_EXPERIMENTAL=1 is required.")
+		flags.StringVar(&options.invoke, "invoke", "", "Invoke a command after the build [experimental]")
```
crazy-max, line 504: In the CLI with `docker buildx build --help` users will not know how to enable it without looking at docs. Maybe we should also be consistent for the CLI with:

```suggestion
		flags.StringVar(&options.invoke, "invoke", "", "Invoke a command after the build. BUILDX_EXPERIMENTAL=1 env var required.")
```

jedevc: I've added a little note in the docs page to instruct users how to enable the `[experimental]` flags - which I think should maybe be enough?

The arguments shouldn't appear in a normal `--help`, unless `BUILDX_EXPERIMENTAL` is already set, so I'm not sure we need to mention it in the `--help`, it feels like something to call out in docs?

crazy-max: > The arguments shouldn't appear in a normal `--help`, unless `BUILDX_EXPERIMENTAL` is already set,

Ah good point :+1:

crazy-max (approved):
LGTM, PTAL @tonistiigi

tonistiigi (approved):

Issue #1172, opened 2022-06-16
Unable to use multiple build definition files with `bake`


Apakottur:
I am unable to use `docker buildx bake` to build a docker image based on two docker-compose files.

To reproduce, create three files, all in the same directory:
```dockerfile
# Dockerfile
FROM ubuntu:latest
```
```yml
# docker-compose.yml
services:
  test:
    build:
      context: .
    entrypoint: echo 1
```
```yml
# docker-compose.prod.yml
services:
  test:
    entrypoint: echo 2
```

The idea is to have the core configuration of an image in the base yml file, and then have another one with overrides (for a prod environment, for example).
I can then build and run the base/prod images like so:
```shell
$ docker-compose build
$ COMPOSE_FILE=docker-compose.yml docker-compose run --rm test
1
$ COMPOSE_FILE=docker-compose.yml:docker-compose.prod.yml docker-compose run --rm test
2
```

The [buildx docs](https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md#-specify-a-build-definition-file--f---file) say that `If multiple files are specified they are all read and configurations are combined`.

However, when I try to use both files with buildx:
```shell
docker buildx bake -f docker-compose.yml -f docker-compose.prod.yml test
```
I am getting:
```shell
error: service "test" has neither an image nor a build context specified: invalid compose project
```

Am I missing something, or is this a bug?

crazy-max:
As I can see, consistency check added in https://github.com/docker/buildx/pull/87 is already handled by the library: https://github.com/compose-spec/compose-go/blob/c45b40b31e66dbc3966642803d133bba33680b21/loader/validate.go#L31-L33

Now with compose v2, I think we can skip the consistency check and remove the dup check in bake as well and just use the default values. In this case it would produce:

```yaml
# docker-compose.yml
services:
  test:
    build:
      context: .
      tags:
        - docker.io/username/webapp:latest
    entrypoint: echo 1
```


```yaml
# docker-compose.prod.yml
services:
  test:
    entrypoint: echo 2
```

```console
$ docker buildx bake -f docker-compose.yml test --print
{
  "group": {
    "default": {
      "targets": [
        "test"
      ]
    }
  },
  "target": {
    "test": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/webapp:latest"
      ]
    }
  }
}
```

```console
$ docker buildx bake -f docker-compose.prod.yml test --print
{
[...]

PR #1277, opened 2022-08-17
bake(compose): fix unskipped services without build context

### LINKED ISSUE #1276: Bake Error: "Failed to read Dockerfile" since 0.9.0 update
Hello!

First of all, thanks for all your hard work.

Since updating to v0.9.0 released yesterday, i get this error whenever I try to build with bake.

```
docker buildx bake -f docker-compose.yaml --load
[+] Building 0.0s (4/4) FINISHED                                                                                                                                                                         
 => [db internal] load build definition from Dockerfile               0.0s
 => => transferring dockerfile: 2B                                              0.0s
 => [db internal] load .dockerignore                                         0.0s
 => => transferring context: 2B                                                   0.0s
 => CANCELED [app internal] load .dockerignore                            0.0s
 => => transferring context: 2B                                                         0.0s
 => CANCELED [app internal] load build definition from Dockerfile                    0.0s
ERROR: failed to solve: failed to read dockerfile: open /tmp/buildkit-mount1176813038/Dockerfile: no such file or directory
```

I can reproduce it both locally and in my CD/CI pipelines, and can confirm that the problem doesn't exist with version 0.8.2.

From comparing what is produced with the `--print` option from a simple Dockerfile with a custom app and a DB, it seems like buildx tries to build every single service, including ones that should not be built (i.e., in the docker-compose.yaml definitions, there is only an image specified for the service, and no "build" key). Then, when it can't find the Dockerfile for services that aren't meant to be built, it fails.

<details><summary>Output of `--print` on 0.8.2</summary>

```
docker buildx bake -f docker-compose.yaml --print
{
  "group": {
    "default": {
      "targets": [
        "app"
      ]
    }
  },
  "target": {
    "app: {
      "context": "app",
      "dockerfile": "Dockerfile",
      "tags": [
         "myapp"
      ]
    }
  }
}
```
</details>

<details><summary> Output of `--print` on 0.9.0</summary>

```
docker buildx bake -f docker-compose.yaml --print
{
  "group": {
    "default": {
      "targets": [
        "app",
        "db"
      ]
    }
  },
  "target": {
    "app: {
      "context": "app",
      "dockerfile": "Dockerfile",
      "tags": [
         "myapp"
      ]
    },
    "db": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "postgres:14"
      ]
    }
  }
}
```

</details>

<details><summary>Files</summary>

docker-compose.yaml
```
version: "3.9"

services:
  app:
    image: myapp
    build:
      context: ./app

  db:
    image: postgres:14-alpine
    volumes:
      - db:/var/lib/postgresql/data

```

app/Dockerfile
```
FROM python:3.10-slim-bullseye-slim

RUN echo "Hello, World!"
```

</details>

<details><summary>My Docker Info</summary>

```
❯ docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.9.0)
  compose: Docker Compose (Docker Inc., v2.6.0)
  scan: Docker Scan (Docker Inc., v0.17.0)

Server:
 Containers: 8
  Running: 8
  Paused: 0
  Stopped: 0
 Images: 11
 Server Version: 20.10.17
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Runtimes: io.containerd.runtime.v1.linux runc sysbox-runc io.containerd.runc.v2
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 0197261a30bf81f1ee8e6a4dd2dea0ef95d67ccb
 runc version: v1.1.3-0-g6724737
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: default
  cgroupns
 Kernel Version: 5.18.0-4-amd64
 Operating System: Debian GNU/Linux bookworm/sid
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 62.23GiB
 Name: [redacted]
 ID: [redacted]
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Default Address Pools:
   Base: 172.25.0.0/16, Size: 24
```
</details>

Thank you for your help!

crazy-max:
fixes #1276 

while fixing compose consistency check in https://github.com/docker/buildx/pull/1181 a regression was introduced where services without build context were taken into account.

it was not detected because our test `TestNoBuildOutOfTreeService` was not checking the targets length: https://github.com/docker/buildx/blob/611329fc7f1365556789bff4747f608d40cdc8a9/bake/compose_test.go#L71-L82

by fixing this issue we now remove the ability to merge compose files without a build context https://github.com/docker/buildx/issues/1172. To fix this we now parse compose files all together so the compose-go library can handle a proper sanity check on its side.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -200,15 +200,15 @@ func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error)
 	}()
 
 	var c Config
-	var fs []*hcl.File
+	var composeFiles []File
+	var hclFiles []*hcl.File
 	for _, f := range files {
-		cfg, isCompose, composeErr := ParseComposeFile(f.Data, f.Name)
+		isCompose, composeErr := validateComposeFile(f.Data, f.Name)
 		if isCompose {
 			if composeErr != nil {
 				return nil, composeErr
 			}
-			c = mergeConfig(c, *cfg)
-			c = dedupeConfig(c)
+			composeFiles = append(composeFiles, f)
 		}
 		if !isCompose {
 			hf, isHCL, err := ParseHCLFile(f.Data, f.Name)
@@ -216,7 +216,7 @@ func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error)
 				if err != nil {
 					return nil, err
 				}
-				fs = append(fs, hf)
+				hclFiles = append(hclFiles, hf)
 			} else if composeErr != nil {
 				return nil, fmt.Errorf("failed to parse %s: parsing yaml: %v, parsing hcl: %w", f.Name, composeErr, err)
 			} else {
@@ -225,8 +225,17 @@ func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error)
 		}
 	}
 
-	if len(fs) > 0 {
-		if err := hclparser.Parse(hcl.MergeFiles(fs), hclparser.Opt{
+	if len(composeFiles) > 0 {
+		cfg, cmperr := ParseComposeFiles(composeFiles)
+		if cmperr != nil {
+			return nil, errors.Wrap(cmperr, "failed to parse compose file")
+		}
+		c = mergeConfig(c, *cfg)
+		c = dedupeConfig(c)
+	}
+
+	if len(hclFiles) > 0 {
+		if err := hclparser.Parse(hcl.MergeFiles(hclFiles), hclparser.Opt{
 			LookupVar:     os.LookupEnv,
 			Vars:          defaults,
 			ValidateLabel: validateTargetName,
@@ -234,18 +243,25 @@ func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error)
 			return nil, err
 		}
 	}
+
 	return &c, nil
 }
 
 func dedupeConfig(c Config) Config {
 	c2 := c
+	c2.Groups = make([]*Group, 0, len(c2.Groups))
+	for _, g := range c.Groups {
+		g1 := *g
+		g1.Targets = dedupSlice(g1.Targets)
+		c2.Groups = append(c2.Groups, &g1)
+	}
 	c2.Targets = make([]*Target, 0, len(c2.Targets))
-	m := map[string]*Target{}
+	mt := map[string]*Target{}
 	for _, t := range c.Targets {
-		if t2, ok := m[t.Name]; ok {
+		if t2, ok := mt[t.Name]; ok {
 			t2.Merge(t)
 		} else {
-			m[t.Name] = t
+			mt[t.Name] = t
 			c2.Targets = append(c2.Targets, t)
 		}
 	}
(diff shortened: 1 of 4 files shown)


bake/hcl.go
```diff
@@ -1,41 +1,11 @@
...
-		return f, true, err
-	}
-	if strings.HasSuffix(fn, ".hcl") {
-		f, diags := hclparse.NewParser().ParseHCL(dt, fn)
-		if diags.HasErrors() {
-			err = diags
-		}
-		return f, true, err
-	}
-	f, diags := hclparse.NewParser().ParseHCL(dt, fn+".hcl")
```
jedevc, line 28: I think in the switch this logic gets dropped? I think we need to keep separate `case "hcl"` and `default` cases in the `switch filepath.Ext(strings.ToLower(f.Name)) {` above.

bake/bake.go
```diff
@@ -200,52 +201,63 @@ func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error)
...
 	return &c, nil
 }
 
 func dedupeConfig(c Config) Config {
 	c2 := c
+	c2.Groups = make([]*Group, 0, len(c2.Groups))
+	for _, g := range c.Groups {
+		g.Targets = dedupSlice(g.Targets)
+		c2.Groups = append(c2.Groups, g)
+	}
```
jedevc, line 253: As I read this, we need to make a non-destructive copy of `c`? If that's the case, we should deref and copy `g` before making changes to `g.Targets`.

bake/hcl.go
crazy-max: I was wondering why we have this fallback to json if hcl parsing fails. Looks like a best effort if user don't use a known extension. I'll bring it back.

bake/bake.go
```diff
@@ -256,24 +272,23 @@ func ParseFile(dt []byte, fn string) (*Config, error) {
 	return ParseFiles([]File{{Data: dt, Name: fn}}, nil)
 }
 
-func ParseComposeFile(dt []byte, fn string) (*Config, bool, error) {
+func ParseComposeFile(dt []byte, fn string) (bool, error) {
```
jedevc, line 275: I like the idea of keeping a smaller footprint for this patch, but as this doesn't return a `Config` struct anymore, I think this might make more sense as `Validate`/etc instead of `Parse`?

May also be worth keeping as a private function IMO.

jedevc (approved):

tonistiigi (approved):

PR #998, opened 2022-03-14
ls: display buildkit version of the nodes

crazy-max:
~needs https://github.com/moby/buildkit/pull/2725~

vendor buildkit to bring https://github.com/moby/buildkit/pull/2725: https://github.com/moby/buildkit/compare/3e38a2d34830...86c33b66e176

```console
$ docker buildx ls
NAME/NODE              DRIVER/ENDPOINT             STATUS                 BUILDKIT             PLATFORMS
builder2               docker-container
  builder20            unix:///var/run/docker.sock running                N/A                  linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
builder3               docker-container
  builder30            unix:///var/run/docker.sock running                86c33b6              linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
builder4 *             docker-container
  builder40            unix:///var/run/docker.sock running                v0.10.0-91-g86c33b66 linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
xenodochial_davinci    docker-container
  xenodochial_davinci0 unix:///var/run/docker.sock running                N/A                  linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
desktop-linux                                      protocol not available
default                docker
  default              default                     running                N/A                  linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```

Maybe we could display this info with a `--no-trunc` flag and not by default. cc @thaJeztah 

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #1279, opened 2022-08-18
inspect: add buildkit version information to command output

jedevc:
Follow-up to #998 - this adds the buildkit version information to the `buildx inspect` command if it's available.


crazy-max (approved):
LGTM to have this in next patch release as it's missing from a released feature in 0.9.0.

tonistiigi (approved):

PR #1280, opened 2022-08-18
rm: ensure context builder is not removed with all-inactive flag

crazy-max:
forgot to check if a context builder is being removed when using `all-inactive` flag (can happen with remote context).

### DIFF
--- a/commands/rm.go
+++ b/commands/rm.go
@@ -149,12 +149,22 @@ func rmAllInactive(ctx context.Context, txn *store.Txn, dockerCli command.Cli, i
 		builders[i] = &nginfo{ng: ng}
 	}
 
+	ctxbuilders, err := dockerCli.ContextStore().List()
+	if err != nil {
+		return err
+	}
+
 	eg, _ := errgroup.WithContext(ctx)
 	for _, b := range builders {
 		func(b *nginfo) {
 			eg.Go(func() error {
 				if err := loadNodeGroupData(ctx, dockerCli, b); err != nil {
 					return errors.Wrapf(err, "cannot load %s", b.ng.Name)
+				}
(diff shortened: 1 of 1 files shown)


jedevc:
It's possible we could break this down a bit differently, there's some already with the `rmAllInactive` method. We could potentially refactor to get a list of all inactive builders, and then `rm` them all, so that we only have one set of context checking steps.

crazy-max:
Was thinking doing that in a follow-up with another refactoring about builder nodes.

commands/rm.go
```diff
@@ -149,13 +149,23 @@ func rmAllInactive(ctx context.Context, txn *store.Txn, dockerCli command.Cli, i
...
 	eg, _ := errgroup.WithContext(ctx)
 	for _, b := range builders {
 		func(b *nginfo) {
 			eg.Go(func() error {
 				if err := loadNodeGroupData(ctx, dockerCli, b); err != nil {
 					return errors.Wrapf(err, "cannot load %s", b.ng.Name)
 				}
+				for _, cb := range ctxbuilders {
+					if b.ng.Driver == "docker" && len(b.ng.Nodes) == 1 && b.ng.Nodes[0].Endpoint == cb.Name {
+						return errors.Errorf("context builder cannot be removed, run `docker context rm %s` to remove this context", cb.Name)
```
jedevc, line 166: I wonder if we should just `return` to skip over the context builders here, instead of returning an error.

crazy-max: I see what you mean but I think the user would expect this builder to be removed and if we don't return an error he will be confused imo.

crazy-max:
As discussed with @tonistiigi this is not needed as `txn.List()` only returns instances from buildx store and not context ones.

PR #1289, opened 2022-08-22
monitor: add common shell features to dev prompt

tonistiigi:
This makes the monitor mode from —invoke more
user friendly by enabling features user might expect
from dev shell.

Interactive mode now has colors, history and keyboard
control for movement (arrows, moving to beginning/end,
by word, emacs controls etc).

Opening as draft. If this solution (and base library it uses) seems desirable then some updates are needed in `go-prompt` (or fork) to make it more configurable (currently made changes in vendor).

@ktock


monitor/monitor.go
```diff
@@ -104,6 +94,36 @@ func RunMonitor(ctx context.Context, containerConfig build.ContainerConfig, relo
...
+						Fn: func(b *prompt.Buffer) {
+							b.CursorRight(1 + b.Document().FindEndOfCurrentWordWithSpace())
+						},
+					},
+					prompt.ASCIICodeBind{
+						ASCIICode: []byte{0x1b, 'b'},
+						Fn: func(b *prompt.Buffer) {
+							b.CursorLeft(len(b.Document().TextBeforeCursor()) - b.Document().FindStartOfPreviousWordWithSpace())
+						},
+					},
```
jedevc, line 124: What do these do? :eyes: I'm not familiar with these key combos. Are they particularly common?

monitor/monitor.go
```diff
@@ -104,6 +94,36 @@ func RunMonitor(ctx context.Context, containerConfig build.ContainerConfig, relo
 					fmt.Fprint(stdout, helpMessage)
 				}
 			}
+			completer := func(d prompt.Document) []prompt.Suggest {
+				return []prompt.Suggest{}
+			}
+
+			prompt.New(exec, completer, prompt.OptionHistory([]string{
+				"reload", "rollback", "exit", "help",
```
jedevc, line 102: This sets the history at the beginning? I think we probably should just leave this empty (maybe a future enhancement to save the history to `~/.buildx/monitor_history` or similar?)

jedevc (commented):
This would be super useful :tada: The approach LGTM :smile: 

I like the splash of color in it - ideally we'd sync up with the buildkit colors in some way. Autocompletion would also be excellent to add at some point once we've stabilized what all the monitor mode commands would be.

jedevc:
@tonistiigi what changes would be needed in go-prompt? Would it be feasible to try and take this for v0.11?

tonistiigi:
see the changes in vendor https://github.com/docker/buildx/pull/1289/commits/6e05092055d20af2bded2b5c9fdca9a7b53512cb

monitor/monitor.go
tonistiigi: Move cursor with `word` increments

jedevc:
Looking through go-prompt, the maintainer seems inactive, so it seems likely we'd probably have to fork if we need changes.

crazy-max:
> Looking through go-prompt, the maintainer seems inactive, so it seems likely we'd probably have to fork if we need changes.

https://github.com/erikgeiser/promptkit and https://github.com/charmbracelet/bubbletea looks to be reliable alternatives.

jsternberg:
Closing this as the PR is from 2022. @tonistiigi contact me if you want me to revisit the implementation here to help update this PR and carry it over the line. If it's something not very important we can drop it otherwise we can create an issue and reprioritize this so it gets done.

PR #1294, opened 2022-08-25
s3 cache client-side support

crazy-max:
follow-up https://github.com/moby/buildkit/pull/2824
fixes https://github.com/docker/build-push-action/issues/647

[s3 cache](https://github.com/moby/buildkit#s3-cache-experimental) client-side support.

cc @bpaquet


util/buildflags/cache.go
```diff
@@ -79,3 +80,24 @@ func addGithubToken(ci *client.CacheOptionsEntry) bool {
...
+		if v, ok := os.LookupEnv("AWS_ACCESS_KEY_ID"); ok {
+			ci.Attrs["access_key_id"] = v
+		}
+	}
+	if _, ok := ci.Attrs["secret_access_key"]; !ok {
+		if v, ok := os.LookupEnv("AWS_SECRET_ACCESS_KEY"); ok {
+			ci.Attrs["secret_access_key"] = v
+		}
+	}
+	if _, ok := ci.Attrs["session_token"]; !ok {
```
crazy-max, line 98: https://github.com/moby/buildkit/pull/3065

docs/reference/buildx_build.md
```diff
@@ -202,15 +206,17 @@ More info about cache exporters and available attributes: https://github.com/mob
 ```
 
 Export build cache to an external cache destination. Supported types are
-`registry`, `local`, `inline` and `gha`.
+`registry`, `local`, `inline`, `gha` and `s3`.
```
jedevc, line 209: :tada: Good to have some more docs! I think we probably want to update the `README.md` as well alongside this?

util/buildflags/cache.go
```diff
@@ -79,3 +80,24 @@ func addGithubToken(ci *client.CacheOptionsEntry) bool {
 	}
 	return ci.Attrs["token"] != "" && ci.Attrs["url"] != ""
 }
+
+func addAwsCredentials(ci *client.CacheOptionsEntry) {
+	if ci.Type != "s3" {
+		return
+	}
+	if _, ok := ci.Attrs["access_key_id"]; !ok {
```
bpaquet, line 88: This should be done automatically by the AWS SDK. Are you sure we really need to do this?

bpaquet: I mean we the SDK should be able to configure itself without any manual mapping, For example it's working with AWS IAM Profile. We should look on how to do the same with these env variables.

docs/reference/buildx_build.md
crazy-max: Ah yes looks like it's missing

crazy-max:
Using the sdk just to load default config brings a lot :worried:

image

util/buildflags/cache.go
bpaquet: Spent some time to think about it.
Your solution will work only for simple authent (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`). The SDK offers plenty of solutions to authent (example by using `AWS_PROFILE`), but I do not think there is a simple way to authenticate inside the client and forward creds to the server. Or it do not worth it. So dealing only with simple authent is fine for me, if it's clear in the doc.

tonistiigi (approved):

PR #1297, opened 2022-08-29
Add git provenance labels

### LINKED ISSUE #1290: Proposal: Attach Git information via image labels
As the industry attempts to move security futher to the left, it becomes increasingly important to link built and pushed container images to their origin; ideally via Git metadata. This enables numerous use cases for tools along the software supply chain. Here are just a few examples:

* Inform contributors about updates to base images via workflows on GitHub
* Trace images that get deployed to production to their Git source
* Remediate CVEs by sending PRs to GitHub repositories updating `apt-get install` etc lines

We are aware that there is currently a lot of great work underway to securly and verifiably attach such provenance information to container images via signed attestations. Additionally similar provenance data is recorded when using remote Git contexts with buildx/buildkit. This work is still very early.

Ideally we would have a very pragmatic solution that could work today without requiring users of `docker build` to every build.

Therefore, we'd like to propose that buildx starts to record the following pieces of provenance when being run with a context that has a `.git` directory:

* Git commit SHA of current checked out HEAD
* Include flag to indicate state of local clone (dirty)
* Dockerfile path

From a privacy perspective we believe that storing Git commit SHAs shouldn't represent a concern given what we see other tools doing (eg. `go build` [storing very similar information](https://utcc.utoronto.ca/~cks/space/blog/programming/GoVersionOfYourSource) without requiring opt-in) and images from private repositories can be pushed to private registries. 

We propose to store this information in image labels following the naming conventions set out by https://specs.opencontainers.org/image-spec/annotations/ accepting the way label inheritance can complicate things. This convention is already widely adopted by vendors and projects:

* Snyk: https://docs.snyk.io/products/snyk-container/scan-your-dockerfile/automatically-link-between-dockerfile-and-images-using-labels and https://snyk.io/blog/how-and-when-to-use-docker-labels-oci-container-annotations/
* Atomist: https://docs.atomist.com/integration/linking-images/
* docker/metadata-action: https://github.com/docker/metadata-action

Initially, we'd want to make storing the Git information by buildx opt-in via an environment variable switch. 

There's a POC with the suggested changes at: https://github.com/cdupuis/buildx/commit/cb8253f759b29e3109ed474e253b683aa6bb046e

Before raising a pull request with the proposed additions, we'd wanted to raise this issue for community feedback.

cdupuis:
As discussed in #1290, this PR adds opt-in labels for Git provenance.

The following labels will be added when `BUILDX_GIT_LABELS=full` is set:

```json
      "Labels": {
        "com.docker.image.source.entrypoint": "Dockerfile",
        "org.opencontainers.image.revision": "5734329c6af43c2ae295010778cd308866b95d9b",
        "org.opencontainers.image.source": "git@github.com:atomisthq/kipz-docker-test.git"
      }
```

With only `BUILDX_GIT_LABELS =1` the repository label will not be included:

```json
      "Labels": {
        "com.docker.image.source.entrypoint": "Dockerfile",
        "org.opencontainers.image.revision": "5734329c6af43c2ae295010778cd308866b95d9b"
      }
```

If the cloned source is not clean, the `revision` label value will get a `-dirty` suffix:

```json
      "Labels": {
        "com.docker.image.source.entrypoint": "Dockerfile",
        "org.opencontainers.image.revision": "5734329c6af43c2ae295010778cd308866b95d9b-dirty"
      }
```

fixes #1290

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -786,6 +786,18 @@ func BuildWithResultHandler(ctx context.Context, drivers []DriverInfo, opt map[s
 
 	eg, ctx := errgroup.WithContext(ctx)
 
+	for _, opt := range opt {
+		gitLabels, err := addGitProvenance(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath)
+		if err != nil {
+			return nil, err
+		}
+		for n, v := range gitLabels {
+			if _, ok := opt.Labels[n]; !ok {
+				opt.Labels[n] = v
+			}
+		}
+	}
+
 	for k, opt := range opt {
 		multiDriver := len(m[k]) > 1
 		hasMobyDriver := false
--- a/build/git.go
+++ b/build/git.go
@@ -1,0 +1,98 @@
+package build
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const DockerfileLabel = "com.docker.image.source.entrypoint"
+
+func addGitProvenance(ctx context.Context, contextPath string, dockerfilePath string) (map[string]string, error) {
+	v, ok := os.LookupEnv("BUILDX_GIT_LABELS")
+	if !ok || contextPath == "" {
+		return nil, nil
+	}
+	labels := make(map[string]string, 0)
+
+	// figure out in which directory the git command needs to run in
+	var wd string
+	if filepath.IsAbs(contextPath) {
+		wd = contextPath
+	} else {
+		cwd, _ := os.Getwd()
+		wd, _ = filepath.Abs(filepath.Join(cwd, contextPath))
+	}
+
+	// check if inside git working tree
+	cmd := exec.CommandContext(ctx, "git", "rev-parse", "--is-inside-work-tree")
+	cmd.Dir = wd
+	err := cmd.Run()
+	if err != nil {
+		logrus.Warnf("Unable to determine Git information")
+		return nil, nil
+	}
+
+	// obtain Git sha of current HEAD
+	cmd = exec.CommandContext(ctx, "git", "rev-parse", "HEAD")
+	cmd.Dir = wd
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, errors.Wrap(err, "error obtaining git head")
+	}
+	sha := strings.TrimSpace(string(out))
+
+	// check if the current HEAD is clean
+	cmd = exec.CommandContext(ctx, "git", "status", "--porcelain", "--ignored")
+	cmd.Dir = wd
+	out, err = cmd.Output()
+	if err != nil {
+		return nil, errors.Wrap(err, "error obtaining git status")
+	}
+	if len(strings.TrimSpace(string(out))) != 0 {
+		sha += "-dirty"
+	}
+	labels[ocispecs.AnnotationRevision] = sha
+
+	// add a remote url if full Git details are requested; if there aren't any remotes don't fail
+	if v == "full" {
+		cmd = exec.CommandContext(ctx, "git", "ls-remote", "--get-url")
+		cmd.Dir = wd
+		out, _ := cmd.Output()
+		if len(out) > 0 {
+			labels[ocispecs.AnnotationSource] = strings.TrimSpace(string(out))
+		}
+	}
+
+	// add Dockerfile path; there is no org.opencontainers annotation for this
+	if dockerfilePath == "" {
+		dockerfilePath = filepath.Join(wd, "Dockerfile")
+	}
+
+	// obtain Git root directory
+	cmd = exec.CommandContext(ctx, "git", "rev-parse", "--show-toplevel")
+	cmd.Dir = wd
+	out, err = cmd.Output()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get git root")
+	}
+	root := strings.TrimSpace(string(out))
+
+	// record only Dockerfile paths that are within the Git root
+	if !filepath.IsAbs(dockerfilePath) {
+		cwd, _ := os.Getwd()
+		dockerfilePath = filepath.Join(cwd, dockerfilePath)
+	}
+	dockerfilePath, _ = filepath.Rel(root, dockerfilePath)
+	if !strings.HasPrefix(dockerfilePath, "..") {
+		labels[DockerfileLabel] = dockerfilePath
+	}
+
+	return labels, nil
+}
--- a/build/git_test.go
+++ b/build/git_test.go
@@ -1,0 +1,115 @@
+package build
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/stretchr/testify/assert"
+)
+
+var repoDir string
+
+func setupTest(tb testing.TB) func(tb testing.TB) {
+	repoDir = tb.TempDir()
+	// required for local testing on mac to avoid strange /private symlinks
+	if runtime.GOOS == "darwin" {
+		repoDir, _ = filepath.EvalSymlinks(repoDir)
+	}
+	cmd := exec.Command("git", "init")
+	cmd.Dir = repoDir
+	err := cmd.Run()
+	assert.Nilf(tb, err, "failed to init git repo: %v", err)
+
+	df := []byte("FROM alpine:latest\n")
+	err = os.WriteFile(filepath.Join(repoDir, "Dockerfile"), df, 0644)
+	assert.Nilf(tb, err, "failed to write file: %v", err)
+
+	cmd = exec.Command("git", "add", "Dockerfile")
+	cmd.Dir = repoDir
+	err = cmd.Run()
+	assert.Nilf(tb, err, "failed to add file: %v", err)
+
+	cmd = exec.Command("git", "config", "user.name", "buildx")
+	cmd.Dir = repoDir
+	err = cmd.Run()
+	assert.Nilf(tb, err, "failed to set git user.name: %v", err)
+
+	cmd = exec.Command("git", "config", "user.email", "buildx@docker.com")
+	cmd.Dir = repoDir
+	err = cmd.Run()
+	assert.Nilf(tb, err, "failed to set git user.email: %v", err)
+
+	cmd = exec.Command("git", "commit", "-m", "Initial commit")
+	cmd.Dir = repoDir
+	err = cmd.Run()
+	assert.Nilf(tb, err, "failed to commit: %v", err)
+
+	return func(tb testing.TB) {
+		os.Unsetenv("BUILDX_GIT_LABELS")
+		os.RemoveAll(repoDir)
+	}
+}
+
+func TestAddGitProvenanceDataWithoutEnv(t *testing.T) {
+	defer setupTest(t)(t)
+	labels, err := addGitProvenance(context.Background(), repoDir, filepath.Join(repoDir, "Dockerfile"))
+	assert.Nilf(t, err, "No error expected")
+	assert.Nilf(t, labels, "No labels expected")
+}
+
+func TestAddGitProvenanceDataWithoutLabels(t *testing.T) {
+	defer setupTest(t)(t)
+	os.Setenv("BUILDX_GIT_LABELS", "full")
+	labels, err := addGitProvenance(context.Background(), repoDir, filepath.Join(repoDir, "Dockerfile"))
+	assert.Nilf(t, err, "No error expected")
+	assert.Equal(t, 2, len(labels), "Exactly 2 git provenance labels expected")
+	assert.Equal(t, "Dockerfile", labels[DockerfileLabel], "Expected a dockerfile path provenance label")
+
+	cmd := exec.Command("git", "rev-parse", "HEAD")
+	cmd.Dir = repoDir
+	out, _ := cmd.Output()
+	assert.Equal(t, strings.TrimSpace(string(out)), labels[ocispecs.AnnotationRevision], "Expected a sha provenance label")
+}
+
+func TestAddGitProvenanceDataWithLabels(t *testing.T) {
+	defer setupTest(t)(t)
+	// make a change to test dirty flag
+	df := []byte("FROM alpine:edge\n")
+	os.Mkdir(filepath.Join(repoDir, "dir"), 0755)
+	os.WriteFile(filepath.Join(repoDir, "dir", "Dockerfile"), df, 0644)
+	// add a remote
+	cmd := exec.Command("git", "remote", "add", "origin", "git@github.com:docker/buildx.git")
+	cmd.Dir = repoDir
+	cmd.Run()
+
+	os.Setenv("BUILDX_GIT_LABELS", "full")
+	labels, err := addGitProvenance(context.Background(), repoDir, filepath.Join(repoDir, "Dockerfile"))
+	assert.Nilf(t, err, "No error expected")
+	assert.Equal(t, 3, len(labels), "Exactly 3 git provenance labels expected")
+	assert.Equal(t, "Dockerfile", labels[DockerfileLabel], "Expected a dockerfile path provenance label")
+	assert.Equal(t, "git@github.com:docker/buildx.git", labels[ocispecs.AnnotationSource], "Expected a remote provenance label")
+
+	cmd = exec.Command("git", "rev-parse", "HEAD")
+	cmd.Dir = repoDir
+	out, _ := cmd.Output()
+	assert.Equal(t, fmt.Sprintf("%s-dirty", strings.TrimSpace(string(out))), labels[ocispecs.AnnotationRevision], "Expected a sha provenance label")
+}
+
+func TestAddGitProvenanceDataOutsideOfGitRepository(t *testing.T) {
+	defer setupTest(t)(t)
+	os.Setenv("BUILDX_GIT_LABELS", "full")
+	parentDir := filepath.Dir(repoDir)
+	cwd, _ := os.Getwd()
+	os.Chdir(parentDir)
+	labels, err := addGitProvenance(context.Background(), filepath.Base(repoDir), "")
+	assert.Nilf(t, err, "No error expected")
+	assert.Equal(t, "Dockerfile", labels[DockerfileLabel], "Expected a dockerfile path provenance label")
+	os.Chdir(cwd)
+}


tianon:
There should probably be another annotation for the context directory path, right?

AkihiroSuda:
> There should probably be another annotation for the context directory path, right?

Could you elaborate "annotation for the context directory path"?

cdupuis:
Sorry @tianon, somehow I missed your comment. Apologies for the late response. 

> There should probably be another annotation for the context directory path, right?

We are resolving the `Dockerfile` path against the root of the Git repository. 

Meaning if you build with `docker buildx build somepath -f Dockerfile -t some:tag` then we would record `somepath/Dockerfile` if the current working directory is the root of Git repository. 

Is that enough for what you had in mind?

commands/build.go
```diff
@@ -643,6 +649,77 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+		// obtain Git sha of current HEAD
+		cmd := exec.Command("git", "rev-parse", "HEAD")
+		cmd.Dir = wd
+		sha, err := cmd.Output()
+		if err != nil {
+			return nil, fmt.Errorf("error obtaining git head: %w", err)
+		}
+
+		// check if the current HEAD is clean
+		cmd = exec.Command("git", "diff", "--quiet", "--exit-code")
```
jedevc, line 676: We should try and avoid the git porcelain commands I think, and try and use the plumbing equivalent: https://stackoverflow.com/questions/2657935/checking-for-a-dirty-index-or-untracked-files-with-git.

I think we also should check for staged files? Not just the unstaged ones.

commands/build.go
```diff
@@ -643,6 +649,77 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+
+		// add the origin url if full Git details are requested
+		if v == "full" {
+			cmd = exec.Command("git", "config", "--get", "remote.origin.url")
+			cmd.Dir = wd
+			remote, err := cmd.Output()
+			if err != nil {
+				return nil, err
+			}
+			labels = append(labels, []string{fmt.Sprintf("org.opencontainers.image.source=%s", strings.TrimSpace(string(remote)))}...)
```
jedevc, line 693: These label strings should be `const`s.

tianon:
I mean something like `docker build -f dir1/Dockerfile dir2` (where both `dir1/Dockerfile` _and_ `dir2` are relevant bits of information, and both are required to figure out how the build happened).

Alternatively, something like `docker build -f ../foo/Dockerfile .` or `docker build -f docker/Dockerfile .` (which is probably a more common case) -- basically any instance where the `Dockerfile` does not sit at the root of the build context.

commands/build.go
```diff
@@ -643,6 +649,77 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
 	return f, nil
 }
 
+func addGitProvenance(labels []string, contextPath string, dockerfilePath string) ([]string, error) {
+	if v, ok := os.LookupEnv("BUILDX_GIT_INFO"); ok && contextPath != "" {
```
jedevc, line 653: nit: we should have this be an early exit, so we don't need to indent the rest of this function

commands/build.go
```diff
@@ -643,6 +649,77 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+		if len(labels) == 0 {
+			labels = make([]string, 0)
+		}
+
+		// figure out in which directory the git command needs to run
+		var wd string
+		cwd, _ := os.Getwd()
```
jedevc, line 660: nit: merge cwd + wd.

jedevc (changes requested):

commands/build.go
tonistiigi: https://github.com/docker/buildx/blob/master/hack/dockerfiles/vendor.Dockerfile#L33 what we do for vendor checks. @corhere suggestions?

cdupuis: Sorry, I'm confused a bit here: @jedevc you said to _avoid the git porcelain_ command but then @tonistiigi points at `git status --porcelain`. Not sure what to do ;-)

cdupuis:
@tianon, we are currently not storing the context path. So far this wasn't really needed for any of our use cases; we aren't trying to store all args etc to recreate the build command. It would be interesting to understand if you had a use case in mind for this? 

> I mean something like `docker build -f dir1/Dockerfile dir2` (where both `dir1/Dockerfile` _and_ `dir2` are relevant bits of information, and both are required to figure out how the build happened).

In this case we would store `dir1/Dockerfile` in the `com.docker.image.dockerfile.path` label, given current working directory is the root of the Git repository.

> Alternatively, something like `docker build -f ../foo/Dockerfile .` or `docker build -f docker/Dockerfile .` (which is probably a more common case) -- basically any instance where the `Dockerfile` does not sit at the root of the build context.

We would store `foo/Dockerfile` or `docker/Dockerfile` in the `com.docker.image.dockerfile.path` label, given `.` is also the root of the Git repository.

tianon:
The best I can come up with are use cases similar to wanting the `Dockerfile` path - imagine `COPY foo.sh /...` being able to link directly back to (the correct) `foo.sh` :sweat_smile:

(I don't feel strongly about this, that's just the only thing I noticed while looking at the annotations :+1:)

commands/build.go
tonistiigi: Actually both `diff` and `status` like in the vendor script are wrong for this case. Diff commit as it is defined atm does not show any untacked(new) files are all. Status does that but still ignores the files in `.gitignore`. `git status --porcelain --ignored` seems to do the trick but there might be a better way.

commands/build.go
```diff
@@ -43,6 +44,10 @@ import (
 	"google.golang.org/grpc/codes"
 )
 
+const ImageSourceLabel = "org.opencontainers.image.source"
+const ImageRevisionLabel = "org.opencontainers.image.revision"
```
tonistiigi, line 48: https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/annotations.go#L39

commands/build.go
```diff
@@ -43,6 +44,10 @@ import (
...
+const DockerfileLabel = "com.docker.image.dockerfile.path"
```
tonistiigi, line 49: Looking at the slsa definition maybe this should be `com.docker.image.source.entrypoint` . Might be better to avoid Dockerfile just because it doesn't actually need to be a Dockerfile as Buildkit supports many formats.

commands/build.go
```diff
@@ -643,6 +653,81 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+		cwd, _ := os.Getwd()
+		wd, _ = filepath.Abs(filepath.Join(cwd, contextPath))
+	}
+
+	// obtain Git sha of current HEAD
+	cmd := exec.Command("git", "rev-parse", "HEAD")
+	cmd.Dir = wd
+	sha, err := cmd.Output()
+	if err != nil {
+		return nil, fmt.Errorf("error obtaining git head: %w", err)
```
tonistiigi, line 680: In the build repos we only use `pkg/errors` as stdlib does not preserve stacktraces.

commands/build.go
```diff
@@ -643,6 +653,81 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+	}
+
+	// make the Dockerfile path relative to the Git root
+	if !filepath.IsAbs(dockerfilePath) {
+		cwd, _ := os.Getwd()
+		dockerfilePath = filepath.Join(cwd, dockerfilePath)
+	}
+	dockerfilePath, err = filepath.Rel(strings.TrimSpace(string(root)), dockerfilePath)
+	if err != nil {
+		return nil, err
```
tonistiigi, line 724: Dockerfile path does not need to be relative to the context directory. This should not fail the build. And path should never be `../../smth`

commands/build_test.go
```diff
@@ -0,0 +1,71 @@
...
+		t.Error("No error expected")
+	}
+	if labels != nil {
+		t.Error("No labels expected")
+	}
+}
+
+func TestAddGitProvenanceDataWithoutLabels(t *testing.T) {
+	os.Setenv("BUILDX_GIT_LABELS", "full")
+	labels, err := addGitProvenance(nil, ".", "")
```
tonistiigi, line 28: What is the "." directory here? Test should create its own testing repo and not expect to be run at a specific location(or am I missing something?).

tonistiigi (commented):
Should `buildx bake` commands also pick up these labels? In that case the code should be moved to the build pkg from commands(or a shared package that can be called from `commands/build` and `bake` pkg). And an additional test added to `bake/bake_test` that checks that target picks up these labels.

commands/build.go
```diff
@@ -643,6 +653,81 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+	// add the origin url if full Git details are requested
+	if v == "full" {
+		cmd = exec.Command("git", "config", "--get", "remote.origin.url")
+		cmd.Dir = wd
+		remote, err := cmd.Output()
+		if err != nil {
+			return nil, fmt.Errorf("failed to obtain remote.origin.url: %w", err)
+		}
+		labels = append(labels, []string{fmt.Sprintf("%s=%s", ImageSourceLabel, strings.TrimSpace(string(remote)))}...)
+	}
```
corhere, line 702: [The OCI image spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md) defines the `org.opencontainers.image.source` annotation as
> URL to get source code for building the image (string)

Presumably the intention is to specify an [RFC 1738](https://www.rfc-editor.org/rfc/rfc1738) URL. However, the `remote.<name>.url` git config value is expected to be a [Git URL](https://git-scm.com/docs/git-fetch#URLS), which include the decidedly non-URL "scp-like" and local path syntaxes. Could this be a problem for consumers of this label?

What about if the build context is a local repository? Naming the clone-source remote `origin` is merely a convention. There is no guarantee that a remote of that name exists, or that the `HEAD` commit ever existed in that remote. A repository may not even have _any_ remotes configured! Even in the happy case where `HEAD` is the head of a remote-tracking branch, the remote it references may not be named `origin`.

Rather than trying to be clever and magical with introspecting the repository, and potentially getting the wrong answer, have you considered setting this label iff the build context is a remote URL, setting it to that URL?

commands/build.go
cdupuis: Ah, makes sense. I'll use those.

commands/build.go
cdupuis: Sounds good to me.

commands/build.go
```diff
@@ -643,6 +653,81 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+func addGitProvenance(labels []string, contextPath string, dockerfilePath string) ([]string, error) {
+	v, ok := os.LookupEnv("BUILDX_GIT_LABELS")
+	if !ok || contextPath == "" {
+		return labels, nil
+	}
+
+	if len(labels) == 0 {
+		labels = make([]string, 0)
+	}
+
```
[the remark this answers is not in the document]
cdupuis: Indeed.

commands/build.go
```diff
@@ -643,6 +653,81 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+	// check if the current HEAD is clean
+	cmd = exec.Command("git", "diff", "--quiet", "--exit-code")
+	cmd.Dir = wd
+	_, err = cmd.Output()
+	if err == nil {
+		labels = append(labels, []string{fmt.Sprintf("%s=%s", ImageRevisionLabel, strings.TrimSpace(string(sha)))}...)
+	} else {
+		labels = append(labels, []string{fmt.Sprintf("%s=%s-dirty", ImageRevisionLabel, strings.TrimSpace(string(sha)))}...)
+	}
```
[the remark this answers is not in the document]
cdupuis: Thanks for spotting those.

commands/build.go
```diff
@@ -643,6 +653,81 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+	var wd string
+	if filepath.IsAbs(contextPath) {
+		wd = contextPath
+	} else {
+		cwd, _ := os.Getwd()
+		wd, _ = filepath.Abs(filepath.Join(cwd, contextPath))
+	}
+
+	// obtain Git sha of current HEAD
+	cmd := exec.Command("git", "rev-parse", "HEAD")
```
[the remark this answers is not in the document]
cdupuis: Added.

commands/build.go
cdupuis: Good point!

commands/build.go
cdupuis: > [The OCI image spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md) defines the `org.opencontainers.image.source` annotation as
> 
> > URL to get source code for building the image (string)
> 
> Presumably the intention is to specify an [RFC 1738](https://www.rfc-editor.org/rfc/rfc1738) URL. However, the `remote.<name>.url` git config value is expected to be a [Git URL](https://git-scm.com/docs/git-fetch#URLS), which include the decidedly non-URL "scp-like" and local path syntaxes. Could this be a problem for consumers of this label?

So far it hasn't been a problem and I kinda wanted to avoid trying to _translate_ those fetch urls into more human friendly ones. My assumption is that the value of `org.opencontainers.image.source` should be usable to get the source code for building which by using the Git urls seems correct, as one can take them and run `git clone <>`.

> What about if the build context is a local repository? Naming the clone-source remote `origin` is merely a convention. There is no guarantee that a remote of that name exists, or that the `HEAD` commit ever existed in that remote. A repository may not even have _any_ remotes configured! Even in the happy case where `HEAD` is the head of a remote-tracking branch, the remote it references may not be named `origin`.

Yes, that's a good point. I was too optimistic here. Changed that to search the remotes for one named `origin`; if not found we now take one of the configured ones. Also it won't fail now if there aren't any remotes configured for a local repository.

commands/build.go
```diff
@@ -643,6 +653,81 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+	cmd.Dir = wd
+	sha, err := cmd.Output()
+	if err != nil {
```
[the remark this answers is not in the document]
cdupuis: @tonistiigi @jedevc, what's your take on this? Should we or should we not fail if `BUILDX_GIT_LABELS` is enabled but the current context isn't a Git repository? I'm torn on that.

commands/build.go
cdupuis: Fixed!

commands/build_test.go
cdupuis: Creating my own test git repo now.

commands/build.go
jedevc: `git status --porcelain --ignored` sounds good to me: from `git-status`:
> Give the output in an easy-to-parse format for scripts. This is similar to the short output, but will remain stable across Git versions and regardless of user configuration. See below for details.

cdupuis: I've switched to that for now.

commands/build.go
```diff
@@ -643,6 +652,101 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
 	return f, nil
 }
 
+func addGitProvenance(ctx context.Context, labels []string, contextPath string, dockerfilePath string) ([]string, error) {
```
jedevc, line 655: nit: I would rather if we could avoid taking `labels` as an argument here (since we don't modify any of the contents, except for adding more), and append it at the call site.

commands/build.go
```diff
@@ -643,6 +652,101 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+	}
+	if len(strings.TrimSpace(string(out))) == 0 {
+		labels = append(labels, fmt.Sprintf("%s=%s", annotations.AnnotationRevision, sha))
+	} else {
+		labels = append(labels, fmt.Sprintf("%s=%s-dirty", annotations.AnnotationRevision, sha))
+	}
+
+	// add a remote url if full Git details are requested; if there aren't any remotes don't fail
+	if v == "full" {
+		cmd = exec.CommandContext(ctx, "git", "remote")
```
jedevc, line 694: I *think* we can use `git ls-remote --get-url` to get the URL of the current remote: https://git-scm.com/docs/git-ls-remote#Documentation/git-ls-remote.txt---get-url, so we shouldn't need to interact with `git config` manually.

commands/build.go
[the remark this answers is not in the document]
jedevc: I think a failure needs to be visible to the user - we definitely shouldn't silently fail. If the builder is requested to add git labels, but can't, then at the very least we should log a failure in the build output, but I don't have a strong opinion on a hard/soft failure.

commands/build.go
cdupuis: Done

commands/build.go
cdupuis: Done

commands/build.go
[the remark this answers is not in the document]
cdupuis: @jedevc what is the best way to get something into the build log?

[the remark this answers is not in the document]
jedevc: Hm, actually I'm not sure if it's possible to easily get something into the build log from here - we could just have a warning message before/after the build starts/completes.

commands/build.go
```diff
@@ -643,6 +653,90 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+		logrus.Warnf("Unable to determine Git information")
+		return nil, nil
+	}
+
+	// obtain Git sha of current HEAD
+	cmd = exec.CommandContext(ctx, "git", "rev-parse", "HEAD")
+	cmd.Dir = wd
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, errors.Errorf("error obtaining git head: %v", err)
```
tonistiigi, line 686: `errors.Wrap(err, "error obtaining.. `

Same for the rest of the errors

commands/build.go
```diff
@@ -643,6 +653,90 @@ func parsePrintFunc(str string) (*build.PrintFunc, error) {
...
+	sha := strings.TrimSpace(string(out))
+
+	// check if the current HEAD is clean
+	cmd = exec.CommandContext(ctx, "git", "status", "--porcelain", "--ignored")
+	cmd.Dir = wd
+	out, err = cmd.Output()
+	if err != nil {
+		return nil, errors.Errorf("error obtaining git status: %v", err)
+	}
+	if len(strings.TrimSpace(string(out))) == 0 {
```
tonistiigi, line 697: I think 

```
if len(strings.TrimSpace(string(out))) != 0 {
   sha += "-dirty"
}
```

reads better. (btw is the trimspace really required?)

commands/build_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+var repoDir string
+
+func setupTest(tb testing.TB) func(tb testing.TB) {
+	repoDir = tb.TempDir()
+	cmd := exec.Command("git", "init")
+	cmd.Dir = repoDir
+	err := cmd.Run()
+	if err != nil {
+		tb.Errorf("failed to init git repo: %v", err)
```
tonistiigi, line 23: btw, we do use `stretchr/testify` in the tests. It is not really a requirement if the tests are simple and don't require additional library but FYI in case you didn't notice.

tonistiigi (commented):
I think you missed https://github.com/docker/buildx/pull/1297#pullrequestreview-1092473232 comment

commands/build.go
cdupuis: Ah got it. I was just trying to use the same pattern that is used elsewhere in `build.go`. Every other location uses `errors.Errorf`.

Happy to change my code though.

commands/build.go
cdupuis: 👍🏽

cdupuis:
> Should `buildx bake` commands also pick up these labels? In that case the code should be moved to the build pkg from commands(or a shared package that can be called from `commands/build` and `bake` pkg). And an additional test added to `bake/bake_test` that checks that target picks up these labels.

@tonistiigi does this need to happen? In this PR or can this come after? If so, could you give some pointers?

jedevc:
@cdupuis yup it should happen ideally - we want to avoid bake and buildx producing different results.

My recommendation for how to split it up:
- Add the git helpers to `build/git.go`, and the tests to `build/git_test.go`
- Call `addGitProvenance` from `BuildWithResultHandler` and modify all the opts structures in the opts `map[string]BuildOpts` with the correct labels, using the `BuildOpts.Inputs`

The code should pretty much stay the same, though the additional test in `bake_test.go` with an e2e test would be super useful (happy to take a look at that if you let me know).

cdupuis:
Thanks @jedevc. I've made those changes. Let me know what you think.

jedevc (approved):
LGTM.

@tonistiigi I'm not sure how we'd do a test here, since the existing tests test for `ReadTargets`  - we don't actually have an easy way to check the opts sent to the builder, either in `BuildWithResultHandler` or in `bake` and `commands/build` - we don't have any tests for this at all anywhere in buildx. Loading them in `ReadTargets` definitely doesn't feel right, since then the `LABEL`s would appear in the `--print` results. I'd be happy to merge without the test for now (we could revisit later).

build/git.go
```diff
@@ -0,0 +1,98 @@
...
+		cmd.Dir = wd
+		out, _ := cmd.Output()
+		if len(out) > 0 {
+			labels[ocispecs.AnnotationSource] = strings.TrimSpace(string(out))
+		}
+	}
+
+	// add Dockerfile path; there is no org.opencontainers annotation for this
+	if dockerfilePath == "" {
+		dockerfilePath = "Dockerfile"
```
jedevc, line 75: Had a quick test, I think the path manipulation is wrong here.

```suggestion
		dockerfilePath = path.Join(contextPath, "Dockerfile")
```

If I do a build in a context that's not current directory and don't set the Dockerfile explicitly, then the `Dockerfile` won't be added, e.g. `docker buildx build ~/target` if I'm currently in `~/someplaceelse`.

build/build.go
```diff
@@ -786,6 +786,16 @@ func BuildWithResultHandler(ctx context.Context, drivers []DriverInfo, opt map[s
 
 	eg, ctx := errgroup.WithContext(ctx)
 
+	for _, opt := range opt {
+		gitLabels, err := addGitProvenance(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath)
+		if err != nil {
+			return nil, err
+		}
+		for n, v := range gitLabels {
+			opt.Labels[n] = v
```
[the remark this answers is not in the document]
cdupuis: I was wondering about that, too. Given it is opt-in, should we print a warning if either of the labels are set?

build/git.go
cdupuis: Yes, you're right. Fixing.

tonistiigi:
> Loading them in ReadTargets definitely doesn't feel right, since then the LABELs would appear in the --print results. 

Why isn't it correct to show these in `--print` results? That should show the configuration of the build that is sent to the daemon, doesn't matter if it is loaded from files, `--set` or from environment variables.

build/build.go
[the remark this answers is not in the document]
tonistiigi: Looks this has not been added?

tonistiigi (commented):
I think the `bake` part is wrong and should be handed in `ReadTargets` so users see what config `bake` produces. If these labels are already set by the config itself, then the environment variables do not override them. I'm fine with merging the current implementation and creating follow-up PR for the `bake` refactor and test.

cdupuis:
> I think the `bake` part is wrong and should be handed in `ReadTargets` so users see what config `bake` produces. If these labels are already set by the config itself, then the environment variables do not override them. I'm fine with merging the current implementation and creating follow-up PR for the `bake` refactor and test.

@tonistiigi, I'm happy to work on this in a subsequent PR.

build/build.go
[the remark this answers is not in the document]
tonistiigi: I'm not sure if a warning is even needed, but I don't think we should not override the values if the user set them manually.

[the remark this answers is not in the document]
cdupuis: > I don't think we should not override the values if the user set them manually.

Double negation always confuses me 😕 , are you saying we should overwrite the labels even though labels might have been set by the user without a warning log?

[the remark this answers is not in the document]
tonistiigi: If user set `build --label org.opencontainers.image.revision=123` then we should not replace that value with the results of the git commands we ran in the background. Then the `123` value should be that gets built.

build/build.go
```diff
@@ -786,6 +786,18 @@ func BuildWithResultHandler(ctx context.Context, drivers []DriverInfo, opt map[s
 
 	eg, ctx := errgroup.WithContext(ctx)
 
+	for _, opt := range opt {
+		gitLabels, err := addGitProvenance(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath)
+		if err != nil {
+			return nil, err
+		}
+		for n, v := range gitLabels {
+			if opt.Labels[n] == "" {
```
tonistiigi, line 795: this should be

```
if _, ok := opt.Labels[n]; !ok {
```
```

cdupuis: ah, once again. I need to learn ....

tonistiigi (approved):

Issue #1058, opened 2022-04-09
`bake --print` show wrong docker-bake.json when we target a group which targets groups


eitsupi:
When printing a group whose target is a group, it appears that the json is printed with a target name that does not exist in the json file for the target.

For example, a hcl file like that:

```hcl
group "default" {
    targets = ["other-group"]
}

group "other-group" {
    targets = ["app"]
}

target "app" {
    dockerfile = "Dockerfile"
    tags = ["docker.io/username/app"]
}
```

```shell
$ docker buildx bake --print -f docker-bake.hcl
[+] Building 0.0s (0/0)                                                                          
{
  "group": {
    "default": {
      "targets": [
        "other-group"
      ]
    }
  },
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/app"
      ]
    }
  }
}
```

The groups included in the default group must be included in the json.

jedevc:
This behavior seems to be explicitly tested for in [bake_test.go](https://github.com/docker/buildx/blob/master/bake/bake_test.go#L1049-L1089)

@crazy-max do you know why the targets aren't recursively resolved?

eitsupi:
Any update?

jedevc:
Sorry for the delay :tada: 

So I discussed with @crazy-max briefly internally, and the current behavior of the group key not appearing in the output JSON is intended. As I understand, there are two possible behaviors and logic behind them:

- `--print` prints all the groups that are directly part of the name specified by the command line argument (or `default` if not specified), and also prints all the targets that are directly part of the name specified. If you want programmable use, you can invoke the command again to traverse the nested group structures of the bake file.
- `--print` prints all the groups that are recursively part of the command line argument, and all the targets that all of those groups contain. If you want programmable use, all of the data is possible to consume directly, however, it doesn't just print that target.

    One issue with printing targets recursively, is that targets and groups may have non-trivial dependencies. If we were to list all the targets, then if we used the new `target:` syntax for overriding a build context, should that dependency target be listed in the output?

However, actually, we don't do either of these at the moment :eyes: We always return the `default` group name in the JSON, even if we specify a different set of targets:

```
$ docker buildx bake --print validate
[+] Building 0.0s (0/0)                                                                                                                     
{
  "group": {
    "default": {
      "targets": [
        "lint",
        ...
```

[...]

PR #1303, opened 2022-09-02
bake: ensure ReadTargets returns the original group

jedevc:
This implements the first option from #1058 - it's also left open to now be possible to implement recursive group resolution, though that hasn't been added yet. Since we already resolve targets recursively, I think we probably should resolve all the groups recursively as well, but that's something to consider in a different PR.

With this patch, there's a couple of implications. With the file:

```hcl
group "default" {
    targets = ["other-group"]
}

group "other-group" {
    targets = ["app"]
}

target "app" {
    dockerfile = "Dockerfile"
    tags = ["docker.io/username/app"]
}
```

We get the following different behavior:
```bash
$ docker buildx bake --print other-group
{
  "group": {
    "other-group": { <-- this is other-group, not default
      "targets": [
        "app"
      ]
    }
  },
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/app"
      ]
    }
  }
}

$ docker buildx bake --print app
{  <-- app is not part of the default group, so don't print a group here
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/app"
      ]
    }
  }
}
```

In the latter, `app` is not part of a selected group, so the field is left empty.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -84,7 +84,7 @@ func ReadLocalFiles(names []string) ([]File, error) {
 	return out, nil
 }
 
-func ReadTargets(ctx context.Context, files []File, targets, overrides []string, defaults map[string]string) (map[string]*Target, []*Group, error) {
+func ReadTargets(ctx context.Context, files []File, targets, overrides []string, defaults map[string]string) (map[string]*Target, map[string]*Group, error) {
 	c, err := ParseFiles(files, defaults)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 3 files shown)


crazy-max:
```console
$ docker buildx bake --print app
{  <-- app is not part of the default group, so don't print a group here
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/app"
      ]
    }
  }
}
```

I agree that it makes things more consistent with HCL but users might just want to reuse this output without specifying the target name:

```json
{
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/app"
      ]
    }
  }
}
```
> app.json

```console
$ docker buildx bake -f app.json --print
ERROR: failed to find target default
```

That's why we currently also prints a group named like the target:

```json
{
  "group": {
    "default": {
      "targets": [
        "app"
      ]
    }
  },
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/app"
      ]
    }
  }
}
```
> app.json

```console
$ docker buildx bake -f app.json --print
{
  "group": {
    "default": {
      "targets": [
        "app"
      ]
    }
  },
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/app"
      ]
    }
  }
}
```

tonistiigi:
Initially we didn't print group information at all. The reason we added was that people wanted to do.

```
docker buildx bake foobar --print > foobar.json
docker buildx bake -f foobar.json
```

If it is the original group name, then this pattern does not work

jedevc:
> Initially we didn't print group information at all. The reason we added was that people wanted to do.

Aha, thanks for that :) I hadn't realized that intended use case. To support that case with nested groups, we definitely want to include all the sub-groups of groups of the selected group - though there's the option to flatten them, or not.

Will close this PR, and open another one with a reworked fix along those lines.

PR #1305, opened 2022-09-02
progress: add prefix to vertex progress group

### LINKED ISSUE #1195: Bake does not show target stage for `COPY --link`
When I build multiple targets (e.g. via a group or explicitly) it's not often clear which target a build(ing) stage corresponds to

[sample log](https://gist.github.com/ciaranmcnulty/ad664329a303ed85d73d4f48735db884)
[Dockerfile](https://github.com/ciaranmcnulty/phpqa/blob/535fe6f857d4b774dc34b9f15d378d0043e2c9c8/Dockerfile)

The build steps that don't have a stage name are the ones that do `COPY --link`, but not all of the them are missing it. 

I'm not clear what is happening behind the scenes to cause this!

jedevc:
Fixes #1195. CC @ciaranmcnulty.

As buildkit now uses progress groups for the COPY --link instruction we need to ensure that we additionally prefix the progress group name, or the target name will be left off in bake commands with more than one target.

### DIFF
--- a/util/progress/multiwriter.go
+++ b/util/progress/multiwriter.go
@@ -24,6 +24,9 @@ func (p *prefixed) Write(v *client.SolveStatus) {
 	if p.force {
 		for _, v := range v.Vertexes {
 			v.Name = addPrefix(p.pfx, v.Name)
+			if v.ProgressGroup != nil {
+				v.ProgressGroup.Name = addPrefix(p.pfx, v.ProgressGroup.Name)
+			}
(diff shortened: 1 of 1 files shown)


tonistiigi:
@sipsma

tonistiigi (approved):

crazy-max (approved):
Looks also related to https://github.com/moby/buildkit/issues/3323

PR #1313, opened 2022-09-12
bake: recursively resolve groups

jedevc:
Fixes https://github.com/docker/buildx/issues/1058.

Groups that contained other groups were not recursively resolved by `ReadTargets`, which prevented output from `--print` from being useable as a self-contained bake file. e.g.

```hcl
group "default" {
    targets = ["other-group"]
}

group "other-group" {
    targets = ["app"]
}

target "app" {
    dockerfile = "Dockerfile"
    tags = ["docker.io/username/app"]
}
```
is used as 
```bash
$ buildx bake --print -f docker-bake.hcl
[+] Building 0.0s (0/0)                                                                                                    
{
  "group": {
    "default": {
      "targets": [
        "other-group"
      ]
    },
    "other-group": {
      "targets": [
        "app"
      ]
    }
  },
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "docker.io/username/app"
      ]
    }
  }
}

```

This patch ensures that all groups that are referenced inside the bake file are actually defined under the groups field. This has required a substantial refactor, as previously only a single group was returned from `ReadTargets`, notably, returning a map of groups, instead of a slice.

This does introduce a small behavior change to the behavior of `--print` - while previously, passing a group name to bake would return all the targets of that group back as the default group, now only the name of that group will be inserted into the default group, keeping the original group intact. The impact of this can be observed in some of the changes to the `bake_test.go` file.

e.g. previously with buildx's docker-bake.hcl:
```
$ docker buildx bake --print validate
[+] Building 0.0s (0/0)                                                                                                    
{
  "group": {
    "default": {
      "targets": [
        "lint",
        "validate-vendor",
        "validate-docs"
      ]
    }
  },
```
vs now:
```
$ buildx bake --print validate
[+] Building 0.0s (0/0)                                                                                                    
{
  "group": {
    "default": {
      "targets": [
        "validate"
      ]
    },
    "validate": {
      "targets": [
        "lint",
        "validate-vendor",
        "validate-docs"
      ]
    }
  },
```

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -84,7 +84,7 @@ func ReadLocalFiles(names []string) ([]File, error) {
 	return out, nil
 }
 
-func ReadTargets(ctx context.Context, files []File, targets, overrides []string, defaults map[string]string) (map[string]*Target, []*Group, error) {
+func ReadTargets(ctx context.Context, files []File, targets, overrides []string, defaults map[string]string) (map[string]*Target, map[string]*Group, error) {
 	c, err := ParseFiles(files, defaults)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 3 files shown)


crazy-max (approved):
LGTM

crazy-max:
PTAL @tonistiigi

PR #1316, opened 2022-09-13
docs: add cache storage backend docs

jedevc:
This fleshes out the documentation to include information along with examples and detailed use cases for the cache backends. The general format and style follows from the style of the build driver docs.

Eventually, these docs will be included on docs.docker.com along with the rest of the build docs. This docs are designed to complement the introduction in https://github.com/docker/docker.github.io/pull/15603, and after both are merged, we should be able to cross-link between them.

The `s3` and `azblob` backends are both relatively sparse on detail, since I don't have the knowledge to write those best, but would be happy to address these in a follow-up (unless there are any glaring issues wrong with them atm).

### DIFF
--- a/docs/guides/cache/azblob.md
+++ b/docs/guides/cache/azblob.md
@@ -30,15 +30,15 @@ $ docker buildx build . --push -t <registry>/<image> \
 Common parameters:
 
 - `name`: the name of the cache image.
-
-Parameters for `--cache-to`:
-
 - `account_url`: the base address of the blob storage account, for example:
   `https://myaccount.blob.core.windows.net`. See
   [authentication](#authentication).
 - `secret_access_key`: specifies the
   [Azure Blob Storage account key](https://docs.microsoft.com/en-us/azure/storage/common/storage-account-keys-manage),
   see [authentication](#authentication).
+
+Parameters for `--cache-to`:
+
 - `mode`: specify cache layers to export (default: `min`), see
   [cache mode](./index.md#cache-mode)
 
--- a/docs/guides/cache/azblob.md
+++ b/docs/guides/cache/azblob.md
@@ -2,61 +2,57 @@
 
 > **Warning**
 >
-> The `azblob` cache is currently unreleased. You can use it today, by using
-> the `moby/buildkit:master` image in your Buildx driver.
+> This cache backend is unreleased. You can use it today, by using the
+> `moby/buildkit:master` image in your Buildx driver.
 
 The `azblob` cache store uploads your resulting build cache to
 [Azure's blob storage service](https://azure.microsoft.com/en-us/services/storage/blobs/).
 
 > **Note**
 >
-> The `azblob` cache storage backend requires using a different driver than
-> the default `docker` driver - see more information on selecting a driver
-> [here](../drivers/index.md). To create a new docker-container driver (which
-> can act as a simple drop-in replacement):
+> This cache storage backend requires using a different driver than the default
+> `docker` driver - see more information on selecting a driver
+> [here](../drivers/index.md). To create a new driver (which can act as a simple
+> drop-in replacement):
 >
 > ```console
 > docker buildx create --use --driver=docker-container
 > ```
 
-To import and export your cache using the `azblob` storage backend we use the
-`--cache-to` and `--cache-from` flags and point it to our desired blob using
-the required `account_url` and `name` parameters:
+## Synopsis
 
 ```console
-$ docker buildx build --push -t <user>/<image> \
-  --cache-to type=azblob,account_url=https://myaccount.blob.core.windows.net,name=my_image \
-  --cache-from type=azblob,account_url=https://myaccount.blob.core.windows.net,name=my_image
+$ docker buildx build . --push -t <registry>/<image> \
+  --cache-to type=azblob,name=<cache-image>[,parameters...] \
+  --cache-from type=azblob,name=<cache-image>[,parameters...]
 ```
+
+Common parameters:
+
+- `name`: the name of the cache image.
+
+Parameters for `--cache-to`:
+
+- `account_url`: the base address of the blob storage account, for example:
+  `https://myaccount.blob.core.windows.net`. See
+  [authentication](#authentication).
+- `secret_access_key`: specifies the
+  [Azure Blob Storage account key](https://docs.microsoft.com/en-us/azure/storage/common/storage-account-keys-manage),
+  see [authentication](#authentication).
+- `mode`: specify cache layers to export (default: `min`), see
+  [cache mode](./index.md#cache-mode)
 
 ## Authentication
 
-To authenticate to Azure to read from and write to the cache, the following
-parameters are required:
-
-* `secret_access_key`: secret access key
-  * specifies the primary or secondary account key for your Azure Blob
-    Storage account. [Azure Blob Storage account keys](https://docs.microsoft.com/en-us/azure/storage/common/storage-account-keys-manage)
-
-While these can be manually provided, if left unspecified, then the credentials
-for Azure will be pulled from the BuildKit server's environment following the
-environment variables scheme for the [Azure Go SDK](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication).
-
-> **Warning**
->
-> These environment variables **must** be specified on the BuildKit server, not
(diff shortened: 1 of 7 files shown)


docs/guides/cache/azblob.md
```diff
@@ -0,0 +1,57 @@
+# Azure Blob Storage cache storage
+
+The `azblob` cache store uploads your resulting build cache to
+[Azure's blob storage service](https://azure.microsoft.com/en-us/services/storage/blobs/).
+
+> *Note*
+>
+> The `azblob` cache storage backend requires using a different driver than
+> the default `docker` driver - see more information on selecting a driver
+> [here](../drivers/index.md). To create a new docker-container driver (which
```
crazy-max, line 10: looking at the new hierarchy of our docs, a relative link lgtm here.

docs/guides/cache/gha.md
```diff
@@ -0,0 +1,101 @@
...
+      -
+        name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: true
+          tags: user/app:latest
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+```
```
crazy-max, line 87: cross link to new `ci` section in follow-up when available.

docs/guides/cache/azblob.md
```diff
@@ -0,0 +1,57 @@
...
+environment variables scheme for the [Azure Go SDK](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication).
+
+> *Warning*
+>
+> These environment variables **must** be specified on the BuildKit server, not
+> the `buildx` client.
+
+## Cache options
+
+The `azblob` cache has lots of parameters to adjust its behavior.
```
crazy-max, line 47: We can link to buildkit readme section or copy paste options available here. I have a preference to copy content here so users are not confused about buildctl vs. buildx.

docs/guides/cache/gha.md
```diff
@@ -0,0 +1,101 @@
...
+
+The `gha` cache has lots of parameters to adjust its behavior.
+
+### Cache mode
+
+See [Registry - Cache mode](./registry.md#cache-mode) for more information.
+
+## Further reading
+
+For an introduction to caching see [Optimizing builds with cache management](https://docs.docker.com/build/cache).
```
crazy-max, line 99: would be `/build/building/cache`.

docs/guides/cache/index.md
```diff
@@ -0,0 +1,97 @@
...
+
+BuildKit supports multiple cache backends, however, they do support different
+sets of features.
+
+| Feature                             | `inline` | `registry` | `local` |    `gha`     |    `s3`    |  `azblob`  |
+|:------------------------------------|:--------:|:----------:|:-------:|:------------:|:----------:|:----------:|
+| **Status**                          |  stable  |   stable   | stable  | experimental | unreleased | unreleased |
+| **Compatible with `docker` driver** |    ✅     |     ❌      |    ❌    |      ❌       |     ❌      |     ❌      |
+| **Compatible with all exporters**   |    ❌     |     ✅      |    ✅    |      ✅       |     ✅      |     ✅      |
+| **External authentication**         |    ❌     |     ❌      |    ❌    |      ✅       |     ✅      |     ✅      |
```
crazy-max, line 97: `External authentication` looks confusing to me. `inline` and `registry` auth is not required when fetching cache but exporting is.

Maybe we should remove this table and stipulate support and auth mode on each cache exporter page. WDYT?

docs/guides/cache/local.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Cache compression
+
+See [Registry - Cache compression](./registry.md#cache-compression) for more information.
+
+## Further reading
+
+For an introduction to caching see [Optimizing builds with cache management](https://docs.docker.com/build/cache).
+
+For more information on the `local` cache backend, see the [BuildKit README](https://github.com/moby/buildkit#local-directory-1).
```
crazy-max, line 87: For this exporter, maybe add a note/warning saying caches are copied over the existing cache so it [keeps growing](https://github.com/moby/buildkit/issues/1896).

crazy-max (commented):
small changes, otherwise docs layout lgtm. I can take a look at `s3` example in a follow-up.

docs/guides/cache/azblob.md
jedevc: Yeah, I've tried to summarize the options neatly for the `registry` / `local` / `gha` cache backends, the only reason I haven't put the whole list here is that the `azblob`/`s3` pages are not fully complete yet. I can just paste them in for now, but I'd prefer to do it with some explanation text as well in a follow-up for these?

docs/guides/cache/index.md
jedevc: Agreed, not sure the table works very well, we've also got the support information above in the list, can duplicate it on the pages.

docs/guides/cache/local.md
jedevc: Will add this in the cache versioning section where there is a mention of how the cache grows, will also link to that issue.

docs/guides/cache/gha.md
jedevc: Have added an inline FIXME :+1:

docs/guides/cache/azblob.md
crazy-max: Sure let's see in a follow-up

crazy-max (approved):
LGTM, PTAL @tonistiigi

docs/guides/cache/registry.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Cache compression
+
+Since `registry` cache is exported separately from the main build result, you
+can specify separate compression parameters for it (which are similar to the
+options provided by the `image` exporter). While the defaults have been
+selected to provide a good out-of-the-box experience, you may wish to tweak the
+parameters to optimize for storage vs compute costs.
+
+To select the compression algorithm, you can use the `compression=<uncompressed|gzip|estargz|zstd>`
```
nanake, line 83: does this require a `oci-mediatypes=true`? https://github.com/moby/buildkit#:~:text=estargz%20should%20be%20used%20with%20oci%2Dmediatypes%3Dtrue

jedevc: `oci-mediatypes` will be implicitly set in these cases for estargz. I'm not sure if we need to call this out here though, I'm on the fence about how much to discuss `oci-mediatypes`, since ideally, at some point we'd switch to using this as the default.

docs/guides/cache/registry.md
```diff
@@ -0,0 +1,106 @@
...
+option. For example, to build the cache with `compression=zstd`:
+
+```console
+$ docker buildx build --push -t <user>/<image> \
+  --cache-to type=registry,ref=<user>/<cache-image>,compression=zstd \
+  --cache-from type=registry,ref=<user>/<cache-image> .
+```
+
+The `compression-level=<value>` option can be used alongside the `compression`
```
nanake, line 92: if this option was not specified, what was the default `compression-level` value for `estargz` and `zstd`?

jedevc: These are properties of the underlying libraries we use - for `gzip`/`estargz` this is `compress/gzip`, while for `zstd` this is `github.com/klauspost/compress/zstd`. Since these constants could change at any time, I don't think we should list them here.

Unfortunately, not sure if we should link to upstream for these: https://pkg.go.dev/compress/gzip doesn't have neat docs on what the default is, though https://github.com/klauspost/compress/tree/master/zstd#status does (under the hood it doesn't use the same numbering level for zstd though).

I'll put the ranges of valid values in, and add links to the projects that do the compression, so users can at least find this info.

docs/guides/cache/registry.md
jedevc: I've gone and added a note about the `oci-mediatypes` option for the `registry` and `local` exporters, but left off the details about `estargz` for now. Later, we'll cross-link to `../exporters/image.md` (which will have much more information on that interaction) once I get round to writing those :smile:

dvdksn:
hey! I did some change suggestions in this PR—please review: https://github.com/jedevc/buildx/pull/2

dvdksn (approved):

crazy-max (approved):
LGTM!

PR #1317, opened 2022-09-13
docs: fix link in driver guide

crazy-max:
While working on merging buildx overview page in main overview one, I found out a link was broken in the drivers guide:

```
#16 0.099 - ./_site/build/buildx/drivers/docker/index.html
#16 0.099   *  internally linking to ../index.md, which does not exist (line 147)
#16 0.099      <a href="../index.md">other drivers</a>
```

Looks like we linked _others drivers_ to the buildx overview page which is also broken on GitHub: https://github.com/docker/buildx/blob/master/docs/guides/index.md.

@jedevc I guess you wanted to link to the drivers guide index page.

Also update links to driver reference. About left `FIXME` I don't think we should use relative links actually as we could easily break pages when moving them between sections in docs repo. Also with the `fix_url` jekyll plugin on docs repo we are safe to use absolute urls.

### DIFF
--- a/docs/guides/drivers/docker-container.md
+++ b/docs/guides/drivers/docker-container.md
@@ -71,5 +71,3 @@ REPOSITORY                       TAG               IMAGE ID       CREATED       
 ## Further reading
 
 For more information on the docker-container driver, see the [buildx reference](https://docs.docker.com/engine/reference/commandline/buildx_create/#driver).
-
-<!--- FIXME: for 0.9, make reference link relative --->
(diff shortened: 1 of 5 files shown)


jedevc (approved):
Yup that was an oversight, my bad.

R.e. relative links, makes sense, but it is frustrating to have links in markdown that link to the docs site, even though the reference content is in this repo :thinking:

crazy-max:
> R.e. relative links, makes sense, but it is frustrating to have links in markdown that link to the docs site, even though the reference content is in this repo 🤔

Relative links to internal content is fine to me so users can still navigate in markdown docs on the repo but don't think we have the choice for content outside the scope of the repo atm.

jedevc:
> Relative links to internal content is fine to me so users can still navigate in markdown docs on the repo but don't think we have the choice for content outside the scope of the repo atm.

Agreed - but aren't the buildx reference pages within the scope of this repo? :eyes: e.g. from [here](https://github.com/docker/buildx/pull/1317/files#diff-882013533859f16e2750eb51df03a04a95e873ed2784860fededebffd76f88a5L48)

PR #1324, opened 2022-09-20
invoke: load defaults from image config

jedevc:
If user does not specify image certain container parameters, we can load them from the exporter metadata :tada:

Additionally, we introduce a new "default" value for the --invoke flag, that keeps all of the default parameters (since cobra does not have an easy way of accepting an optional flag argument).

Example with overriding the `USER`:

```dockerfile
#syntax=docker/dockerfile:1
FROM ubuntu:22.04
RUN useradd -ms /bin/bash foo
USER foo
```
```bash
$ BUILDX_EXPERIMENTAL=1 docker buildx build --invoke sh .
...
$ whoami
foo
```

I'm not sure about the interaction of `CMD` and `ENTRYPOINT` here, it doesn't feel right, but I can't quite find the logic of how they're joined together in moby.

CC @ktock


build/build.go
```diff
@@ -703,11 +706,49 @@ func Invoke(ctx context.Context, cfg ContainerConfig) error {
...
+		}
+		cwd := ""
+		if cfg.Cwd != nil {
+			cwd = *cfg.Cwd
+		} else if img != nil {
+			cwd = img.Config.WorkingDir
+		}
+		env := cfg.Env
+		if env == nil && img != nil {
+			env = img.Config.Env
```
tonistiigi, line 733: I think image env should be always set. If user sets own env they are added on top of image like in `docker run`

build/build.go
```diff
@@ -703,11 +706,49 @@ func Invoke(ctx context.Context, cfg ContainerConfig) error {
...
+		cwd := ""
+		if cfg.Cwd != nil {
+			cwd = *cfg.Cwd
+		} else if img != nil {
+			cwd = img.Config.WorkingDir
+		}
+		env := append([]string{}, cfg.Env...)
+		if img != nil {
+			env = append(env, img.Config.Env...)
+		}
```
ktock, line 734: I think environment variables in `ContainerConfig` should be prioritized.
Current behaviour looks like envvars set by user aren't set if the image already has them.

e.g. `foo=bar` isn't set:

```dockerfile
FROM busybox
RUN echo hello > /hello
ENV foo=foo
```

```console
$ BUILDX_EXPERIMENTAL=1 /tmp/out/buildx build --invoke env=foo=bar,env=foo2=bar2 /tmp/ctx
[+] Building 4.1s (6/6) FINISHED                                                                                                   
 => [internal] load .dockerignore                                                                                             0.0s
 => => transferring context: 2B                                                                                               0.0s
 => [internal] load build definition from Dockerfile                                                                          0.0s
 => => transferring dockerfile: 86B                                                                                           0.0s
 => [internal] load metadata for docker.io/library/busybox:latest                                                             3.6s
 => [auth] library/busybox:pull token for registry-1.docker.io                                                                0.0s
 => [1/2] FROM docker.io/library/busybox@sha256:ad9bd57a3a57cc95515c537b89aaa69d83a6df54c4050fcf2b41ad367bec0cd5              0.3s
 => => resolve docker.io/library/busybox@sha256:ad9bd57a3a57cc95515c537b89aaa69d83a6df54c4050fcf2b41ad367bec0cd5              0.0s
 => => sha256:729ce43e2c915c3463b620f3fba201a4a641ca5a282387e233db799208342a08 772.99kB / 772.99kB                            0.2s
 => => extracting sha256:729ce43e2c915c3463b620f3fba201a4a641ca5a282387e233db799208342a08                                     0.0s
 => [2/2] RUN echo hello > /hello                                                                                             0.1s
/ # env
OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc
SHLVL=1
foo=foo
HOME=/root
foo2=bar2
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=unix:///dev/otel-grpc.sock
PWD=/
OTEL_TRACES_EXPORTER=otlp
```

jedevc: :tada: oops, good catch, will fixup that up :smile:

ktock (approved):

tonistiigi (approved):

PR #1326, opened 2022-09-21
cli: update help message for --push

jedevc:
The default behavior of `--push` has changed from `--output type=registry` to `--output type=image,push=true`, with `type=registry` now otherwise undocumented.

As a cleanup for docs, we should update the flag inline with the true behavior.


crazy-max (approved):

tonistiigi:
These should be equivalent. Was there really any change of behavior?

jedevc:
Yup, sorry, they are equivalent in behavior - this patch just brings the `--help` changes inline with how it works internally.

The logic here is in starting to write docs for all the exporters: `image`, `local`, `tar`, etc. Given we don't actually mention `type=registry` anywhere currently in the docs, and it falls back to `type=image,push=true` internally, I think it would make sense to keep it out of the docs and `--help`, and encourage users to use `type=image` wherever possible, while keeping it for backwards-compatibility.

tonistiigi:
It is mentioned in https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#registry . It is an alias so it should be allowed to use both variants interchangeably. If we already have an alias that is easier to understand and write for the user, then I don't see why we shouldn't also promote it in the help messages.

PR #1334, opened 2022-09-29
Driver docs update

dvdksn:
Updating driver documentation, making editorial and style changes, and aligning with documentation for caching backends.

### DIFF
--- a/docs/guides/drivers/docker-container.md
+++ b/docs/guides/drivers/docker-container.md
@@ -17,7 +17,10 @@ Run the following command to create a new builder, named `container`, that uses
 the Docker container driver:
 
 ```console
-$ docker buildx create --name container --driver docker-container
+$ docker buildx create \
+  --name container \
+  --driver=docker-container \
+  --driver-opt=[key=value,...]
 container
 ```
 
@@ -26,13 +29,14 @@ pass to `--driver-opt`:
 
 | Parameter       | Value  | Default          | Description                                                                                |
 | --------------- | ------ | ---------------- | ------------------------------------------------------------------------------------------ |
-| `image`         | string |                  | Sets the image to use for running BuildKit.                                                |
-| `network`       | string |                  | Sets the network mode for running the BuildKit container.                                  |
-| `cgroup-parent` | string | `/docker/buildx` | Sets the cgroup parent of the BuildKit container if Docker is using the `cgroupfs` driver. |
+| `image`         | String |                  | Sets the image to use for running BuildKit.                                                |
+| `network`       | String |                  | Sets the network mode for running the BuildKit container.                                  |
+| `cgroup-parent` | String | `/docker/buildx` | Sets the cgroup parent of the BuildKit container if Docker is using the `cgroupfs` driver. |
 
 ## Usage
 
-When you run a build, Buildx pulls the specified `image` (by default, [`moby/buildkit`](https://hub.docker.com/r/moby/buildkit))
+When you run a build, Buildx pulls the specified `image` (by default,
+[`moby/buildkit`](https://hub.docker.com/r/moby/buildkit))
 [Docker Hub](https://hub.docker.com/u/moby/buildkit). When the container has
 started, Buildx submits the build submitted to the containerized build server.
 
@@ -99,6 +103,3 @@ $ docker buildx build . \
 
 For more information on the Docker container driver, see the
 [buildx reference](https://docs.docker.com/engine/reference/commandline/buildx_create/#driver).
-
-If want to explore builders running on a remote server, see the
-[Kubernetes driver](./kubernetes.md) and the [Remote driver](./remote.md).
--- a/docs/guides/drivers/docker-container.md
+++ b/docs/guides/drivers/docker-container.md
@@ -1,40 +1,43 @@
 # Docker container driver
 
-The buildx docker-container driver allows creation of a managed and
-customizable BuildKit environment inside a dedicated Docker container.
+The buildx Docker container driver allows creation of a managed and customizable
+BuildKit environment in a dedicated Docker container.
 
-Using the docker-container driver has a couple of advantages over the basic
-docker driver. Firstly, we can manually override the version of buildkit to
-use, meaning that we can access the latest and greatest features as soon as
-they're released, instead of waiting to upgrade to a newer version of Docker.
-Additionally, we can access more complex features like multi-architecture
-builds and the more advanced cache exporters, which are currently unsupported
-in the default docker driver.
+Using the Docker container driver has a couple of advantages over the default
+Docker driver. For example:
 
-We can easily create a new builder that uses the docker-container driver:
+- Specify custom BuildKit versions to use.
+- Build multi-arch images, see [QEMU](#qemu)
+- Advanced options for
+  [cache import and export](https://docs.docker.com/build/building/cache/)
+
+## Synopsis
+
+Run the following command to create a new builder, named `container`, that uses
+the Docker container driver:
 
 ```console
 $ docker buildx create --name container --driver docker-container
 container
 ```
 
-We should then be able to see it on our list of available builders:
+The following table describes the available driver-specific options that you can
+pass to `--driver-opt`:
+
+| Parameter       | Value  | Default          | Description                                                                                |
+| --------------- | ------ | ---------------- | ------------------------------------------------------------------------------------------ |
+| `image`         | string |                  | Sets the image to use for running BuildKit.                                                |
+| `network`       | string |                  | Sets the network mode for running the BuildKit container.                                  |
+| `cgroup-parent` | string | `/docker/buildx` | Sets the cgroup parent of the BuildKit container if Docker is using the `cgroupfs` driver. |
+
+## Usage
+
+When you run a build, Buildx pulls the specified `moby/buildkit` image from
+[Docker Hub](https://hub.docker.com/u/moby/buildkit). When the container has
+started, Buildx submits the build submitted to the containerized build server.
 
 ```console
-$ docker buildx ls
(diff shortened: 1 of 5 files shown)


docs/guides/drivers/docker-container.md
```diff
@@ -1,40 +1,42 @@
...
-Additionally, we can access more complex features like multi-architecture
-builds and the more advanced cache exporters, which are currently unsupported
-in the default docker driver.
+Using the Docker container driver has a couple of advantages over the default
+Docker driver. For example:
 
-We can easily create a new builder that uses the docker-container driver:
+- Specify custom BuildKit versions to use.
+- Build multi-arch images
+- Advanced options for [cache import and export](../cache/index.md)
```
crazy-max, line 11:
```suggestion
- Advanced options for [cache import and export](index.md)
```

dvdksn: I wanted to link to the cache docs overview. Bad idea? :)

crazy-max: > I wanted to link to the cache docs overview.

Ah I thought your were linking to the same folder.

> Bad idea? :)

Yes atm we can't link to pages outside the current folder for remote pages. You have to use absolute links:

```suggestion
- Advanced options for [cache import and export](https://docs.docker.com/build/building/cache/)
```

docs/guides/drivers/docker-container.md
```diff
@@ -45,29 +48,57 @@ WARNING: No output specified with docker-container driver. Build result will onl
...
+## Loading to local image store
+
+Unlike when using the default `docker` driver, images built with the
+`docker-container` driver must be explicitly loaded into the local image store.
+Use the `--load` flag:
 
-```console
-$ docker buildx build --load -t <image> --builder=container .
+```bash
+$ docker buildx build . --load -t <image> --builder=container
```
AkihiroSuda, line 58: Why?

dvdksn: oops yeah I was not getting syntax highlighting in my editor with `console` so I used bash while editing, I will change it back...

dvdksn: Or maybe you meant why I changed the position of the build context arg :-) 

only to align with the cache docs where many examples are multiline and the dot at the end becomes awkward and sometimes missed

docs/guides/drivers/docker-container.md
```diff
@@ -1,40 +1,43 @@
...
+| `image`         | string |                  | Sets the image to use for running BuildKit.                                                |
+| `network`       | string |                  | Sets the network mode for running the BuildKit container.                                  |
+| `cgroup-parent` | string | `/docker/buildx` | Sets the cgroup parent of the BuildKit container if Docker is using the `cgroupfs` driver. |
 
-If we trigger a build, the appropriate `moby/buildkit` image will be pulled
-from [Docker Hub](https://hub.docker.com/u/moby/buildkit), the image started,
-and our build submitted to our containerized build server.
+## Usage
+
+When you run a build, Buildx pulls the specified `moby/buildkit` image from
```
jedevc, line 35:
```suggestion
When you run a build, Buildx pulls the specified `image` (by default, [`moby/buildkit`](https://hub.docker.com/r/moby/buildkit))
```
Oops my original writing wasn't super clear here :eyes:

docs/guides/drivers/docker-container.md
```diff
@@ -45,29 +48,57 @@ WARNING: No output specified with docker-container driver. Build result will onl
...
+> compression/decompression will likely take a large performance hit.
+
 ## Further reading
 
-For more information on the docker-container driver, see the [buildx reference](https://docs.docker.com/engine/reference/commandline/buildx_create/#driver).
+For more information on the Docker container driver, see the
+[buildx reference](https://docs.docker.com/engine/reference/commandline/buildx_create/#driver).
+
+If want to explore builders running on a remote server, see the
+[Kubernetes driver](./kubernetes.md) and the [Remote driver](./remote.md).
```
jedevc, line 104: Ahhh this isn't exactly true :see_no_evil: `docker-container` "can" do things on a remote server by using a different docker context. I can't find where we document this explicitly anywhere, but we should explain more of the interaction between `buildx` and `docker context` in the `docker.md` page in the future, and we could cross-link to there from here.

docs/guides/drivers/index.md
```diff
@@ -1,39 +1,78 @@
...
+prioritizes simplicity and ease of use. It has limited support for advanced
+features like caching and output formats, and isn't configurable. Other drivers
+provide more flexibility and are better at handling advanced scenarios. The
+`kubernetes` and `remote` drivers specifically aim to enable remote builders.
+
+The following table outlines some of the differences between drivers.
+
+| Feature                      |  `docker`   | `docker-container` | `kubernetes` |      `remote`      |
+| :--------------------------- | :---------: | :----------------: | :----------: | :----------------: |
+| **Automatically load image** |     Yes     |         No         |      No      |         No         |
```
jedevc, line 22: I see the emojis are gone :eyes:

Any particular reason, and is there any chance we could have both the words and the colored emoji? I find the colors easier to quickly scan visually, but if there's a good reason we can definitely leave them out :smile:

docs/guides/drivers/index.md
```diff
@@ -1,39 +1,78 @@
...
+| **Cache export**             | Inline only |        Yes         |     Yes      |        Yes         |
+| **Remote builders**          |     No      |         No         |     Yes      |        Yes         |
+| **Tarball output**           |     No      |        Yes         |     Yes      |        Yes         |
+| **Multi-arch images**        |     No      |        Yes         |     Yes      |        Yes         |
+| **BuildKit configuration**   |     No      |        Yes         |     Yes      | Managed externally |
+
+## List available drivers
```
jedevc, line 29:
```suggestion
## List available builders
```

docs/guides/drivers/kubernetes.md
```diff
@@ -107,28 +83,54 @@ kube0-6977cdcb75-rkc6b   1/1     Running   0          8s
...
+options. You must wrap both `nodeSelector` and `tolerations` in double quotes.
+For example:
+
+```console
+$ docker buildx create \
+  --bootstrap \
+  --name=kube \
+  --driver=kubernetes \
+  --driver-opt="nodeselector=label=value","tolerations=key=key1,value=value1"
+```
```
jedevc, line 120: Hm this isn't right.

The final line should be `--driver-opt='"nodeselector=label=value","tolerations=key=key1,value=value1"'` (single quotes around the option value), or even `'--driver-opt="nodeselector=label=value","tolerations=key=key1,value=value1"'` (single quotes around the whole option).

Otherwise, the double quotes will actually get removed by `/bin/sh`, so buildkit sees the same values as before. We need `/bin/sh` to remove the single quotes, so buildkit actually gets the double quoted values correctly (yes quoting rules are horrible :tada:).

Phrasing this is hard as well - these options don't have explicitly different syntax from the others, it's that because they may contain commas, these would otherwise get parsed incorrectly. The CSV parser lets us explicitly group a field that may contain commas by using double quotes around the entire field.

I wonder if we might want to rework the code for some of the details of this syntax to cover the edge cases here, and at least be a little more permissive with how we do the parsing (making sure we don't silently drop ignored csv options would also be a good step to help error correct here).

docs/guides/drivers/remote.md
```diff
@@ -14,157 +14,173 @@ $ docker buildx create \
...
-only creates a ClusterIP service). To configure the builder to be accessible
-remotely, you can use an appropriately configured Ingress, which is outside the
-scope of this guide.
-
-To access the service remotely, we can use the port forwarding mechanism in
-kubectl:
+Remember that you need to use the `--load` flag if you want to load the build
+result into the Docker daemon.
+
+## Guide: Remote BuildKit in Docker container
```
jedevc, line 84: Not so sure about calling these "Guides" since these will actually appear in the "Manuals" section on https://docs.docker.com :thinking: We could *potentially* split these out into their own files and have them live in the "Guides" section of the docs, but they feel a little short-form and light on details for that kind of setup.

Maybe "Example" could work better? Not sure.

docs/guides/drivers/index.md
```diff
@@ -1,39 +1,78 @@
...
-To create a new builder that uses one of the above drivers, you can use the
-[`docker buildx create`](https://docs.docker.com/engine/reference/commandline/buildx_create/) command:
+Buildx drivers are configurations for how and where the BuildKit backend runs.
+Driver settings are customizable and allows fine-grained control of the builder.
+Buildx supports the following drivers:
+
+- `docker`: uses the BuildKit library bundled into the Docker daemon.
+- `docker-container`: creates a dedicated BuildKit container using Docker.
+- `kubernetes`: creates BuildKit pods in a Kubernetes cluster.
+- `remote`: connects directly to a manually managed BuildKit daemon.
```
jedevc, line 10: As a follow-up, we should also update the cache parameters?

docs/guides/drivers/index.md
dvdksn: I removed them because I thought they looked out of place on the [docs site][link] (font and color issue). But, we haven't taken a stance on the use of emojis in the style guide. Let me bring up the topic with the docs team in our next huddle.

[link]: https://docs.docker.com/build/building/drivers/

dvdksn: I will add them back in the meantime until we have a verdict 🤓

jedevc (approved):

docs/guides/drivers/docker-container.md
crazy-max: Having context set there is not common and does not reflect the usage:

```
```

This is what we use atm across all our documentation and examples. Would prefer to keep it as it is.

dvdksn:
PS, I don't have merge permissions on this repo so feel free to merge it someone :)

Issue #1342, opened 2022-10-04
`panic: assignment to entry in nil map` where using `BUILDX_GIT_LABELS=full` with bake


dgageot:
Here's my setup:

On [master branch](https://github.com/docker/buildx/commit/105c214d15802358e2cd02622a20c5803aa1ec5e):
 + run `make` to enter shell mode.
 + `export BUILDX_GIT_LABELS=full`
 + `cd compose`
 + `git init`
 + `git -c user.name="John" -c user.email="jo@email.org" commit --allow-empty -m "wip"`
 + `docker buildx bake`

Here's the error:

```
[+] Building 0.0s (0/0)
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/docker/buildx/build.BuildWithResultHandler({0x1d81000, 0x40000b5580}, {0x4000302150, 0x1, 0x1}, 0x40000bb170, {0x1d62780, 0x40002e04c0}, {0x40001f7b18, 0x14}, ...)
	/src/build/build.go:804 +0x15b8
github.com/docker/buildx/build.Build(...)
	/src/build/build.go:741
github.com/docker/buildx/commands.runBake({0x1d91b98, 0x40004a62a0}, {0x400045d6e0, 0x0, 0x2}, {{0x400028b370, 0x1, 0x1}, {0x0, 0x0, ...}, ...})
	/src/commands/bake.go:148 +0xaac
github.com/docker/buildx/commands.bakeCmd.func1(0x400007ec80?, {0x400045d6e0, 0x0, 0x2})
	/src/commands/bake.go:182 +0x198
github.com/spf13/cobra.(*Command).execute(0x400007ec80, {0x40000b5960, 0x2, 0x2})
	/src/vendor/github.com/spf13/cobra/command.go:872 +0x4e8
github.com/spf13/cobra.(*Command).ExecuteC(0x400057a500)
	/src/vendor/github.com/spf13/cobra/command.go:990 +0x360
github.com/spf13/cobra.(*Command).Execute(...)
	/src/vendor/github.com/spf13/cobra/command.go:918
github.com/docker/cli/cli-plugins/plugin.RunPlugin(0x1a1dd00?, 0x400007e000, {{0x1a1bf6c, 0x5}, {0x1a261b7, 0xb}, {0x1d61610, 0x13}, {0x0, 0x0}, ...})
	/src/vendor/github.com/docker/cli/cli-plugins/plugin/plugin.go:51 +0x154
[...]

PR #1343, opened 2022-10-04
[1342] Fix assignment to nil map

dgageot:
Make sure the labels map is not empty before setting values

Fixes https://github.com/docker/buildx/issues/1342

Signed-off-by: David Gageot <david.gageot@docker.com>


crazy-max (approved):
LGTM thanks

PR #1129, opened 2022-05-14
check context builder endpoint

crazy-max:
while making some tests with @tonistiigi about an issue with `rm --all-inactive` that doesn't effectively remove some inactive nodes we saw that there is an underlying issue with docker context being detected as a docker-container builder:

```
$ docker context create --docker host=ssh://mini remote-buildkit
remote-buildkit

$ docker context ls
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                               KUBERNETES ENDPOINT   ORCHESTRATOR
default *           moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock                                         swarm
desktop-linux       moby                                                          npipe:////./pipe/dockerDesktopLinuxEngine
desktop-windows     moby                                                          npipe:////./pipe/dockerDesktopWindowsEngine
remote-buildkit     moby                                                          ssh://mini

$ docker buildx ls
NAME/NODE         DRIVER/ENDPOINT                        STATUS                 BUILDKIT PLATFORMS
builder           docker-container
  builder0        unix:///var/run/docker.sock            running                394ccf8  linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
  mac-mini-m1     tcp://mac-mini-m1.home.foobar.com:2376 running                394ccf8  linux/arm64*, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
  sifive          tcp://sifive.home.foobar.com:2375      running                394ccf8  linux/riscv64*
builder2 *        docker-container
  builder20       unix:///var/run/docker.sock            running                394ccf8  linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
fervent_raman     docker-container
  fervent_raman0  unix:///var/run/docker.sock            running                v0.10.3  linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
desktop-linux                                            protocol not available
[...]

PR #1345, opened 2022-10-04
bake: do not fail printing definition if instance unavailable

### LINKED ISSUE #1344: buildx bake with --print now requires a daemon
I'm using buildx bake in a docker image without a daemon in a CI process for preparing a docker-bake.json for future build.
In the image, I installed docker buildx with : 

```
# usefull for doing a "docker-buildx bake --print" which does not require docker deamon to run
COPY --from=docker/buildx-bin /buildx /usr/local/sbin/docker-buildx
```

This does not work anymore : 
```
ERROR: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

crazy-max:
fixes #1344 

Since 0.9 we are checking the context builder endpoint to make sure it's reachable before returning the instance: https://github.com/docker/buildx/pull/1129#issuecomment-1195542481. By doing so we are not soft-failing as before when loading `driversForNodeGroup` and therefore `bake --print` now fails if no instance is available.

This PR will not return an error if we are printing local bake files. An instance is still required when building or loading remote bake files.

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -86,13 +86,17 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions) (err error
 		}
 	}()
 
-	dis, err := getInstanceOrDefault(ctx, dockerCli, in.builder, contextPathHash)
-	if err != nil {
-		return err
-	}
-
+	var dis []build.DriverInfo
 	var files []bake.File
 	var inp *bake.Input
+
+	// instance only needed for reading remote bake files or building
+	if url != "" || !in.printOnly {
+		dis, err = getInstanceOrDefault(ctx, dockerCli, in.builder, contextPathHash)
+		if err != nil {
+			return err
+		}
+	}
 
 	if url != "" {
 		files, inp, err = bake.ReadRemoteFiles(ctx, dis, url, in.files, printer)


crazy-max:
@tonistiigi With this repro https://github.com/docker/buildx/issues/1344#issuecomment-1266628617 I found out that completing the trace span for the bake command on return timed out (10s). Looks to be this defer: https://github.com/docker/buildx/blob/105c214d15802358e2cd02622a20c5803aa1ec5e/commands/bake.go#L35-L37 calling: https://github.com/docker/buildx/blob/105c214d15802358e2cd02622a20c5803aa1ec5e/util/tracing/trace.go#L26

commands/bake.go
```diff
@@ -87,7 +87,7 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions) (err error
 	}()
 
 	dis, err := getInstanceOrDefault(ctx, dockerCli, in.builder, contextPathHash)
-	if err != nil {
+	if err != nil && (url != "" || !in.printOnly) {
```
thaJeztah, line 90: IIUC, in this case;

- we don't need the instance
- but we're still calling`getInstanceOrDefault()` (but ignore if it fails)

Could we make the condition so that we skip capping `getInstanceOrDefault()` if `url == ""` and `in.printOnly`? My concern would be that calling `getInstanceOrDefault()` (if the instance is not accessible) would result in a delay (buildx trying to access it, which may result in a "timeout" after XX seconds).

thaJeztah (changes requested):
left a question 😅 

Also would be good to have an integration test for this (setting a "bogus" / unreachable instance, then verify it works)

commands/bake.go
crazy-max: > would result in a delay (buildx trying to access it, which may result in a "timeout" after XX seconds).

It might if context is set to a remote docker endpoint but not for a socket one.

> * but we're still calling`getInstanceOrDefault()` (but ignore if it fails)

Yes that's the same behavior in 0.8.2 when default instance falls back to a docker-container one if not available (https://github.com/docker/buildx/pull/1129#issuecomment-1195542481) and therefore silently fails but still prints with bake.

But agree that we could just skip returning an instance if not required.

crazy-max:
> Also would be good to have an integration test for this (setting a "bogus" / unreachable instance, then verify it works)

That's a broader subject but yeah we need integration tests for the cli and mock some components like the docker/cli.

commands/bake.go
crazy-max: @thaJeztah Should be good

thaJeztah (approved):
thanks for updating

LGTM

commands/bake.go
```diff
@@ -86,14 +86,18 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions) (err error
...
-	if err != nil {
-		return err
-	}
-
+	var dis []build.DriverInfo
 	var files []bake.File
 	var inp *bake.Input
 
+	// instance only needed for reading remote bake files or building
+	if url != "" || !in.printOnly {
```
jedevc, line 94: I wonder if a better way to write this would be to fetch the instance "lazily"? So that we only call `getInstanceOnDefault` to get `dis` at the point of use?

jedevc: Yeah sounds good :tada:

The logic around drivers keeps getting hairier and hairier, we should definitely look into some cleanup soon :eyes:

jedevc (approved):

commands/bake.go
crazy-max: I have another branch to mutualize the builder/driver logic so we could lazily load an instance. Not yet ready but you can take a look: https://github.com/docker/buildx/compare/master...crazy-max:buildx:builder-pkg

PR #1356, opened 2022-10-13
go.mod: matttproud/golang_protobuf_extensions v1.0.2 (use tag)

thaJeztah:
it's the same commit: https://github.com/matttproud/golang_protobuf_extensions/compare/v1.0.2...c182affec369


thaJeztah:
Not sure if it will automatically update in that case. The _actual_ culprit is in k8s.io/common, so I'm manually updating the versions to get them out of the dependency tree (in case go mod will prefer "untagged" over "tagged" for the same commit)

crazy-max (approved):

PR #1366, opened 2022-10-19
lint: remove structcheck

jedevc:
`structcheck` is abandoned, and has been replaced by the `unused` linter. See https://golangci-lint.run/usage/linters/#structcheck for more information.

> The owner seems to have abandoned the linter. Replaced by unused.

See similar in https://github.com/moby/buildkit/pull/3196

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -81,10 +81,7 @@ type commonOptions struct {
 	progress     string
 	pull         *bool
 
-	// golangci-lint#826
-	// nolint:structcheck
 	exportPush bool
-	// nolint:structcheck
 	exportLoad bool
 }
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,7 @@ linters:
     - revive
     - staticcheck
     - typecheck
+    - nolintlint
   disable-all: true
 
 linters-settings:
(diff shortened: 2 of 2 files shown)


AkihiroSuda (approved):

commands/build.go
```diff
@@ -81,10 +81,9 @@ type commonOptions struct {
 	progress     string
 	pull         *bool
 
-	// golangci-lint#826
```
thaJeztah, line 84: Looks to link to https://github.com/golangci/golangci-lint/issues/826 (about this linter not being good)

commands/build.go
```diff
@@ -81,10 +81,9 @@ type commonOptions struct {
...
-	// nolint:structcheck
+	//nolint:structcheck
 	exportPush bool
-	// nolint:structcheck
+	//nolint:structcheck
```
thaJeztah, line 86: These should no longer be needed if we disable the linter? Or is it still running in the default set? (if so, probably an update to 1.49 or 1.50 is needed); https://github.com/docker/buildx/blob/1bb375fe5ccad33244bcc3ede169e1cabcc7fcc1/hack/dockerfiles/lint.Dockerfile#L7

jedevc: Yikes, I screwed up the commit there, I could have *sworn* I removed that line :thinking:

Anyways, have properly removed them now.

thaJeztah (approved):
LGTM

PR #1368, opened 2022-10-21
Userns to driver

jedevc:
Follow-up to https://github.com/docker/buildx/pull/887 (see https://github.com/docker/buildx/pull/887#issuecomment-1285292603 for more context).

This resolves https://github.com/docker/buildx/discussions/1358, so that a docker-container builder can be created even in the absence of the socket - it shouldn't fail until it is started, either with `--bootstrap` or by attempting a build.

The issue that caused this flow to stop working was introduced in https://github.com/docker/buildx/pull/1206, which was released in [v0.9](https://github.com/docker/buildx/releases/tag/v0.9.0) - I wonder if it's worth cherry-picking?

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -36,7 +36,6 @@ const (
 type Driver struct {
 	driver.InitConfig
 	factory      driver.Factory
-	userNSRemap  bool // true if dockerd is running with userns-remap mode
 	netMode      string
 	image        string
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

AkihiroSuda:
(Eventually we should have CI to cover userns, but it can be another PR)

crazy-max (approved):

PR #1375, opened 2022-10-27
formatting, structure, style updates

dvdksn:
Restructured and rewrote to the best of my understanding :-)

- Consolidated similar exporters on a single page
- Rewrote parts of the index page


dvdksn:
I tried to create this pr towards `jedevc:exporter-docs` using the gh cli but I failed miserably, so I had to use the GUI in the end 💀

PR #1392, opened 2022-11-04
docs: refactored file and directory structure

dvdksn:
Signed-off-by: David Karlsson <david.karlsson@docker.com>

- Changed the directory structure in docs/ to match the structure used on docs.docker.com
- Updated links


dvdksn:
Hmm, this depends on docker/docs#16090 -- any way we could make the validator run with that ref instead of master?

crazy-max (approved):
LGTM with https://github.com/docker/docs/pull/16090

PR #1397, opened 2022-11-07
Correct documentation on gha scopes

markhildreth-gravity:
I believe the documentation is misleading in saying that specifying a scope in a gha cache declaration will `ensure each branch gets its own cache`. I believe this should say that each build or each image will get its own cache.


dvdksn:
LGTM, cc @jedevc

jedevc (approved):

PR #1403, opened 2022-11-10
Docs consistency fixes

jedevc:
:hammer_and_wrench: Fixes https://github.com/docker/setup-buildx-action/issues/179#issuecomment-1309977562
:hammer_and_wrench: Also ensures we use consistent parameter table headers with "Type" instead of "Value", to be inline with GitHub actions (e.g. [build-push-action](https://github.com/docker/build-push-action#inputs))

### DIFF
--- a/docs/manuals/exporters/image-registry.md
+++ b/docs/manuals/exporters/image-registry.md
@@ -19,7 +19,7 @@ $ docker buildx build --output type=registry[,parameters] .
 The following table describes the available parameters that you can pass to
 `--output` for `type=image`:
 
-| Parameter              | Value                                  | Default | Description                                                                                                                                                                                                                         |
+| Parameter              | Type                                   | Default | Description                                                                                                                                                                                                                         |
 | ---------------------- | -------------------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `name`                 | String                                 |         | Specify image name(s)                                                                                                                                                                                                               |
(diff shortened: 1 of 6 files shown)


crazy-max (approved):

docs/manuals/exporters/image-registry.md
```diff
@@ -19,7 +19,7 @@ $ docker buildx build --output type=registry[,parameters] .
...
 | ---------------------- | -------------------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `name`                 | String                                 |         | Specify image name(s)                                                                                                                                                                                                               |
 | `push`                 | `true`,`false`                         | `false` | Push after creating the image.                                                                                                                                                                                                      |
```
crazy-max, line 25: Maybe Type `Bool` for `push` would make more sense?

dvdksn (approved):

docs/manuals/exporters/image-registry.md
jedevc: WDYT @dvdksn? Our actions seem to use `Bool`, these docs use `true,false` - any preference? I think `true,false` is more explicit in the options that are available, while `Bool` is shorter and more succinct.

dvdksn: I was secretly hoping y'all had a strong preference for either because I don't 🙈 But IMO let's stick with `true`,`false` for now since it is more explicit.

PR #1406, opened 2022-11-14
Synchronise access to the map when printing.

felixdesouza:
If you happen to run:

    docker buildx imagetools inspect --format '{{. | json}}' imageRef

Where `imageRef` points to a manifest list with multiple platforms.

You can sometimes run into `concurrent map write` errors as there are multiple goroutines trying to write to the `imageInfo` and `buildInfo` maps.

NB: I'm not a Go developer, if there is a nicer way to do this, happy to make edits.

### DIFF
--- a/util/imagetools/printers.go
+++ b/util/imagetools/printers.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync"
 	"text/tabwriter"
 	"text/template"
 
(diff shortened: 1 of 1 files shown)


jedevc:
Hey, nice find! :tada: Thanks for the PR as well :heart:

Ideally we should change the map to an array, since then we can assign to it in parallel, using the index of the loop as the index into the resulting array (similar to how we do it in [exporter/containerimage/writer.go](https://github.com/moby/buildkit/blob/d3f26c63389c2ad66600ae5c884d6cbde627d6d8/exporter/containerimage/writer.go#L243-L260)). But the logic here seems a little fiddly, so I think we should just be able to have locks.

With that, a couple notes:
- We only need `sync.Mutex` (instead of `sync.RWMutex`)
- We don't need to take the address of with `&`
- Ideally, we could have two mutexes, one for each map that we write to

CC @crazy-max, this was added in https://github.com/docker/buildx/pull/972, should we cherry-pick? Probably to 0.9 at least?

felixdesouza:
Thanks for the review!

jedevc (approved):

jedevc:
@felixdesouza could you sign-off your second commit as well?

PR #1407, opened 2022-11-15
go.mod: golang.org/x/crypto v0.1.0 (Fix `ssh: parse error in message type 27` with OpenSSH >= 8.9)

AkihiroSuda:
ref:
- https://github.com/moby/buildkit/pull/3274


thaJeztah:
If we need this back ported to v0.9, I wonder if we should use `v0.1.0` as an intermediate step (as golang.org/x/crypto v0.1.0 looks to have the needed fix)

crazy-max (approved):

PR #1420, opened 2022-11-22
Support new `tar` option for oci+docker exporters

jedevc:
:arrow_up: Follow-up to https://github.com/moby/buildkit/pull/3162

https://github.com/moby/buildkit/pull/3162 requires changes to buildx, as it requires improved validation in the CLI to correctly determine whether it should pass a file or a directory to the buildkit server. Additionally, we need to update the buildkit vendored packages for the updated client-side validation.


crazy-max (changes requested):
Replace directive for moby needs to match buildkit one: https://github.com/docker/buildx/blob/60e2029e70ab8c1dad5b48b51efc27c5bd4042c4/go.mod#L171

```
replace github.com/docker/docker => github.com/docker/docker v20.10.3-0.20221006005007-99aa9bb766b5+incompatible // 22.06 branch (v22.06-dev)
```

crazy-max:
Get this one in, as follow-up we should have unit tests for this package: https://github.com/docker/buildx/blob/master/util/buildflags

PR #1430, opened 2022-11-28
Refactor builder and drivers info logic

crazy-max:
Moves builder and drivers info logic from `commands` pkg to a dedicated client pkg so we can easily interact with builders and drivers. Refactoring reduced latency for operations when loading drivers data or checking already validated endpoints:

```shell
# before
$ time docker buildx ls
...
real    0m5.396s
user    0m0.180s
sys     0m0.796s

# after
$ time docker buildx ls
real    0m3.460s
user    0m0.176s
sys     0m0.881s
```

Added some TODOs as there is still room for improvements.


builder/builder.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	if b.NodeGroup.Name == "default" && len(b.NodeGroup.Nodes) == 1 {
+		b.NodeGroup.Name = b.NodeGroup.Nodes[0].Endpoint
+	}
+
+	return b, nil
+}
+
+// Validate validates builder context
+func (b *Builder) Validate() error {
```
jedevc, line 98: IMO, this should be called as part of `New` if possible.

builder/driver.go
```diff
@@ -0,0 +1,221 @@
...
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+)
+
+type Driver struct {
+	Name        string
+	Driver      driver.Driver
```
jedevc, line 24: :thinking: this is confusing :cry: 

I think we should avoid having `driver.Driver` and `builder.Driver` - I guess that was the reason for `driverInfo` before. I think it's ok for now, but maybe we should revisit later.

jedevc (commented):
I like the refactor :tada: - left a couple comments.

I'm curious, I can't quite see where the performance improvement has come from :eyes:

builder/driver.go
```diff
@@ -0,0 +1,221 @@
...
+	ImageOpt    imagetools.Opt
+	ProxyConfig map[string]string
+	Version     string
+	Err         error
+}
+
+// LoadDrivers loads drivers for this builder.
+func (b *Builder) LoadDrivers(ctx context.Context, withData bool) error {
+	eg, _ := errgroup.WithContext(ctx)
+	b.Drivers = make([]Driver, len(b.NodeGroup.Nodes))
```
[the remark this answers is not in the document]
crazy-max: Yes first option looks good

builder/builder.go
crazy-max: Yes in my initial attempt it was a client opt but then I thought it could be handy if we want this client to also create/update a builder. Will change that for now.

crazy-max:
> I'm curious, I can't quite see where the performance improvement has come from 👀

The builder client is lazy and will load data, resolve endpoints when needed. Need more work as I miss some cases.

builder/driver.go
crazy-max: Yes, ideally it should be directly carried by the factory.

builder/builder.go
jedevc: Yeah, we can keep it as a separate method as well, I just think we should call it in `New` by default.

builder/driver.go
crazy-max: Yeah sounds better, changes to `Node`.

build/build.go
```diff
@@ -109,28 +110,15 @@ type NamedContext struct {
...
-	ProxyConfig map[string]string
-}
-
-type DockerAPI interface {
-	DockerAPI(name string) (dockerclient.APIClient, error)
-}
-
-func filterAvailableDrivers(drivers []DriverInfo) ([]DriverInfo, error) {
-	out := make([]DriverInfo, 0, len(drivers))
+func filterAvailableDrivers(nodes []builder.Node) ([]builder.Node, error) {
```
jedevc, line 113:
```suggestion
func filterAvailableNodes(nodes []builder.Node) ([]builder.Node, error) {
```

jedevc (approved):
LGTM, just a minor function rename :)

builder/node.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	imageopt, err := b.ImageOpt()
+	if err != nil {
+		return nil, err
+	}
+
+	for i, n := range b.NodeGroup.Nodes {
+		func(i int, n store.Node) {
+			eg.Go(func() error {
```
errordeveloper, line 61: The code would be easier to read if at least one of these annonymous function was tunred into a mehtod.

crazy-max: Yes agree, I put a comment on this method to we can track this as follow-up: https://github.com/docker/buildx/pull/1430/files#diff-7c6ac5bb2a9c2322f716d7b66113042e64298fbdc5cba600abed692a0c2a6570R38

errordeveloper (approved):
Thanks a lot, this is a significant improvement! I might follow up with a few more things, as I'm trying to use some of the same functions in another project.

commands/ls.go
```diff
@@ -32,52 +31,24 @@ func runLs(dockerCli command.Cli, in lsOptions) error {
...
+	eg, _ := errgroup.WithContext(timeoutCtx)
 	for _, b := range builders {
-		func(b *nginfo) {
+		func(b *builder.Builder) {
 			eg.Go(func() error {
-				err = loadNodeGroupData(ctx, dockerCli, b)
-				if b.err == nil && err != nil {
-					b.err = err
-				}
+				_, _ = b.LoadNodes(timeoutCtx, true)
```
tonistiigi, line 51: Why is this ignoring the error instead of setting it to the builder?

Issue #1346, opened 2022-10-05
bakefile: mark variables as required


tnaroska:
### Current
There seems to be no way in current hcl bakefile syntax to mark a variable as required. There is a way to specify defaults:
```
# docker-bake.hcl
variable "TAG" {
  default = "latest"
}
```
Or without explicit default the variable defaults to empty:
```
# docker-bake.hcl
variable "TAG" {
}
```

### Proposed Enhancement
I'm trying to define a variable that is required to be set by the caller (i.e. through environment variable). Something like:
```
# docker-bake.hcl
variable "TAG" {
  required = true
}
```
Where the user needs to specify the tag when executing `buildx bake` through either env-var or by passing multiple bake files (where the variable needs to be set to a value in at least one of the bake files).

**Expectation:**
```
# docker-bake.hcl
variable "TAG" {
  required = true
}


$ buildx bake              # -> error: required variable "TAG" not defined
$ TAG=latest buildx bake   # -> success
```

crazy-max:
Having something as suggested in https://github.com/docker/buildx/pull/491#issue-773693934 with type constraints and validation like [terraform](https://www.terraform.io/docs/configuration/variables.html#declaring-an-input-variable) does would be more aligned I think:

```hcl
variable "slugs" {
  type = list(string)
  default = [
    "crazymax/diun",
    "ghcr.io/crazy-max/diun"
  ]
  validation {
    condition = length(var.slugs) > 0
    error_message = "At least one slug is required."
  }
}
```

tnaroska:
💯 that would be a nicer, more general approach!

PR #1434, opened 2022-11-29
Resource interpolation support

jedevc:
:hammer_and_wrench: Fixes https://github.com/docker/buildx/issues/445.

This took a while :smile: To do this, I needed to substantially rework how we parse HCL, and hack around some of the HCL library limitations that make it more complex to perform the partial block evaluations that we need to be able to have blocks that reference themselves as in the example from the feature request:

```hcl
target "front" {
  context = target.front.name
  tags    = [target.front.context]
}
```

The core of the PR reworks block resolution to be lazy, similar to how we evaluate global attributes today. Then we also need to ensure that we can perform partial evaluation correctly.

Some of the error messages aren't *quite* right - it seems that `hcl.Diagnostics` don't have a wrapping behavior, so we need to detect these properly to provide good error messages in the case of typos, etc.


crazy-max:
Looks great! Was thinking we could have a common interface for `Blocks`, `Attributes` and maybe `Validators` if we implement this in the future: https://github.com/docker/buildx/issues/1346#issuecomment-1268314923

tonistiigi (commented):
One thing I discovered was not handling default values. Eg. you can't do `target.foo.dockerfile` while if you do `bake foo --print` then the dockerfile value is filled in with default value.

tonistiigi (approved):

PR #1439, opened 2022-11-30
store: set nodegroup last activity

crazy-max:
adds last activity property that will be updated on build or when a node is updated.

```console
$ docker buildx inspect
Name:          builder2
Last Activity: 2022-11-30 13:56:51 +0100 CET

Nodes:
Name:           builder20
Endpoint:       unix:///var/run/docker.sock
Driver Options: env.BUILDKIT_STEP_LOG_MAX_SPEED="10485760" env.JAEGER_TRACE="localhost:6831" image="moby/buildkit:master" network="host" env.BUILDKIT_STEP_LOG_MAX_SIZE="10485760"
Status:         running
Flags:          --debug --allow-insecure-entitlement security.insecure --allow-insecure-entitlement network.host
Buildkit:       7615120
Platforms:      linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
```

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"time"
 
 	"github.com/docker/docker/pkg/ioutils"
 	"github.com/gofrs/flock"
@@ -15,6 +16,7 @@ import (
 const (
 	instanceDir = "instances"
 	defaultsDir = "defaults"
+	activityDir = "activity"
 )
 
 func New(root string) (*Store, error) {
@@ -22,6 +24,9 @@ func New(root string) (*Store, error) {
 		return nil, err
 	}
 	if err := os.MkdirAll(filepath.Join(root, defaultsDir), 0700); err != nil {
+		return nil, err
+	}
+	if err := os.MkdirAll(filepath.Join(root, activityDir), 0700); err != nil {
 		return nil, err
 	}
 	return &Store{root: root}, nil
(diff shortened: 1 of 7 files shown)


store/store.go
```diff
@@ -121,15 +138,42 @@ func (t *Txn) SetCurrent(key, name string, global, def bool) error {
...
 		}
 	} else {
-		os.RemoveAll(filepath.Join(t.s.root, "defaults", h)) // ignore error
+		os.RemoveAll(filepath.Join(t.s.root, defaultsDir, h)) // ignore error
 	}
 	return nil
 }
 
+func (t *Txn) UpdateLastActivity(ng *NodeGroup) error {
+	return ioutils.AtomicWriteFile(filepath.Join(t.s.root, activityDir, ng.Name), []byte(fmt.Sprintf("%d", time.Now().Unix())), 0600)
```
jedevc, line 151: Any reason for using the Unix timestamp in the file? Could we instead use `time.RFC3339` as the format?

Would prefer something that's vaguely human-readable if possible, unless we have precedent for a different format?

store/store.go
```diff
@@ -89,19 +104,21 @@ func (t *Txn) Save(ng *NodeGroup) error {
...
+	return ioutils.AtomicWriteFile(filepath.Join(t.s.root, instanceDir, name), dt, 0600)
 }
 
 func (t *Txn) Remove(name string) error {
 	name, err := ValidateName(name)
 	if err != nil {
 		return err
 	}
-	return os.RemoveAll(filepath.Join(t.s.root, "instances", name))
+	defer t.RemoveLastActivity(name)
```
jedevc, line 120: We should probably warn/error if this fails?

store/store.go
```diff
@@ -89,19 +104,21 @@ func (t *Txn) Save(ng *NodeGroup) error {
 	if err != nil {
 		return err
 	}
+	defer t.UpdateLastActivity(ng)
```
jedevc, line 107: We should probably error if this fails, given we do for the builds?

jedevc (approved):

PR #1440, opened 2022-11-30
builder: add worker gc policies and labels

crazy-max:
~needs #1430~

Adds builder GC Policies with worker labels. Will be displayed when using `inspect` command:

```console
$ docker buildx inspect
Name:          builder2
Last Activity: 2022-11-30 12:42:47 +0100 CET

Nodes:
Name:           builder20
Endpoint:       unix:///var/run/docker.sock
Driver Options: env.BUILDKIT_STEP_LOG_MAX_SPEED="10485760" env.JAEGER_TRACE="localhost:6831" image="moby/buildkit:latest" network="host" env.BUILDKIT_STEP_LOG_MAX_SIZE="10485760"
Status:         running
Flags:          --debug --allow-insecure-entitlement security.insecure --allow-insecure-entitlement network.host
BuildKit:       v0.10.6
Platforms:      linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
Labels:
 org.mobyproject.buildkit.worker.executor:         oci
 org.mobyproject.buildkit.worker.hostname:         docker-desktop
 org.mobyproject.buildkit.worker.network:          host
 org.mobyproject.buildkit.worker.oci.process-mode: sandbox
 org.mobyproject.buildkit.worker.selinux.enabled:  false
 org.mobyproject.buildkit.worker.snapshotter:      overlayfs
GC Policy rule#0:
 All:           false
 Filters:       type==source.local,type==exec.cachemount,type==source.git.checkout
 Keep Duration: 48h0m0s
 Keep Bytes:    488.3MiB
GC Policy rule#1:
 All:           false
 Keep Duration: 1440h0m0s
 Keep Bytes:    24.21GiB
GC Policy rule#2:
 All:        false
 Keep Bytes: 24.21GiB
GC Policy rule#3:
 All:        true
 Keep Bytes: 24.21GiB
```

### DIFF
--- a/builder/node.go
+++ b/builder/node.go
@@ -10,6 +10,7 @@ import (
 	"github.com/docker/buildx/util/dockerutil"
 	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/buildx/util/platformutil"
+	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/util/grpcerrors"
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -23,6 +24,8 @@ type Node struct {
 	Driver      driver.Driver
 	DriverInfo  *driver.Info
 	Platforms   []ocispecs.Platform
+	GCPolicy    []client.PruneInfo
+	Labels      map[string]string
 	ImageOpt    imagetools.Opt
 	ProxyConfig map[string]string
 	Version     string
@@ -182,8 +185,12 @@ func (n *Node) loadData(ctx context.Context) error {
 		if err != nil {
 			return errors.Wrap(err, "listing workers")
 		}
-		for _, w := range workers {
+		for idx, w := range workers {
 			n.Platforms = append(n.Platforms, w.Platforms...)
+			if idx == 0 {
+				n.GCPolicy = w.GCPolicy
+				n.Labels = w.Labels
+			}
 		}
 		n.Platforms = platformutil.Dedupe(n.Platforms)
 		inf, err := driverClient.Info(ctx)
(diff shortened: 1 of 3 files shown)


builder/driver.go
```diff
@@ -0,0 +1,233 @@
...
+		workers, err := driverClient.ListWorkers(ctx)
+		if err != nil {
+			return errors.Wrap(err, "listing workers")
+		}
+		for idx, w := range workers {
+			d.Platforms = append(d.Platforms, w.Platforms...)
+			if idx == 0 {
+				d.GCPolicy = w.GCPolicy
+				d.Labels = w.Labels
+			}
```
crazy-max, line 217: Atm it returns GC Policies and worker labels only for the default worker.

jedevc:
Hm, this does make the output a bit noisier, and the policies are slightly tricky to mentally parse. I wonder if the GC policies should be hidden behind a verbose flag? Not sure about this, it's definitely useful information, but probably not to the average user.

I think the labels should be good without that, they're not too long.

crazy-max:
> Hm, this does make the output a bit noisier, and the policies are slightly tricky to mentally parse.

Yeah that's noisy but still relevant and for parsing we should do smth like https://github.com/docker/buildx/pull/830#issuecomment-1348729303

I will move `Platforms` before `Labels`.

jedevc:
LGTM, but we should make sure to get the follow-up to tidy the output and parsing into the same release as this.

jedevc (approved):

crazy-max:
Yes and also another follow-up for our actions toolkit: https://github.com/docker/actions-toolkit/blob/8cb1698b203083ee1c3996cbd2a0d197cae79d9f/src/buildx/builder.ts#L57

jedevc:
@crazy-max @tonistiigi, anything blocking this?

PR #1444, opened 2022-12-02
imagetools inspect: handle provenance and sboms

crazy-max:
Supports provenance and sbom for `imagetools inspect` command.

```console
$ docker buildx imagetools inspect crazymax/buildkit:attest --format "{{json .}}"
```

https://gist.github.com/crazy-max/5bbc2e60c58263a0162bbf3b920d3d09

### DIFF
--- a/util/imagetools/loader.go
+++ b/util/imagetools/loader.go
@@ -1,0 +1,357 @@
+package imagetools
+
+// TODO: replace with go-imageinspect library when public
+
+import (
+	"context"
+	"encoding/json"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/platforms"
+	"github.com/containerd/containerd/remotes"
+	"github.com/docker/distribution/reference"
+	"github.com/moby/buildkit/util/contentutil"
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	annotationReference = "vnd.docker.reference.digest"
+)
+
+type contentCache interface {
+	content.Provider
+	content.Ingester
+}
+
+type loader struct {
+	resolver remotes.Resolver
+	cache    contentCache
+}
+
+type manifest struct {
+	desc     ocispec.Descriptor
+	manifest ocispec.Manifest
+}
+
+type index struct {
+	desc  ocispec.Descriptor
+	index ocispec.Index
+}
+
+type asset struct {
+	config *ocispec.Image
+	sbom   *sbomStub
+	slsa   *slsaStub
+}
+
+type result struct {
+	mu        sync.Mutex
+	indexes   map[digest.Digest]index
+	manifests map[digest.Digest]manifest
+	images    map[string]digest.Digest
+	refs      map[digest.Digest][]digest.Digest
+
+	platforms []string
+	assets    map[string]asset
+}
+
+func newLoader(resolver remotes.Resolver) *loader {
+	return &loader{
+		resolver: resolver,
+		cache:    contentutil.NewBuffer(),
+	}
+}
+
+func (l *loader) Load(ctx context.Context, ref string) (*result, error) {
+	named, err := parseRef(ref)
+	if err != nil {
+		return nil, err
+	}
+
+	_, desc, err := l.resolver.Resolve(ctx, named.String())
+	if err != nil {
+		return nil, err
+	}
+
+	canonical, err := reference.WithDigest(named, desc.Digest)
+	if err != nil {
+		return nil, err
+	}
+
+	fetcher, err := l.resolver.Fetcher(ctx, canonical.String())
+	if err != nil {
+		return nil, err
+	}
+
+	r := &result{
+		indexes:   make(map[digest.Digest]index),
+		manifests: make(map[digest.Digest]manifest),
+		images:    make(map[string]digest.Digest),
+		refs:      make(map[digest.Digest][]digest.Digest),
+		assets:    make(map[string]asset),
+	}
+
+	if err := l.fetch(ctx, fetcher, desc, r); err != nil {
+		return nil, err
+	}
+
+	for platform, dgst := range r.images {
+		r.platforms = append(r.platforms, platform)
+
+		mfst, ok := r.manifests[dgst]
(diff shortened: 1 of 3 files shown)


util/imagetools/loader.go
```diff
@@ -0,0 +1,415 @@
...
+	desc     ocispec.Descriptor
+	manifest ocispec.Manifest
+}
+
+type index struct {
+	desc  ocispec.Descriptor
+	index ocispec.Index
+}
+
+type mat struct {
```
jedevc, line 50: `mat`?

util/imagetools/loader.go
```diff
@@ -0,0 +1,415 @@
...
+		res[p] = m.provenance
+	}
+	return res
+}
+
+func (r *result) SBOMs() map[string]json.RawMessage {
+	if len(r.mats) == 0 {
+		return nil
+	}
+	res := make(map[string]json.RawMessage)
```
jedevc, line 407: Hm, I wonder if we could do a best-effort parse and process this as a `map[string]interface{}`? That way we could do `{{ .SBOM.Name }}` or similar to access individual fields on the SBOM, while not tying the client directly to a single struct definition.

util/imagetools/printers.go
```diff
@@ -31,53 +27,46 @@ type Printer struct {
...
 
-	dt, manifest, err := resolver.Get(ctx, name)
+	ref, err := parseRef(name)
 	if err != nil {
 		return nil, err
 	}
 
-	var index ocispecs.Index
-	if err = json.Unmarshal(dt, &index); err != nil {
+	dt, mfst, err := resolver.Get(ctx, ref.String())
```
jedevc, line 50: As I read this, are we fetching the image index/manifest here as well as in `insres`? Not quite sure I understand what `insres` is for, the name doesn't indicate a lot.

util/imagetools/loader.go
```diff
@@ -0,0 +1,415 @@
...
+
+func (l *loader) scanSBOM(ctx context.Context, fetcher remotes.Fetcher, r *result, refs []digest.Digest, mat *mat) error {
+	ctx = remotes.WithMediaTypeKeyPrefix(ctx, "application/vnd.in-toto+json", "intoto")
+	for _, dgst := range refs {
+		mfst, ok := r.manifests[dgst]
+		if !ok {
+			return errors.Errorf("referenced image %s not found", dgst)
+		}
+		for _, layer := range mfst.manifest.Layers {
+			if layer.MediaType == "application/vnd.in-toto+json" && layer.Annotations["in-toto.io/predicate-type"] == "https://spdx.dev/Document" {
```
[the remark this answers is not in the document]
crazy-max: See https://github.com/docker/buildx/pull/1444#discussion_r1047416950

util/imagetools/loader.go
```diff
@@ -0,0 +1,415 @@
...
+		}
+
+		r.mats[platform] = m
+	}
+
+	sort.Strings(r.platforms)
+	return r, nil
+}
+
+func (l *loader) fetch(ctx context.Context, fetcher remotes.Fetcher, desc ocispec.Descriptor, r *result) error {
```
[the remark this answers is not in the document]
crazy-max: See https://github.com/docker/buildx/pull/1444#discussion_r1047416950

jedevc:
Is there a way we can support multiple SBOMs? e.g. in the case of a dockerfile where multiple stages have been scanned, we should have a way to get multiple.

Maybe `.SBOM` for the "first" SBOM, and `.SBOMS[idx]` for others?

Side note: the issue with ordering here is that the "most relevant" SBOM is currently not guaranteed to be first (though I think I have a solution for fixing that buildkit-side).

jedevc:
Aside from the above, LGTM :tada:

util/imagetools/loader.go
```diff
@@ -0,0 +1,407 @@
...
+					return err
+				}
+				as.sbom = sbom
+			}
+		}
+	}
+	return nil
+}
+
+type provenance struct { // TODO: this is only a stub, to be refactored later
```
tonistiigi, line 287: Could we just show the raw SLSA in here. Just `json.RawMessage`.

tonistiigi (commented):
The SBOM part seems ok but I don't think we should invent another temporary provenance type atm.

tonistiigi:
If we possibly want to add our own typed structs for SBOM/Provenance in the future then maybe it is better if we leave the names of the raw structs as `SPDX` and `SLSA` atm.

tonistiigi:
Did some testing with https://github.com/docker/buildx/compare/master...tonistiigi:buildx:inspect-provenance?expand=1 branch. Mostly seems to work with raw values but commands are really slow(when blobs are being pulled I believe). Maybe this can be optimized or should show debug logs. Especially because some of the SBOM values can be really big they should only be pulled when user really asks for the data.

tonistiigi:
> If we possibly want to add our own typed structs for SBOM/Provenance in the future then maybe it is better if we leave the names of the raw structs as SPDX and SLSA atm.

wdyt @crazy-max @jedevc

crazy-max:
> If we possibly want to add our own typed structs for SBOM/Provenance in the future then maybe it is better if we leave the names of the raw structs as SPDX and SLSA atm.

Not sure about the SLSA terminology here. It can be [VSA](https://slsa.dev/verification_summary/v0.2) or [Provenance](https://slsa.dev/provenance/v0.2) atm os I think Provenance still makes sense. Maybe SLSAProvenance?

jedevc:
The combination seems odd to me - maybe we could go with `{{ .SBOM.SPDX }}` and `{{ .Provenance.SLSA }}`? Those can be the raw data structures, leaving us open in the future to additional formats such as `{{ .SBOM.CycloneDX }}`, as well as utility fields for our own typed structs in the future, e.g. `{{ .SBOM.Packages[0].Name }}`.

tonistiigi:
>  maybe we could go with {{ .SBOM.SPDX }} and {{ .Provenance.SLSA }}?

I'm not sure if it is possible to have `{{ .SBOM.SPDX }}` be the spdx and `{{ .SBOM }}` itself a clear package list later without all the spdx garbage. If this is possible then lgtm. Otherwise `SBOMSPDX` seems fine to me as well. Although that would also mess up `{{ json . }}`. I think what we really need is just a custom format handling for these special keys (eg like atm there some commands have `--format table`). Content is only loaded when the format is exactly the key. In `{{ json . }}` there would be no SBOM fields.

tonistiigi (approved):

tonistiigi:
I merged this in but if we still want to make changes to naming we still can before the GA.

Issue #1362, opened 2022-10-18
Bake segfault when setting variable defaulting to null from env


skycaptain:
When setting a variable which defaults to `null`, I'm getting the following segfault.

`docker-bake.hcl`:

```hcl
variable "FOO" { default = null }
target "default" { }
```

Error:

```shell
$ FOO="bar" docker buildx bake --print
[+] Building 0.0s (0/0)                                                                                                                                                                                                                           
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x845567]

goroutine 1 [running]:
github.com/docker/buildx/bake/hclparser.(*parser).resolveValue(0xc000531720, {0xc000908980, 0x3})
        github.com/docker/buildx/bake/hclparser/hclparser.go:244 +0xe27
github.com/docker/buildx/bake/hclparser.Parse({0x221aec8, 0xc000895698}, {0x1ff3f90?, 0xc000759d70?, 0x1ff0a50?}, {0x1d408e0?, 0xc000759da0})
        github.com/docker/buildx/bake/hclparser/hclparser.go:358 +0xbcd
github.com/docker/buildx/bake.ParseFiles({0xc0004e05a0, 0x1, 0x19?}, 0x18?)
        github.com/docker/buildx/bake/bake.go:238 +0xa45
github.com/docker/buildx/bake.ReadTargets({0x13?, 0x0?}, {0xc0004e05a0?, 0xc000056004?, 0x59?}, {0xc000837a20, 0x1, 0xc00079f801?}, {0x0, 0x0, ...}, ...)
        github.com/docker/buildx/bake/bake.go:88 +0x52
github.com/docker/buildx/commands.runBake({0x222a098, 0xc00017a2a0}, {0xc0003c2520, 0x0, 0x1}, {{0x3145700, 0x0, 0x0}, {0x0, 0x0, ...}, ...})
        github.com/docker/buildx/commands/bake.go:106 +0xb2c
github.com/docker/buildx/commands.bakeCmd.func1(0xc000733900?, {0xc0003c2520, 0x0, 0x1})
        github.com/docker/buildx/commands/bake.go:182 +0x205
github.com/spf13/cobra.(*Command).execute(0xc000733900, {0xc000409400, 0x1, 0x1})
        github.com/spf13/cobra@v1.5.0/command.go:872 +0x694
github.com/spf13/cobra.(*Command).ExecuteC(0xc000366500)
[...]

PR #1449, opened 2022-12-03
bake: support null arg and label value

crazy-max:
follow-up https://github.com/docker/buildx/issues/1362#issuecomment-1287397312

```hcl
variable "GO_VERSION" {
  default = ""
}
target "default" {
  args = {
    GO_VERSION = GO_VERSION
  }
}
```

```dockerfile
ARG GO_VERSION="1.18"
FROM golang:${GO_VERSION}
```

```console
$ docker buildx bake --print
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "GO_VERSION": ""
      }
    }
  }
}
```

In this example, `GO_VERSION` arg will always be sent in the request and overwrites the one in the Dockerfile. Therefore build would fail as `GO_VERSION` will be empty.

With this PR, `null` type is now handled and will set the arg only if set:

```hcl
variable "GO_VERSION" {
  default = null
}
target "default" {
  args = {
    GO_VERSION = GO_VERSION
  }
}
```

```console
$ docker buildx bake --print
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile"
    }
  }
}
```

```console
$ GO_VERSION=1.19 docker buildx bake --print
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "GO_VERSION": "1.19"
      }
    }
  }
}
```

### DIFF
--- a/bake/hclparser/expr.go
+++ b/bake/hclparser/expr.go
@@ -83,7 +83,7 @@ func appendJSONFuncCalls(exp hcl.Expression, m map[string]struct{}) error {
 
 	// hcl/v2/json/ast#stringVal
 	val := src.FieldByName("Value")
-	if val.IsZero() {
+	if !val.IsValid() || val.IsZero() {
 		return nil
 	}
 	rng := src.FieldByName("SrcRange")
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -559,24 +559,24 @@ type Target struct {
 	Attest   []string `json:"attest,omitempty" hcl:"attest,optional" cty:"attest"`
 	Inherits []string `json:"inherits,omitempty" hcl:"inherits,optional" cty:"inherits"`
 
-	Context          *string           `json:"context,omitempty" hcl:"context,optional" cty:"context"`
-	Contexts         map[string]string `json:"contexts,omitempty" hcl:"contexts,optional" cty:"contexts"`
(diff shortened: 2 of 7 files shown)


crazy-max:
Including a null value in a string template is not yet supported:

```hcl
variable "GO_VERSION" {
  default = null
}
target "default" {
  args = {
    GO_VERSION = "${GO_VERSION}-alpine"
  }
}
```

This will lead to:

```
Invalid template interpolation value; The expression result is null. Cannot include a null value in a string template.
```

bake/hclparser/expr.go
```diff
@@ -83,11 +83,11 @@ func appendJSONFuncCalls(exp hcl.Expression, m map[string]struct{}) error {
 
 	// hcl/v2/json/ast#stringVal
 	val := src.FieldByName("Value")
-	if val.IsZero() {
+	if !val.IsValid() || val.IsZero() {
 		return nil
 	}
 	rng := src.FieldByName("SrcRange")
-	if val.IsZero() {
+	if rng.IsZero() {
```
crazy-max, line 90: Looks like `SrcRange` was not correctly checked. Can open another PR for this one.

tonistiigi (approved):

PR #1435, opened 2022-11-30
vendor: github.com/moby/buildkit v0.11.0-rc1.0.20221207183946-5993b526de65

thaJeztah:
- drops the replace-rule for github.com/aws/aws-sdk-go-v2/config (as it no longer was replacing anything)
- drops the replace-rules for docker/cli and docker/docker (at least as long as we continue using tagged releases)
- removes the github.com/docker/docker/pkg/stringid package (which was redundant)

full diff: https://github.com/moby/buildkit/compare/9624ab4710dd..5993b526de65dc62e70978c90e3dd6086b1db6e7

PR #1452, opened 2022-12-06
Remove stale replace directive

errordeveloper:
This replace directive slipped in as part of 57d22a7bd, however it wasn't needed in the first place as it points to the latest version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -170,7 +170,6 @@ require (
 )
 
 replace (
-	github.com/aws/aws-sdk-go-v2/config => github.com/aws/aws-sdk-go-v2/config v1.15.5 // same as buildkit
 	github.com/docker/cli => github.com/docker/cli v20.10.3-0.20221124184145-c0fa00e6142d+incompatible // v23.0.0-dev
 	github.com/docker/docker => github.com/docker/docker v20.10.3-0.20221124164242-a913b5ad7ef1+incompatible // 22.06 branch (v23.0.0-dev)
 	k8s.io/api => k8s.io/api v0.22.4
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -170,7 +170,6 @@ require (
 )
 
 replace (
-	github.com/aws/aws-sdk-go-v2/config => github.com/aws/aws-sdk-go-v2/config v1.15.5 // same as buildkit
```
crazy-max, line 173: This was on purpose to be consistent with BuildKit one: https://github.com/moby/buildkit/blob/4451e1be0e6889ffc56225e54f7e26bd6fdada54/go.mod#L12 so we don't upgrade inadvertently this dependency.

errordeveloper: In that case, let's add a more specific comment.. why do we suspect this particular dependency can break on upgrade? I was under impression that AWS SDK is not the kind of project that evolves rapidly, quite the opposite...

errordeveloper: > > In that case, let's add a more specific comment..
> 
> `// same as buildkit` looks clear enough to me but feel free to do so.

I read that as if buildkit repo is supposed have same replace directive, but it hasn't got one. So what you need to say is that the reason for replace statement is to pin the dependency and prevent upgrades.

> >  I was under impression that AWS SDK is not the kind of project that evolves rapidly, quite the opposite...
> 
> With [252 releases](https://github.com/aws/aws-sdk-go-v2) as of today I don't think so :sweat_smile:

98% of the SDK is generated code, most of the commits are additions for new API calls or new fields in existing API calls. Release count is probably very artificial also.

crazy-max:
> I was also removing these in #1435

Yes looks fine after all :+1:

PR #1456, opened 2022-12-07
build: refactor reference parsing for oci image layouts

jedevc:
:arrow_up: Follow up to #1173.

We allow any valid image reference format for the oci-layout, not just limiting to name@digest, we additionally allow images of the form name:tag@digest now.

The name of the reference is used to find the local directory to lookup the store in, while the tag and digest are attached to a random identity to generate the dummy reference sent to the oci-layout context.

This separation of the target to replace and the value to replace it with ensures that any tag or digest set in the client is properly sent across to the server. The tag, while not relevant now, may become relevant at some point in the future for the server, so we can should send it across for now.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -36,8 +37,10 @@ import (
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/client/llb"
+	"github.com/moby/buildkit/client/ociindex"
 	"github.com/moby/buildkit/exporter/containerimage/exptypes"
 	gateway "github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/moby/buildkit/identity"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/upload/uploadprovider"
 	"github.com/moby/buildkit/solver/errdefs"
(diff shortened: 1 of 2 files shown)


crazy-max (approved):

tonistiigi (commented):
Why the handwritten parsing and not just using the `reference` library?

jedevc:
> Why the handwritten parsing and not just using the reference library?

From the early implementation in https://github.com/docker/buildx/pull/1173#discussion_r901172626: `need to add "dummy.io", because reference.Parse*() funcs require we do not start with a '/'`. The reference library doesn't allow a valid path as the name, so we can only include OCI images in the current directory :sweat_smile:

tonistiigi (commented):
We should still do the validation of digest on client side. If it is a tag, we should resolve it on the client side as well.

tonistiigi (commented):
Requiring digest is very painful for the user. We should just support directory (optional tag), and find the digest on client side if possible.

jedevc:
Have added logic for doing some basic tag resolution as well as some digest verification. Unfortunately, the hand-written parsing is still required.

build/build.go
```diff
@@ -1478,26 +1481,53 @@ func LoadInputs(ctx context.Context, d driver.Driver, inp Inputs, pw progress.Wr
...
 			pathAlone := strings.TrimPrefix(v.Path, "oci-layout://")
-			parts := strings.SplitN(pathAlone, "@", 2)
-			if len(parts) != 2 {
-				return nil, errors.Errorf("invalid oci-layout context %s, must be oci-layout:///path/to/layout@sha256:hash", v.Path)
+			localPath := pathAlone
+			localPath, dig, hasDigest := strings.Cut(localPath, "@")
+			localPath, tag, hasTag := strings.Cut(localPath, ":")
+			if !hasDigest {
+				if !hasTag {
+					tag = "latest"
```
tonistiigi, line 1489: If `index` only contains one record then the `tag==latest` constraint shouldn't be required.

jedevc: Have added another branch to check for this case.

jedevc:
cc @deitch (thought you might be interested in this one)

deitch:
Thanks for cc'ing me @jedevc . I am interested.

I always have to remind myself of the current behaviour, just to evaluate what this does. 😁 

The prior was `oci-layout:///path/to/image@sha256:<hash>`. It would then look in `/path/to/image/blobs/sha256/<hash>` to find the manifest/index.

This changes it to also support `oci-layout:///path/to/image:tag`, so it can read `index.json` for the tag, and then find the manifest/index from there? That is much nicer.

Based on the code, it looks like any form will work, e.g. I have seen many that "override" the `index.json` to include many images, not just many tags, and I think that would work here?

deitch:
Should we add some examples to the docs/README for it?

jedevc:
Ah, yup, have updated the reference docs for it no, to explain the new tags behavior :+1: It's a lot neater now that we don't have to explain about requiring a digest in the name :tada:

tonistiigi (approved):

PR #1460, opened 2022-12-11
e2e: load to docker store

crazy-max:

### DIFF
--- a/hack/test-driver
+++ b/hack/test-driver
@@ -131,6 +131,14 @@ buildxCmd build ${buildPlatformFlag} \
   "${context}"
 cat "${context}/metadata-build.json"
 
+# load to docker store
+if [ "$DRIVER" != "docker" ]; then
+  buildxCmd build \
+    --output="type=docker,name=buildx-test-load" \
+    --builder="${builderName}" \
+    "${context}"
+fi
+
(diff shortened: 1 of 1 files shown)


crazy-max:
https://github.com/docker/buildx/actions/runs/3670274233/jobs/6204713067#step:11:205

```
#12 exporting to docker image format
#12 ERROR: docker exporter does not currently support exporting manifest lists
------
 > exporting to docker image format:
------
ERROR: failed to solve: docker exporter does not currently support exporting manifest lists
```

@jedevc I think we need changes [here](https://github.com/moby/buildkit/blob/01c0d6748e84eec86f8a61efd8cedf1f4ac345f3/exporter/oci/export.go#L118-L120) to properly handle multi-platform for OCI exporter like we do for container one.

jedevc:
@crazy-max I think this should be solved by https://github.com/moby/buildkit/pull/3380.

It's a result of buildx attempting to attach provenance attestations by default, but, we need to filter out the inline attestations for the docker exporter just like we do for the local and tar exporter.

crazy-max:
@jedevc Nice, didn't catch that PR.

jedevc (approved):

PR #1462, opened 2022-12-13
build: set provenance vcs details

crazy-max:
follow-up https://github.com/moby/buildkit/pull/3240#issuecomment-1296613640

Sets provenance vcs details.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -595,6 +595,10 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		so.FrontendAttrs["attest:provenance"] = "mode=min,inline-only=true"
 	}
 
+	for k, v := range getGitAttributes(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath) {
+		so.FrontendAttrs[k] = v
+	}
+
 	// set platforms
 	if len(opt.Platforms) != 0 {
 		pp := make([]string, len(opt.Platforms))
@@ -845,21 +849,6 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 	}()
 
 	eg, ctx := errgroup.WithContext(ctx)
-
-	for _, opt := range opt {
-		gitLabels, err := addGitProvenance(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath)
-		if err != nil {
-			return nil, err
-		}
-		for n, v := range gitLabels {
-			if _, ok := opt.Labels[n]; !ok {
-				if opt.Labels == nil {
-					opt.Labels = map[string]string{}
-				}
-				opt.Labels[n] = v
-			}
-		}
-	}
 
 	for k, opt := range opt {
 		multiDriver := len(m[k]) > 1
--- a/build/git.go
+++ b/build/git.go
@@ -3,23 +3,41 @@ package build
 import (
 	"context"
 	"os"
-	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 
-	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
-	"github.com/pkg/errors"
+	"github.com/docker/buildx/util/gitutil"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
 )
 
 const DockerfileLabel = "com.docker.image.source.entrypoint"
 
-func addGitProvenance(ctx context.Context, contextPath string, dockerfilePath string) (map[string]string, error) {
-	v := os.Getenv("BUILDX_GIT_LABELS")
-	if (v != "1" && v != "full") || contextPath == "" {
-		return nil, nil
+func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath string) (res map[string]string) {
+	res = make(map[string]string)
+	if contextPath == "" {
+		return
 	}
-	labels := make(map[string]string, 0)
+
+	setGitLabels := false
+	if v, ok := os.LookupEnv("BUILDX_GIT_LABELS"); ok {
+		if v == "full" { // backward compatibility with old "full" mode
+			setGitLabels = true
+		} else if v, _ := strconv.ParseBool(v); v {
+			setGitLabels = v
+		}
+	}
+	setGitInfo := true
+	if v, ok := os.LookupEnv("BUILDX_GIT_INFO"); ok {
+		if v, _ := strconv.ParseBool(v); v {
+			setGitInfo = v
+		}
+	}
+
+	if !setGitLabels && !setGitInfo {
+		return
+	}
 
 	// figure out in which directory the git command needs to run in
 	var wd string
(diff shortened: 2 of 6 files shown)


build/build.go
```diff
@@ -595,6 +595,28 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
...
+		so.FrontendAttrs["label:"+specs.AnnotationSource] = gitprv.Source
+		so.FrontendAttrs["vcs:source"] = gitprv.Source
+	}
+	if gitprv.Revision != "" {
+		so.FrontendAttrs["label:"+specs.AnnotationRevision] = gitprv.Revision
+		so.FrontendAttrs["vcs:revision"] = gitprv.Revision
+	}
+	if gitprv.DockerfilePath != "" {
+		so.FrontendAttrs["label:"+DockerfileLabel] = gitprv.DockerfilePath
+	}
```
crazy-max, line 618: Keep the current behavior that sets git provenance labels introduced in https://github.com/docker/buildx/pull/1297.

build/git.go
```diff
@@ -3,23 +3,32 @@ package build
...
+}
+
+func getGitProvenance(ctx context.Context, mode string, contextPath string, dockerfilePath string) (*gitProvenance, error) {
+	switch mode {
+	case "min":
+		mode = "1"
+	case "max":
+		mode = "full"
+	}
+	if (mode != "1" && mode != "full") || contextPath == "" {
```
crazy-max, line 29: @tonistiigi @jedevc Atm I just translate provenance mode to the one we currently have for labels. WDYT?

crazy-max: I think provenance vcs attributes should be set only for `mode=max`.

jedevc: > Atm I just translate provenance mode to the one we currently have for labels. WDYT?

We should do the reverse IMO. Let's use the `min` and `max` here, and then translate to the "legacy" labels mode.

Sidenote: we should also use `strconv.ParseBool` instead of just a check for `1`.

> I think provenance vcs attributes should be set only for `mode=max`.

Agreed I think VCS only for `mode=max` is a good call and makes sense with the rest of how `min` and `max` are setup. `min` shouldn't "leak" information, `max` should provide everything that it is possible to provide.

build/build.go
```diff
@@ -595,6 +595,28 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
...
+			so.FrontendAttrs["vcs:source"] = gitprv.Source
+		}
+		if gitprv.Revision != "" {
+			so.FrontendAttrs["label:"+specs.AnnotationRevision] = gitprv.Revision
+			so.FrontendAttrs["vcs:revision"] = gitprv.Revision
+		}
+		if gitprv.DockerfilePath != "" {
+			so.FrontendAttrs["label:"+DockerfileLabel] = gitprv.DockerfilePath
+		}
+	}
```
jedevc, line 618: I'm not sure about this :thinking: 

I don't think setting `--provenance mode=max` should mean that we set additional labels. Labels control should be set *exclusively* when the `BUILDX_GIT_LABELS` env variable is set. Though I think `BUILDX_GIT_LABELS` probably should set the vcs opts?

I think we could refactor `getProvenanceMode` to not do the env variable lookup, we could just do it right in this function, then, if that variable is set, set the labels.

crazy-max: My thought was to have the same behavior for provenance mode and BUILDX_GIT_LABELS env vars so users can still set vcs attributes. I agree that BUILDX_GIT_LABELS should just set labels. I think we should also add a deprecation notice if this env var is being used and tell user to switch to provenance.

jedevc: I think my confusion is that in `getGitProvenance` we have string values of "full" through the function. Ideally, we'd pass both the `BUILDX_GIT_LABELS` and extracted mode into `getGitProvenance`, and then parse what we want to do at the top - then we don't have to have magic strings through it.

> I think we should also add a deprecation notice if this env var is being set and tell user to switch to provenance.

Yeah, can we do this at a higher level? Otherwise, we'd get it on every call to `toSolveOpt` :scream:

crazy-max: > Yeah, can we do this at a higher level? Otherwise, we'd get it on every call to `toSolveOpt` 😱

Yes should be at same level where we display the "no output" warn in BuildWithResultHandler.

tonistiigi (commented):
This PR shouldn't care about provenance attestations, just set the vcs options. Provenance is only defined by #1412

build/build.go
```diff
@@ -595,6 +595,41 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		so.FrontendAttrs["attest:provenance"] = "mode=min,inline-only=true"
 	}
 
+	var prvmode string
+	if v, ok := opt.Attests["attest:provenance"]; ok {
+		if prvmode, err = getProvenanceMode(*v); err != nil {
+			return nil, nil, err
+		}
+	}
+	gitlabels := os.Getenv("BUILDX_GIT_LABELS")
```
tonistiigi, line 604: `BUILDX_GIT_LABELS` should remain as it is you can add a new env `BUILDX_GIT_INFO` that defaults to true.

If `BUILDX_GIT_LABELS` is true then set labels.
If `BUILDX_GIT_INFO` is not false then set `vcs`. Do not check anything about provenance mode.

crazy-max: Changed the behavior with your suggestion

build/git.go
```diff
@@ -30,69 +32,38 @@ func addGitProvenance(ctx context.Context, contextPath string, dockerfilePath st
 		wd, _ = filepath.Abs(filepath.Join(cwd, contextPath))
 	}
 
-	// check if inside git working tree
-	cmd := exec.CommandContext(ctx, "git", "rev-parse", "--is-inside-work-tree")
-	cmd.Dir = wd
-	err := cmd.Run()
-	if err != nil {
+	gitc, _ := gitutil.New(gitutil.WithContext(ctx), gitutil.WithWorkingDir(wd))
+	if !gitc.IsInsideWorkTree() {
```
tonistiigi, line 36: This is being accessed without error check? Possible panic?

build/git.go
```diff
@@ -30,69 +32,38 @@ func addGitProvenance(ctx context.Context, contextPath string, dockerfilePath st
...
-	// add a remote url if full Git details are requested; if there aren't any remotes don't fail
-	if v == "full" {
-		cmd = exec.CommandContext(ctx, "git", "ls-remote", "--get-url")
-		cmd.Dir = wd
-		out, _ := cmd.Output()
-		if len(out) > 0 {
-			labels[ocispecs.AnnotationSource] = strings.TrimSpace(string(out))
+	if sha, err := gitc.FullCommit(); err == nil && sha != "" {
+		gitdetails.Revision = sha
+		if gitc.IsDirty() {
```
tonistiigi, line 45: It looks like this increases the amount of `git` calls compared to the previous implementation? Can be handled in follow up, but every invocation counts.

build/git.go
```diff
@@ -30,69 +32,38 @@ func addGitProvenance(ctx context.Context, contextPath string, dockerfilePath st
...
-
-	// record only Dockerfile paths that are within the Git root
-	if !filepath.IsAbs(dockerfilePath) {
-		cwd, _ := os.Getwd()
-		dockerfilePath = filepath.Join(cwd, dockerfilePath)
-	}
-	dockerfilePath, _ = filepath.Rel(root, dockerfilePath)
-	if !strings.HasPrefix(dockerfilePath, "..") {
-		labels[DockerfileLabel] = dockerfilePath
+	if root, err := gitc.RootDir(); err == nil && root != "" {
```
tonistiigi, line 54: This part isn't needed unless labels I think.

tonistiigi (commented):
cc @cdupuis

build/git.go
crazy-max: Ah yes right only for labels

build/git.go
crazy-max: Good catch I forgot to update the interface after a change.

cdupuis:
Finally this code is getting cleaned up. Thanks @crazy-max and @tonistiigi.

tonistiigi (approved):

PR #1463, opened 2022-12-13
e2e: use native k3s installation script

crazy-max:
[`debianmaster/actions-k3s` action](https://github.com/debianmaster/actions-k3s/) triggers warnings in our e2e workflow: https://github.com/docker/buildx/actions/runs/3654304646

image

Switch to a native script inside our workflow so we don't depend on this action anymore.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -20,6 +20,7 @@ on:
 
 env:
   DESTDIR: "./bin"
+  K3S_VERSION: "v1.21.2-k3s1"
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -133,20 +134,67 @@ jobs:
       -
         name: Install k3s
         if: matrix.driver == 'kubernetes'
-        uses: debianmaster/actions-k3s@b9cf3f599fd118699a3c8a0d18a2f2bda6cf4ce4
-        id: k3s
+        uses: actions/github-script@v6
         with:
-          version: v1.21.2-k3s1
+          script: |
+            const fs = require('fs');
```
AkihiroSuda, line 140: Is there any specific reason to avoid shell scripting?

jedevc:
Could you make a note in the commit message, or in the source code that points to the upstream https://github.com/debianmaster/actions-k3s/blob/master/index.js?

Took me a moment to realize that the code is imported directly, so LGTM to me once we reference the origin of the import.

jedevc (approved):

Issue #1377, opened 2022-10-28
Multiple targets bake is slow to send context even if all targets use the same.


dgageot:
I've got a bake build that builds around 20 targets concurrently.
All of the targets share the same context, a Go project of around 100Mb.
When I run buildx bake, it's visible that it sends the same context to the buildkit, for each target.
I thought it was the logs that made me believe it was so but that underneath it was all fine and sent only once.

So I created a "minimal" [test project](https://github.com/dgageot/bake-contexts) with 40 targets building the same
Go sources. I added a 100Mb useless file to the context.

`docker buildx bake -f slow.hcl --load` takes multiple minutes to complete and we can see that the context is being sent 40 times.

I tested another option where I add a `base` target to the bakefile, copy the whole context into that image and made the Dockerfile copy from this target instead of directly from .
Now, the build is way less verbose and finished in 19s with:

`BASE="target:base" docker buildx bake -f slow.hcl --load`

By the way, this workaround requires changing all my Dockerfiles. Every `COPY x y` becomes `COPY --from=base x y`. It would be much simpler if we could redefine the "." context and set it to "target:base" using build-contexts flags.

Another thing interesting is how the build takes 19s and if you sum up the build logs, you've got 4.7s.

```
[+] Building 18.5s (173/173) FINISHED
 => [base internal] load build definition from Dockerfile                                                            0.0s
 => => transferring dockerfile: 59B                                                                                  0.0s
 => [base internal] load .dockerignore                                                                               0.0s
 => => transferring context: 2B                                                                                      0.0s
 => [pkg26 internal] load build context                                                                              0.0s
 => => transferring context: 7.50kB                                                                                  0.0s
 => CACHED [pkg26 1/1] COPY . .                                                                                      0.1s
 => [pkg26] exporting to image                                                                                       0.7s
 => => exporting layers                                                                                              0.0s
 => => writing image sha256:b028daff45f597ec1bf4a6a090391c5dce0fff854def9057b4ee005ee79ff8ea                         0.0s
 => => writing image sha256:82f0fd8aa9e650107ba5a6909752e26afe17d183af92d54ec28396ad44af70e5                         0.0s
 => [pkg04 internal] load build definition from Dockerfile                                                           0.0s
 => => transferring dockerfile: 360B                                                                                 0.0s
 => [pkg12 internal] load build definition from Dockerfile                                                           0.0s
 => => transferring dockerfile: 32B                                                                                  0.0s
[...]

PR #1466, opened 2022-12-13
Refactor build package

errordeveloper:
#1430 moved code was moved from `build/build.go`, however it is still a very long file with deeply nested logic. This change attemts to make a few improvements.


builder/node.go
```diff
@@ -34,9 +34,27 @@ func (b *Builder) Nodes() []Node {
...
+	if len(out) > 0 {
+		return out, nil
+	}
+	return nil, err
+}
+
 // LoadNodes loads and returns nodes for this builder.
 // TODO: this should be a method on a Node object and lazy load data for each driver.
-func (b *Builder) LoadNodes(ctx context.Context, withData bool) (_ []Node, err error) {
+func (b *Builder) LoadNodes(ctx context.Context, withData, onlyAvailable bool) (_ []Node, err error) {
```
jedevc, line 57: Ideally, I'd like to not modify this to take another boolean.

`AvailableNodes` can be renamed to `LoadAvailableNodes` - then `LoadNodes` and `LoadAvailableNodes` can call `LoadNodes` internally, and do the filtering on that directly.

errordeveloper: I think that makese sense, thanks 👍

errordeveloper: PTAL [32c9bf0](https://github.com/docker/buildx/pull/1466/commits/32c9bf0683cf25f2d472c2fe939dce7daf9d774e)

build/build.go
```diff
@@ -356,7 +81,7 @@ func toRepoOnly(in string) (string, error) {
 	return strings.Join(out, ","), nil
 }
 
-func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Options, bopts gateway.BuildOpts, configDir string, pw progress.Writer, dl dockerLoadCallback) (solveOpt *client.SolveOpt, release func(), err error) {
+func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt options.Options, bopts gateway.BuildOpts, configDir string, pw progress.Writer, dl dockerLoadCallback) (solveOpt *client.SolveOpt, release func(), err error) {
```
errordeveloper, line 84: I'd like to move this function into `options` package, but it looks like that would be more invloved, so I'll leave it for another PR.

options/options.go
```diff
@@ -0,0 +1,61 @@
+package options
```
tonistiigi, line 1: This should not be in a separate package. If there is some import case where only types are needed then it should be types pkg under `build`. But I don't see how these cases should be possible as these are just the types that `build` pkg users for its functions and not for anything else.

tonistiigi (commented):
Not sure how much this collides with #1296, but if it does we should get that in before moving a lot of things around.

options/options.go
errordeveloper: > Not sure how much this collides with #1296, but if it does we should get that in before moving a lot of things around.

That PR only touches 3 lines in `build/build.go`.

errordeveloper:
> Not sure how much this collides with #1296, but if it does we should get that in before moving a lot of things around.

That PR only touches 3 lines in `build/build.go`.

options/options.go
errordeveloper: Actually, the intention was to also move `toSolveOpts` and decouple it from the driver abstraction, which I'm planning to embark on next. The intention would be to make a cleaner abstractions we can continue to iterrate on, right now `build` package has lots of things leaky pieces. Eventully such `options` package could provide a more high-level `SovleOpts` replacement.

tonistiigi: In addition to https://github.com/docker/buildx/issues/1483#issuecomment-1362011227

I'm quite skeptical of the `toSolveOpt` abstraction that wouldn't just be a gimmick that only sets some of the properties, leaving the complicated ones to `build` pkg internals. There are lots of places where build invocations directly interact with driver interface (booting, caps detection, artifact loading, result chaining, multi-node result joining etc.). What we don't want is that there is an extra abstraction that Buildx itself doesn't really use and now when we need to implement some important user-facing features like https://github.com/docker/buildx/issues/1377 or https://github.com/docker/buildx/issues/1104 (that I think both change quite a lot how `build` pkg is called), we find out that this extra abstraction does not work in this cases and does not let us implement these features.

Assuming the abstraction that doesn't have the described maintenance issues exists and it is worth investing in writing one, it should happen within the `build` pkg. The problem isn't with splitting bigger files into smaller ones or changing function signatures where it makes sense, but introducing new pkg boundaries where the pkg is not usable on its own. If after the `build` pkg refactoring, we end up in the state where public functions(files) can be split in a way they become useful independently and without sacrificing future dev cost, then it is cheap to do the pkg split.

crazy-max:
With the controller implementation I think we can close this one?

PR #1477, opened 2022-12-15
build: lookup the right git binary on WSL

crazy-max:
closes https://github.com/docker/buildx/pull/1470

Windows drives in WSL version 2 uses the 9P protocol which the WSL instance connects to. When running git operations on a 9p fs using git from `/usr/bin/git`, it can be quite slow:

```console
$ cd /mnt/c/github.com/moby/moby
$ time git -c log.showSignature=false status --porcelain --ignored >/dev/null
real    1m5.620s
user    0m0.387s
sys     0m59.010s
```

But using `git.exe` is quite fast:

```console
$ cd /mnt/c/github.com/moby/moby
$ time git.exe -c log.showSignature=false status --porcelain --ignored >/dev/null
real    0m0.147s
user    0m0.003s
sys     0m0.009s
```

This change will detect if git operations occur on WSL and on a 9p fs and in this case lookup for `git.exe`, otherwise fallback to `git`.

### DIFF
--- a/build/git.go
+++ b/build/git.go
@@ -3,18 +3,19 @@ package build
 import (
 	"context"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
 
 	"github.com/docker/buildx/util/gitutil"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
-	"github.com/sirupsen/logrus"
+	"github.com/pkg/errors"
 )
 
 const DockerfileLabel = "com.docker.image.source.entrypoint"
 
-func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath string) (res map[string]string) {
+func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath string) (res map[string]string, _ error) {
 	res = make(map[string]string)
 	if contextPath == "" {
 		return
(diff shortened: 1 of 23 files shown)


build/git.go
```diff
@@ -48,9 +48,14 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 		wd, _ = filepath.Abs(filepath.Join(cwd, contextPath))
 	}
 
-	gitc := gitutil.New(gitutil.WithContext(ctx), gitutil.WithWorkingDir(wd))
+	gitc, err := gitutil.New(gitutil.WithContext(ctx), gitutil.WithWorkingDir(wd))
+	if err != nil {
+		logrus.Warnf("Failed to initialize git: %v", err)
```
tonistiigi, line 53: This should only warn if there is a `.git` dir

build/git.go
```diff
@@ -48,9 +48,14 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
...
+		return
+	}
+
 	if !gitc.IsInsideWorkTree() {
```
tonistiigi, line 57: False in here should not be a warning. If we get error from `git` then we should show it. Or say that `--debug` is needed to see the error.

build/git.go
tianon: Probably also want to be careful of Git worktree support (where `.git` will exist, but is a file that contains something like `gitdir: /full/path/to/.git/worktrees/foo`).

tianon: (if this is client-side, it'll "just work" when you run `git` commands but if it's server-side then you want to probably warn that `.git` exists but it's a file or something like that since you won't be able to get the original `.git` directory source in that case)

crazy-max: It client-side before sending the build request yes

build/git.go
```diff
@@ -50,30 +51,45 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 
 	gitc, err := gitutil.New(gitutil.WithContext(ctx), gitutil.WithWorkingDir(wd))
 	if err != nil {
-		logrus.Warnf("Failed to initialize git: %v", err)
+		if st, err := os.Stat(path.Join(wd, ".git")); err == nil && st.IsDir() {
+			return nil, errors.New("git not found in PATH, but .git directory exists")
```
tonistiigi, line 55: Still a warning, not a failing error.

```
WARN: No git was found in the system. Current commit information was not captured by the build.
```
```

File: build/git.go [hunk 4]
```diff
@@ -50,30 +51,45 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
...
+		}
 		return
 	}
 
 	if !gitc.IsInsideWorkTree() {
-		logrus.Warnf("Unable to determine git information")
-		return
+		return nil, errors.New("not inside a git repository")
```
tonistiigi, line 61: Not an error. If no git repo then just continue.

build/git.go
tonistiigi: I didn't notice that warning is on caller side. But update the message.

tonistiigi (approved):

PR #1482, opened 2022-12-21
Propagate SOURCE_DATE_EPOCH from the client env

AkihiroSuda:
See https://reproducible-builds.org/docs/source-date-epoch/

Expected to be used with BuildKit v0.11 or later (https://github.com/moby/buildkit/issues?q=label%3Aarea%2Freproducible-builds+)

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -579,6 +579,13 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		}
 	}
 
+	// Propagate SOURCE_DATE_EPOCH from the client env
+	if v := os.Getenv("SOURCE_DATE_EPOCH"); v != "" {
+		if _, ok := so.FrontendAttrs["build-arg:SOURCE_DATE_EPOCH"]; !ok {
+			so.FrontendAttrs["build-arg:SOURCE_DATE_EPOCH"] = v
+		}
+	}
+
 	if len(opt.Attests) > 0 {
 		if !bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations")) {
 			return nil, nil, errors.Errorf("attestations are not supported by the current buildkitd")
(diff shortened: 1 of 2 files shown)


tonistiigi:
Could we do it also on `bake` level so the `bake --print` would show that these arguments get passed.

build/build.go
```diff
@@ -579,6 +579,13 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		}
 	}
 
+	// Propagate SOURCE_DATE_EPOCH from the client env
+	if v, ok := os.LookupEnv("SOURCE_DATE_EPOCH"); ok {
```
tonistiigi, line 583: I think this should be `ok && v != ""` as `SOURCE_DATE_EPOCH=` could be a common way to reset the behavior.

AkihiroSuda: To support `SOURCE_DATE_EPOCH=`, the condition cannot contain `v != ""` IIUC

AkihiroSuda:
> Could we do it also on `bake` level so the `bake --print` would show that these arguments get passed.

updated

build/build.go
tonistiigi: wdym? I'm thinking of the case where environment has `SOURCE_DATE_EPOCH` defined but the user doesn't want to forward it to the buildkit(user wants buildkit to just use the current time). What should they set?

AkihiroSuda: When the Dockerfile already has `ARG SOURCE_DATE_EPOCH=<non empty value>`, I'd expect `export SOURCE_DATE_EPOCH=` to set the build arg to an empty string.

tonistiigi: Dockerfile defining `SOURCE_DATE_EPOCH` is only one possibility. The user might just want to control the timestamps in the config metadata and not the `RUN` commands.

AkihiroSuda: fixed

tonistiigi (approved):

AkihiroSuda:
I guess we can safely backport this to buildctl too:
- https://github.com/moby/buildkit/issues/4227

PR #1487, opened 2022-12-27
add system proxy config support for cli requests

tonistiigi:
This adds support for system HTTP_PROXY config for the requests that are made directly by the CLI. When previously user needed to call `HTTPS_PROXY=x docker buildx build` then now they can define these variables in `~/.docker/buildx/proxy.json` where they are loaded automatically.

Note that this is different from the proxy config in the Docker CLI config that buildx also loads. That config is per host and forwarded to the VM side, so it can be a completely different configuration.

@djs55 @thaJeztah 

### DIFF
--- a/cmd/buildx/main.go
+++ b/cmd/buildx/main.go
@@ -56,6 +56,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := configureProxy(cmd); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+
 	if plugin.RunningStandalone() {
 		err = runStandalone(cmd)
 	} else {
--- a/cmd/buildx/proxy.go
+++ b/cmd/buildx/proxy.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/docker/buildx/util/confutil"
+	"github.com/docker/cli/cli/command"
+	"github.com/pkg/errors"
+)
+
+func configureProxy(cmd *command.DockerCli) error {
+	fp := filepath.Join(confutil.ConfigDir(cmd), "proxy.json")
+	dt, err := os.ReadFile(fp)
+	if err != nil {
+		if os.IsNotExist(err) || os.IsPermission(err) {
+			return nil
+		}
+		return err
+	}
+	env := map[string]string{}
+	if err := json.Unmarshal(dt, &env); err != nil {
+		return errors.Wrapf(err, "failed to parse proxy config %s", fp)
+	}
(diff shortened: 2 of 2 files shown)


thaJeztah (changes requested):
WHYY???

So now;

- `docker buildx build`  will use  `~/,docker/buidx/proxy.json`
- `docker build` will use `HTTPS_PROXY` or `HTTP_PROXY`
- `docker compose build` will use????
- `docker manifest inspect` will use ???
- `docker scan` will use ???

Great. Unified experience.

tonistiigi:
> docker build will use HTTPS_PROXY or HTTP_PROXY

No, `docker build` forwards to buildx. Also, this doesn't change anything about HTTPS_PROXY or HTTP_PROXY envs.

> docker compose build will use????

Compose build should also forward to buildx (proxy matters only if they do requests).

> docker manifest inspect will use ???

Should forward to `imagetools inspect` (or remove)

> docker scan will use ???

That is for `docker scan` authors to decide.

> WHYY???

So that people who have this issue atm have a solution.

thaJeztah:
> So that people who have this issue atm have a solution.

What is the issue?


My issue is that buildx should not have this config; if we _do_ want proxy settings for the CLIs to be configurable in a configuration file, it should not be a buildx config, but a CLI configuration.

tonistiigi:
> What is the issue?

The system proxy can be configured for dockerd and Docker Desktop. People who rely on this(and for example can't dial to hosts without the proxy settings) with buildx hit an issue with their builds because buildkit does not send the registry credentials as plaintext to the daemon for security reasons, but client will ask for a short-lived token directly instead. To make this work user needs to always manually set `HTTP_PROXY` when calling buildx, or add it to .bashrc but then when it changes it can easily break. This allows that when such user modifies the proxy settings in desktop settings, the builds do not break. There are some other HTTP requests that buildx can do in advanced cases, but these are much less important.

> if we do want proxy settings for the CLIs to be configurable in a configuration file, it should not be a buildx config, but a CLI configuration.

If Docker CLI ever adds a similar feature defined in some other config file I see no issue with buildx also looking up that config.

tonistiigi:
We still don't have any solution for this.

jedevc (approved):

jedevc:
@thaJeztah we're not blocked on this, and can do the release without it - but this in the milestone. Is there anything we should do to move this forward, or should we try and do this upstream in cli?

jankap:
I do have exactly the mentioned problems. Proxies are handled via env vars, I have no control over the `docker buildx` call and its builders. Using the already set env vars or docker config vars should be the right way.

See https://github.com/docker/buildx/issues/2667#issuecomment-3033367177 and https://github.com/microsoft/vscode-remote-release/issues/11032.

Any chance to get this solved?

PR #1498, opened 2023-01-09
Improved attestation inspect

jedevc:
:arrow_up: Follow up to #1444 

This includes 3 main changes:
- Break after first found SPDX document (prioritizes the first attestations to take advantage of https://github.com/moby/buildkit/pull/3479)
- Re-work stub structs to have `interface{}` types instead of `json.RawMessage`; this allows them to be accessed using go template notation, e.g. `--format "{{ range .SBOM.SPDX.packages }}{{ println .name .versionInfo }}{{ end }}"`.
  - We strip away the `Predicate` field to directly access the predicate of the attestation. In the future, we should to sanity check the subject of the attestation against the pulled image, as well as the predicate type - this should be the responsibility of the go-imageinspect library.
  - If we want to provide access to the raw attestation contents (including subject, predicate-type, etc) in the future, we likely want to do that through a simple `.Attestations` field, but it seems an edge use-case for now.
- Allow access to multiple SPDX documents (created using the `BUILDKIT_SBOM_SCAN` args). The first SPDX document is added to `.SPDX`, while all documents are appended to `.SPDXs` - this can allow tools to search through dependencies used in build tooling.

### DIFF
--- a/util/imagetools/loader.go
+++ b/util/imagetools/loader.go
@@ -255,8 +255,8 @@ func (l *loader) scanConfig(ctx context.Context, fetcher remotes.Fetcher, desc o
 }
 
 type sbomStub struct {
-	SPDX  interface{}   `json:",omitempty"`
-	SPDXs []interface{} `json:",omitempty"`
+	SPDX            interface{}   `json:",omitempty"`
+	AdditionalSPDXs []interface{} `json:",omitempty"`
 }
 
 func (l *loader) scanSBOM(ctx context.Context, fetcher remotes.Fetcher, r *result, refs []digest.Digest, as *asset) error {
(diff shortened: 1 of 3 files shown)


util/imagetools/loader.go
```diff
@@ -285,8 +291,8 @@ func (l *loader) scanSBOM(ctx context.Context, fetcher remotes.Fetcher, r *resul
 	return nil
 }
 
-type slsaStub struct {
-	Provenance json.RawMessage `json:",omitempty"`
+type provenanceStub struct {
+	SLSA interface{} `json:",omitempty"`
 }
```
crazy-max, line 299: Provenance > SLSA? I thought it was SLSA > Provenance to keep common denominator first? Like in the future we can have SLSA > VSA.

jedevc: I think it makes sense to have Type > Format. We have SBOM > SPDX, so I think it makes sense to have Provenance > SLSA.

VSA isn't provenance, so we'd have another "type" for that I think.

crazy-max: Makes sense :+1:

util/imagetools/loader.go
```diff
@@ -275,17 +276,27 @@ func (l *loader) scanSBOM(ctx context.Context, fetcher remotes.Fetcher, r *resul
 				if err != nil {
 					return err
 				}
-				as.sbom = &sbomStub{
-					SPDX: dt,
+				var spdx struct {
+					Predicate interface{} `json:"predicate"`
```
tonistiigi, line 280: Bit concerned that there isn't any way to extract the actual attestation. But I guess we can add some special case/command for that later.

tonistiigi (approved):

PR #1502, opened 2023-01-10
fix the directory of the buildx binary

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

As I saw in the [docker-bake.hcl](https://github.com/docker/buildx/blob/master/docker-bake.hcl#L104), the binary is outputted into the folder `./bin/build/<binary>,` but in the README, it says `./bin/<binary>,` so, this PR aims to fix this.


crazy-max (approved):
LGTM thanks!

Issue #1497, opened 2023-01-09
panic: runtime error: invalid memory address or nil pointer dereference in BuildWithResultHandler


djs55:
In a Docker Desktop test on Windows I saw a Dockerfile
```
# syntax=docker/dockerfile:1
FROM alpine:3.16
RUN --network=host
```

and a build log
```
docker.exe buildx build --no-cache -t tag .
#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 105B done
#1 DONE 0.0s
#2 [internal] load .dockerignore
#2 transferring context: 2B done
#2 DONE 0.1s
#3 resolve image config for docker.io/docker/dockerfile:1
#3 ...
#4 [auth] docker/dockerfile:pull token for registry-1.docker.io
#4 DONE 0.0s
#3 resolve image config for docker.io/docker/dockerfile:1
#3 DONE 1.3s
#5 docker-image://docker.io/docker/dockerfile:1@sha256:9ba7531bd80fb0a858632727cf7a112fbfd19b17e94c4e84ced81e24ef1a0dbc
#5 resolve docker.io/docker/dockerfile:1@sha256:9ba7531bd80fb0a858632727cf7a112fbfd19b17e94c4e84ced81e24ef1a0dbc 0.0s done
#5 sha256:1e8a16826fd1c80a63fa6817a9c7284c94e40cded14a9b0d0d3722356efa47bd 2.37kB / 2.37kB done
#5 sha256:1328b32c40fca9bcf9d70d8eccb72eb873d1124d72dadce04db8badbe7b08546 0B / 9.94MB 0.1s
#5 sha256:9ba7531bd80fb0a858632727cf7a112fbfd19b17e94c4e84ced81e24ef1a0dbc 2.00kB / 2.00kB done
#5 sha256:ad87fb03593d1b71f9a1cfc1406c4aafcb253b1dabebf569768d6e6166836f34 528B / 528B done
#5 sha256:1328b32c40fca9bcf9d70d8eccb72eb873d1124d72dadce04db8badbe7b08546 9.94MB / 9.94MB 0.3s done
#5 extracting sha256:1328b32c40fca9bcf9d70d8eccb72eb873d1124d72dadce04db8badbe7b08546 0.1s done
#5 DONE 0.5s
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x1094a18]
       
goroutine 28 [running]:
[...]

PR #1504, opened 2023-01-10
Refactor BuildWithResultHandler to simplify concurrency usage

jedevc:
:hammer_and_wrench: Fixes https://github.com/docker/buildx/issues/1497.

Using the error group concurrency primitive we can avoid needing to create a separate wait group.
    
This allows us to sidestep the issue described in https://github.com/docker/buildx/issues/1497#issuecomment-1375905222 where the wait group could be completed, but the build invocation functions had not terminated - if one of the functions was to terminate with an error, then it was possible to encounter a race condition, where the result handling code would begin executing, despite an error.
    
The refactor to use a separate error group which more elegantly handles the concept of function returns and errors, ensures that we can't encounter this issue.

For clarity, this fix is split in to three separate commits, the first two refactoring to simplify the fix, the final one the actual fix, to make the diffs easier to review.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -946,10 +946,10 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 			if multiTarget {
 				span, ctx = tracing.StartSpan(ctx, k)
 			}
+			baseCtx := ctx
 
 			res := make([]*client.SolveResponse, len(dps))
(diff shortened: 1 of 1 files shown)


tonistiigi:
cc @ktock

tonistiigi (approved):
This looks bit scary (for cherry-pick), but I can't find any reason why it shouldn't work.

PR #1511, opened 2023-01-11
build: error when using docker exporter and attestations

jedevc:
Bug reported by @crazy-max:

> it seems image is not loaded anymore: https://github.com/docker/build-push-action/actions/runs/3892759918/jobs/6644553386#step:5:107
>
> We have importing to docker in the logs but does not look to be imported actually

This occurs when attestations are explicitly requested, and buildkit generates an index - which docker will not import. We should explicitly error in this case, and not allow the user to `--load` an image that contains explicit attestation flags.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -588,18 +588,22 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		}
 	}
 
-	if len(opt.Attests) > 0 {
-		if !bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations")) {
-			return nil, nil, errors.Errorf("attestations are not supported by the current buildkitd")
-		}
-		for k, v := range opt.Attests {
-			if v == nil {
-				continue
-			}
-			so.FrontendAttrs[k] = *v
+	attests := make(map[string]string)
+	for k, v := range opt.Attests {
+		if v != nil {
+			attests[k] = *v
 		}
 	}
-	if _, ok := opt.Attests["attest:provenance"]; !ok {
+	supportsAttestations := bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations"))
+	if len(attests) > 0 {
+		if !supportsAttestations {
+			return nil, nil, errors.Errorf("attestations are not supported by the current buildkitd")
+		}
+		for k, v := range attests {
+			so.FrontendAttrs[k] = v
+		}
+	}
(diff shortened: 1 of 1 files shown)


commands/build.go
```diff
@@ -231,6 +231,15 @@ func runBuild(dockerCli command.Cli, in buildOptions) (err error) {
 	if err != nil {
 		return err
 	}
+	for _, out := range outputs {
+		if out.Type == "docker" {
```
tonistiigi, line 235: this should use `driver.MultiPlatform`

tonistiigi: Or I guess because this is container driver with a Docker endpoint for exporting that endpoint should be checked similarly as `driver.MultiPlatform` is detected. If this gets complicated and doesn't work for regular multi-platform atm then it can be follow-up.

jedevc: I've reworked the patch to be handled in `toSolveOpt`, where we handle the exporter and multi-platform case like this today - this means it'll also work in bake.

I think you're right, we should check multi-platform capabilities in `toSolveOpt`, to see if we need to error with `"docker exporter does not currently support exporting manifest lists"` for either attestations or multi-platform?

But we can do that as a follow-up, it seems orthogonal to this - it makes sense for the behavior for attestations and multi-platform should be linked together in the same place (that both generate indexes).

crazy-max (approved):

build/build.go
```diff
@@ -447,6 +447,25 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
...
+	if len(attests) > 0 {
+		if !supportsAttestations {
+			return nil, nil, errors.Errorf("attestations are not supported by the current buildkitd")
+		}
+		for k, v := range attests {
+			so.FrontendAttrs[k] = v
+		}
+	}
+	if _, ok := opt.Attests["attest:provenance"]; !ok && supportsAttestations {
+		so.FrontendAttrs["attest:provenance"] = "mode=min,inline-only=true"
```
tonistiigi, line 466: I'm not sure I understand what stops the inline attestation to be added when loading to docker? Or did we block this in buildkit side?

jedevc: Yup, this is blocked buildkit side - https://github.com/moby/buildkit/pull/3466/files#diff-93b81ebb2d74a7ff89a643dfd137603e309b87754076de895b0c15dd4626a169R76.

We will attach inline attestations - but only if we would are already exporting an image index. For cases where we need to transform the manifest into the index, we only perform that upgrade for the `image` exporter.

PR #1512, opened 2023-01-11
docs: small tweak to opentelemetry

nicks:
The buildx builder persists across Desktop restarts. So when we start up Jaeger, we should recommend that people configure it to persist across restarts as well.

Signed-off-by: Nick Santos <nick.santos@docker.com>

### DIFF
--- a/docs/guides/opentelemetry.md
+++ b/docs/guides/opentelemetry.md
@@ -6,7 +6,7 @@ To capture the trace to [Jaeger](https://github.com/jaegertracing/jaeger), set
 First create a Jaeger container:
 
 ```console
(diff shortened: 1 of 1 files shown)


crazy-max:
Not related to this PR but docs-upstream failed with latest changes from master on docs repo: https://github.com/docker/buildx/actions/runs/3895035448/jobs/6649869726#step:8:3732

```
#0 0.043 - ./_site/engine/reference/commandline/plugin_ls/index.html
#0 0.043   *  linking to internal hash #filter that does not exist (line 142)
#0 0.043      <a href="#filter">filtering</a>
```

Ring a bell @thaJeztah?

crazy-max (approved):

PR #1534, opened 2023-01-20
docs: add note about minimum docker version

jedevc:
https://github.com/moby/buildkit/issues/3490 seems to have been caused by an old docker version - we should add a note to the Buildx README about minimum version requirements (we now need OCI support in the docker daemon now that moby/buildkit uses an OCI image), as well as adding a warning for what users can expect if they go ahead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,13 @@ Key features:
 
 # Installing
 
-Using `buildx` as a docker CLI plugin requires using Docker 19.03 or newer.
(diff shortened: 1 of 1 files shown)


crazy-max (approved):

crazy-max:
Needs also docs follow-up in https://docs.docker.com/build/install-buildx/ (cc @dvdksn)

PR #1549, opened 2023-01-24
imagetools: silence intoto warnings

jedevc:
As reported by @tonistiigi:

> using imagetools create it shows WARNING: reference for unknown type: application/vnd.in-toto+json . the command works but adds bogus output

This is the buildx equivalent of https://github.com/moby/buildkit/pull/3063.


util/imagetools/create.go
```diff
@@ -151,8 +153,9 @@ func (r *Resolver) Combine(ctx context.Context, srcs []*Source) ([]byte, ocispec
 }
 
 func (r *Resolver) Push(ctx context.Context, ref reference.Named, desc ocispec.Descriptor, dt []byte) error {
-	ref = reference.TagNameOnly(ref)
+	ctx = remotes.WithMediaTypeKeyPrefix(ctx, intoto.PayloadType, "intoto")
```
thaJeztah, line 156: As we're only consuming a single `const` from this; would it make sense to copy it and define it locally? I expect the mediatype would never change, so copying "should" be fine (as long as we reference where it came from); https://github.com/in-toto/in-toto-golang/blob/38d125cbf1c6ac87210df9c8366e9d65d916262a/in_toto/model.go#L51-L52

thaJeztah: (5K+ lines of vendored code and indirect dependencies added seemed like a bit much for a single `const` 😂)

jedevc: Agreed :laughing:

Somehow I missed we were already using this as an inline literal string :facepalm:

thaJeztah: Yeah, I was curious if there was more to that const than just a string, so I checked it out locally, and from the looks of it it was "just that" (a string), so it seemed pretty low risk that it would ever change (or at least a "calculated risk"®) 😅

thaJeztah (approved):
LGTM (but more eyes never hurt in case I missed something around the `WithMediaTypePrefix()` parts 😅)

thaJeztah:
@tonistiigi PTAL

tonistiigi (approved):

PR #1557, opened 2023-01-26
go.mod: bump to Go 1.19

crazy-max:
Go 1.20 will be there soon, I think it's time to move our go.mod to latest stable. We can then remove the compat in our vendor.Dockerfile

Downstream projects like compose or github.com/linuxkit/linuxkit should not be affected.

cc @rumpl 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/docker/buildx
 
-go 1.17
+go 1.19
 
 require (
 	github.com/aws/aws-sdk-go-v2/config v1.15.5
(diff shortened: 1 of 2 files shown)


jedevc (approved):

jedevc:
Would we also be interested in bumping in buildkit? I know we use 1.18 (for generics).

crazy-max:
> Would we also be interested in bumping in buildkit? I know we use 1.18 (for generics).

BuildKit is a tricky one, it's used by many projects that are still on Go 1.18. Maybe wait for 1.20 being GA.

crazy-max:
PTAL @thaJeztah

thaJeztah:
Do we _require_ 1.19 anywhere? Otherwise bumping to go1.18 (as "minimum known version"), and removing the `--compat` (or `--compat=go1.18`) probably would work?

PR #1563, opened 2023-01-28
ci: generate releases.json on release event

crazy-max:
We are currently using the [GitHub API in our setup-buildx-action](https://github.com/docker/setup-buildx-action/blob/0648fd6fd6801dcae14b6299cc1b23187d3cb797/src/github.ts#L10-L41) to check for latest and tagged releases to make sure they exist before download. But this requires using a token to avoid rate-limit. It's fine for public runners but GHES runners [don't have the `github.token` populated](https://github.com/docker/setup-buildx-action/blob/0648fd6fd6801dcae14b6299cc1b23187d3cb797/action.yml#L47-L55) automatically (see https://github.com/docker/setup-buildx-action/issues/194). They need to create a PAT.

This PR will solve this issue by generating and pushing a `releases.json` file in this repo when we publish a GitHub Release, see https://github.com/docker/buildx/actions/runs/4033493181/jobs/6934050707#step:3:5

This file will then be fetched through `raw.githubusercontent.com` endpoint on `setup-buildx-action` repo. This endpoint is better served for our purpose with 5000 requests per hour compared to the GitHub API endpoint that is limited to 60 requests per hour (unauth) and 1000 request per hour when authenticated.

Also ignore `.github/releases.json` file on pull request event as an action in a workflow run can't trigger a new workflow run anyway. See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#example-using-more-than-one-event

We could also generate this file within the setup-buildx-action repo if we don't want that in Buildx repo. Let me know what you think about it.


jedevc (approved):
This SGTM - since this is the only GitHub API we use as part of the action, it makes sense to access the actual release data in the same way as we access the releases themselves. There is going to always be some inconsistency here though - e.g. when tagging a new release, the binaries for that release can't be in the JSON file in the commit we tag (since I think the artifact URL is unpredictable?). That feels kind of annoying, so we should work out how we manage the file - maybe we should only allow modifying it on the master branch (and verify this through CI)?

We could maybe bring this into our actions utility tools when we do that? It seems generic enough that we might also want to use it in some of our other actions, e.g. for buildkit.

Slightly frustrating that this change was made right before the weekend 😢

crazy-max:
> There is going to always be some inconsistency here though - e.g. when tagging a new release, the binaries for that release can't be in the JSON file in the commit we tag (since I think the artifact URL is unpredictable?). That feels kind of annoying, so we should work out how we manage the file - maybe we should only allow modifying it on the master branch (and verify this through CI)?

This will be fine. Currently we are creating a draft GitHub Release:

https://github.com/docker/buildx/blob/a718d07f6414bfe960899050c0107e6baa86dcba/.github/workflows/build.yml#L201

So the workflow updating the `releases.json` will not be triggered until we publish the release. Will look like this: https://github.com/docker/buildx/compare/master...crazy-max:buildx:ci-releases-json

crazy-max:
Updated to remove the script, bake target and new Dockerfile so we only use GitHub Actions to generate the releases JSON file.

.github/workflows/releases-json.yml
```diff
@@ -0,0 +1,55 @@
...
+        name: Commit changes
+        run: |
+          git add -A .
+      -
+        name: Create PR
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
+        with:
+          base: master
+          branch: releases-json/${{ github.event.release.name }}
+          commit-message: "update .github/releases.json"
```
jedevc, line 49:
```suggestion
          commit-message: "github: update .github/releases.json"
```

Or some other prefix, for consistency.

.github/workflows/releases-json.yml
```diff
@@ -0,0 +1,55 @@
...
+  release:
+    types:
+      - released
+  pull_request:
+    paths-ignore:
+      - '.github/releases.json'
+
+jobs:
+  generate:
+    uses: crazy-max/.github/.github/workflows/releases-json.yml@002654044825b3c2b9856af61b8a2aaf389706b1
```
jedevc, line 17: Nice, good to have something reusable :sweat_smile: 

Tiny nit - could we potentially have this produce indented JSON, it would be useful if it was human-readable.

.github/workflows/releases-json.yml
crazy-max: `github` prefix sounds good

.github/workflows/releases-json.yml
crazy-max: > Tiny nit - could we potentially have this produce indented JSON, it would be useful if it was human-readable.

Yes might be better, done.

jedevc (approved):

PR #1564, opened 2023-01-28
bake: SOURCE_DATE_EPOCH: fix `panic: assignment to entry in nil map`

### LINKED ISSUE #1562: bake: possible panic when using SOURCE_DATE_EPOCH
Passing `SOURCE_DATE_EPOCH` environment variable causes `buildx bake` to crash if no build-args dictionary is defined in the hcl.
Reproducible with buildx 0.10.0 and 0.10.1.

docker-bake.hcl :
```
target "default" {
  dockerfile-inline = "FROM alpine"
}
```

Crash when passing SOURCE_DATE_EPOCH:
```
$ SOURCE_DATE_EPOCH=123 docker buildx bake
[+] Building 0.0s (0/0)                                                                                                 
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/docker/buildx/bake.ReadTargets({0x106279300?, 0x140006bdec0?}, {0x140004840f0?, 0x13?, 0x1065a7d40?}, {0x14000741a58, 0x1, 0x4?}, {0x0, 0x0, ...}, ...)
	github.com/docker/buildx/bake/bake.go:148 +0x4f8
github.com/docker/buildx/commands.runBake({0x1065b60c0, 0x14000514800}, {0x10752b508, 0x0, 0x0}, {{0x10752b508, 0x0, 0x0}, {0x0, 0x0, ...}, ...})
	github.com/docker/buildx/commands/bake.go:132 +0xae0
github.com/docker/buildx/commands.bakeCmd.func1(0x140000bf800?, {0x10752b508, 0x0, 0x0})
	github.com/docker/buildx/commands/bake.go:202 +0x1a8
github.com/spf13/cobra.(*Command).execute(0x140000bf800, {0x140001ac010, 0x0, 0x0})
	github.com/spf13/cobra@v1.6.1/command.go:916 +0x5c8
github.com/spf13/cobra.(*Command).ExecuteC(0x140000bf200)
	github.com/spf13/cobra@v1.6.1/command.go:1044 +0x35c
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.6.1/command.go:968
main.runStandalone(0x105d3570d?)
	github.com/docker/buildx/cmd/buildx/main.go:40 +0x7c
main.main()
	github.com/docker/buildx/cmd/buildx/main.go:60 +0xec
```
Offending code:
 https://github.com/docker/buildx/blob/a718d07f6414bfe960899050c0107e6baa86dcba/bake/bake.go#L145-L150

Workaround:
Add an `args` map to the hcl file:
```
target "default" {
  args = { DUMMY = "" }
  dockerfile-inline = "FROM alpine"
}
```

AkihiroSuda:
Fix #1562

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -144,6 +144,9 @@ func ReadTargets(ctx context.Context, files []File, targets, overrides []string,
 	// The logic is purposely duplicated from `build/build`.go for keeping this visible in `bake --print`.
 	if v := os.Getenv("SOURCE_DATE_EPOCH"); v != "" {
 		for _, f := range m {
+			if f.Args == nil {
+				f.Args = make(map[string]*string)
+			}
 			if _, ok := f.Args["SOURCE_DATE_EPOCH"]; !ok {
 				f.Args["SOURCE_DATE_EPOCH"] = &v
 			}


bake/bake.go
```diff
@@ -144,6 +144,9 @@ func ReadTargets(ctx context.Context, files []File, targets, overrides []string,
 	// The logic is purposely duplicated from `build/build`.go for keeping this visible in `bake --print`.
 	if v := os.Getenv("SOURCE_DATE_EPOCH"); v != "" {
 		for _, f := range m {
+			if f.Args == nil {
+				f.Args = make(map[string]*string)
```
jedevc, line 148: I think it's only the assignment to the map that causes the panic, the if condition is valid even if f.Args is nil - maybe we should only create a new map inside the condition?

jedevc (approved):

bake/bake.go
AkihiroSuda: Not a huge fan of dereferencing the nil slice even if it is valid. Just my preference though.

jedevc: SGTM

PR #1568, opened 2023-01-30
build: silently fail if git remote not found

crazy-max:
reported by @tonistiigi 

> Got this warning `WARNING: failed to get git remote url: no remote URL found for either origin or upstream` from by xx checkout. As a regular user I wouldn't really understand what it is about. Not even clear it came from buildx.

As discussed we should just silently fail if remote Git URL is not found. Also improve warning message by adding `buildx:` prefix.

### DIFF
--- a/build/git.go
+++ b/build/git.go
@@ -52,20 +52,20 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 	gitc, err := gitutil.New(gitutil.WithContext(ctx), gitutil.WithWorkingDir(wd))
 	if err != nil {
 		if st, err := os.Stat(path.Join(wd, ".git")); err == nil && st.IsDir() {
-			return res, errors.New("git was not found in the system. Current commit information was not captured by the build")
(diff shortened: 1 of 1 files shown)


build/git.go
```diff
@@ -52,20 +52,20 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 	gitc, err := gitutil.New(gitutil.WithContext(ctx), gitutil.WithWorkingDir(wd))
 	if err != nil {
 		if st, err := os.Stat(path.Join(wd, ".git")); err == nil && st.IsDir() {
-			return res, errors.New("git was not found in the system. Current commit information was not captured by the build")
+			return res, errors.New("buildx provenance: git was not found in the system. Current commit information was not captured by the build")
```
tonistiigi, line 55: I think just `buildx:` would be enough for prefix?

crazy-max: I'm also fine with it

tonistiigi (approved):

PR #1574, opened 2023-01-31
ci: fix pr branch name in releases-json workflow

crazy-max:
```
  /usr/bin/git checkout --progress -B releases-json/ 9f0b2a21-fceb-4f3a-832c-19ad87af7a3f --
  fatal: 'releases-json/' is not a valid branch name
  Error: The process '/usr/bin/git' failed with exit code 128
```

Well, on `workflow_dispatch` there is no release name. Let's just put a fixed branch name.

### DIFF
--- a/.github/workflows/releases-json.yml
+++ b/.github/workflows/releases-json.yml
@@ -46,7 +46,7 @@ jobs:
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
         with:
           base: master
-          branch: releases-json/${{ github.event.release.name }}
+          branch: bot/releases-json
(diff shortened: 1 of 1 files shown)


.github/workflows/releases-json.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
         with:
           base: master
-          branch: releases-json/${{ github.event.release.name }}
+          branch: bot/releases-json
```
jedevc, line 49: Would prefer if we could do something like a date? Just to disambiguate between multiple.

jedevc (approved):

PR #1577, opened 2023-01-31
monitor: add shell escape feature

jedevc:
Debuggers like GDB support prefixing a command with an exclamation mark "!" to pass the rest of the command to the shell. We can mimic this functionality.

Additionally, we can enable shlex parsing easily by using the google/shlex utility.

### DIFF
--- a/monitor/monitor.go
+++ b/monitor/monitor.go
@@ -2,8 +2,11 @@ package monitor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"os/exec"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -12,6 +15,7 @@ import (
 	"github.com/containerd/console"
 	controllerapi "github.com/docker/buildx/commands/controller/pb"
 	"github.com/docker/buildx/util/ioset"
+	"github.com/google/shlex"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/term"
 )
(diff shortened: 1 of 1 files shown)


monitor/monitor.go
```diff
@@ -255,3 +276,10 @@ type nopCloser struct {
 }
 
 func (c nopCloser) Close() error { return nil }
+
+func defaultShell(os string) []string {
```
thaJeztah, line 280: Is this what's used to run in the container?

- Should this ideally be shared with what's defined in [BuildKit](https://github.com/moby/buildkit/blob/9b28f40b8e80244e27cc2935c1522f2f63e7f55e/frontend/dockerfile/dockerfile2llb/defaultshell.go#L3-L8)?
- Also curious what should happen if the image actually has a different shell configured (`SHELL`), and perhaps expects something different than the default

jedevc: This isn't inside the container - that's handled separately.

This is to just run something on the client system, outside of the debugging session.

thaJeztah: Ah, yeah, if it's not inside the container then definitely doesn't need to be shared 👍

ktock (approved):
Thank you!

tonistiigi (commented):
For what would you use this?

jedevc:
I often use something like this when working with GDB to work out what's in the environment - listing files, reading their contents of them, etc. Maybe using shell to construct a specific input that might cause a bug, and then rerun the entire program, keeping previous breakpoints, etc, without needing to restart the GDB session, or open a different shell session.

This is possibly even more useful for working with debugging scenarios - e.g. you might want to create, view, edit or remove files using the shell that are used in your build. I can imagine using it like so:

```
(buildx) !make vendor # or some other simple change
(buildx) reload # rerun the build
```

jedevc:
I'd still be interested in taking this, or something like this - even though it's not in keeping with our style in buildx, it feels like it's more in style with other debuggers: it's kind of just about force-of-habit.

Maybe it would be good enough if we support CTRL+z to background the process to easily allow getting back to the host terminal, at least from monitor mode?

tonistiigi:
>  it feels like it's more in style with other debuggers:

Other debuggers debug your host software. We should stick to only processes defined using containers.

PR #1581, opened 2023-02-01
monitor: resolve paths arguments in client

ktock:
Following up for https://github.com/docker/buildx/pull/1296#issuecomment-1410470087

Quoted from @jedevc 's comment (https://github.com/docker/buildx/pull/1296#issuecomment-1410470087):

> ```
> BUILDX_EXPERIMENTAL=1 buildx build . --builder=dev --invoke=sh
> INFO: connecting to buildx server
> [+] Building 0.0s (2/2) FINISHED                                                                                                                                  
>  => [internal] load build definition from Dockerfile                                                                                                         0.0s
>  => => transferring dockerfile: 2B                                                                                                                           0.0s
>  => [internal] load .dockerignore                                                                                                                            0.0s
>  => => transferring context: 2B                                                                                                                              0.0s
> ERROR: failed to build: rpc error: code = Unknown desc = failed to solve: failed to read dockerfile: open /tmp/buildkit-mount3665150422/Dockerfile: no such file or directory
> ```

This commit allows using relative paths on `buildx build` arguments by fixing the controller client to resolve paths on the client side.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
@@ -28,6 +29,7 @@ import (
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	dockeropts "github.com/docker/cli/opts"
+	"github.com/docker/docker/builder/remotecontext/urlutil"
 	"github.com/docker/docker/pkg/ioutils"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/exporter/containerimage/exptypes"
@@ -514,6 +516,13 @@ func launchControllerAndRunBuild(dockerCli command.Cli, options buildOptions) er
 	}
 
 	// Start build
+	// NOTE: buildx server has the current working directory different from the client
+	// so we need to resolve paths to abosolute ones in the client.
+	optsP, err := resolvePaths(&opts)
+	if err != nil {
+		return err
+	}
+	opts = *optsP
 	ref, resp, err := c.Build(ctx, opts, pr, os.Stdout, os.Stderr, progress)
 	if err != nil {
 		return errors.Wrapf(err, "failed to build") // TODO: allow invoke even on error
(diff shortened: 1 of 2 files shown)


jedevc:
Hm, I'm fairly sure we should look for a better way to go about this :thinking:

As is, this looks *really* tricky to maintain over time, and I think it's best to avoid modifying the user-provided options wherever possible. I think we need to push down the file resolution to the components that actually open the files, though those are scattered in BuildKit's client code as well as in Buildx.

For example, in Buildkit, we need to handle it here:

https://github.com/docker/buildx/blob/6506166f02c3c856507a1b1708e1cd6fac2bf614/vendor/github.com/moby/buildkit/client/solve.go#L193-L201

To get the information of the right path through - I think we need to add a current working directory field to the `SolveOpt`. Then that gives us the ability to handle the paths used in the cache importers/exporters, as well as the dockerfile and contexts.

For things like the metadata file, and secrets, we need to handle those buildx-side - so we'd need to add the same working directory property to the `controllerapi.BuildOptions`.

I think this would work better as a long-term approach? The buildkit-side patches shouldn't be too complex, and should just affect the client, so we wouldn't need new capabilities or anything.

@tonistiigi are we missing an easier option here?

jedevc:
With #1614 merged, @ktock, could you rebase this with those most recent changes?

ktock:
@jedevc Thank you for working on #1614. Rebased this PR on that commit.

commands/build.go
```diff
@@ -649,3 +657,154 @@ func dockerUlimitToControllerUlimit(u *dockeropts.UlimitOpt) *controllerapi.Ulim
 	}
 	return &controllerapi.UlimitOpt{Values: values}
 }
+
+// resolvePaths resolves all paths contained in controllerapi.BuildOptions
+// and replaces them to absolute paths.
+func resolvePaths(options *controllerapi.BuildOptions) (_ *controllerapi.BuildOptions, err error) {
+	if options.ContextPath != "-" {
+		options.ContextPath, err = absIfLocal(options.ContextPath)
```
jedevc, line 665: I'm not sure if we should do `os.Stat` here as part of `absIfLocal`? I think we should just do `filepath.Abs`, and similarly below.

We know at this point it's definitely a path, and if the file doesn't exist, we should let the controller throw the error (which may or may not be a file not found error, we shouldn't assume the implementation at this point).

ktock: Thank you for the comment. Yes, we don't need os.Stat here because the options are already strongly typed. Fixed the patch.

jedevc: I just pushed a commit on top to handle the new oci-layout named context (which can also be a path). Does that seem good to you? If so, I think this is ready-to-go :tada:

jedevc (approved):

commands/build.go
ktock: Thank you for the following-up! LGTM.

PR #1582, opened 2023-02-01
Linting tidy-ups

jedevc:
Adds two new linting rules, gosec (to catch potential security issues, similar to https://github.com/moby/buildkit/pull/3224), and forbidigo (to catch use of `fmt.Errorf`, similar to https://github.com/moby/buildkit/pull/3305).

The gosec fixups did catch a couple of loop reference issues in the bake code, where we continued looping after capturing a reference from the loop - which would invalidate that reference. @crazy-max can you confirm these look alright to you now?


crazy-max (approved):
LGTM

PR #1591, opened 2023-02-03
update github.com/compose-spec/compose-go to v1.9.0

crazy-max:
https://github.com/compose-spec/compose-go/compare/v1.6.0...v1.9.0

### DIFF
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -185,7 +185,7 @@ func loadDotEnv(curenv map[string]string, workingDir string) (map[string]string,
 		return nil, err
 	}
 
-	envs, err := dotenv.UnmarshalBytes(dt)
+	envs, err := dotenv.UnmarshalBytesWithLookup(dt, nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


bake/compose.go
```diff
@@ -185,7 +185,7 @@ func loadDotEnv(curenv map[string]string, workingDir string) (map[string]string,
 		return nil, err
 	}
 
-	envs, err := dotenv.UnmarshalBytes(dt)
+	envs, err := dotenv.UnmarshalBytesWithLookup(dt, nil)
```
crazy-max, line 188: signature changed: https://github.com/compose-spec/compose-go/commit/784a7e77d00efbc046bc9b3c19b8a5936ffe1606#diff-4c44c44d24dee73102acafac9247f9522055c6f12509b32912544f239f314b8dL136-L139

jedevc (approved):

bake/compose.go
thaJeztah: Will that be problematic (e.g. compose using both `buildx 0.10` (which doesn't have this patch) and `compose-go 1.9.0` ? Do we need (deprecated) compatibility wrappers in compose-go? ./cc @ndeloof

crazy-max: I don't think so, compose-go in buildx is only used for bake and compose only use the `build` package in buildx afaik.

PR #1592, opened 2023-02-03
build: check reachable git commits

### LINKED ISSUE #1587: Git warning being displayed when building in a git directory with no commits yet
Encountered with `buildx: Docker Buildx (Docker Inc., v0.10.0)`

To reproduce (in a new dir):
```shell
git init
echo "FROM scratch" > Dockerfile
docker buildx build .
```

Output:
```
 => => transferring dockerfile: 50B                                        0.0s
 => [internal] load .dockerignore                                          0.0s
 => => transferring context: 2B                                            0.0s
 => exporting to image                                                     0.0s
 => => writing image sha256:54ac6bb475f018dd78804ac491a2dc0e0e3331f8b6347  0.0s
WARNING: failed to get git commit: fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

crazy-max:
fixes #1587 

We should check if the working tree has git commits before retrieving the git commit so it would not fail. I was also thinking of just silently fail. Let me know what sounds best.

### DIFF
--- a/build/git.go
+++ b/build/git.go
@@ -64,7 +64,7 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 		return res, nil
 	}
 
-	if sha, err := gitc.FullCommit(); err != nil {
+	if sha, err := gitc.FullCommit(); err != nil && !gitutil.IsUnknownRevision(err) {
 		return res, errors.Wrapf(err, "buildx: failed to get git commit")
 	} else if sha != "" {
 		if gitc.IsDirty() {
--- a/util/gitutil/gitutil.go
+++ b/util/gitutil/gitutil.go
@@ -3,6 +3,7 @@ package gitutil
 import (
 	"bytes"
 	"context"
+	"os"
 	"os/exec"
 	"strings"
 
@@ -115,6 +116,9 @@ func (c *Git) run(args ...string) (string, error) {
 	if c.wd != "" {
 		cmd.Dir = c.wd
 	}
+
+	// Override the locale to ensure consistent output
+	cmd.Env = append(os.Environ(), "LC_ALL=C")
 
 	stdout := bytes.Buffer{}
 	stderr := bytes.Buffer{}
--- a/util/gitutil/gitutil.go
+++ b/util/gitutil/gitutil.go
@@ -138,3 +138,12 @@ func (c *Git) clean(out string, err error) (string, error) {
 	}
 	return out, err
 }
+
+func IsUnknownRevision(err error) bool {
+	if err == nil {
+		return false
+	}
+	// https://github.com/git/git/blob/a6a323b31e2bcbac2518bddec71ea7ad558870eb/setup.c#L204
+	errMsg := strings.ToLower(err.Error())
+	return strings.Contains(errMsg, "unknown revision or path not in the working tree") || strings.Contains(errMsg, "bad revision")
+}
--- a/util/gitutil/gitutil_test.go
+++ b/util/gitutil/gitutil_test.go
@@ -46,6 +46,18 @@ func TestGitShortCommit(t *testing.T) {
 	require.Equal(t, 7, len(out))
 }
 
+func TestGitFullCommitErr(t *testing.T) {
+	Mktmp(t)
+	c, err := New()
+	require.NoError(t, err)
+
+	GitInit(c, t)
+
+	_, err = c.FullCommit()
+	require.Error(t, err)
+	require.True(t, IsUnknownRevision(err))
+}
+
 func TestGitTagsPointsAt(t *testing.T) {
 	Mktmp(t)
 	c, err := New()


jedevc (approved):

build/git.go
```diff
@@ -64,17 +64,19 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
...
-	} else if sha != "" {
-		if gitc.IsDirty() {
-			sha += "-dirty"
-		}
-		if setGitLabels {
-			res["label:"+specs.AnnotationRevision] = sha
-		}
-		if setGitInfo {
-			res["vcs:revision"] = sha
+	if gitc.HasCommit() {
```
tonistiigi, line 67: Could we do this as an error handler for `FullCommit` instead? 99.9% of builds will not have this special no-commit repo case and don't need the extra `git` binary call.

util/gitutil/gitutil.go
```diff
@@ -134,3 +134,7 @@ func (c *Git) clean(out string, err error) (string, error) {
 	}
 	return out, err
 }
+
+func IsUnknownRevision(err error) bool {
+	return err != nil && (strings.Contains(err.Error(), "unknown revision or path not in the working tree") || strings.Contains(err.Error(), "bad revision"))
```
thaJeztah, line 139: Is this error produced by git itself? Would it make sense to link to the source code where this originates from (in case this error would ever change; that way we know where to look for the updated error string).

(I did a quick search, and found other projects also referring to this error, so I "assume" it's an error produced by git https://grep.app/search?q=%22unknown%20revision%20or%20path%20not%20in%20the%20working%20tree%22)

thaJeztah: LOL, looks like others [generate this errors with a capital `U`](https://github.com/github/hub/blob/38bcd4ae469e5f53f01901340b715c7658ab417a/git/git.go#L136) (do we need to to `strtolower` the error, or match on `nknown revision` (skipping the first letter)? 

I also wonder if this string can be localised on Windows; looking at this line, (using `test_i18ngrep `) I fear that may be possible; https://github.com/git/git/blob/a6a323b31e2bcbac2518bddec71ea7ad558870eb/t/t1506-rev-parse-diagnosis.sh#L112

thaJeztah: Maybe we need to call `git` with a specific locale 🤔 https://stackoverflow.com/a/13044660/1811501

crazy-max: Might be better to use `LC_ALL=C` for git commands then https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html to override all the other localization settings.

thaJeztah: LOL, yeah, whatever works (and whatever doesn't _require_ locale configs to be present) as long as we get something consistent that we can parse.

My eye dropped on it, and I recall users on Windows reporting tickets and in those tickets various commands were translated, so I anticipated "that could be problematic if we're looking for a specific string"

jedevc (approved):

util/gitutil/gitutil.go
```diff
@@ -116,6 +117,9 @@ func (c *Git) run(args ...string) (string, error) {
 		cmd.Dir = c.wd
 	}
 
+	// Override the locale to ensure consistent output
+	cmd.Env = append(os.Environ(), "LC_ALL=C")
```
thaJeztah, line 121: Before I facepalm myself for not asking; did someone look (or do we need a tracking ticket) on the BuildKit side if we should also set this option there? (I think BuildKit also shells out to `git` - it may be in a more predictable environment, but possibly could still be a good thing to do)

util/gitutil/gitutil.go
```diff
@@ -134,3 +138,10 @@ func (c *Git) clean(out string, err error) (string, error) {
 	}
 	return out, err
 }
+
+func IsUnknownRevision(err error) bool {
+	// https://github.com/git/git/blob/a6a323b31e2bcbac2518bddec71ea7ad558870eb/setup.c#L204
+	return err != nil &&
+		(strings.Contains(strings.ToLower(err.Error()), "unknown revision or path not in the working tree") ||
+			strings.Contains(strings.ToLower(err.Error()), "bad revision"))
+}
```
thaJeztah, line 147: Mostly a nit (the whole condition is becoming a bit lengthy); wondering if splitting up, and using an early return makes things easier to read.

```go
func IsUnknownRevision(err error) bool {
	if err == nil {
		return false
	}
	// https://github.com/git/git/blob/a6a323b31e2bcbac2518bddec71ea7ad558870eb/setup.c#L204
	errMsg := strings.ToLower(err.Error())
	return strings.Contains(errMsg, "unknown revision or path not in the working tree") || strings.Contains(errMsg, "bad revision")
}
```

thaJeztah (approved):
LGTM, but left some thoughts 😂

util/gitutil/gitutil.go
crazy-max: It might be useful on BuildKit too so we have a consistent output when users report issues.

thaJeztah (approved):
LGTM

Issue #766, opened 2021-09-13
Setting Service Account on Kubernetes Deployment


apjoseph:
`docker buildx create` appears to have no driver specific option or mechanism for the Kubernetes driver to specify the service account the deployment should use. As far as I can tell there is no way to set this other than to manually modify the deployment itself after it as been created by applying a patch like this:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: example-buildkit0
  name: example-buildkit0
spec:
  selector:
    matchLabels:
      app: example-buildkit0
  template:
    spec:
      serviceAccountName: buildx-builder
      containers:
        - name: buildkitd
          image: moby/buildkit:buildx-stable-1
```   
          
It would be highly beneficial to allow the ability to set serviceAccounts on creation as these are often needed to access external resources such as a storage bucket using WorkloadIdentity. This is a fairly fundamental need for builds.

tonistiigi:
@morlay

odinsy:
any news?

PR #1597, opened 2023-02-04
kubernetes driver: add serviceaccount opt

robertlestak:
Addresses #766

Adds a new (optional) `serviceaccount` opt. When provided, it will be set as the created pod's `serviceAccountName`.

Example usage:

```bash
docker buildx create --name example \
    --driver kubernetes \
    --driver-opt=serviceaccount=example-sa
```

### DIFF
--- a/docs/reference/buildx_create.md
+++ b/docs/reference/buildx_create.md
@@ -163,6 +163,7 @@ No driver options.
 - `requests.memory` - Sets the request memory value specified in bytes or with a valid suffix. Example `requests.memory=500Mi`, `requests.memory=4G`
 - `limits.cpu` - Sets the limit CPU value specified in units of Kubernetes CPU. Example `limits.cpu=100m`, `limits.cpu=2`
 - `limits.memory` - Sets the limit memory value specified in bytes or with a valid suffix. Example `limits.memory=500Mi`, `limits.memory=4G`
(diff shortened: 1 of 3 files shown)


crazy-max:
PTAL @morlay @AkihiroSuda

AkihiroSuda (approved):

crazy-max (dismissed):
Can you update reference docs please in https://github.com/docker/buildx/blob/master/docs/reference/buildx_create.md#-set-additional-driver-specific-options---driver-opt for the kubernetes driver?

@dvdksn Would also need follow-up in docs repo: https://docs.docker.com/build/drivers/kubernetes/

dvdksn:
side note @crazy-max, I think the file you linked to is actually redundant, now that we have the full opt descriptions for each driver in the docs. If you agree, I can create a PR to clean up  `buildx_create.md`

crazy-max:
> side note @crazy-max, I think the section in the file you linked to is actually redundant, now that we have the full opt descriptions for each driver in the docs. If you agree, I can create a PR to clean up `buildx_create.md`

Ah sure yeah sounds good

jedevc:
@crazy-max @dvdksn for the time being, I pushed a commit to update `buildx_create.md` - we can clean it up later.

Anything else blocking on merging this, so we can get it in v0.11?

jedevc (approved):

Issue #745, opened 2021-08-24
bake --push throws error when image field is not defined within docker-compose file


dudicoco:
Within a docker-compose file, you might want to build services while only pushing some of them to a registry, for example:

```yaml
services:
  service1:
    build:
      context: ./service1
  service2:
    image: ${DOCKER_REGISTRY}/myservice:latest
    build:
      context: ./service2
```

Running `docker compose push` with the above configuration would skip service1 and only push service2.
Running `docker buildx bake --push` throws an error: `error: tag is needed when pushing to registry`.

The `bake` command should be able to skip pushing images which do not have a tag defined.

tonistiigi:
Global flags like `--push` apply to all targets. If you want to build specific ones then `docker buildx bake service2 servicefoo`

dudicoco:
@tonistiigi I want to build all targets, but push only the targets which has an image tag defined. I believe it's reasonable to expect this from bake considering that this is the behavior with docker compose.

crazy-max:
Looking at your compose file:

```
docker buildx bake --push --print
{
  "group": {
    "default": {
      "targets": [
        "service2",
        "service1"
      ]
    }
  },
  "target": {
    "service1": {
      "context": "service1",
      "dockerfile": "Dockerfile",
      "output": [
        "type=image,push=true"
      ]
    },
    "service2": {
      "context": "service2",
      "dockerfile": "Dockerfile",
      "tags": [
        "/myservice:latest"
      ],
      "output": [
        "type=image,push=true"
      ]
    }
  }
}
```

If you want to disable push for some services you can add [`output: type=cacheonly` field with `x-bake`](https://docs.docker.com/build/bake/compose-file/#extension-field-with-x-bake) so `--push` will not override this service:

```yaml
services:
  service1:
    build:
      context: ./service1
      x-bake:
        output: type=cacheonly
  service2:
    image: ${DOCKER_REGISTRY}/myservice:latest
    build:
      context: ./service2
```

```
docker buildx bake --push --print
{
  "group": {
    "default": {
      "targets": [
        "service2",
        "service1"
      ]
    }
  },
  "target": {
    "service1": {
      "context": "service1",
      "dockerfile": "Dockerfile",
      "output": [
        "type=cacheonly"
      ]
    },
    "service2": {
      "context": "service2",
      "dockerfile": "Dockerfile",
      "tags": [
        "/myservice:latest"
      ],
      "output": [
        "type=image,push=true"
      ]
    }
  }
}
```

PR #1600, opened 2023-02-05
bake: Make image field optional with compose

mcraek:
Potentially fixes https://github.com/docker/buildx/issues/745

This allows for the `image` field in compose files to be optional when using the `--push` option with `bake`. 

Builds still occur similar to how `docker compose --build` handles this situation, and a warning is output indicating what wasn't pushed due to the `image` field not being specified. This is similar to the warning that appears when `--push` is not specified with `bake`.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -360,6 +360,15 @@ func toRepoOnly(in string) (string, error) {
 	return strings.Join(out, ","), nil
 }
 
+func isBake() bool {
+	for _, i := range os.Args {
+		if i == "bake" {
+			return true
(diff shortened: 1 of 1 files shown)


crazy-max:
There should not be anything related to bake in the build package. Bake is an implementation detail. Like @tonistiigi said in https://github.com/docker/buildx/issues/745 you should just specify the services you want to build and push.

If you want to disable push for some services see my workaround: https://github.com/docker/buildx/issues/745#issuecomment-1443403038

PR #1612, opened 2023-02-08
Add env var to disable default attestations

cpuguy83:
For certain cases we need to build with `--provenance=false`. However not all build envs (especially in the OSS ethos) have the latest buildx so just blanket setting `--provenance=false` will fail in these cases.

Having an env var allows people to set the value without having to worry about if the buildx version has the `--provenance` flag.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -462,8 +462,19 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 			so.FrontendAttrs[k] = v
 		}
 	}
+
 	if _, ok := opt.Attests["attest:provenance"]; !ok && supportsAttestations {
-		so.FrontendAttrs["attest:provenance"] = "mode=min,inline-only=true"
+		const noAttestEnv = "BUILDX_NO_DEFAULT_ATTESTATIONS"
+		var noProv bool
+		if v, ok := os.LookupEnv(noAttestEnv); ok {
+			noProv, err = strconv.ParseBool(v)
+			if err != nil {
+				return nil, nil, errors.Wrap(err, "invalid "+noAttestEnv)
+			}
+		}
+		if !noProv {
+			so.FrontendAttrs["attest:provenance"] = "mode=min,inline-only=true"
+		}
 	}
 
 	switch len(opt.Exports) {


build/build.go
```diff
@@ -463,7 +463,16 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		}
 	}
 	if _, ok := opt.Attests["attest:provenance"]; !ok && supportsAttestations {
-		so.FrontendAttrs["attest:provenance"] = "mode=min,inline-only=true"
+		var noProv bool
+		if v, ok := os.LookupEnv("BUILDX_NO_DEFAULT_PROVENANCE"); ok {
```
crazy-max, line 467: I recall we have https://github.com/docker/buildx/blob/78058ce5f301fb5f7acb31eaee16fc85e37911dc/build/build.go#L386 to inline cache metadata to image config. I wonder if the var should be named `BUILDKIT_NO_DEFAULT_PROVENANCE` instead for consistency even if `BUILDKIT_INLINE_CACHE` should have been `BUILDX_INLINE_CACHE` imo as this is var is not handled in BuildKit but Buildx.

jedevc: Yeah, we also have [`BUILDX_NO_DEFAULT_LOAD`](https://github.com/docker/buildx/blob/04804ff3551b44d2d930be669aceaf2dd41b1f8b/build/build.go#L1698-L1708) though :thinking: 

Consistency isn't great here, though I'd be tempted to go with the `BUILDX_` prefix if possible.

I wonder if we should make the variable name more generic though, something like `BUILDX_NO_DEFAULT_ATTESTATIONS` - just in case we ever went the route of making some other attestations default as well (currently not planned, but better to have the breathing room IMHO).

jedevc: @cpuguy83 wdyt?

crazy-max: `BUILDX_NO_DEFAULT_ATTESTATIONS` lgtm

cpuguy83: Updated.

jedevc (approved):
LGTM, @tonistiigi?

Probably worth cherry-picking to 0.10 as well?

tonistiigi (approved):

jedevc (approved):

PR #1614, opened 2023-02-09
controller: strongly type the controller api

jedevc:
Strongly typing the API allows us to perform all command line parsing fully on the client-side, where we have access to the client local directory and all the client environment variables, which may not be available on the remote server.

Additionally, the controller api starts to look a lot like build.Options, so at some point in the future there may be an oppportunity to merge the two, which would allow both build and bake to execute through the controller, instead of needing to maintain multiple code paths. Maybe? I'm not sure - I think it depends on how stdin is handled.

@ktock, I think this would make something like https://github.com/docker/buildx/pull/1581 *significantly* easier to implement, am I right? We wouldn't need to dance around parsing and formatting CSVs, we could just edit the structs directly.


build/build.go
```diff
@@ -84,12 +84,13 @@ type Options struct {
...
 	Platforms     []specs.Platform
 	Pull          bool
-	Session       []session.Attachable
 	ShmSize       opts.MemBytes
 	Tags          []string
 	Target        string
 	Ulimits       *opts.UlimitOpt
 
+	Session []session.Attachable
+
```
ktock, line 93: nit: Why do we need to change the order?

ktock (commented):
Thank you. Overall LGTM.

build/build.go
jedevc: There's no real reason for this - except to try and match up with the [`controllerapi.BuildOptions` struct](https://github.com/docker/buildx/pull/1614/files/aead0c66e5e070d3a1882ff31860b2088bca18a8#diff-0cae9bac155d49c8975ab4e5c60d7bb85d98dc5226def18dc0daaeb66497e7e5R30-R49), that I'd like to eventually merge into the controller API. 

In conversation with @tonistiigi, we were talking about the possibility of having all buildx builds go through the controller interface, including with bake, etc, so as to reduce the number of possible code paths, though I think we need to keep playing around to try and see if that's possible.

ktock (approved):

controller/pb/controller.proto
```diff
@@ -25,32 +25,54 @@ message BuildOptions {
...
-  repeated string CacheTo = 8;
-  string CgroupParent = 9;
-  repeated string Contexts = 10;
-  repeated string ExtraHosts = 11;
-  string ImageIDFile = 12;
-  repeated string Labels = 13;
-  string NetworkMode = 14;
-  repeated string NoCacheFilter = 15;
-  repeated string Outputs = 16;
+  map<string, string> NamedContexts = 4;
```
tonistiigi, line 28: We should try to group the things that belong together. Either in substructs where they make sense or at least reserved ID ranges.

Currently, everything is randomly mixed up. Parameters sent to frontend, input contexts with print flags, imageID/quiet that are client-side options(quiet probably shouldn't be here at all. Nice if things that rely on session would be separate as well. CommonOpts is not a good name and I don't understand the property split there.

controller/pb/controller.proto
```diff
@@ -25,32 +25,54 @@ message BuildOptions {
...
+  map<string, string> Attrs = 2;
+  string Destination = 3;
+}
+
+message CacheOptionsEntry {
+  string Type = 1;
+  map<string, string> Attrs = 2;
+}
+
+message SSH {
```
tonistiigi, line 65: We need to think how the things that use session API work. Should everything in here have (optional) sessionID defined or at least one in `BuildOptions`.

controller/pb/controller.proto
jedevc: > We should try to group the things that belong together. Either in substructs where they make sense or at least reserved ID ranges.

Agreed. Can we do this as a follow-up, so as to not block https://github.com/docker/buildx/pull/1581? Substructs would be my preference wherever possible.

> imageID/quiet that are client-side options(quiet probably shouldn't be here at all)

Have pushed a commit that removes the quiet option entirely from the controller api - we can also extract the imageID file writing from the controller as well as a follow-up. To do this, we have the api return the solve response of the build, so we can extract the image id (which is needed because `--quiet` prints the image id, functionality that was previously broken on the remote controller).

> CommonOpts is not a good name and I don't understand the property split there.

Currently, this is because these are command line options and are entirely shared with bake. This is another thing to refactor, but it's kind of all rolled up into "how do we get bake calling the controller api" - would prefer to address this a bit later, but it's on my todo list!

controller/pb/controller.proto
jedevc: I think for now, this is *probably* alright as is? AFAICT, each invocation to the build API should be in a separate session (except where we can perform future optimizations to merge groups of requests into a single session to improve performance)?

Maybe I'm missing something about future enhancements, but IMO not worth blocking on for now.

controller/control/controller.go
```diff
@@ -6,14 +6,15 @@ import (
 
 	"github.com/containerd/console"
 	controllerapi "github.com/docker/buildx/controller/pb"
+	"github.com/moby/buildkit/client"
 )
 
 type BuildxController interface {
-	Invoke(ctx context.Context, ref string, options controllerapi.ContainerConfig, ioIn io.ReadCloser, ioOut io.WriteCloser, ioErr io.WriteCloser) error
-	Build(ctx context.Context, options controllerapi.BuildOptions, in io.ReadCloser, w io.Writer, out console.File, progressMode string) (ref string, err error)
+	Build(ctx context.Context, options controllerapi.BuildOptions, in io.ReadCloser, w io.Writer, out console.File, progressMode string) (ref string, resp *client.SolveResponse, err error)
```
tonistiigi, line 13: Why is progressmode part of controller interface.

Maybe this is semantics. Is `Controller` interface that different implementation implement (eg. direct/remote)? Or is it a high-level client interface a caller would use to invoke builds. I don't think it is the same thing.

controller/pb/controller.proto
```diff
@@ -25,30 +25,56 @@ message BuildOptions {
...
-  CommonOptions Opts = 25;
+  repeated string Platforms = 16;
+  repeated Secret Secrets = 17;
+  int64 ShmSize = 18;
+  repeated SSH SSH = 19;
+  repeated string Tags = 20;
+  string Target = 21;
+  UlimitOpt Ulimits = 22;
+
+  CommonOptions Opts = 24;
```
tonistiigi, line 49: Something needs to be done with this `Opts` inside `Opts`

tonistiigi (approved):
Comments for follow-ups

controller/control/controller.go
jedevc: cc @ktock, do you have additional context for this?

ktock: This is because remote/local controllers have different implementations for status handling (local: printed by `(controller/build).RunBuild` vs remote: printed by `(controller/remote).Client.Build()`). But I think we can now refactor this API to forward `chan *client.SolveStatus` instead of passing `progressMode` via that arg.

ktock: Opened https://github.com/docker/buildx/pull/1671

PR #1617, opened 2023-02-10
imagetools: process `com.docker.reference.*` annotations

jedevc:
See https://github.com/moby/buildkit/issues/3496 for more context.

To give us the option later down the road of producing recommended OCI names in BuildKit (using com instead of vnd, woops), we need to update Buildx to be able to process both.

Ideally, if a Buildx/BuildKit release hadn't been made we could just switch over, but since we have, we'd need to support both (at least for a while, eventually we could consider deprecating+removing the vnd variant).

We should follow this up with a BuildKit update to produce the `com` annotation (though less convinced we'd want to cherry-pick that one).

### DIFF
--- a/util/imagetools/loader.go
+++ b/util/imagetools/loader.go
@@ -21,8 +21,11 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-const (
-	annotationReference = "vnd.docker.reference.digest"
+var (
+	annotationReferences = []string{
+		"com.docker.reference.digest",
+		"vnd.docker.reference.digest", // TODO: deprecate/remove after migration to new annotation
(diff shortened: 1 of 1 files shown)


util/imagetools/loader.go
```diff
@@ -21,8 +21,11 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-const (
-	annotationReference = "vnd.docker.reference.digest"
+var (
+	annotationReferences = []string{
+		"com.docker.reference.digest",
+		"vnd.docker.reference.digest",
```
crazy-max, line 27: Can you add a TODO for later removal?

PR #1620, opened 2023-02-10
Improvements for remote controller code

jedevc:
A few commits with various improvements for the remote controller code.

- Don't kill the remote controller immediately after build. We should keep the server alive, since the client can't know exactly what's running on the controller - also we want the behavior for debugging or not to remain relatively similar. At some point in the future, we could introduce logic to the server where it can shut itself down, if no builds are currently being run, however, the server should dictate this, instead of each individual client.
- Add more informative server exit messages, so that the reasons for exit are always available in a log.
- Add signal handling for SIGTERM (in addition to the existing SIGINT). SIGKILL can be caused by pkill and similar programs, which is useful for debugging, etc.
- Switch to using `context.Context` for timeout handling with GRPC instead of a custom loop - this simplifies the code, makes it more configurable, and gives us better performance.
- Use unique files for each buildx version. Instead of only having a single controller server, we allow for having multiple, one for each version of buildx, so we should never have any version clashes, meaning user interaction is less necessary. This means we can dramatically change the protocol between buildx versions, without needing to worry about any form of compatibility.
  - We also should keep separate log files for each server - but I'm not quite sure how that should interact with a `log_file` specified in `config.toml`. Ideas for this welcome :smile:

### DIFF
--- a/controller/remote/controller.go
+++ b/controller/remote/controller.go
@@ -35,6 +35,12 @@ const (
 	serveCommandName = "_INTERNAL_SERVE"
 )
 
+var (
+	defaultLogFilename    = fmt.Sprintf("buildx.%s.log", version.Revision)
+	defaultSocketFilename = fmt.Sprintf("buildx.%s.sock", version.Revision)
+	defaultPIDFilename    = fmt.Sprintf("buildx.%s.pid", version.Revision)
+)
+
 type serverConfig struct {
 	// Specify buildx server root
 	Root string `toml:"root"`
@@ -55,7 +61,7 @@ func NewRemoteBuildxController(ctx context.Context, dockerCli command.Cli, opts 
 
 	// connect to buildx server if it is already running
 	ctx2, cancel := context.WithTimeout(ctx, 1*time.Second)
-	c, err := newBuildxClientAndCheck(ctx2, filepath.Join(serverRoot, "buildx.sock"))
+	c, err := newBuildxClientAndCheck(ctx2, filepath.Join(serverRoot, defaultSocketFilename))
 	cancel()
 	if err != nil {
 		if !errors.Is(err, context.DeadlineExceeded) {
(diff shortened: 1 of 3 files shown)


controller/remote/controller.go
```diff
@@ -35,6 +35,12 @@ const (
 	serveCommandName = "_INTERNAL_SERVE"
 )
 
+var (
+	defaultLogFilename    = fmt.Sprintf("buildx.%s.log", version.Revision)
+	defaultSocketFilename = fmt.Sprintf("buildx.%s.sock", version.Revision)
+	defaultPIDFilename    = fmt.Sprintf("buildx.%s.pid", version.Revision)
```
ktock, line 41: nit: Should be encoded not to include "/" (e.g. sha256?)

ktock (commented):
Thank you, overall LGTM

controller/remote/controller.go
jedevc: I *think* these revisions are already sha256-encoded, they're derived from the git revision.

@crazy-max does this make sense to you? Is there a different version identifier that might make more sense here / do we need to add a round of sha256 to remove special characters?

jedevc:
Added an updated commit to ensure that we take the absolute path of the buildx binary, since on exec we change the working directory to `/`, which is likely different to the working directory (so relative paths will resolve to the wrong location).

ktock (approved):
Thank you. LGTM.

controller/remote/controller.go
crazy-max: There is an edge case where `version.Revision` can be empty if not built from the Dockerfile. I think we could use the same pattern as the kubernetes driver with `contextPathHash` if that makes sense: https://github.com/docker/buildx/blob/74f64f88a7a94af52d4e64530d79bf0303524fbf/controller/build/build.go#L182-L186

crazy-max: As discussed I got confused as this is for buildx server and not build commands. We could use https://pkg.go.dev/runtime/debug#ReadBuildInfo and read `vcs.revision`.

jedevc: Added a follow-up in https://github.com/docker/buildx/pull/1637.

crazy-max (approved):

PR #1637, opened 2023-02-16
version: add fallback to fill revision from BuildInfo

jedevc:
:hammer_and_wrench: As mentioned in https://github.com/docker/buildx/pull/1620#discussion_r1108406794.

If Revision is not set by the linker, fallback to load the revision in the same format as generated by ./hack/git-meta.

Unfortunately, we have no access to the tags, so we can't have a fallback for the Version field.


version/version.go
```diff
@@ -11,3 +16,24 @@ var (
 	// the program at linking time.
 	Revision = ""
 )
+
+func init() {
+	if info, ok := debug.ReadBuildInfo(); ok {
```
crazy-max, line 21: Could we check first `Revision` is empty to avoid extra call to build info?

jedevc: Done :tada:

crazy-max (approved):
LGTM, PTAL @thaJeztah

version/version.go
```diff
@@ -1,13 +1,47 @@
 package version
 
+import (
+	"runtime/debug"
+	"strconv"
+)
+
+const (
+	defaultVersion = "0.0.0+unknown"
```
crazy-max, line 9: nit: I think we should have had a proper suffix for the default version:

```suggestion
	defaultVersion = "v0.0.0+unknown"
```

jedevc: You're right :+1:

We should also apply a similar fix in buildkit imo.

crazy-max: needs rebase since #1670

version/version.go
thaJeztah: Does this information gets used when buildx is used as a module? Because in that case it would be showing version from "whatever project used it as a module"

jedevc: :scream: ah! I think you're right, ugh.

We could try and read through the modules first, and look for buildx in there, so we don't hit that? Frustratingly, `BuildInfo.Main.Version` doesn't get populated, so we can't pull from there.

Maybe we should drop this PR, and avoid clever guessing, since using buildx as a module is  definitely a valid use case (e.g. in compose, in docker desktop, etc).

crazy-max (approved):

version/version.go
crazy-max: I think `vcs.*` is only populated at build time

jedevc: Yeah - but I think it's populated by whatever vcs information is in the build.

So, if compose uses buildx as a module, then with this patch, we'd populate `Revision` with compose's revision.

crazy-max: Ah right yeah

PR #1640, opened 2023-02-21
monitor: add `debug-shell` and `on-error`

ktock:
Related to #1104

> PR7
> docker buildx build --invoke=debug-shell should go directly to monitor mode where processes can be created with exec. We can also have docker buildx debug-shell to start monitor mode without a specific container context.

> PR8
> Add docker buildx build --invoke=on-error. In this mode, if build ends with an error debug shell will be opened from the error location. The error returned by buildkit is typed and contains the references to the state of the error and state in the beginning of the failed step. Monitor commands allow to switch in-between of these states. Error also includes source map that can be shown to the user.

## `debug-shell` subcommand

This directly starts the monitor without performing the build.
Users can also attach to other sessions (if any) and perform reload, rollback, etc... in that session.

On terminal 1:

```
$ BUILDX_EXPERIMENTAL=1 /tmp/out/buildx build --invoke sh /tmp/ctx3
```

On terminal 2:

```
$ BUILDX_EXPERIMENTAL=1 /tmp/out/buildx debug-shell
Interactive container was restarted with process "spc3i3vpqzezebdfun503tt13". Press Ctrl-a-c to switch to the new container
Switched IO
(buildx) list
ID				CURRENT_SESSION
b3ylh0tj7mzb3l8x35tdsko78	false
(buildx) attach b3ylh0tj7mzb3l8x35tdsko78
Attached to process "". Press Ctrl-a-c to switch to the new container
(buildx) ps
PID				CURRENT_SESSION	COMMAND
hf5ne0940cnlqdlo6khdn8t11	false		[sh]
(buildx) attach hf5ne0940cnlqdlo6khdn8t11
Attached to process "hf5ne0940cnlqdlo6khdn8t11". Press Ctrl-a-c to switch to the new container
(buildx) Switched IO

/ # ls /
a      dev    hi     lib    proc   sys    usr
bin    etc    home   lib64  root   tmp    var
/ # Switched IO
(buildx) rollback
invoke error: context canceled
Interactive container was restarted with process "whuvsxe4lizynxli7rhrga3jk". Press Ctrl-a-c to switch to the new container
(buildx) Switched IO
/ # ls
a      dev    hi     lib    proc   sys    usr
bin    etc    home   lib64  root   tmp    var
/ # Switched IO
(buildx) reload
invoke error: rpc error: code = Canceled desc = Canceled: exit code: 4294967295: context canceled
[+] Building 0.9s (6/6) FINISHED                                                             
 => [internal] load .dockerignore                                                       0.0s
 => => transferring context: 2B                                                         0.0s
 => [internal] load build definition from Dockerfile                                    0.0s
 => => transferring dockerfile: 87B                                                     0.0s
 => [internal] load metadata for docker.io/library/busybox:latest                       0.7s
 => [1/3] FROM docker.io/library/busybox@sha256:c118f538365369207c12e5794c3cbfb7b042d9  0.0s
 => => resolve docker.io/library/busybox@sha256:c118f538365369207c12e5794c3cbfb7b042d9  0.0s
 => CACHED [3/3] RUN echo aaaa > /a                                                     0.0s
 => CACHED [2/3] RUN echo hi > /hi                                                      0.0s
Interactive container was restarted with process "2utxvq9ij8yrwjiiqr3qtte71". Press Ctrl-a-c to switch to the new container
(buildx) Switched IO
/ # ls /
a      dev    hi     lib    proc   sys    usr
bin    etc    home   lib64  root   tmp    var
```

## `--invoke=debug-shell`

This directly starts the monitor without performing the build.
Users can also attach to other sessions (if any) and perform reload, rollback, etc... in that session.

The difference between this and `debug-shell` subcommand is that `--invoke=debug-shell` allows performing `reload` with the specified arguments of the `buildx build` and starting a new session based on that result.

```
$ BUILDX_EXPERIMENTAL=1 /tmp/out/buildx build --invoke debug-shell /tmp/ctx3
Interactive container was restarted with process "ltqny4fj99df1cgyjkvmbyyst". Press Ctrl-a-c to switch to the new container
Switched IO
(buildx) list
ID	CURRENT_SESSION
(buildx) reload
[+] Building 0.9s (6/6) FINISHED                                                             
 => [internal] load build definition from Dockerfile                                    0.0s
 => => transferring dockerfile: 87B                                                     0.0s
 => [internal] load .dockerignore                                                       0.1s
 => => transferring context: 2B                                                         0.0s
 => [internal] load metadata for docker.io/library/busybox:latest                       0.7s
 => [1/3] FROM docker.io/library/busybox@sha256:c118f538365369207c12e5794c3cbfb7b042d9  0.0s
 => => resolve docker.io/library/busybox@sha256:c118f538365369207c12e5794c3cbfb7b042d9  0.0s
 => CACHED [2/3] RUN echo hi > /hi                                                      0.0s
 => CACHED [3/3] RUN echo aaaa > /a                                                     0.0s
Interactive container was restarted with process "p3uhdcmimfhzmmpy0okrvurae". Press Ctrl-a-c to switch to the new container
(buildx) Switched IO
/ # Switched IO
(buildx) list
ID				CURRENT_SESSION
15pqy92xvlibmz66aac0doag9	true
(buildx) Switched IO

/ # ls /
a      dev    hi     lib    proc   sys    usr
bin    etc    home   lib64  root   tmp    var
```

## `--invoke=on-error`

The container is invoked based on the state of the error returned by BuildKit. 

Buildx controller saves that state and returns a typed error that contains the reference to that state and the client(monitor) uses it for starting containers/processes. For the typed error, we use the same logic as [BuildKit does in `errdefs.SolveError` etc](https://github.com/moby/buildkit/blob/v0.11.3/solver/errdefs/solve.go).

`rollback --init` command allows to start the container on the initial state of the failed step instead of the result state.

```Dockerfile
FROM busybox
RUN echo hi > /hi
RUN echo failed > /failed && echo fail > /
```

```
$ BUILDX_EXPERIMENTAL=1 /tmp/out/buildx build --invoke=on-error /tmp/ctx2
[+] Building 1.7s (7/7) FINISHED                                                             
 => [internal] load .dockerignore                                                       0.0s
 => => transferring context: 2B                                                         0.0s
 => [internal] load build definition from Dockerfile                                    0.1s
 => => transferring dockerfile: 111B                                                    0.0s
 => [internal] load metadata for docker.io/library/busybox:latest                       1.3s
 => [auth] library/busybox:pull token for registry-1.docker.io                          0.0s
 => [1/3] FROM docker.io/library/busybox@sha256:c118f538365369207c12e5794c3cbfb7b042d9  0.0s
 => => resolve docker.io/library/busybox@sha256:c118f538365369207c12e5794c3cbfb7b042d9  0.0s
 => CACHED [2/3] RUN echo hi > /hi                                                      0.0s
 => ERROR [3/3] RUN echo failed > /failed && echo fail > /                              0.2s
------
 > [3/3] RUN echo failed > /failed && echo fail > /:
#0 0.122 /bin/sh: can't create /: Is a directory
------
Interactive container was restarted with process "pk87itbolpib2frm851tdr6fq". Press Ctrl-a-c to switch to the new container
/ # ls /
bin     etc     hi      lib     proc    sys     usr
dev     failed  home    lib64   root    tmp     var
/ # cat failed
failed
/ # Switched IO
(buildx) rollback --init
invoke error: context canceled
Interactive container was restarted with process "bl6shbfd1hgnmj4syv5g8i16u". Press Ctrl-a-c to switch to the new container
(buildx) Switched IO
/ # ls
bin    etc    home   lib64  root   tmp    var
dev    hi     lib    proc   sys    usr
/ # cat failed
cat: can't open 'failed': No such file or directory
```


ktock:
Ready for review about codes but it seems that documents didn't pass CI but `make docs` doesn't fix this. How can we pass the docs CI on this PR? https://github.com/docker/buildx/actions/runs/4230634394/jobs/7348218338

commands/build.go
```diff
@@ -521,3 +549,49 @@ func (s *shmSize) Set(v string) error {
 func (s *shmSize) Type() string {
 	return s.org.Type()
 }
+
+func addDebugShellCommand(cmd *cobra.Command, dockerCli command.Cli) {
+	cmd.AddCommand(
+		debugShellCmd(dockerCli),
+	)
+}
```
crazy-max, line 557: I think the issue with `docs-upstream` workflow is because command is not added to the root one so it would look for `docker buildx build debug-shell`. I will take a closer look.

crazy-max: Is `debug-shell` supposed to be a sub command of `build`?

ktock: @crazy-max Thank you for looking at this.

> Is debug-shell supposed to be a sub command of build?

No, it' supposed to be a root command (`docker buildx debug-shell`) and it's actually implemented as a root command as described in the above PR description.

commands/root.go
```diff
@@ -89,6 +89,7 @@ func addCommands(cmd *cobra.Command, dockerCli command.Cli) {
 	)
 	if isExperimental() {
 		remote.AddControllerCommands(cmd, dockerCli)
+		addDebugShellCommand(cmd, dockerCli)
```
crazy-max, line 92: Argh doesn't seem to work. @ktock I will take a look on my side and push related changes to your branch if you're ok?

ktock: @crazy-max 

> I will take a look on my side and push related changes to your branch if you're ok?

Yes. Thank you for taking a look at this.

crazy-max: Ok it seems that's because it's a new command and we don't have the page available on docs repo: https://github.com/docker/docs/tree/main/engine/reference/commandline

I will make some changes to the `docs-upstream` workflow to handle that. Keep you posted.

crazy-max: @ktock Should be fixed with https://github.com/docker/buildx/pull/1651. Can you rebase?

ktock: Passed CI, thanks!

ktock:
CI failure seems unrelated to the change. https://github.com/docker/buildx/actions/runs/4301437650/jobs/7498699046#step:10:63

> ERROR: failed to solve: busybox: failed to do request: Head "https://registry-1.docker.io/v2/library/busybox/manifests/latest": dial tcp: lookup registry-1.docker.io on 10.43.0.10:53: read udp 10.42.0.2:36373->10.43.0.10:53: i/o timeout

ktock:
All green :tada:

ktock:
@jedevc @tonistiigi @crazy-max Rebased, PTAL :pray:

ktock:
Could we move this PR and related ones forward?

build/result.go
```diff
@@ -0,0 +1,371 @@
...
+func execOpFromError(solveErr *errdefs.SolveError) (*pb.ExecOp, error) {
+	if solveErr == nil {
+		return nil, errors.Errorf("no error is available")
+	}
+	switch op := solveErr.Solve.Op.GetOp().(type) {
+	case *pb.Op_Exec:
+		return op.Exec, nil
+	default:
+		return nil, errors.Errorf("invoke: unsupported error type")
+	}
```
jedevc, line 370: Do we want to support this on other OpTypes? e.g. if a copy fails, would we want to be able to break there?

controller/build/build.go
```diff
@@ -379,3 +382,19 @@ func controllerUlimitOpt2DockerUlimit(u *controllerapi.UlimitOpt) *dockeropts.Ul
 	}
 	return dockeropts.NewUlimitOpt(&values)
 }
+
+type ResultContextError struct {
+	ResultContext *build.ResultContext
+	error
+}
```
jedevc, line 389: Not sure I understand the purpose of this - doesn't `build.ResultContext` already contain an error type?

jedevc (commented):
Sorry, I had some pending comments and forgot to post them - I'll try and take a more in-depth look asap.

build/build.go
```diff
@@ -800,8 +774,241 @@ func Invoke(ctx context.Context, cfg ContainerConfig) error {
...
+
+	processCfg := gateway.StartRequest{
+		Args:   args,
+		Env:    env,
+		User:   user,
+		Cwd:    cwd,
+		Tty:    cfg.Tty,
+		Stdin:  cfg.Stdin,
+		Stdout: cfg.Stdout,
+		Stderr: cfg.Stderr,
```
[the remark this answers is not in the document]
ktock: refactored and introduced `populateX` funcs.

commands/build.go
```diff
@@ -400,9 +411,26 @@ func launchControllerAndRunBuild(dockerCli command.Cli, options buildOptions) er
...
 func parseInvokeConfig(invoke string) (cfg controllerapi.ContainerConfig, err error) {
 	cfg.Tty = true
-	if invoke == "default" {
+	switch invoke {
+	case "default", "debug-shell":
+		return cfg, nil
+	case "on-error":
+		// NOTE: we overwrite the command to run because the original one should fail on the failed step.
+		// TODO: make this configurable.
+		cfg.Cmd = []string{"/bin/sh"}
```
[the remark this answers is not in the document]
ktock: SGTM

commands/build.go
```diff
@@ -345,29 +346,39 @@ func launchControllerAndRunBuild(dockerCli command.Cli, options buildOptions) er
...
-	ref, err := c.Build(ctx, options.BuildOptions, pr, os.Stdout, os.Stderr, options.progress)
-	if err != nil {
-		return errors.Wrapf(err, "failed to build") // TODO: allow invoke even on error
-	}
-	if err := pw.Close(); err != nil {
-		logrus.Debug("failed to close stdin pipe writer")
-	}
-	if err := pr.Close(); err != nil {
-		logrus.Debug("failed to close stdin pipe reader")
+	if options.invoke != "debug-shell" {
```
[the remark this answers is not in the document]
ktock: refactored to avoid explicit checking

build/result.go
ktock: SGTM. It looks like that the error doesn't propagates the execution information (e.g. args,usr,cwd,envvar,...) on non-exec steps so we need to modify BuildKit maybe. We also need to think about the UI for how to make COPY's source and/or destination visible to the user.

controller/build/build.go
ktock: Added explanation of this struct as a comment in the code.

```
// ResultContextError is an error type used for passing ResultContext from this package
// to the caller of RunBuild. This is only used when RunBuild fails with an error.
// When it succeeds without error, ResultContext is returned via non-error returned value.
//
// Caller can extract ResultContext from the error returned by RunBuild as the following:
//
//	resp, res, buildErr := cbuild.RunBuild(ctx, req.Options, inR, statusChan)
//	var re *cbuild.ResultContextError
//	if errors.As(buildErr, &re) && re.ResultContext != nil {
//		res = re.ResultContext
//	}
```

jedevc:
Thanks for the quick response, sorry it's taking so long to look through thinks on my side :cry: 

I think this looks mostly good to me :tada: I do have one ask though - would it be possible to split up this PR (and related PRs) into a few separate commits (not separate PRs, just separate commits in the PR)? Especially as the scope of the debugging work grows, it's getting hard to track what's changing and what's moving where, which means it takes longer to scan through and properly review the code.

I think https://github.com/moby/buildkit/pull/3339/commits is a good example of split commits: it's a complex new feature, but the work is split down into multiple commits, each which is quite easy to test and review in isolation - the entire PR changes thousands of lines, but each commit does one single thing. The general ordering of commits tells a story as well: first there's a few bug fixes + general tidy ups, then some background work to prepare for the main changes, then the main changes, and finally some additional add-on parts of the main feature.

For example, in this PR I can spot a few things that might be good to split out:
- Using `grpcerrors` for GRPC connections
- Moving code out of invoke.go into result.go
- Making changes to the user-interface and the cli
- Changes to the monitor
- Adding the new controller.Inspect method

Let me know if that sounds alright to you :tada: (and sorry for not having raised this earlier, it probably would've helped moved things forwards a bit faster!)

ktock:
Thanks for the comment. Created separate commits for changes.

jedevc (commented):
Thanks for splitting up commits a bit more, it's significantly easier to look through a review :tada: Looks like the split has a few chunks in the wrong commits though, each commit should be self-contained and compile by itself.

Aside from that, just a couple minor comments, otherwise I'm happy :tada:

controller/build/build.go
ktock: Thanks for the comment. Fixed the code not to use ResultContextError.

build/build.go
```diff
@@ -920,7 +920,12 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 							}
 							results.Set(resultKey(dp.driverIndex, k), res)
 							if resultHandleFunc != nil {
-								resultHandleFunc(dp.driverIndex, &ResultContext{cc, res})
+								resultCtx, err := NewResultContext(cc, so, res)
```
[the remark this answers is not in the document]
ktock: Fixed.

build/result.go
```diff
@@ -0,0 +1,371 @@
...
+	case err := <-errCh:
+		return nil, err
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
+}
+
+// ResultContext is a build result with the client that built it.
+type ResultContext struct {
+	client   *client.Client
```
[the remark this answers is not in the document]
ktock: Removed unused fields.

build/result.go
```diff
@@ -0,0 +1,371 @@
...
+		}
+	}
+	return processCfg, nil
+}
+
+func containerConfigFromResult(ctx context.Context, res *gateway.Result, c gateway.Client, cfg controllerapi.InvokeConfig) (*gateway.NewContainerRequest, error) {
+	if res.Ref == nil {
+		return nil, errors.Errorf("no reference is registered")
+	}
+	if cfg.Initial {
```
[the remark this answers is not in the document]
ktock: Fixed.

jedevc (approved):

jedevc:
I would say that before we merge #1656, I'd like to see a couple follow-ups to this:

- We should support breaking on other failed commands, like COPY - atm, `on-error` stops in the right place, but I can't spawn any session. Even if we just drop into a default shell, with no extra args, environment variables, that's alright for now. This will be necessary to provide a good experience with #1656 as well.
- The commands in the monitor are getting long - we really need to split these up, out of a switch statement, and ideally into something a bit more dynamic. Something like implementations of a `monitor.Command` interface?

  As inspiration:

  ```go
  type Command interface {
      // executes the command
      Exec(args []string /* anything else we might need*/)
  
      // returns command information
      Info() Info
  }
  
  type Info struct {
      // a usage string, e.g. "rollback [--init] [entrypoint] [cmd...]"
      Usage string
  
      // a short description, e.g. "re-runs the interactive container with initial rootfs contents"
      Description string
  
      // a longer description, describes in detail what the command does, the
      // behavior of each arg, etc.
      Help string
  }
  ```

ktock:
Thank you for the suggestion. SGTM.

jedevc:
@ktock I quickly hit a couple of issues just now (though this seems to be an issue before this PR tbf):

- The second `.Build` request we create always seems to exit with context cancelled - this means the buildkit logs are full of `time="2023-04-18T12:02:37Z" level=error msg="/moby.buildkit.v1.Control/Solve returned error: rpc error: code = Canceled desc = context canceled"`. This now happens on every build, since we always spin up a second `Build` right away in `getResultAt`, instead of right away.

  Reproducer: `buildx build . --builder=dev --detach=false`

  Seemingly related to this, we don't seem to get this for the remote controller when `--detach=true` and `--sbom=true`, instead, in this case we get a BuildKit panic :cry:

  Reproducer: `buildx build . --builder=dev --detach=true --sbom=true`

  ```
  panic: runtime error: invalid memory address or nil pointer dereference [recovered]
  [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x125f688]
  
  goroutine 21695 [running]:
  go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End.func1()
	  /src/vendor/go.opentelemetry.io/otel/sdk/trace/span.go:383 +0x2a
  go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End(0xc000c3c300, {0x0, 0x0, 0x2652920?})
	  /src/vendor/go.opentelemetry.io/otel/sdk/trace/span.go:421 +0xa89
  panic({0x1697f20, 0x2624f80})
	  /usr/local/go/src/runtime/panic.go:884 +0x213
  github.com/moby/buildkit/solver/llbsolver/proc.SBOMProcessor.func1({0x1b58a08, 0xc0034f5f50}, 0xc004f9f8d0, 0xc00036a960, 0xc002729ec0)
	  /src/solver/llbsolver/proc/sbom.go:41 +0x448
  github.com/moby/buildkit/solver/llbsolver.(*Solver).Solve(0xc00036a960, {0x1b58a08, 0xc0034f5f50}, {0xc00086a720, _}, {_, _}, {0x0, 0x0, {0x0, ...}, ...}, ...)
	  /src/solver/llbsolver/solver.go:511 +0x12ad
  github.com/moby/buildkit/control.(*Controller).Solve(0xc00016cee0, {0x1b58a08, 0xc0034f5f50}, 0xc002611b80)
	  /src/control/control.go:421 +0x1208
  github.com/moby/buildkit/api/services/control._Control_Solve_Handler.func1({0x1b58a08, 0xc0034f5f50}, {0x18b48e0?, 0xc002611b80})
	  /src/api/services/control/control.pb.go:2438 +0x78
  github.com/moby/buildkit/util/grpcerrors.UnaryServerInterceptor({0x1b58a08, 0xc0034f5f50}, {0x18b48e0?, 0xc002611b80?}, 0xc00055b508?, 0x4dcff7?)
	  /src/util/grpcerrors/intercept.go:14 +0x3d
  github.com/grpc-ecosystem/go-grpc-middleware.ChainUnaryServer.func1.1.1({0x1b58a08?, 0xc0034f5f50?}, {0x18b48e0?, 0xc002611b80?})
	  /src/vendor/github.com/grpc-ecosystem/go-grpc-middleware/chain.go:25 +0x3a
  go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.UnaryServerInterceptor.func1({0x1b58960, 0xc0027a6d70}, {0x18b48e0, 0xc002611b80}, 0xc001c710c0, 0xc001c71100)
	  /src/vendor/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go:342 +0x528
  main.unaryInterceptor.func1({0x1b58a08?, 0xc0034f5e00?}, {0x18b48e0, 0xc002611b80}, 0xc001c710c0, 0xc001c71100)
	  /src/cmd/buildkitd/main.go:576 +0x1b5
  github.com/grpc-ecosystem/go-grpc-middleware.ChainUnaryServer.func1.1.1({0x1b58a08?, 0xc0034f5e00?}, {0x18b48e0?, 0xc002611b80?})
	  /src/vendor/github.com/grpc-ecosystem/go-grpc-middleware/chain.go:25 +0x3a
  github.com/grpc-ecosystem/go-grpc-middleware.ChainUnaryServer.func1({0x1b58a08, 0xc0034f5e00}, {0x18b48e0, 0xc002611b80}, 0xc002baea50?, 0x1696c00?)
	  /src/vendor/github.com/grpc-ecosystem/go-grpc-middleware/chain.go:34 +0xbe
  github.com/moby/buildkit/api/services/control._Control_Solve_Handler({0x1868080?, 0xc00016cee0}, {0x1b58a08, 0xc0034f5e00}, 0xc0029f4b60, 0xc00031bb60)
	  /src/api/services/control/control.pb.go:2440 +0x138
  google.golang.org/grpc.(*Server).processUnaryRPC(0xc0005723c0, {0x1b617e0, 0xc003fc1040}, 0xc002c7fb00, 0xc00028d500, 0x2637458, 0x0)
	  /src/vendor/google.golang.org/grpc/server.go:1336 +0xd33
  google.golang.org/grpc.(*Server).handleStream(0xc0005723c0, {0x1b617e0, 0xc003fc1040}, 0xc002c7fb00, 0x0)
	  /src/vendor/google.golang.org/grpc/server.go:1704 +0xa36
  google.golang.org/grpc.(*Server).serveStreams.func1.2()
	  /src/vendor/google.golang.org/grpc/server.go:965 +0x98
  created by google.golang.org/grpc.(*Server).serveStreams.func1
	  /src/vendor/google.golang.org/grpc/server.go:963 +0x28a
  ```

  This seems to be caused by getting an empty result? We should fix this case in buildkit, but I just want to make sure that hitting this is intended?

- This call now seems suspicious to me: https://github.com/docker/buildx/blob/ba6e5cddb043d1a36b57211fbef50db70d8d0d41/build/result.go#L80-L83. Because `Evaluate: true` is set, we don't actually get any logs until we return from the parent function. We need to make sure we keep lazy evaluation here.

jedevc:
Another thing, we seem to keep getting tripped up here:

https://github.com/docker/buildx/blob/ba6e5cddb043d1a36b57211fbef50db70d8d0d41/build/result.go#L32

If the result is empty, because we did a `buildx build --print`, or if it has multiple elements, because it had multiple `--platform`s, then we'll panic/fail.

Issue #1623, opened 2023-02-13
Improve the Git VCS collection logic


tonistiigi:
Since v0.10 https://github.com/docker/buildx/pull/1462 buildx sends basic Git info with the build request from a local directory so that builder can use it to link build result with the source. This info is Git origin address, current commit, and an indicator if the local directory is dirty.

Issues that have appeared with this:

- When using `bake`, all targets run their Git commands individually, even if they check the same directory. When using lots of targets, this results in a considerable slowdown. Warnings are also duplicated. These queries should be synchronized. Probably a simple memoized function with `flightcontrol` would do.
- The git info warnings and the TTY progressbar to not synchronize properly. I think there is a race condition. Often the behavior is that the program appears to be stuck while Git commands are running, then build starts, and only after it has completed we get a list of warnings from the very beginning of the command. This process always runs before actual buildkit requests, so it should not conflict with build progress.
- Even though we have already tried to optimize this, in bigger projects getting the "dirty checkout value" still takes considerable time. It is easy to make it take 5-10sec that could be a very considerable slowdown of the build(especially if a lot of it is cached). The program looking like it is stuck doesn't help either. I think once users figure out why this is broken for them, they will opt-out of the whole VCS feature as a whole. Considering that we don't have good ideas how to improve this speed significantly, I think we should consider making the dirty-bit opt-in. Repo URL and revision would be sent automatically as these Git requests are quick, but if the user wants a dirty bit they need to set an environment variable.

@cdupuis @crazy-max @jedevc

crazy-max:
> * When using `bake`, all targets run their Git commands individually, even if they check the same directory. When using lots of targets, this results in a considerable slowdown. Warnings are also duplicated. These queries should be synchronized. Probably a simple memoized function with `flightcontrol` would do.

Sounds good.

> The git info warnings and the TTY progressbar to not synchronize properly. I think there is a race condition. Often the behavior is that the program appears to be stuck while Git commands are running, then build starts, and only after it has completed we get a list of warnings from the very beginning of the command. This process always runs before actual buildkit requests, so it should not conflict with build progress.

Yes we have similar issues with other logs like https://github.com/docker/buildx/blob/1f73f4fd5d6da87c081f3b1b95175e3fda3a976c/build/build.go#L844 or https://github.com/docker/buildx/blob/1f73f4fd5d6da87c081f3b1b95175e3fda3a976c/build/build.go#L1705

[...]

PR #1650, opened 2023-02-24
build: makes git dirty check opt-in

crazy-max:
partially solves https://github.com/docker/buildx/issues/1623


cdupuis (approved):
It is unfortunate that we have to make this opt-in from a supply chain perspective but it is probably for the greater performance good.

tonistiigi (approved):

PR #1671, opened 2023-03-09
controller: followup refactoring on Build API

ktock:
This PR addresses the following:
- Do not pass `progressMode` to `controller.Build()` for https://github.com/docker/buildx/pull/1614#discussion_r1116091511
- Extract nested `CommonOptions` on controller API for https://github.com/docker/buildx/pull/1614#discussion_r1116093569


controller/local/controller.go
```diff
@@ -34,13 +33,13 @@ type localController struct {
...
 
-func (b *localController) Build(ctx context.Context, options controllerapi.BuildOptions, in io.ReadCloser, w io.Writer, out console.File, progressMode string) (string, *client.SolveResponse, error) {
+func (b *localController) Build(ctx context.Context, options controllerapi.BuildOptions, in io.ReadCloser, statusChan chan *client.SolveStatus) (string, *client.SolveResponse, error) {
 	if !b.buildOnGoing.CompareAndSwap(false, true) {
 		return "", nil, errors.New("build ongoing")
 	}
 	defer b.buildOnGoing.Store(false)
 
-	resp, res, err := cbuild.RunBuild(ctx, b.dockerCli, options, in, progressMode, nil)
+	resp, res, err := cbuild.RunBuild(ctx, b.dockerCli, options, in, "quiet", statusChan)
```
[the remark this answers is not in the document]
ktock: Fixed to print progress/warnings on the client side. Now `(controller/build).RunBuild` is configurable about the method to print the progress/warnings. 

```console
$ BUILDX_EXPERIMENTAL=1 /tmp/out/buildx build /tmp/ctx3
[+] Building 0.9s (6/6) FINISHED                  docker-container:testbuilder
 => [internal] load .dockerignore                                         0.0s
 => => transferring context: 2B                                           0.0s
 => [internal] load build definition from Dockerfile                      0.0s
 => => transferring dockerfile: 107B                                      0.0s
 => WARN: Empty continuation line found in: RUN echo hi > /hi     echo "  0.0s
 => [internal] load metadata for docker.io/library/busybox:latest         0.6s
 => [1/3] FROM docker.io/library/busybox@sha256:c118f538365369207c12e579  0.0s
 => => resolve docker.io/library/busybox@sha256:c118f538365369207c12e579  0.0s
 => CACHED [2/3] RUN echo hi > /hi     echo "hello"                       0.0s
 => CACHED [3/3] RUN echo aaaa > /a                                       0.0s

 - Empty continuation line found in: RUN echo hi > /hi     echo "hello"
```

ktock:
@jedevc Thank you for your time for reviewing :pray: Fixed the patch based on the comments.

commands/bake.go
```diff
@@ -29,7 +28,11 @@ type bakeOptions struct {
 	printOnly  bool
 	sbom       string
 	provenance string
-	controllerapi.CommonOptions
```
jedevc, line 32: Would it be possible to split the changes to the common options out into a separate PR so we can merge those in asap? Those changes look good to me, and then we can unblock some other stuff.

I'm not 100% sure about the progress changes here, the progress config doesn't seem like it'll catch everything on the remote side.

I *think* what we want is a way of exposing a `progress.Writer` over the network, so that the client can pass a `progress.Writer` through `controller.Build`. Then we don't also have to pass a statusChan, since we can use `progress.Tee` if we need the individual solve status events. Then everything that gets written to that progress.Writer, either on the client, the local controller, or the remote controller will end up printed (even if it's contents written directly to the progress writer by e.g. the driver loader).

jedevc: Essentially, something like: https://github.com/docker/buildx/compare/master...jedevc:buildx:ktock-progress-printer (the remote side isn't properly implemented, and just creates a dummy printer, and then `Tee`s out the channel to transmit over the connection).

We should create the progress printer outside of each controller call, and pass it into the interface. Then we can lift `printWarnings` out of the controller entirely :tada: (should make things a bit neater).

Only thing I'm not sure about is how to handle the description: https://github.com/jedevc/buildx/blob/ktock-progress-printer/controller/build/build.go#L185-L193 - we need to have this info ready when we create the progress writer, so we need to load the drivers client before calling `RunBuild` (so maybe this is blocked on doing that?).

jedevc: We could also supplement `progress.Printer` with a new `Println` method or similar, that just prints out the content after progress has completed - we can then use this to handle the `PrintFunc` case.

jedevc:
Closing this in preference of https://github.com/docker/buildx/pull/1737 and https://github.com/docker/buildx/pull/1755.

PR #1675, opened 2023-03-14
SOURCE_DATE_EPOCH build arg injection fixes

jedevc:
This PR pushes all `SOURCE_DATE_EPOCH` parsing into the command layer. This fixes two issues (in separate commits):
- For the `build` command, we pull out `SOURCE_DATE_EPOCH` early, and set it in the build-args - then in the case of the remote controller (which connects to a buildx server), we can still set `SOURCE_DATE_EPOCH` correctly.
- For the `bake` command, we pull out `SOURCE_DATE_EPOCH` early, and set it as an override - this fixes https://github.com/NixOS/nixpkgs/pull/221023#discussion_r1134601194, where if `SOURCE_DATE_EPOCH` is set in the test environment (I think this is standard on Nix, though I'm not familiar) we generate additional build args. By splitting into a separate override in the bake command, tests don't end up accidentally reading the contents of the environment.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -604,13 +604,6 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		}
 	}
 
-	// Propagate SOURCE_DATE_EPOCH from the client env
-	if v := os.Getenv("SOURCE_DATE_EPOCH"); v != "" {
-		if _, ok := so.FrontendAttrs["build-arg:SOURCE_DATE_EPOCH"]; !ok {
-			so.FrontendAttrs["build-arg:SOURCE_DATE_EPOCH"] = v
-		}
-	}
-
 	// set platforms
 	if len(opt.Platforms) != 0 {
 		pp := make([]string, len(opt.Platforms))
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -149,6 +149,19 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
 		return err
 	}
 
+	if v := os.Getenv("SOURCE_DATE_EPOCH"); v != "" {
+		// TODO: extract env var parsing to a method easily usable by library consumers
+		for _, t := range tgts {
+			if _, ok := t.Args["SOURCE_DATE_EPOCH"]; ok {
+				continue
+			}
+			if t.Args == nil {
+				t.Args = map[string]*string{}
+			}
+			t.Args["SOURCE_DATE_EPOCH"] = &v
+		}
+	}
+
 	// this function can update target context string from the input so call before printOnly check
 	bo, err := bake.TargetsToBuildOpt(tgts, inp)
 	if err != nil {
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -140,19 +140,6 @@ func ReadTargets(ctx context.Context, files []File, targets, overrides []string,
 		}
 	}
 
-	// Propagate SOURCE_DATE_EPOCH from the client env.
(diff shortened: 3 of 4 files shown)


build/build.go
```diff
@@ -604,13 +604,6 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		}
 	}
 
-	// Propagate SOURCE_DATE_EPOCH from the client env
-	if v := os.Getenv("SOURCE_DATE_EPOCH"); v != "" {
-		if _, ok := so.FrontendAttrs["build-arg:SOURCE_DATE_EPOCH"]; !ok {
-			so.FrontendAttrs["build-arg:SOURCE_DATE_EPOCH"] = v
-		}
-	}
```
crazy-max, line 612: For the case where buildx is used as lib (compose), it is not going to work anymore I guess?

jedevc: Mm you're right. I'm not quite sure if that's a bug or a feature though :thinking: 

I'm leaning towards it's probably best that when used as a library buildx should avoid reading from the environment as much as possible. We also have some minor precedent here - `DOCKER_DEFAULT_PLATFROM`.

Though I wonder about cases like `BUILDX_NO_DEFAULT_ATTESTATIONS`. On the one hand, it makes sense to propagate them, on the other, it could be confusing to the library consumer if the environment overrides the options that they have passed to Buildx.

IMO, in the future we should push all environment variables that affect the build solve request (so not including buildx driver env vars, etc) into the command layer, and then we should provide a utility method to parse environment variables and modify the build options from them:

https://github.com/docker/buildx/blob/4a73abfd645457ee8819855cb1a337ab23f833dc/controller/pb/controller.proto#L47-L73

Does that seem somewhat reasonable? If so, I'll add a TODO in the code - it's tricky to handle right now, since I think with the controller code, I'd expect we'd try and get library consumers of Buildx to use that API instead of using `build.Build`?

crazy-max: SGTM :+1:

crazy-max (approved):

build/build.go
thaJeztah: ./cc FYI @glours @ndeloof (for compose)

commands/bake.go
```diff
@@ -84,6 +84,10 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
 	if in.provenance != "" {
 		overrides = append(overrides, fmt.Sprintf("*.attest=%s", buildflags.CanonicalizeAttest("provenance", in.provenance)))
 	}
+	if v := os.Getenv("SOURCE_DATE_EPOCH"); v != "" {
+		// TODO: extract env var parsing to a method easily usable by library consumers
+		overrides = append(overrides, fmt.Sprintf("*.args.SOURCE_DATE_EPOCH=%s", v))
```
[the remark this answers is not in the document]
jedevc: While this does change the previous behavior - I think the override behavior makes more sense (if we'd be ok with making a breaking change). That way, we can have all user-provided options that aren't in the file, override the file (instead of having different behavior for environment variables or command line options).

If we don't want to break, I can find another way to capture and push the env vars down.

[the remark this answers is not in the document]
tonistiigi: I don't think `SOURCE_DATE_EPOCH` is explicitly "override what I have in the file". It is more like "auto-load from the current environment because I can't find any other value". If the user wants override behavior, they can use `--set` or define a variable block in the bake file.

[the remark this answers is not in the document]
crazy-max: Ah right that makes sense for the override behavior with bake.

AkihiroSuda (approved):

tonistiigi (approved):

PR #1680, opened 2023-03-15
Use controller option in bake

jedevc:
:spiral_notepad: Marking as draft until we merge https://github.com/docker/buildx/pull/1671.

This PR continues the move to attempt to merge the `build.Options` struct into the `controllerapi.Options` message.
    
To do this, we extract all the input parameters into a dedicated message (adding the missing ones, except for the `InStream` parameter which will require some additional fiddling around). We also rework the `NamedContexts` to allow containing States (by transmitting them as `DefinitionOps`), and adding a `Linked` field to the common options.

With these changes, we can have bake generate `controllerapi.BuildOptions`, and then convert those to `build.Options` using the `controller/build` package.
    
This is an intermediate patch, designed to allow us to clean up some shared logic between both build and bake. The next step will be to modify bake to use the controller API directly, and completely skip the `build.Options` generation step.


controller/build/build.go
```diff
@@ -42,41 +43,10 @@ import (
...
-		return nil, nil, errors.Errorf("--no-cache and --no-cache-filter cannot currently be used together")
-	}
-
-	contexts := map[string]build.NamedContext{}
-	for name, path := range in.NamedContexts {
-		contexts[name] = build.NamedContext{Path: path}
-	}
-
-	printFunc, err := parsePrintFunc(in.PrintFunc)
+	opts, err := ToBuildOpts(in, inStream)
```
ktock, line 46: It seems that some fields (e.g. Exports) are processed twice (in ToBuild and in RunBuild), can we try to dedupe them?

ktock (commented):
Thank you for working on it.

ktock:
Thanks, LGTM.

controller/pb/controller.proto
```diff
@@ -118,6 +131,7 @@ message CommonOptions {
   bool Pull = 4;
   bool ExportPush = 5;
   bool ExportLoad = 6;
+  bool Linked = 7;
```
crazy-max, line 134: `Linked` is internal. I'm a bit on the fence to have it in the proto def: https://github.com/docker/buildx/blob/4a73abfd645457ee8819855cb1a337ab23f833dc/build/build.go#L93-L94

jedevc: Agreed - but based on the current usage of it, I think we need to. It's used in `BuildWithResultHandler` which will be called from the controller side, while it's generated from targets before we make the controller request.

I think it's only used here? https://github.com/docker/buildx/blob/4a73abfd645457ee8819855cb1a337ab23f833dc/build/build.go#L688-L690

I'm quite sure how else we can do it, though open to suggestions :thinking:

crazy-max: Ah right let's keep then :+1:

crazy-max (approved):

jsternberg:
I'm going to close this because we deleted the controller in buildx so I don't think this pull request is relevant anymore.

PR #1681, opened 2023-03-15
docs: update since buildinfo removal

crazy-max:
follow-up https://github.com/moby/buildkit/pull/3582

Update docs since buildinfo has been removed from BuildKit. We need to keep some buildinfo logic in buildx for backward compat with old builders. Such as:

https://github.com/docker/buildx/blob/4a73abfd645457ee8819855cb1a337ab23f833dc/build/build.go#L559-L562

~Keeping in draft until BuildKit 0.12.~


jedevc (approved):
:tada:

dvdksn (approved):

Issue #1312, opened 2022-09-11
Proposal: target pattern rule with bake


crazy-max:
Let's take the following definition with repeated and almost similar targets:

```hcl
group "validate" {
  targets = ["validate-lint", "validate-vendor", "validate-docs", "validate-authors"]
}

target "validate-lint" {
  dockerfile = "./hack/dockerfiles/lint.Dockerfile"
  output = ["type=cacheonly"]
}

target "validate-vendor" {
  dockerfile = "./hack/dockerfiles/vendor.Dockerfile"
  target = "validate"
  output = ["type=cacheonly"]
}


target "validate-docs" {
  dockerfile = "./hack/dockerfiles/docs.Dockerfile"
  target = "validate"
  output = ["type=cacheonly"]
}

target "validate-authors" {
  dockerfile = "./hack/dockerfiles/authors.Dockerfile"
  target = "validate"
  output = ["type=cacheonly"]
}
```

A common pattern with bake is the ability to group and run targets in parallel. Atm we don't have a nice way of defining implicit target by using a _pattern rule_ like [GNU Make](https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html).

A target pattern contains the character `%` (exactly one of them) in the target; otherwise, it looks exactly like an ordinary target. The `%` matches any nonempty substring, while other characters match only themselves. A target pattern is composed of a `%` between a prefix and a suffix, either or both of which may be empty. The text between the prefix and the suffix is called the _stem_. Thus, when the pattern `validate-%` matches the target name `validate-vendor`, the stem is `vendor`.

Like [GNU Make automatic vairables](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html#Automatic-Variables) we can also have values computed afresh for each rule that is executed, based on the target and prerequisites of the rule within its own recipe. By taking the example above we could then write:

```hcl
group "validate" {
  targets = ["validate-lint", "validate-vendor", "validate-docs", "validate-authors"]
}

target "validate-%" {
  dockerfile = "./hack/dockerfiles/${*}.Dockerfile"
  target = "${*}" != "lint" ? "validate" : ""
  output = ["type=cacheonly"]
}
```

* `${*}` defines the _stem_ with which an implicit rule matches as explained above. (e.g. for `validate-vendor` would be `vendor`)
* `${@}` defines the target name. (e.g. for `validate-vendor` would be `validate-vendor`)

tonistiigi:
Seems to be same issue as #1150

PR #1690, opened 2023-03-21
Implement matrix for bake targets

### LINKED ISSUE #1150: Bake; placeholders or templating targets?
I have a job that builds a matrix of PHP versions, base distros (and some other stuff IRL). I want to combine it into one bake, but still source the versions from env vars. I'm not sure if there's a way to do this in bake, or if I need to generate HCL/JSON separately?

 I could generate the default group pretty easily if I set an env var like `PHP_VERSIONS=7,8 DISTROS =alpine,ubuntu`:

```hcl
variable PHP_VERSIONS {}
variable DISTROS {}

group "default" {
  targets = flatten(
    [for d in split(",", "${DISTROS}"):
      [for v in split(",", "${PHP_VERSIONS}"):
        "${d}-php-${v}"
      ]
    ]
  )
}
```

However I can't see how to avoid explicitly enumerating all the targets:
```hcl
target "alpine-php-8" {}
[...etc...]
```
 Ideally I'd have some way to template the targets, is there anything possible in bake's HCL?

jedevc:
:hammer: Fixes #1150 and #1312

This PR implements the proposal from https://github.com/docker/buildx/issues/1150#issuecomment-1146204434, setting a matrix property on targets which can be used to easily parameterize builds such as https://github.com/docker/buildx/issues/1312#issue-1369020589:

```hcl
group "validate" {
  targets = ["validate-lint", "validate-vendor", "validate-docs", "validate-authors"]
}

target "validate-lint" {
  dockerfile = "./hack/dockerfiles/lint.Dockerfile"
  output = ["type=cacheonly"]
}

target "validate-vendor" {
  dockerfile = "./hack/dockerfiles/vendor.Dockerfile"
  target = "validate"
  output = ["type=cacheonly"]
}


target "validate-docs" {
  dockerfile = "./hack/dockerfiles/docs.Dockerfile"
  target = "validate"
  output = ["type=cacheonly"]
}

target "validate-authors" {
  dockerfile = "./hack/dockerfiles/authors.Dockerfile"
  target = "validate"
  output = ["type=cacheonly"]
}
```

With this PR, the above example can be simplified to:

```hcl
group "validate" {
  targets = ["validate-lint", "validate-vendor", "validate-docs", "validate-authors"]
}

target "validate" {
  matrix = {
    tgt = ["lint", "vendor", "docs", "authors"]
  }
  name = "validate-${tgt}"
  dockerfile = "./hack/dockerfiles/${tgt}.Dockerfile"
  target = "validate"
  output = ["type=cacheonly"]
}
```

See the added tests for more information on the functionality - I'll add some docs once it's clear there's some consensus on the design + implementation.

There's still some work to be done around the matrix feature, though I'd be happy to do them as follow-ups, I think there's enough functionality to review in this PR. Specifically, we should consider:
- :heavy_check_mark: Can we provide a shortcut to easily refer to all targets from a matrix without needing to manually specify them (see in the above example, where the `["lint", "vendor", "docs", "authors"]` is effectively duplicated, once in the matrix definition and once in the group definition)? We could implicitly turn the old label from the target (`validate` in this case) into a group that contains all the targets (so the `validate` group would be automatically created). This feels like the simplest option, but I'm not sure about the semantics of automatically creating groups :thinking: 
- Should we allow for a GitHub-matrix like syntax, to `include` or `exclude` specific matrix combinations? I think this would be useful, though I'm not sure of the exact syntax we'd want for this.
- Should we support matrixes for groups as well? There's no reason why not, and adds consistency, but I struggle to see a concrete use case.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -246,13 +246,28 @@ func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error)
 	}
 
 	if len(hclFiles) > 0 {
-		if err := hclparser.Parse(hcl.MergeFiles(hclFiles), hclparser.Opt{
+		renamed, err := hclparser.Parse(hcl.MergeFiles(hclFiles), hclparser.Opt{
 			LookupVar:     os.LookupEnv,
 			Vars:          defaults,
 			ValidateLabel: validateTargetName,
-		}, &c); err.HasErrors() {
+		}, &c)
+		if err.HasErrors() {
 			return nil, err
 		}
+
+		for _, renamed := range renamed {
+			for oldName, newNames := range renamed {
+				newNames = dedupSlice(newNames)
+				if len(newNames) == 1 && oldName == newNames[0] {
+					continue
+				}
+				c.Groups = append(c.Groups, &Group{
+					Name:    oldName,
+					Targets: newNames,
+				})
+			}
+		}
+		c = dedupeConfig(c)
 	}
 
 	return &c, nil
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -24,6 +24,7 @@ import (
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/convert"
 )
 
 var (
@@ -612,6 +613,11 @@ type Target struct {
 	linked bool
 }
 
+var _ hclparser.WithEvalContexts = &Target{}
+var _ hclparser.WithGetName = &Target{}
+var _ hclparser.WithEvalContexts = &Group{}
+var _ hclparser.WithGetName = &Group{}
+
 func (t *Target) normalize() {
 	t.Attest = removeDupes(t.Attest)
 	t.Tags = removeDupes(t.Tags)
@@ -795,7 +801,20 @@ func (t *Target) AddOverrides(overrides map[string]Override) error {
 	return nil
 }
 
-func (t *Target) EvalContexts(ectx *hcl.EvalContext, block *hcl.Block, loadDeps func(hcl.Expression) hcl.Diagnostics) ([]*hcl.EvalContext, error) {
+func (g *Group) GetEvalContexts(ectx *hcl.EvalContext, block *hcl.Block, loadDeps func(hcl.Expression) hcl.Diagnostics) ([]*hcl.EvalContext, error) {
+	content, _, err := block.Body.PartialContent(&hcl.BodySchema{
+		Attributes: []hcl.AttributeSchema{{Name: "matrix"}},
+	})
+	if err != nil {
+		return nil, err
+	}
+	if _, ok := content.Attributes["matrix"]; ok {
+		return nil, errors.Errorf("matrix is not supported for groups")
+	}
+	return []*hcl.EvalContext{ectx}, nil
+}
+
+func (t *Target) GetEvalContexts(ectx *hcl.EvalContext, block *hcl.Block, loadDeps func(hcl.Expression) hcl.Diagnostics) ([]*hcl.EvalContext, error) {
 	content, _, err := block.Body.PartialContent(&hcl.BodySchema{
 		Attributes: []hcl.AttributeSchema{{Name: "matrix"}},
 	})
@@ -841,6 +860,53 @@ func (t *Target) EvalContexts(ectx *hcl.EvalContext, block *hcl.Block, loadDeps 
 		ectxs = ectxs2
 	}
 	return ectxs, nil
+}
+
+func (g *Group) GetName(ectx *hcl.EvalContext, block *hcl.Block, loadDeps func(hcl.Expression) hcl.Diagnostics) (string, error) {
+	content, _, diags := block.Body.PartialContent(&hcl.BodySchema{
+		Attributes: []hcl.AttributeSchema{{Name: "name"}, {Name: "matrix"}},
+	})
+	if diags != nil {
+		return "", diags
+	}
+
+	if _, ok := content.Attributes["name"]; ok {
+		return "", errors.Errorf("name is not supported for groups")
+	}
+	if _, ok := content.Attributes["matrix"]; ok {
+		return "", errors.Errorf("matrix is not supported for groups")
+	}
+	return block.Labels[0], nil
+}
+
+func (t *Target) GetName(ectx *hcl.EvalContext, block *hcl.Block, loadDeps func(hcl.Expression) hcl.Diagnostics) (string, error) {
+	content, _, diags := block.Body.PartialContent(&hcl.BodySchema{
+		Attributes: []hcl.AttributeSchema{{Name: "name"}, {Name: "matrix"}},
+	})
+	if diags != nil {
+		return "", diags
+	}
+
+	attr, ok := content.Attributes["name"]
+	if !ok {
+		return block.Labels[0], nil
+	}
+	if _, ok := content.Attributes["matrix"]; !ok {
+		return "", errors.Errorf("name requires matrix")
+	}
+	if diags := loadDeps(attr.Expr); diags.HasErrors() {
+		return "", diags
+	}
+	value, diags := attr.Expr.Value(ectx)
+	if diags != nil {
+		return "", diags
+	}
+
+	value, err := convert.Convert(value, cty.String)
+	if err != nil {
+		return "", err
+	}
+	return value.AsString(), nil
 }
 
 func TargetsToBuildOpt(m map[string]*Target, inp *Input) (map[string]build.Options, error) {
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -23,6 +23,7 @@ import (
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/pkg/errors"
+	"github.com/zclconf/go-cty/cty"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


crazy-max:
> * Can we provide a shortcut to easily refer to all targets from a matrix without needing to manually specify them (see in the above example, where the `["lint", "vendor", "docs", "authors"]` is effectively duplicated, once in the matrix definition and once in the group definition)? We could implicitly turn the old label from the target (`validate` in this case) into a group that contains all the targets (so the `validate` group would be automatically created). This feels like the simplest option, but I'm not sure about the semantics of automatically creating groups 🤔

I don't think it's related to the matrix proposal but https://github.com/docker/buildx/issues/1312 to use target pattern rules with _stem_:

```hcl
group "validate" {
  targets = ["validate-lint", "validate-vendor", "validate-docs", "validate-authors"]
}

target "validate-%" {
  dockerfile = "./hack/dockerfiles/${*}.Dockerfile"
  target = "${*}" != "lint" ? "validate" : ""
  output = ["type=cacheonly"]
}
```

tonistiigi:
@ciaranmcnulty @cpuguy83

tonistiigi:
> Can we provide a shortcut to easily refer to all targets from a matrix without needing to manually specify them 

Yes, please. There should not be a need to define an extra group with the same matrix values repeated as targets. When calling a target that defines matrix, all matrix combinations are built.

jedevc:
> > Can we provide a shortcut to easily refer to all targets from a matrix without needing to manually specify them
> 
> Yes, please. There should not be a need to define an extra group with the same matrix values repeated as targets. When calling a target that defines matrix, all matrix combinations are built.

Done in 080687026111e6a03bc5d1030d9d231db03f74d5.

For the input:

```hcl
# no separate group necessary!

target "validate" {
  matrix = {
    tgt = ["lint", "vendor", "docs", "authors"]
  }
  name = "validate-${tgt}"
  dockerfile = "./hack/dockerfiles/${tgt}.Dockerfile"
  target = "${tgt}" != "lint" ? "validate" : ""
  output = ["type=cacheonly"]
}
```

```
$ buildx bake --print validate
[+] Building 0.0s (0/0)                                                                                                                                           
{
  "group": {
    "default": {
      "targets": [
        "validate"
      ]
    },
    "validate": {
      "targets": [
        "validate-lint",
        "validate-vendor",
        "validate-docs",
        "validate-authors"
      ]
    }
  },
  "target": {
    "validate-authors": {
      "context": ".",
      "dockerfile": "./hack/dockerfiles/authors.Dockerfile",
      "target": "validate",
      "output": [
        "type=cacheonly"
      ]
    },
    "validate-docs": {
      "context": ".",
      "dockerfile": "./hack/dockerfiles/docs.Dockerfile",
      "target": "validate",
      "output": [
        "type=cacheonly"
      ]
    },
    "validate-lint": {
      "context": ".",
      "dockerfile": "./hack/dockerfiles/lint.Dockerfile",
      "target": "",
      "output": [
        "type=cacheonly"
      ]
    },
    "validate-vendor": {
      "context": ".",
      "dockerfile": "./hack/dockerfiles/vendor.Dockerfile",
      "target": "validate",
      "output": [
        "type=cacheonly"
      ]
    }
  }
}
```

If the group is defined separately, it will be correctly merged with the previous target - the underlying representation is a little, but I think it's what the user expects.

bake/hcl_test.go
```diff
@@ -634,6 +634,348 @@ func TestHCLMultiFileAttrs(t *testing.T) {
...
+		switch g.Name {
+		case "x":
+			require.Equal(t, []string{"a", "b"}, g.Targets)
+		case "y":
+			require.Equal(t, []string{"c", "d"}, g.Targets)
+		}
+	}
+}
+
+func TestHCLMatrixArgs(t *testing.T) {
```
tonistiigi, line 927: Add a test for

```
variable "ABC" {
  default = "def"
}

target "bar" {
  matrix = {
   aa = split(",", ABC)
  }
  name = "bar-${aa}"
  args = {
    foo = aa
  }
}
```
`ABC=11,22,33 docker buildx bake`

bake/bake.go
```diff
@@ -245,13 +246,28 @@ func ParseFiles(files []File, defaults map[string]string) (_ *Config, err error)
...
 			LookupVar:     os.LookupEnv,
 			Vars:          defaults,
 			ValidateLabel: validateTargetName,
-		}, &c); err.HasErrors() {
+		}, &c)
+		if err.HasErrors() {
 			return nil, err
 		}
+
+		for _, renamed := range renamed {
```
tonistiigi, line 258: Does this need to be outside of the `hclparser` pkg? Why not just do it as part of `Parse()` ?

ciaranmcnulty:
My main use case for this sort of thing is this sort of thing:

`PHP_VERSIONS=81,82 FLAVOURS=debian,alpine docker buildx bake`

It looks good for the basic cases 👍

bake/hcl_test.go
```diff
@@ -634,6 +634,348 @@ func TestHCLMultiFileAttrs(t *testing.T) {
...
+
+	require.Equal(t, 1, len(c.Targets))
+	require.Equal(t, "app", c.Targets[0].Name)
+	require.Equal(t, "y", *c.Targets[0].Dockerfile)
+}
+
+func TestHCLRenameTarget(t *testing.T) {
+	dt := []byte(`
+		target "abc" {
+			name = "xyz"
```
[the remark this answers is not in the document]
crazy-max: I agree, like GHA for example if we take this example:

```hcl
target "validate" {
  matrix = {
    tgt = ["lint", "vendor", "docs", "authors"]
  }
  dockerfile = "./hack/dockerfiles/${tgt}.Dockerfile"
  target = "${tgt}" != "lint" ? "validate" : ""
  output = ["type=cacheonly"]
}
```

`name` would be by default `validate-${tgt}`.

For:

```hcl
target "validate" {
  matrix = {
    tgt = ["lint", "vendor", "docs", "authors"]
    arg = ["bar", "baz"]
  }
  dockerfile = "./hack/dockerfiles/${tgt}.Dockerfile"
  target = "${tgt}" != "lint" ? "validate" : ""
  output = ["type=cacheonly"]
  args = {
    myarg = arg
  }
}
```

`name` would be by default `validate-${tgt}-${arg}`.

So maybe we don't need to expose `name` at all.

bake/bake.go
jedevc: Yeah, this is unfortunately required. The `bake/hclparser` package parses into a generic struct type, so doesn't actually have any knowledge of `Target`s or `Group`s, which theoretically would make it reusable in other places if we ever wanted to (or would make adding a new block type simple).

Only the `bake` package actually has knowledge of the `Target` and `Group`, and how they interact, so we need the logic here. :cry:

jedevc:
> As it's generating the product of all the lists in matrix, is there any way to remove an item from the resulting generated targets?

Yeah, I think we probably will want that, see my second point:

> Should we allow for a GitHub-matrix like syntax, to include or exclude specific matrix combinations? I think this would be useful, though I'm not sure of the exact syntax we'd want for this.

I'm not *quite* sure on syntax, but it's also definitely a thing for a follow-up PR. The easiest suggestion that springs to mind:

```hcl
target "validate" {
  matrix = {
    tgt = ["lint", "vendor", "docs", "authors"]
  }
  matrix-include = [
    {tgt = "another-option"}  # this option is added to the list of matrix combinations 
  ]
  matrix-exclude = [
    {tgt = "authors"}  # this option is removed from the list of matrix combinations
  ]

  name = "validate-${tgt}"
  dockerfile = "./hack/dockerfiles/${tgt}.Dockerfile"
  target = "${tgt}" != "lint" ? "validate" : ""
  output = ["type=cacheonly"]
}
```

It feels not great though, maybe it would make sense to group all of them under one single property, so that `matrix.args` is the block that actually contains the parameters, and then we could have `matrix.include` and `matrix.exclude`?

I'm not a huge fan of how [GitHub mixes the `include` and `exclude` fields into the top-level](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs), it feels unintuitive to me, but I could be convinced otherwise.

ciaranmcnulty:
It seems to me it's trivial to _add_ to the matrix by simply defining some explicit additional targets that don't use this new mechanism

bake/hcl_test.go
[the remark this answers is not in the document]
tonistiigi: I don't think we should remove the name(or try to generate it); validation is enough. See the examples above with non-string matrix values. You can't automatically determine what the name should be for these cases.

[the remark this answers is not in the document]
jedevc: I've updated with some additional constraints:

- Explicit renaming with the `name` property is *only* permitted when used with the `matrix` property.
- Group does not support either `name` or `matrix` (we may choose to relax this constraint over time if we get a valid use case).
- All generated names must be unique.

I'm not convinced over `name` should contain all `matrix` variables, but all the others sound good to me.

jedevc:
As another option instead of having a way of `exclude`ing, we could potentially at some point in the future add another target option: `skip`. If `skip` is true, then we don't include it in the matrix, if it's false (the default), it is included.

For example:

```hcl
target "test" {
  matrix = {
    driver = ["docker", "docker-container", "kubernetes", "remote"]
    test = ["a", "b", "c"]
  }
  name = "test-${driver}-${test}"
  skip = driver == "kubernetes" && test == "c"
}
```

I'd prefer something like this, instead of another `matrix` property of needing to nest them if possible - but let's hold off on adding anything to resolve this in the current PR.

bake/hclparser/hclparser.go
```diff
@@ -49,29 +51,38 @@ type parser struct {
...
-	progressB map[*hcl.Block]map[string]struct{}
-	doneF     map[string]struct{}
-	doneB     map[*hcl.Block]map[string]struct{}
+	progressV map[uint64]struct{}
+	progressF map[uint64]struct{}
+	progressB map[uint64]map[string]struct{}
+	doneB     map[uint64]map[string]struct{}
+}
+
+type internalGetEvalContexts interface {
```
tonistiigi, line 68: This should be public if we expect caller to implement it. And type checked with `var _ WithEvalContexts = &Group{}`

jedevc: Done. I've type checked for Group + Target for both the interfaces.

tonistiigi (approved):
LGTM but could you update docs as well

jedevc:
We've moved the file definition into https://github.com/docker/docs, so I'll follow up there, cc @dvdksn.

jedevc:
:tada: Started the docs work in https://github.com/docker/docs/pull/17041.

PR #1700, opened 2023-03-28
[v0.10] Fix attestation overrides with duplicate `type`s

jedevc:
:cherries: Cherry-pick of #1699 - not a clean pick because of #1614.

This PR:
- Ports the tests from #1699 (changes required because keys in the map are slightly different and have `attest:` prefix).
- Changes the magic internal value used by attestations to `disabled` (instead `enabled`), which allows us to keep the same tests as in #1699. This behavior is not a breaking change, since the value is entirely undocumented and internal - it should not be used by external users.
- Adds the merging behavior from #1699, a different implementation is required because of changes to the function signature from #1614.

### DIFF
--- a/util/buildflags/attests.go
+++ b/util/buildflags/attests.go
@@ -30,9 +30,16 @@ func ParseAttests(in []string) (map[string]*string, error) {
 
 		k := "attest:" + attestType
 		if _, ok := out[k]; ok {
-			return nil, errors.Errorf("duplicate attestation field %s", attestType)
+			if disabled == nil {
+				return nil, errors.Errorf("duplicate attestation field %s", attestType)
+			}
+			if *disabled {
+				out[k] = nil
+			}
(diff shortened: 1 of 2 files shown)


util/buildflags/attests.go
```diff
@@ -23,57 +23,65 @@ func ParseAttests(in []string) (map[string]*string, error) {
...
+			return "", nil, errors.Errorf("invalid value %s", field)
 		}
 		key = strings.TrimSpace(strings.ToLower(key))
 
 		switch key {
 		case "type":
 			attestType = value
-		case "enabled":
-			enabled, err = strconv.ParseBool(value)
+		case "disabled":
```
crazy-max, line 74: This change is fine by me, as you said it's internal. PTAL @tonistiigi.

crazy-max (approved):

PR #1702, opened 2023-03-29
build: remove buildinfo

jedevc:
:arrow_up: Follow up https://github.com/moby/buildkit/pull/3582.

BuildKit v0.11 has deprecated buildinfo, and BuildKit v0.12 will remove it, so we should remove it from Buildx as well.

We may want to hold on merging this until we vendor the next release of BuildKit - not exactly sure on release timing.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -555,12 +555,6 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 				}
 			}
 		}
-		if e.Type == "docker" || e.Type == "image" || e.Type == "oci" {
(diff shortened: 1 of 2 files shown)


jedevc:
woops, have just seen https://github.com/docker/buildx/pull/1681 :facepalm:

thaJeztah:
> woops, have just seen #1681 🤦

Looks like this PR removes things that are not removed in that other PR? Should those still be removed?

jedevc:
Yeah, we shouldn't remove them, we still need to keep doing them :facepalm:

PR #1177, opened 2022-06-21
build: print instance being used

crazy-max:
~needs https://github.com/moby/buildkit/pull/3378~

Display current builder being used when building. Can help users who need to know the type of builder in use during their build but also for log reports:

```
$ docker buildx build .
[+] Building 4.8s (11/16)                                                               docker-container:builder2
 => [internal] load build definition from Dockerfile                                                         0.1s
 => => transferring dockerfile: 2.93kB                                                                       0.1s 
 => [internal] load .dockerignore                                                                            0.1s 
 => => transferring context: 45B                                                                             0.1s 
 => resolve image config for docker.io/docker/dockerfile-upstream:1.5.0                                      0.6s
 => CACHED docker-image://docker.io/docker/dockerfile-upstream:1.5.0@sha256:bda6ac9f61f2b676331acfd656a07bc  0.0s
 => => resolve docker.io/docker/dockerfile-upstream:1.5.0@sha256:bda6ac9f61f2b676331acfd656a07bcd55b369143a  0.0s 
 => [internal] load metadata for docker.io/tonistiigi/xx:1.1.2                                               2.4s 
 => [internal] load metadata for docker.io/library/golang:1.19-alpine                                        2.4s 
 => [auth] library/golang:pull token for registry-1.docker.io                                                0.0s
 => [auth] tonistiigi/xx:pull token for registry-1.docker.io                                                 0.0s
 => [xx 1/1] FROM docker.io/tonistiigi/xx:1.1.2@sha256:9dde7edeb9e4a957ce78be9f8c0fbabe0129bf5126933cd35748  0.0s
 => => resolve docker.io/tonistiigi/xx:1.1.2@sha256:9dde7edeb9e4a957ce78be9f8c0fbabe0129bf5126933cd3574888f  0.0s 
 => CANCELED [internal] load build context                                                                   1.5s 
 => => transferring context: 20.57kB                                                                         1.5s 
 => [golatest 1/1] FROM docker.io/library/golang:1.19-alpine@sha256:2381c1e5f8350a901597d633b2e517775eeac7a  0.1s 
 => => resolve docker.io/library/golang:1.19-alpine@sha256:2381c1e5f8350a901597d633b2e517775eeac7a6682be392  0.0s 
ERROR: failed to solve: Canceled: context canceled
```

```
$ BUILDKIT_PROGRESS=plain docker buildx build .
#0 building with "builder2" instance using docker-container driver

#1 [internal] load build definition from Dockerfile
[...]

PR #1703, opened 2023-03-29
Bump BuildKit to latest master

jedevc:
The latest BuildKit has removed some build-info fields, so we should update it to the latest master, to prevent drift.

Hopefully this is alright for vendoring? I noticed we already have a non-released version of BuildKit vendored in master, for https://github.com/docker/buildx/pull/1177, so at least this doesn't make things more complex than they already were :scream: (and if we're vendoring off of master, at least better to be explicit about it, and to stay up-to-date, similar to https://github.com/docker/buildx/pull/1198#issue-1295483359).


go.mod
```diff
@@ -31,23 +31,23 @@ require (
...
+	google.golang.org/grpc v1.53.0
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api v0.25.4
-	k8s.io/apimachinery v0.25.4
-	k8s.io/apiserver v0.25.4
-	k8s.io/client-go v0.25.4
+	k8s.io/api v0.26.2
+	k8s.io/apimachinery v0.26.2
+	k8s.io/apiserver v0.26.2
+	k8s.io/client-go v0.26.2
```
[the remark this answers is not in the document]
jedevc: I have just for `client-go` in a `replace` at the bottom of the file - is that what you mean?

cmd/buildx/main.go
```diff
@@ -28,7 +27,6 @@ import (
 )
 
 func init() {
-	seed.WithTimeAndRand()
```
crazy-max, line 31: I think like BuildKit we need to keep it for now: https://github.com/moby/buildkit/blob/eb3c9d44366e1dbb1de4768d271686b69dd43201/cmd/buildkitd/main.go#L77-L78

@AkihiroSuda Is there a better migration path for this?

jedevc: I don't think it's required - it only calls `math`s `rand.Seed`, and we only use that in one place, where we now explicitly construct it for the pod chooser in the kubernetes driver.

AkihiroSuda: Vendored dependencies might be depending on `rand.Seed` too?

jedevc: Ah good point, I'll keep this in for now then - looks like some of the kubernetes code uses it for doing random backoff, and similar.

Issue #1708, opened 2023-03-30
Consolidate logic for detection of remote git urls


jedevc:
:bug: Moves from https://github.com/moby/buildkit/issues/3755

We have a few locations of testing if a URL is a valid Git URL:

https://github.com/docker/buildx/blob/33388d6edef7aba7e51c258f86f1091e427fd485/bake/remote.go#L86-L94

https://github.com/docker/buildx/blob/33388d6edef7aba7e51c258f86f1091e427fd485/bake/bake.go#L1043-L1045

https://github.com/docker/buildx/blob/33388d6edef7aba7e51c258f86f1091e427fd485/build/build.go#L1349

However, since https://github.com/moby/buildkit/commit/8bfeafaaa1c184eecb67ff012cdd6044bc3fa738, the logic for parsing Git URLs has changed, and the corresponding logic in Buildx has not been updated - we should use these BuildKit helpers.

`gitutil.ParseGitRef` (from buildkit) does *not* produce the same results as `urlutil.IsGitURL` (from moby):

- `gitutil.ParseGitRef` does not recognize slashes right after the name such as `git@github.com/docker/buildx.git` (note `/docker` instead of `:docker`), while `urlutil.IsGitURL` (this *use* to work before https://github.com/moby/buildkit/commit/8bfeafaaa1c184eecb67ff012cdd6044bc3fa738, I have no idea why or how though). cc @AkihiroSuda do you know why this behavior would have broken after that commit? I can't identify what part caused the issue :eyes:
- `gitutil.ParseGitRef` recognize `git://` urls such as `ssh://git@github.com:docker/buildx.git`, while `urlutil.IsGitURL` does not seem to.

Because of the disparity in these functions, buildx will incorrectly identify URLs as git-urls, and buildkit will fail to handle them correctly.

PR #1710, opened 2023-04-03
Use buildkit's gitutil package to detect remote git files

jedevc:
:hammer_and_wrench: Fixes https://github.com/docker/buildx/issues/1708.

BuildKit's gitutil package behaves slightly differently than moby's urlutil, so we should rely on BuildKit's gitutil when detecting URLs to avoid cases of accidentally producing invalid build requests that can confuse users.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -18,7 +18,6 @@ import (
 	"github.com/docker/buildx/util/buildflags"
 	"github.com/docker/buildx/util/platformutil"
 	"github.com/docker/cli/cli/config"
-	"github.com/docker/docker/builder/remotecontext/urlutil"
 	hcl "github.com/hashicorp/hcl/v2"
 	"github.com/moby/buildkit/client/llb"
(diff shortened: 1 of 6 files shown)


crazy-max:
I think we also need changes in https://github.com/docker/buildx/blob/6535f16aec98d4e94e3a5f6c9dbe736a0f407f7a/bake/remote.go#L86-L129

jedevc:
I think we should be able to replace the method there with the new `build.IsRemoteURL` function (just pushed a commit to do this).

We probably also want to update `detectGitContext` to align with buildkit, but we should do that by exposing that method to be public first (will open a buildkit-side PR to do that).

crazy-max (approved):

crazy-max:
> We probably also want to update `detectGitContext` to align with buildkit, but we should do that by exposing that method to be public first (will open a buildkit-side PR to do that).

Yes let's do this in a follow-up.

PR #1716, opened 2023-04-04
build(cli): allow passing in-stream using command.Cli

distorhead:
Use command.Cli::In() if it is set, use os.Stdin otherwise.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -312,7 +312,7 @@ func getImageID(resp map[string]string) string {
 }
 
 func runBasicBuild(ctx context.Context, dockerCli command.Cli, opts *controllerapi.BuildOptions, options buildOptions, printer *progress.Printer) (*client.SolveResponse, error) {
-	resp, res, err := cbuild.RunBuild(ctx, dockerCli, *opts, os.Stdin, printer, false)
+	resp, res, err := cbuild.RunBuild(ctx, dockerCli, *opts, dockerCli.In(), printer, false)
 	if res != nil {
 		res.Done()
 	}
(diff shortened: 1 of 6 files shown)


jedevc:
Sorry, I don't have any context for this change :eyes: Could you explain the change a bit more, and what it resolves?

distorhead:
Rebased on actual master branch to resolve conflict.

commands/build.go
```diff
@@ -832,3 +832,10 @@ func printValue(printer printFunc, version string, format string, res map[string
 	}
 	return printer([]byte(res["result.json"]), os.Stdout)
 }
+
+func getInStream(dockerCli command.Cli) io.ReadCloser {
+	if dockerCli.In() != nil {
```
crazy-max, line 837: Don't think this is needed, `In()` is always set in cli: 

* https://github.com/docker/buildx/blob/341fb65f6f568696ed5e2585eb747aedaed7ee1a/cmd/buildx/main.go#L54
* https://github.com/docker/cli/blob/5be21394cbbafc5627bb3e8a54014c108a534b52/cli/command/cli.go#L494
* https://github.com/docker/cli/blob/master/cli/command/cli_options.go#L16C1-L26

commands/build.go
```diff
@@ -324,7 +324,7 @@ func runControllerBuild(ctx context.Context, dockerCli command.Cli, opts *contro
 	var retErr error
 	var resp *client.SolveResponse
 	f := ioset.NewSingleForwarder()
-	f.SetReader(os.Stdin)
+	f.SetReader(getInStream(dockerCli))
```
crazy-max, line 327:
```suggestion
	f.SetReader(dockerCli.In())
```

commands/build.go
distorhead: Thanks for a catch! Actualized code.

distorhead:
@crazy-max Hi,
Done here, just resolved conflict and rebased on actual upstream branch.

jedevc (approved):

crazy-max (approved):

distorhead:
> @distorhead I've pushed some additional commits to apply this same logic across the whole `commands` package. Both `bake` and `debug-shell` should also do the same.

Thanks a lot!

Issue #1072, opened 2022-04-14
Proposal: flag to list bake targets


crazy-max:
Atm the bake definition needs to be opened to find out available targets. This is not practical and should be available through the cli.

Suggest to add a `--targets` or `--list` flag for `bake` cmd that would list the available targets.

Targets would need a `description` field to be displayed:

```hcl
variable "GO_VERSION" {
  default = "1.17"
}

target "_common" {
  args = {
    GO_VERSION = GO_VERSION
  }
}

target "lint" {
  description = "Validate using golangci-lint and yamllint linters"
  inherits = ["_common"]
  dockerfile = "./hack/dockerfiles/lint.Dockerfile"
  output = ["type=cacheonly"]
}

target "test" {
  description = "Runs tests suite" 
  inherits = ["_common"]
  target = "test-coverage"
  output = ["./coverage"]
}

target "binaries" {
  description = "Build binaries for the current platform"
  inherits = ["_common"]
  target = "binaries"
  output = ["./bin"]
  platforms = ["local"]
}

target "binaries-cross" {
  description = "Build binaries for multi platform"
  inherits = ["binaries"]
  platforms = [
    "darwin/amd64",
    "darwin/arm64",
    "linux/amd64",
    "linux/arm/v6",
    "linux/arm/v7",
    "linux/arm64",
    "linux/ppc64le",
    "linux/riscv64",
    "linux/s390x",
    "windows/amd64",
    "windows/arm64"
  ]
}
```

```
$ docker buildx bake --list
lint:           Validate using golangci-lint and yamllint linters
test:           Runs tests suite
binaries:       Build binaries for the current platform
binaries-cross: Build binaries for multi platform
```

Targets that don't specify a `description` will not be displayed.

This flag cannot be used if a target is specified.

For compose it might be tricky as there is no such field in the compose spec that we could use. Maybe a comment can work in this case but not supported with JSON format:

```hcl
# Build binaries for the current platform
target "binaries" {
  inherits = ["_common"]
  target = "binaries"
  output = ["./bin"]
  platforms = ["local"]
}
```

```yml
services:
  binaries: # Build binaries for the current platform
    build:
      target: binaries
```

Or an extra arg to target:

```hcl
target "binaries" "Build binaries for the current platform" {
  inherits = ["_common"]
  target = "binaries"
  output = ["./bin"]
  platforms = ["local"]
}
```

But again not suitable for a JSON representation as well as a compose file.

Open to suggestions.

[...]

PR #1727, opened 2023-04-09
cli: add shell completion

crazy-max:
closes #1674 
partially solves #1072

Add completion to:
* List builder names for `--builder` persistent flag and as arg for some cmds
* Folders (context) for build command
* List bake targets

```
$ docker __completeNoDesc buildx --builder ""
builder
builder2
buildkit0106
buildkit093
remote-builder
default
desktop-linux
docker201016
docker2300
docker2400
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

```
$ docker __completeNoDesc buildx --builder "do"
docker201016
docker2300
docker2400
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

For bake, can be tested within this repo:

```
$ docker __completeNoDesc buildx ""
bake
build
create
du
help
imagetools
inspect
ls
prune
rm
stop
use
version
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

```
$ docker __completeNoDesc buildx bake ""
default
validate
meta-helper
_common
lint
validate-vendor
validate-docs
validate-authors
validate-generated-files
update-vendor
update-docs
update-authors
update-generated-files
mod-outdated
test
binaries
binaries-cross
release
image
image-cross
image-local
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

```
$ docker __completeNoDesc buildx bake "v"
validate
validate-vendor
validate-docs
validate-authors
validate-generated-files
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -23,7 +23,6 @@ import (
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/store/storeutil"
 	"github.com/docker/buildx/util/buildflags"
-	"github.com/docker/buildx/util/cobrautil/completion"
 	"github.com/docker/buildx/util/ioset"
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/buildx/util/tracing"
@@ -254,7 +253,9 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 			}
 			return runBuild(dockerCli, options)
 		},
-		ValidArgsFunction: completion.Disable,
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return nil, cobra.ShellCompDirectiveFilterDirs
+		},
 	}
 
 	var platformsDefault []string
--- a/commands/build.go
+++ b/commands/build.go
@@ -23,6 +23,7 @@ import (
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/store/storeutil"
 	"github.com/docker/buildx/util/buildflags"
+	"github.com/docker/buildx/util/cobrautil/completion"
 	"github.com/docker/buildx/util/ioset"
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/buildx/util/tracing"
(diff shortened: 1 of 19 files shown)


util/cobrautil/completion.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// ValidArgsFn defines a completion func to be returned to fetch completion options
+type ValidArgsFn func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective)
+
+func NoCompletion(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return nil, cobra.ShellCompDirectiveNoSpace
+}
+
+func CompleteBakeTargets(files []string) ValidArgsFn {
```
thaJeztah, line 20: Perhaps bit if a nit, but if these were in a package named "completion", then these could be `completion.BakeTargets` (etc) which may a bit shorter, and more descriptive

util/cobrautil/completion.go
```diff
@@ -0,0 +1,62 @@
...
+			return nil, cobra.ShellCompDirectiveError
+		}
+		var filtered []string
+		if toComplete == "" {
+			return tgts, cobra.ShellCompDirectiveNoFileComp
+		}
+		for _, tgt := range tgts {
+			if strings.HasPrefix(tgt, toComplete) {
+				filtered = append(filtered, tgt)
+			}
```
jedevc, line 37: Could we split out the filtering logic into a separate function to share with other completion funcs as well?

crazy-max: Don't think it's worth it currently. Also targets and builders are not of the same type so would need some reflection.

thaJeztah: Generics! 😂

commands/build.go
```diff
@@ -253,6 +253,9 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 			}
 			return runBuild(dockerCli, options)
 		},
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return nil, cobra.ShellCompDirectiveFilterDirs
+		},
```
jedevc, line 258: Looks like this one was missed from the `completion` package?

crazy-max: As it just returns `cobra.ShellCompDirectiveFilterDirs` I don't think we need a specific func in completion pkg if this is what you mean?

jedevc (approved):

PR #1732, opened 2023-04-11
builder: extra init error handling

milas:
* Return errors from creating the `NodeGroup`
* Ensure that `b.NodeGroup != nil` before reading from it during validation

### Details
This was reported by a Compose user:
* docker/compose#10453

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -108,7 +108,7 @@ func New(dockerCli command.Cli, opts ...Option) (_ *Builder, err error) {
 
 // Validate validates builder context
 func (b *Builder) Validate() error {
-	if b.NodeGroup.DockerContext {
+	if b.NodeGroup != nil && b.NodeGroup.DockerContext {
 		list, err := b.opts.dockerCli.ContextStore().List()
 		if err != nil {
 			return err
--- a/store/storeutil/storeutil.go
+++ b/store/storeutil/storeutil.go
@@ -61,7 +61,10 @@ func GetCurrentInstance(txn *store.Txn, dockerCli command.Cli) (*store.NodeGroup
 		return nil, err
 	}
 	if ng == nil {
-		ng, _ = GetNodeGroup(txn, dockerCli, dockerCli.CurrentContext())
+		ng, err = GetNodeGroup(txn, dockerCli, dockerCli.CurrentContext())
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return ng, nil


builder/builder.go
```diff
@@ -250,7 +258,8 @@ func GetBuilders(dockerCli command.Cli, txn *store.Txn) ([]*Builder, error) {
 	builders := make([]*Builder, len(storeng))
 	seen := make(map[string]struct{})
 	for i, ng := range storeng {
-		b, err := New(dockerCli,
+		b, err := New(
```
jedevc, line 261: Sorry to nitpick, but why has the formatting changed? I agree here, that it should be on a newline :tada: - but I'd like to fix this in the linter/autoformatter, instead of just in one-off cases.

builder/builder.go
```diff
@@ -97,6 +98,9 @@ func New(dockerCli command.Cli, opts ...Option) (_ *Builder, err error) {
 			return nil, err
 		}
 	}
+	if err != nil {
+		return nil, err
+	}
```
milas, line 103: It's worth pointing out that it's possible to get no error but still have `b.NodeGroup == nil`:
https://github.com/milas/buildx/blob/e4a27ec8ab5a15ab12dcdf3bc4806f0852896eef/store/storeutil/storeutil.go#L63-L67

Let me know if you think we should return an error here in that case (`errors.New("no valid builder context")` or something) OR change `GetCurrentInstance` to propagate the error from the fallback call to `GetNodeGroup`

builder/builder.go
milas: Sorry, that was unintentional and just fixed! (I had to recreate buildx project in IntelliJ and it reset to "reformat entire file on save" 🙀)

builder/builder.go
crazy-max: I'm curious of the builder configuration. I guess there is an odd docker context being used as the default builder? Would be nice to have the output of `docker buildx ls` and `docker buildx inspect`.

jedevc: I don't quite understand this error check. How can it be nil at this point, without having already returned from the function?

jedevc: @crazy-max if there are no builders present, I think this is what causes that:

https://github.com/docker/buildx/blob/672eeed9a627d578746438e2bd611772ed7f9288/store/storeutil/storeutil.go#L54-L68

I think the proper fix here is to detect the error on the second lookup, and return that?

milas: Yep, there's two things:

The first, which is _currently_ covered by this PR is the missing error check after `NodeGroup` construction (i.e. either `GetNodeGroup` or `GetCurrentInstance`).

The second, linked a couple times in this thread ⬆️, which is NOT addressed by this PR [right now] is the lack of error propagation in `GetCurrentInstance`'s fallback call to `GetNodeGroup` - will not raise an error / segfault _here_ but will probably return a builder that is unusable and will 💥 on the next call

Given that e.g. `txn.Current()` can also return `nil, nil`, I think the best option is:
```go
	if b.opts.name != "" {
		if b.NodeGroup, err = storeutil.GetNodeGroup(b.opts.txn, dockerCli, b.opts.name); err != nil {
			return nil, err
		}
	} else {
		if b.NodeGroup, err = storeutil.GetCurrentInstance(b.opts.txn, dockerCli); err != nil {
			return nil, err
		}
	}
	if err != nil {
		return nil, err
	}

	// ⬇️  NEW ❗ 
	if b.NodeGroup == nil {
		return nil, errors.New("no builder available")
	}
	// ⬆️  NEW ❗ 


	if b.opts.validate {
		if err = b.Validate(); err != nil {
			return nil, err
		}
	}

	return b, nil
```

milas: (Could also do this in `Validate()` -- i.e. make that return an error if `b.NodeGroup == nil`)

milas:
> I'm curious of the builder configuration

@crazy-max: Got a response here:
```text
$ docker buildx inspect
docker: 'buildx' is not a docker command.
See 'docker --help'

$ docker -v
Docker version 23.0.3, build 3e7cbfd
```
I guess in this case it's Engine but without buildx plugin, which was my suspicion.

So if we can return a predictable error here, Compose can also show a nice error, kind of like `docker/cli` does:
```text
ubuntu@primary:~$ DOCKER_BUILDKIT=1 docker build --help
ERROR: BuildKit is enabled but the buildx component is missing or broken.
       Install the buildx component to build images with BuildKit:
       https://docs.docker.com/go/buildx/
```

crazy-max:
> > I'm curious of the builder configuration
> 
> @crazy-max: Got a response here:
> 
> ```
> $ docker buildx inspect
> docker: 'buildx' is not a docker command.
> See 'docker --help'
> 
> $ docker -v
> Docker version 23.0.3, build 3e7cbfd
> ```

This is not from the same user reporting the issue with Docker 20.10. This is one is with Docker 23 but I'm not able to repro on my side, see https://github.com/docker/compose/issues/10453#issuecomment-1507609958

builder/builder.go
crazy-max: Looking at https://github.com/docker/compose/blob/af6f0ffb9e992502bad030d5c773f523ebd3ebc5/pkg/compose/build_buildkit.go#L36, we are creating a new Builder instance without specifying a name so it would get the current one: https://github.com/docker/buildx/blob/672eeed9a627d578746438e2bd611772ed7f9288/builder/builder.go#L95-L98

Having `NodeGroup` nil means there is no docker context which is not possible with Docker installed with what we know from https://github.com/docker/compose/issues/10453 imo.

I would have a preference for returning the error in https://github.com/docker/buildx/blob/672eeed9a627d578746438e2bd611772ed7f9288/store/storeutil/storeutil.go#L64

@tonistiigi Do you recall why we don't handle an error in this line?

milas:
I managed to reproduce this purely by chance this weekend while working on a personal project in a really silly way:
```shell
docker --context=doesnotexist buildx bake foo
```

I'm doing cursed things that require me to switch back and forth between an AMD64 + ARM64 Docker context so kept explicitly passing Docker context. FWIW I'm using `docker` driver with Engine v23.x.

So making a typo here is an easy way to trigger it at least 🙃

builder/builder.go
thaJeztah: Just glancing over, but does the new error handling in this PR actually capture an error? Because both if/else branches before it seem to be returning an error (if any).

crazy-max:
> I managed to reproduce this purely by chance this weekend while working on a personal project in a really silly way:

Ok that's interesting. With Docker Desktop I'm not able to repro:

```
$ docker --context=doesnotexist buildx bake foo
unable to resolve docker endpoint: context "doesnotexist" does not exist

$ cat ~/.docker/buildx/current
{"Key":"unix:///var/run/docker.sock","Name":"","Global":false}
```

But on a server or using a dind image I repro:

```
$ docker run --rm -it docker:23.0-dind sh

/ # docker --context=doesnotexist buildx bake foo
[+] Building 0.0s (0/0)
[signal SIGSEGV: segmentation violation code=0x1 addr=0x39 pc=0x10ca4c9]

goroutine 1 [running]:
github.com/docker/buildx/builder.(*Builder).Validate(0x238f4c0?)
        github.com/docker/buildx/builder/builder.go:111 +0x29
github.com/docker/buildx/builder.New({0x238f4c0?, 0xc000060800}, {0xc0006e7970, 0x2, 0x2380ac0?})
        github.com/docker/buildx/builder/builder.go:101 +0x27f
github.com/docker/buildx/commands.runBake({0x238f4c0, 0xc000060800}, {0xc0005b60c0, 0x1, 0x1}, {{0x3347d28, 0x0, 0x0}, {0x0, 0x0, ...}, ...})
        github.com/docker/buildx/commands/bake.go:107 +0xc5d
github.com/docker/buildx/commands.bakeCmd.func1(0xc000281800?, {0xc0005b60c0, 0x1, 0x1})
        github.com/docker/buildx/commands/bake.go:202 +0x205
github.com/spf13/cobra.(*Command).execute(0xc000281800, {0xc0004af7e0, 0x1, 0x2})
        github.com/spf13/cobra@v1.6.1/command.go:916 +0x862
github.com/spf13/cobra.(*Command).ExecuteC(0xc00058c900)
        github.com/spf13/cobra@v1.6.1/command.go:1044 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.6.1/command.go:968
github.com/docker/cli/cli-plugins/plugin.RunPlugin(0x1d0af20?, 0xc000281200, {{0x2010c0a, 0x5}, {0x201ae58, 0xb}, {0x2359f50, 0x7}, {0x0, 0x0}, ...})
        github.com/docker/cli@v23.0.0-rc.1+incompatible/cli-plugins/plugin/plugin.go:51 +0x130
main.runPlugin(0x2054ae6?)
        github.com/docker/buildx/cmd/buildx/main.go:45 +0x105
main.main()
        github.com/docker/buildx/cmd/buildx/main.go:62 +0xf9

/ # cat ~/.docker/buildx/current
{"Key":"doesnotexist","Name":"","Global":false}
```

On Docker Desktop if I invoke buildx in standalone mode I'm able to repro:

```
$ DOCKER_CONTEXT=doesnotexist /usr/local/lib/docker/cli-plugins/docker-buildx bake foo
[signal SIGSEGV: segmentation violation code=0x1 addr=0x39 pc=0x10ca4c9]

goroutine 1 [running]:
github.com/docker/buildx/builder.(*Builder).Validate(0x238f4c0?)
        github.com/docker/buildx/builder/builder.go:111 +0x29
github.com/docker/buildx/builder.New({0x238f4c0?, 0xc000148800}, {0xc00081fa90, 0x2, 0x2380ac0?})
        github.com/docker/buildx/builder/builder.go:101 +0x27f
github.com/docker/buildx/commands.runBake({0x238f4c0, 0xc000148800}, {0xc00011f7d0, 0x1, 0x1}, {{0x3347d28, 0x0, 0x0}, {0x0, 0x0, ...}, ...})
        github.com/docker/buildx/commands/bake.go:107 +0xc5d
github.com/docker/buildx/commands.bakeCmd.func1(0xc00023f800?, {0xc00011f7d0, 0x1, 0x1})
        github.com/docker/buildx/commands/bake.go:202 +0x205
github.com/spf13/cobra.(*Command).execute(0xc00023f800, {0xc000120170, 0x1, 0x1})
        github.com/spf13/cobra@v1.6.1/command.go:916 +0x862
github.com/spf13/cobra.(*Command).ExecuteC(0xc00023f200)
        github.com/spf13/cobra@v1.6.1/command.go:1044 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.6.1/command.go:968
main.runStandalone(0x2054ae6?)
        github.com/docker/buildx/cmd/buildx/main.go:40 +0x75
main.main()
        github.com/docker/buildx/cmd/buildx/main.go:60 +0x108

$ cat ~/.docker/buildx/current 
{"Key":"doesnotexist","Name":"","Global":false}
```

So my guess is the proxy on DD does something with the context.

builder/builder.go
milas: Sorry for delay -- added error propagation from the fallback call based on context name as suggested above!

crazy-max (approved):
LGTM thanks

PR #1735, opened 2023-04-13
build: support for saving local state by build ref

tonistiigi:
Some build parameters like directory paths, are never sent to buildkit. But it may still make sense to store them and associate with a unique ref for a build that is used to track other parameters on buildkit side.

@crazy-max

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -25,6 +25,7 @@ import (
 	"github.com/containerd/containerd/platforms"
 	"github.com/docker/buildx/builder"
 	"github.com/docker/buildx/driver"
+	"github.com/docker/buildx/localstate"
 	"github.com/docker/buildx/util/dockerutil"
 	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/buildx/util/progress"
@@ -648,6 +649,12 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		return nil, nil, err
 	} else if len(ulimits) > 0 {
 		so.FrontendAttrs["ulimit"] = ulimits
+	}
+
+	// remember local state like directory path that is not sent to buildkit
+	so.Ref = identity.NewID()
+	if err := saveLocalState(so, opt, node, configDir); err != nil {
+		return nil, nil, err
 	}
 
 	return &so, releaseF, nil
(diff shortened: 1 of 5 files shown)


build/build.go
```diff
@@ -1612,3 +1619,40 @@ func noPrintFunc(opt map[string]Options) bool {
...
+			if err != nil {
+				return err
+			}
+		}
+		ls, err := localstate.New(configDir)
+		if err != nil {
+			return err
+		}
+		if err := ls.SaveRef(node.Builder, node.Name, so.Ref, localstate.State{
+			LocalPath:      lp,
```
crazy-max, line 1650: Do we want to also set the command being invoked?:

```suggestion
            Cmd:            strings.Join(os.Args[1:], " "),
			LocalPath:      lp,
```

This could be handy as we can't easily reconstruct from frontend attrs.

I also wonder if we could set the Buildx version.

tonistiigi: I don't think we should save the raw command. That means someone will try to parse the string again and we don't want that. Also, build can be configured in many other ways and non cmd pkg should not access `os.Args`. If there are more fields that we should capture, we should add them directly in a structured way.

I'm not sure about the version as we don't save the BuildKit version either in history API iirc. But maybe there is an use-case you have in mind.

crazy-max: > That means someone will try to parse the string again and we don't want that. Also, build can be configured in many other ways and non cmd pkg should not access `os.Args`.

Ah right. Main idea was to track back the command issued for a build but frontend attrs should be enough actually.

> I'm not sure about the version as we don't save the BuildKit version either in history API iirc. But maybe there is an use-case you have in mind.

I was thinking of a diagnostic behavior so user could track back what client version was used for this. BuildKit version looks better for this use case actually, that's a good point. It would be useful to have it in the history for the same use case.

crazy-max (approved):

build/build.go
```diff
@@ -1612,3 +1619,40 @@ func noPrintFunc(opt map[string]Options) bool {
...
+
+	if so.Ref == "" {
+		return nil
+	}
+
+	lp := opt.Inputs.ContextPath
+	dp := opt.Inputs.DockerfilePath
+	if lp != "" || dp != "" {
+		if lp != "" {
+			lp, err = filepath.Abs(lp)
```
jedevc, line 1634: :thinking: This doesn't seem right, think we need something more similar to: https://github.com/docker/buildx/blob/b716e48926298a5be3f2381a6fb09d6cd442272f/commands/build.go#L683

jedevc: Ignore me, I hadn't realized this was on `SolveOpt` - so it's definitely a path to a local file at this point.

build/build.go
```diff
@@ -1612,3 +1619,40 @@ func noPrintFunc(opt map[string]Options) bool {
...
+				return err
+			}
+		}
+		ls, err := localstate.New(configDir)
+		if err != nil {
+			return err
+		}
+		if err := ls.SaveRef(node.Builder, node.Name, so.Ref, localstate.State{
+			LocalPath:      lp,
+			DockerfilePath: dp,
```
crazy-max, line 1651:
```suggestion
			LocalPath:      filepath.ToSlash(lp),
			DockerfilePath: filepath.ToSlash(dp),
```

tonistiigi: We do want this to be the actual path that can be looked up from the system. If we do `ToSlash` then it might not be possible to get back the original path.

crazy-max (approved):

PR #1737, opened 2023-04-14
controller: refactor progress api

jedevc:
:hammer: Implements https://github.com/docker/buildx/pull/1671#discussion_r1155875282:

> We should create the progress printer outside of each controller call, and pass it into the interface. Then we can lift printWarnings out of the controller entirely tada (should make things a bit neater).

:no_entry_sign: Replaces https://github.com/docker/buildx/pull/1671 (though this PR still doesn't fix the result printing, happy to do that once I get some initial thoughts on this approach)

Refactor the progress printer creation to the caller-side of the controller api. Then, instead of passing around status channels, we can simply pass around the higher level interface progress.Writer, which allows us to correctly pull warnings out of the remote controller.

WDYT @ktock?

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/csv"
@@ -15,6 +16,7 @@ import (
 
 	"github.com/containerd/console"
 	"github.com/docker/buildx/build"
+	"github.com/docker/buildx/builder"
 	"github.com/docker/buildx/controller"
 	cbuild "github.com/docker/buildx/controller/build"
 	"github.com/docker/buildx/controller/control"
@@ -35,8 +37,11 @@ import (
 	"github.com/docker/docker/pkg/ioutils"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/exporter/containerimage/exptypes"
+	"github.com/moby/buildkit/solver/errdefs"
 	"github.com/moby/buildkit/util/appcontext"
 	"github.com/moby/buildkit/util/grpcerrors"
+	"github.com/moby/buildkit/util/progress/progressui"
+	"github.com/morikuni/aec"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -202,7 +207,28 @@ func runBuild(dockerCli command.Cli, options buildOptions) (err error) {
 		}
 	}
 
+	contextPathHash := options.contextPath
+	if absContextPath, err := filepath.Abs(contextPathHash); err == nil {
+		contextPathHash = absContextPath
+	}
+	b, err := builder.New(dockerCli,
+		builder.WithName(options.builder),
+		builder.WithContextPathHash(contextPathHash),
+	)
+	if err != nil {
+		return err
+	}
+
+	ctx2, cancel := context.WithCancel(context.TODO())
+	defer cancel()
 	progressMode, err := options.toProgress()
+	if err != nil {
+		return err
+	}
+	printer, err := progress.NewPrinter(ctx2, os.Stderr, os.Stderr, progressMode, progressui.WithDesc(
+		fmt.Sprintf("building with %q instance using %s driver", b.Name, b.Driver),
+		fmt.Sprintf("%s:%s", b.Driver, b.Name),
+	))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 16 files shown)


ktock (commented):
Thank you! overall LGTM

controller/pb/progress.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (w *writer) Write(status *client.SolveStatus) {
+	w.ch <- ToControlStatus(status)
+}
+
+func (w *writer) ValidateLogSource(digest.Digest, interface{}) bool {
+	return true
+}
+
+func (w *writer) ClearLogSource(interface{}) {}
```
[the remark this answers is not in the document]
jedevc: I think we should be fine actually

https://github.com/docker/buildx/blob/48b733d6da24265d75f134b6934d67bc657073b5/util/progress/writer.go#L48-L67

They are used by the printer, and since the networked progress writer forwards all data into a printer, we'll actually reuse the logic from there (we might end up doing unnecessary network requests, but it's localhost, so I think it should be fine).

jedevc:
Rebased onto master now that #1640 is merged.

ktock (approved):
Thanks!

commands/build.go
```diff
@@ -200,7 +205,22 @@ func runBuild(dockerCli command.Cli, in buildOptions) error {
...
+		return err
+	}
+	b, err := builder.New(dockerCli,
+		builder.WithName(opts.Builder),
+		builder.WithContextPathHash(opts.ContextPath),
+	)
+	if err != nil {
+		return err
+	}
+	printer, err := progress.NewPrinter(context.TODO(), os.Stderr, os.Stderr, progressMode, progressui.WithDesc(
```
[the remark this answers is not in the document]
crazy-max: I'm fine to make it more self-explanatory

commands/build.go
```diff
@@ -211,10 +231,14 @@ func runBuild(dockerCli command.Cli, in buildOptions) error {
 			return errors.Wrap(err, "removing image ID file")
 		}
 	}
-	resp, _, err := cbuild.RunBuild(ctx, dockerCli, opts, os.Stdin, progress, nil, false)
+	resp, _, err := cbuild.RunBuild(ctx, dockerCli, opts, os.Stdin, printer, false)
+	if err1 := printer.Wait(); err == nil {
```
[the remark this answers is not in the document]
jedevc: Yikes, yup. I've refactored and it makes it a lot clearer how this gets closed.

commands/build.go
```diff
@@ -517,7 +541,22 @@ func launchControllerAndRunBuild(dockerCli command.Cli, options buildOptions) er
 	if err != nil {
 		return err
 	}
-	progress, err := options.toProgress()
+
+	progressMode, err := options.toProgress()
```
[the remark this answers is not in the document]
jedevc: Refactored the common components out to a single function.

jedevc:
Addressed review comments, and also realized I'd completely screwed up the implementation for `reload` on the monitor.

So I've also added some changes to `progress.Printer`, to allow suspending printing and resuming it later on, so we can intersperse a single `progress.Printer` with the contents of the monitor. Essentially, it prevents needing to recreate it over and over with the same parameters and handles all of the synchronization details internally.

Finally, I've tidied up the controller `INFO: launching local buildx controller` messages, and now have them integrated into the progress output:

```
$ BUILDX_EXPERIMENTAL=1 docker buildx build . --detach=false
[+] Building 1.1s (27/27) FINISHED                                                                            docker-container:container
 => [internal] connecting to local controller                     
```

util/progress/printer.go
```diff
@@ -98,12 +103,35 @@ func NewPrinter(ctx context.Context, w io.Writer, out console.File, mode string,
...
+
+			pw.logMu.Lock()
+			pw.logSourceMap = map[digest.Digest]interface{}{}
+			pw.logMu.Unlock()
+
+			close(pw.ready)
+
+			resumeLogs := logutil.Pause(logrus.StandardLogger())
+			// not using shared context to not disrupt display but let is finish reporting errors
+			pw.warnings, pw.err = progressui.DisplaySolveStatus(ctx, c, w, pw.status, solveStatusOpt...)
```
ktock, line 122: `warnings` should be appended here otherwise `Pause`+`Unpause` can discard old warnings from `Warnings()`.

jedevc: Hm, agreed, I actually did try this. Unfortunately, if you run the same build twice (using `reload`), you then get duplicate warnings. I think the right call here is to discard warnings like we currently do, but print the warnings in the monitor, so we still get them?

ktock: SGTM. Or printing warning before starting monitor might be enough for covering the case where the monitor is enabled but the user doesn't call `reload`. As of now, no warning is printed for the user in this case.

jedevc: Hm, this is actually surprisingly difficult to refactor into a way that feels natural - ideally, we'd print the warnings after every single `reload`. But to do that, the warnings logic ends up getting propagated down into the monitor which doesn't seem right to me.

@tonistiigi @crazy-max WDYT about adding a `progressui.WithOnClose()` to be able to execute on finishing the solve display, so we could define the warnings printing logic there, instead of needing to propagate the warnings printing logic throughout? Something like 6590ececd67dc9d7fbbf04320713a7b5eca66647?

crazy-max: The printer object with variadic opts and on close logic looks good.

I wonder if we could also take this to our advantage and make things better for logrus output that is not synchronized properly with the TTY progressbar like: https://github.com/docker/buildx/blob/afcaa8df5fdb9c8c908c8dae325d86a0f8f8c3ef/build/build.go#L693

The Formatter interface would help: https://pkg.go.dev/github.com/sirupsen/logrus#Formatter

crazy-max (approved):

ktock (approved):

tonistiigi (approved):

PR #1745, opened 2023-04-18
monitor: update check for exec args

jedevc:
`exec` does not take a server name, so we can remove the unnecessary check here.

### DIFF
--- a/monitor/monitor.go
+++ b/monitor/monitor.go
@@ -248,7 +248,7 @@ func RunMonitor(ctx context.Context, curRef string, options *controllerapi.Build
 					fmt.Fprintf(stdout, "Attached to process %q. Press Ctrl-a-c to switch to the new container\n", id)
 				case "exec":
 					if len(args) < 2 {
-						fmt.Println("exec: server name must be passed")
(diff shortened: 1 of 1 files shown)


monitor/monitor.go
```diff
@@ -247,10 +247,6 @@ func RunMonitor(ctx context.Context, curRef string, options *controllerapi.Build
 					}
 					fmt.Fprintf(stdout, "Attached to process %q. Press Ctrl-a-c to switch to the new container\n", id)
 				case "exec":
-					if len(args) < 2 {
-						fmt.Println("exec: server name must be passed")
-						continue
-					}
```
ktock, line 253: This can result in panic when the user invokes `exec` without arguments. Could we set default values for `controllerapi.InvokeConfig.Entrypoint` and `controllerapi.InvokeConfig.Cmd` (e.g. `/bin/sh`) when no args passed?

jedevc: Ugh, yeah, my bad - we don't actually want to remove the check here, we just want to rework the error message instead.

ktock (approved):

PR #1747, opened 2023-04-18
controller: only capture the result when invoked through a controller

jedevc:
This ensures that the code used to capture and evaluated a result is only executed when built through the controller. Otherwise, no build result should be recorded.

This ensures that new code added to capture and store the build result for debugging isn't used when BUILDX_EXPERIMENTAL is not set.

See https://github.com/docker/buildx/pull/1640#issuecomment-1513321529 for more information.

**This should be removed at some point in the future** - however, for now, we should have something like this to prevent breaking the state of master while we work on the debugging functionality.

### DIFF
--- a/controller/build/build.go
+++ b/controller/build/build.go
@@ -46,7 +46,7 @@ const defaultTargetName = "default"
 // NOTE: When an error happens during the build and this function acquires the debuggable *build.ResultContext,
 // this function returns it in addition to the error (i.e. it does "return nil, res, err"). The caller can
 // inspect the result and debug the cause of that error.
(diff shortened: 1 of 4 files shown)


controller/build/build.go
```diff
@@ -206,15 +206,20 @@ func buildTargets(ctx context.Context, dockerCli command.Cli, ng *store.NodeGrou
...
-		}
-	})
+	var resp map[string]*client.SolveResponse
+	if generateResult {
+		var mu sync.Mutex
+		var idx int
+		resp, err = build.BuildWithResultHandler(ctx, nodes, opts, dockerutil.NewClient(dockerCli), confutil.ConfigDir(dockerCli), progress.Tee(printer, statusChan), func(driverIndex int, gotRes *build.ResultContext) {
+			mu.Lock()
+			defer mu.Unlock()
+			if res == nil || driverIndex < idx {
```
tonistiigi, line 216: This logic here does not make sense to me. But it is in old code.

tonistiigi (approved):

PR #1750, opened 2023-04-19
Refactor build/controller API to use single `client.Build` call

jedevc:
Depends on #1737.

This PR performs some refactoring after #1640.

To summarize:

- Recombine `Build` and `BuildWithResultHandler` together. They both return partially evaluated results in `ResultContext`. These results must be `Wait`ed to get the actual build response.
- Only use a single `client.Build` call. We only need one gateway session.
- Fix multi-platform and `--print` builds to work over the controller interface.
- Fix `--print` and `--metadata-file` logic to happen on the client side.
- Fix the staggered progress noticed in https://github.com/docker/buildx/pull/1640#issuecomment-1513000048
  - This was the main motivator for using a single `client.Build` session. If we compute the result across two gateway sessions, then we either have to ignore the incoming progress and print out only the final progress results, or we have to merge the two progresses together to display in a single terminal (which... doesn't go well, we get a lot of duplicates).
- Change the controller API: `Build` only performs the parts of the build relevant to actually evaluating the result, to perform the export, `Finalize` must be called with the targeted build ref.
  - There's a good reason for this change - to get all of the invocation progress into a single Build call to BuildKit, we *have* to perform all invoke calls before we perform an export. That's because, to trigger the export, we have to close the `Build` call, which would then leave us unable to invoke in the future using that Build.
  - This is actually a good thing :tada: This means that we won't perform exports before invoke is called, which we were having to do before.
  - The downside here is we have to work out how to "pause" the progress output while we're in monitor mode, so that we don't end up accidentally overflowing into the monitor - not quite sure the best way to do it, I've added a hack into buildkit vendor.


commands/build.go
```diff
@@ -231,9 +241,13 @@ func runBuild(dockerCli command.Cli, in buildOptions) error {
 			return errors.Wrap(err, "removing image ID file")
 		}
 	}
-	resp, _, err := cbuild.RunBuild(ctx, dockerCli, opts, os.Stdin, printer, false)
-	if err1 := printer.Wait(); err == nil {
-		err = err1
+	res, err := cbuild.RunBuild(ctx, dockerCli, opts, os.Stdin, printer)
+	resp, err2 := res.Wait(ctx)
```
ktock, line 244: We should check `res != nil` to avoid panic

build/result.go
```diff
@@ -131,7 +40,39 @@ type ResultContext struct {
 	cleanupsMu sync.Mutex
 }
 
-func (r *ResultContext) Done() {
+func (r *ResultContext) hook(hook func(ctx context.Context) error) {
+	r.hooks = append(r.hooks, hook)
```
ktock, line 44: It looks like called from goroutines so should be protected by lock.

ktock (commented):
Thanks!

monitor/monitor.go
```diff
@@ -145,7 +141,7 @@ func RunMonitor(ctx context.Context, curRef string, options *controllerapi.Build
 						}
 					}
 					var resultUpdated bool
-					ref, _, err := c.Build(ctx, *bo, nil, stdout, stderr, progressMode) // TODO: support stdin, hold build ref
+					ref, err := c.Build(ctx, *bo, nil, progress) // TODO: support stdin, hold build ref
```
ktock, line 144: It seems that no progress is printed on `reload`?

jedevc:
Closing in preference of https://github.com/docker/buildx/pull/1804.

PR #1752, opened 2023-04-21
bake: use default filenames from compose-go

nicksieger:


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/Masterminds/semver/v3 v3.2.0
 	github.com/aws/aws-sdk-go-v2/config v1.15.5
-	github.com/compose-spec/compose-go v1.9.0
```
crazy-max, line 8: Can we have this in a separate commit or maybe another PR if not required for new filenames change?

dvdksn:
might want to update the [compose tests](https://github.com/docker/buildx/blob/master/bake/compose_test.go#L552) to cover other filename variations?

go.mod
nicksieger: The new compose go has some new requirements on load that caused test failures, so probably better to review those separately.

nicksieger: #1753, builds on this one

bake/bake.go
```diff
@@ -42,14 +44,13 @@ type Override struct {
 }
 
 func defaultFilenames() []string {
-	return []string{
-		"docker-compose.yml",  // support app
-		"docker-compose.yaml", // support app
+	composeNames := composecli.DefaultFileNames
+	return append(composeNames, []string{
```
jedevc, line 48: :eyes: will this end up modifying the contents of `composecli.DefaultFileNames`? I *think* the slices will end up sharing the same storage?

I think this would be the fix?

```go
composeNames := append([]string, composecli.DefaultFileNames)
```

nicksieger: It will change the underlying storage but the original `DefaultFileNames` var will still only point to the original slice contents. Probably worth changing though just to eliminate any confusion.

crazy-max (approved):
LGTM thanks

jedevc (approved):

PR #1753, opened 2023-04-21
compose go v1.13.4

nicksieger:


bake/compose.go
```diff
@@ -28,10 +28,14 @@ func ParseComposeFiles(fs []File) (*Config, error) {
...
 
 func ParseCompose(cfgs []compose.ConfigFile, envs map[string]string) (*Config, error) {
+	if envs == nil {
+		envs = make(map[string]string)
+	}
 	cfg, err := loader.Load(compose.ConfigDetails{
 		ConfigFiles: cfgs,
 		Environment: envs,
 	}, func(options *loader.Options) {
+		options.SetProjectName("bake", false)
```
crazy-max, line 38: There is no defaults anymore for the project name? I was thinking it was the name of the working directory?

nicksieger: compose-go was raising a "project name must not be empty" error without this. Looks like the logic changed in compose-spec/compose-go#374. Might have been inadvertent, I'll check with Milas on that.

nicksieger: Ok, good to know. There's nothing that I'm aware of in buildx that's dependent on the latest compose-go, so does it work to file an issue and/or start a PR over there and close this in favor of a new compose-go upgrade PR later on?

bake/compose.go
jedevc, line 1: I think we also want to add support for the new compose features at the same time?

- https://github.com/compose-spec/compose-go/pull/354
- https://github.com/compose-spec/compose-go/pull/365
- I *think* that's it?

They should slot neatly into `compose.go`, with tests in `compose_test.go`:

https://github.com/docker/buildx/blob/43a07f399714c91f9c0131754d771dfc6610b16d/bake/compose.go#L86-L103

bake/compose.go
jedevc: Hmm, is there any possible workaround we could do for this temporarily? I'd like to take the latest compose version in buildx v0.11 (aimed for this week or soon after), but also would be nice to make sure we take a tagged version from `compose-go`.

jedevc (commented):
@milas @nicksieger it would be nice to take this for v0.11 of buildx if we could, any objections?

bake/compose.go
nicksieger: Agreed, maybe they could go in a separate PR after we decide how to handle the project name default?

jedevc: Sure I'd be happy with that.

bake/compose.go
jedevc: @crazy-max @milas @nicksieger: is taking this change as-is with `"bake"` as the project name acceptable for the time being? If so, we can merge this and move on with the v0.11 milestone.

nicksieger: I proposed the workaround by writing the code, so it's an implicit 👍 from me. I don't think the project name is ever really exposed when baking a compose file, right?

jedevc (approved):

PR #1759, opened 2023-04-25
bump go.mod to 1.20

crazy-max:
Basic interfaces satisfy 'comparable' starting with Go 1.20: 

https://github.com/docker/buildx/blob/afcaa8df5fdb9c8c908c8dae325d86a0f8f8c3ef/build/result.go#L32

https://github.com/docker/buildx/blob/afcaa8df5fdb9c8c908c8dae325d86a0f8f8c3ef/vendor/github.com/moby/buildkit/solver/result/result.go#L48

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/docker/buildx
 
-go 1.19
+go 1.20
 
 require (
 	github.com/Masterminds/semver/v3 v3.2.0


tonistiigi:
Not seen this before. How does the version in `go.mod` affect the compilation of generics code?

crazy-max:
> Not seen this before. How does the version in `go.mod` affect the compilation of generics code?

It doesn't as long as you're building with Go 1.20 but building with Go 1.19 will result in:

```
# github.com/moby/buildkit/frontend/gateway/client
vendor/github.com/moby/buildkit/frontend/gateway/client/client.go:18:29: Reference does not implement comparable
note: module requires Go 1.20
```

We should then make clear that the module requires Go 1.20 when used as a library if we want to ba aligned.

This is also reported by some IDEs:

image

jedevc:
Ah yes, this is because of https://github.com/moby/buildkit/pull/3707.

PR #1760, opened 2023-04-26
monitor: Move commands to a separated package

ktock:
Following-up https://github.com/docker/buildx/pull/1640#issuecomment-1512785173

*From @jedevc 's comment:*

> The commands in the monitor are getting long - we really need to split these up, out of a switch statement, and ideally into 
> something a bit more dynamic. Something like implementations of a monitor.Command interface?
> 
> As inspiration:
> ```golang
> type Command interface {
>     // executes the command
>     Exec(args []string /* anything else we might need*/)
> 
>     // returns command information
>     Info() Info
> }
> 
> type Info struct {
>     // a usage string, e.g. "rollback [--init] [entrypoint] [cmd...]"
>     Usage string
> 
>     // a short description, e.g. "re-runs the interactive container with initial rootfs contents"
>     Description string
> 
>     // a longer description, describes in detail what the command does, the
>     // behavior of each arg, etc.
>     Help string
> }
> ```

This commit moves monitor commands to `monior/commands` package. Commands still need access to the `monitor` object and buildx controller so this commit enables this via `Monitor` interface stored in `monitor/types`.

### DIFF
--- a/monitor/commands/reload.go
+++ b/monitor/commands/reload.go
@@ -1,0 +1,76 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	controllererrors "github.com/docker/buildx/controller/errdefs"
+	controllerapi "github.com/docker/buildx/controller/pb"
+	"github.com/docker/buildx/monitor/types"
+	"github.com/docker/buildx/util/progress"
+	"github.com/pkg/errors"
+)
+
+type ReloadCmd struct {
+	m types.Monitor
+
+	stdout   io.WriteCloser
+	progress *progress.Printer
+
+	options      *controllerapi.BuildOptions
+	invokeConfig controllerapi.InvokeConfig
+}
+
+func NewReloadCmd(m types.Monitor, stdout io.WriteCloser, progress *progress.Printer, options *controllerapi.BuildOptions, invokeConfig controllerapi.InvokeConfig) types.Command {
+	return &ReloadCmd{m, stdout, progress, options, invokeConfig}
+}
+
+func (cm *ReloadCmd) Info() types.CommandInfo {
+	return types.CommandInfo{HelpMessage: "reloads the context and build it"}
+}
+
+func (cm *ReloadCmd) Exec(ctx context.Context, args []string) error {
+	var bo *controllerapi.BuildOptions
+	if cm.m.AttachedSessionID() != "" {
+		// Rebuilding an existing session; Restore the build option used for building this session.
+		res, err := cm.m.Inspect(ctx)
+		if err != nil {
+			fmt.Printf("failed to inspect the current build session: %v\n", err)
+		} else {
+			bo = res.Options
+		}
+	} else {
+		bo = cm.options
+	}
+	if bo == nil {
+		return errors.Errorf("no build option is provided")
+	}
+	if cm.m.AttachedSessionID() != "" {
+		if err := cm.m.Disconnect(ctx); err != nil {
+			fmt.Println("disconnect error", err)
+		}
+	}
+	var resultUpdated bool
+	cm.progress.Unpause()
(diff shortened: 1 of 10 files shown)


monitor/types/types.go
```diff
@@ -0,0 +1,75 @@
...
+	"context"
+	"io"
+
+	"github.com/containerd/console"
+	controllerapi "github.com/docker/buildx/controller/pb"
+	"github.com/moby/buildkit/client"
+)
+
+// Monitor provides APIs for attaching and controlling the buildx server.
+type Monitor interface {
```
jedevc, line 13: Ah, I guessed we might need something like this interface, nice :tada:

Since there's a lot of shared methods here with `BuildxController` could we have this interface extend that one? Then we just need to define the methods on top of that.

monitor/commands/reload.go
```diff
@@ -0,0 +1,75 @@
...
+
+	stdout       io.WriteCloser
+	stderr       console.File
+	progressMode string
+
+	options      *controllerapi.BuildOptions
+	invokeConfig controllerapi.InvokeConfig
+}
+
+func NewReloadCmd(m types.Monitor, stdout io.WriteCloser, stderr console.File, progressMode string, options *controllerapi.BuildOptions, invokeConfig controllerapi.InvokeConfig) types.Command {
```
jedevc, line 26: Could we use `ioset.In` to bundle these args together?

monitor/monitor.go
```diff
@@ -301,41 +158,78 @@ func RunMonitor(ctx context.Context, curRef string, options *controllerapi.Build
 	}
 }
 
+func printHelpMessage(out io.Writer, registeredCommands map[string]types.Command) {
+	var names []string
+	for name := range registeredCommands {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	fmt.Fprint(out, "Available commands are:\n\n")
```
jedevc, line 167:
```suggestion
	fmt.Fprint(out, "Available commands are:\n")
```
Let's keep the same format as before.

monitor/monitor.go
```diff
@@ -119,172 +119,29 @@ func RunMonitor(ctx context.Context, curRef string, options *controllerapi.Build
...
+				}
+
+				// Registered commands
+				cmdname := args[0]
+				if cm, ok := registeredCommands[cmdname]; ok {
+					if err := cm.Exec(ctx, args); err != nil {
+						fmt.Fprintf(stdout, "failed %q: %v\n", cmdname, err)
+					}
+				} else {
 					fmt.Printf("unknown command: %q\n", l)
```
jedevc, line 143: For errors that aren't from a specific command, could we prefix them with `monitor:`? That way it kind of mirrors `sh:` errors as opposed to `cat:` errors.

jedevc (commented):
Just a couple of initial thoughts, haven't done a proper deep dive into the command implementations yet, will try and get to that tomorrow :heart:

monitor/types/types.go
ktock: `Monitor` hides `ref` arguments from `BuildxController` and the function signatures are different so it seems that we can't use `BuildxController`'s API for `Monitor`.

monitor/commands/reload.go
ktock: This arguments don't contain stdin and has different stderr type (`stderr console.File`) so maybe we can't use `ioset.In` here.

monitor/types/types.go
ktock: Thank you, SGTM.

ktock: Cherry-picked https://github.com/docker/buildx/commit/8063632c547f206f9fdc66809c5eb41886f666e3 to this PR.

jedevc (approved):
This is good now IMO :tada:

I think we can do some future tidy-ups, but this is a significant improvement from the big switch statement :tada: :tada:

Thanks!

PR #1767, opened 2023-05-02
github: don't auto-add kind/ labels

jedevc:
`kind` labels should only be manually added by maintainers after confirming the issue. Otherwise, it can be tricky to sort through all the issues when filtering for legitimate bugs.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -2,7 +2,6 @@
 name: Bug Report
 description: Report a bug
 labels:
-  - kind/bug
   - status/triage
 
 body:


.github/ISSUE_TEMPLATE/feature.yml
```diff
@@ -2,7 +2,6 @@
 name: Feature request
 description: Missing functionality? Come tell us about it!
 labels:
-  - kind/enhancement
```
thaJeztah, line 5: This should be okay to keep I think? It's a feature / enhancement request.

thaJeztah:
But isn’t the intent of the separate templates to assign bug or enhancement (if one uses “bug” or “enhancement” as template?)

The “triage” label would be for a maintainer to confirm if it’s a  bug (but perhaps a “needs-confirmation” label could be added for bugs, although that could be considered implicit for “triage”)

crazy-max (approved):

PR #1770, opened 2023-05-03
Create buildx integration tests

jedevc:
Buildx currently doesn't have any integration tests - this has actively caused regressions and feature breakages in the past. This is only going to continue as the scope of buildx grows to include more functionality and support more scenarios.

To resolve this, this PR adds a collection of *very* basic integration tests. The idea is that we should test the command line interface, at the point of user access - avoiding testing surfaces that the user doesn't directly interact with. Currently, only a couple of very basic tests for each supported driver are run (`docker`, `docker-container`, `remote`) to test basic build functionality and showcase a working sandbox.

Moving forward, we should work on adding matrix test combinations for experimental features, test high-level bake interactions, multi-platform, etc, etc.


crazy-max:
Can you move the first commit to another PR?

Dockerfile
```diff
@@ -61,6 +76,17 @@ FROM binaries-$TARGETOS AS binaries
 # enable scanning for this stage
 ARG BUILDKIT_SBOM_SCAN_STAGE=true
 
+FROM gobase AS integration-test-base
+RUN apk add --no-cache docker runc containerd
```
crazy-max, line 80: As discussed we should provide custom versions in our matrix when running integration tests. e.g., containerd 1.6/1.7 and moby 23/24. Pretty much like we do in BuildKit: https://github.com/moby/buildkit/blob/81d19ad945cdbf921d182c50fc91169b1375b44d/Dockerfile#L4-L6

Can be done in follow-up

jedevc: @crazy-max any chance you could take a look at this one :pray: :pray:

crazy-max: Will do this in a follow-up instead, there are more changes than I expected. Some in BuildKit https://github.com/moby/buildkit/compare/master...crazy-max:buildkit:itg-dockerd to set extra envs (basically prepend path with right location like we do with containerd) and path to the alt binary.

.github/workflows/test.yml
```diff
@@ -0,0 +1,65 @@
...
+          - docker
+          - docker-container
+          - remote
+        pkg:
+          - ./tests
+        typ:
+          - integration
+        include:
+          - pkg: ./...
+            skip-integration-tests: 1
```
crazy-max, line 39: I think we can remove this for now. We are already running unit tests in the build workflow.

Dockerfile
```diff
@@ -18,6 +21,17 @@ ENV GOFLAGS=-mod=vendor
 ENV CGO_ENABLED=0
 WORKDIR /src
 
+FROM registry:$REGISTRY_VERSION AS registry
+
+FROM moby/buildkit:$BUILDKIT_VERSION AS buildkit
+
+FROM gobase AS gotestsum
```
tonistiigi, line 28: What's the purpose of this?

tests/build.go
```diff
@@ -0,0 +1,99 @@
...
+
+	desc, provider, err := contentutil.ProviderFromRef(target)
+	require.NoError(t, err)
+	imgs, err := testutil.ReadImages(sb.Context(), provider, desc)
+	require.NoError(t, err)
+
+	_ = imgs
+}
+
+func createExample(t *testing.T) string {
```
tonistiigi, line 81: why "Example" ? more like "createTestProject" (more specific ones can have suffixes/numbers).

tests/build.go
```diff
@@ -0,0 +1,99 @@
...
+	t.Helper()
```
tonistiigi, line 82: Is `Helper()` correct in here?

tests/build.go
```diff
@@ -0,0 +1,99 @@
...
+
+	dockerfile := []byte(`
+FROM busybox:latest
+COPY foo /etc/foo
+RUN cp /etc/foo /etc/bar
+
+FROM scratch
+COPY --from=0 /etc/bar /bar
```
tonistiigi, line 90: Although this is just a dummy test, never use `--from=0` numbers. Somebody might read it and think this is what you are supposed to write.

tests/build.go
```diff
@@ -0,0 +1,99 @@
...
+
+	out, err := build(sb, fmt.Sprintf("--output=type=image,name=%s,push=true", target), dir)
+	require.NoError(t, err, string(out))
+
+	desc, provider, err := contentutil.ProviderFromRef(target)
+	require.NoError(t, err)
+	imgs, err := testutil.ReadImages(sb.Context(), provider, desc)
+	require.NoError(t, err)
+
+	_ = imgs
```
tonistiigi, line 78: leftover

tests/integration.go
```diff
@@ -0,0 +1,30 @@
...
+
+func tmpdir(t *testing.T, appliers ...fstest.Applier) (string, error) {
+	tmpdir := t.TempDir()
+	if err := fstest.Apply(appliers...).Apply(tmpdir); err != nil {
+		return "", err
+	}
+	return tmpdir, nil
+}
+
+func buildx(sb integration.Sandbox, args ...string) *exec.Cmd {
```
tonistiigi, line 20: `buildxCmd`

tests/integration_test.go
```diff
@@ -0,0 +1,34 @@
...
+	var tests []func(t *testing.T, sb integration.Sandbox)
+	tests = append(tests, buildTests...)
+	tests = append(tests, inspectTests...)
+	tests = append(tests, lsTests...)
+	testIntegration(t, tests...)
+}
+
+func testIntegration(t *testing.T, funcs ...func(t *testing.T, sb integration.Sandbox)) {
+	mirroredImages := integration.OfficialImages("busybox:latest", "alpine:latest")
+	mirroredImages["moby/buildkit:buildx-stable-1"] = "docker.io/moby/buildkit:buildx-stable-1"
```
tonistiigi, line 29: This can be follow-up but it would be nice if this is configurable.

tonistiigi (commented):
Small naming comments. Generally looks good.

Dockerfile
jedevc: This was pulled in from https://github.com/moby/buildkit/pull/3727, but I wasn't actually using it. I've now enabled test failure annotations as in buildkit, see https://github.com/docker/buildx/actions/runs/5012653318.

tonistiigi (approved):

jedevc:
:tada:

PR #1772, opened 2023-05-04
container-driver: avoid hanging on container panic

tonistiigi:
When buildkit container dies with sigkill while the progress is streaming, it could have left the progress hanging.

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -9,6 +9,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/docker/buildx/driver"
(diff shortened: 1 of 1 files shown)


driver/docker-container/driver.go
```diff
@@ -369,7 +370,11 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
 	}
 
 	var opts []client.ClientOpt
+	var counter int64
 	opts = append(opts, client.WithContextDialer(func(context.Context, string) (net.Conn, error) {
+		if atomic.AddInt64(&counter, 1) > 1 {
+			return nil, net.ErrClosed
```
tonistiigi, line 376: This is just for cleaner error handling. It would work without it as well.

PR #1782, opened 2023-05-09
docs: add initial guide about debugging monitor mode

jedevc:
Adds some very basic early documentation about monitor mode and debugging - the idea of this is to link to it in our release notes for buildx v0.11, so that we can start getting early feedback from users and work out what we need to focus on in the future :tada:

Not sure where this doc should go actually... we've moved the guides into the docs repo - but ideally we should keep it in buildx, at least while we're quickly iterating on this feature.

### DIFF
--- a/docs/guides/debugging.md
+++ b/docs/guides/debugging.md
@@ -1,0 +1,113 @@
+# Debug monitor
+
+To assist with creating and debugging complex builds, Buildx provides a
+debugger to help you step through the build process and easily inspect the
+state of the build environment at any point.
+
+> **Note**
+>
+> The debug monitor is a new experimental feature in recent versions of Buildx.
+> There are rough edges, known bugs, and missing features. Please try it out
+> and let us know what you think!
+
+## Starting the debugger
+
+To start the debugger, first, ensure that `BUILDX_EXPERIMENTAL=1` is set in
+your environment.
+
+```console
+$ export BUILDX_EXPERIMENTAL=1
+```
+
+To start a debug session for a build, you can use the `--invoke` flag with the
+build command to specify a command to launch in the resulting image.
+
+```console
+$ docker buildx build --invoke /bin/sh .
++] Building 4.2s (19/19) FINISHED
+ => [internal] connecting to local controller                                                                                   0.0s
+ => [internal] load build definition from Dockerfile                                                                            0.0s
+ => => transferring dockerfile: 32B                                                                                             0.0s
+ => [internal] load .dockerignore                                                                                               0.0s
+ => => transferring context: 34B                                                                                                0.0s
+ ...
+Launching interactive container. Press Ctrl-a-c to switch to monitor console
+Interactive container was restarted with process "dzz7pjb4pk1mj29xqrx0ac3oj". Press Ctrl-a-c to switch to the new container
+Switched IO
+/ #
+```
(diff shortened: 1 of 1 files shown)


crazy-max:
> Not sure where this doc should go actually... we've moved the guides into the docs repo - but ideally we should keep it in buildx, at least while we're quickly iterating on this feature.

That's fine during the experimental phase but should be in docs repo on GA.

docs/guides/debugging.md
```diff
@@ -0,0 +1,112 @@
...
+/ # ls
+bin    etc    lib    mnt    proc   run    srv    tmp    var
+dev    home   media  opt    root   sbin   sys    usr    work
+```
+
+## Using the monitor
+
+By default, when debugging, you'll be dropped into a shell in the final stage.
+
+By using the `Ctrl-a-c` key combination when in a debugging shell, you'll
```
dvdksn, line 56: I'm guessing `a` is for `Alt`?

```suggestion
When you're in a debug shell, you can use the `Ctrl-Alt-C` key combination to
```

docs/guides/debugging.md
```diff
@@ -0,0 +1,112 @@
...
+toggle between this debug shell and the monitor mode for the debugging session,
+which allows you to run high-level commands to control the debug environment.
```
dvdksn, line 58: I'm guessing `a` is for `Alt`?

```suggestion
When you're in a debug shell, you can use the `Ctrl-Alt-C` key combination to
toggle between the debug shell and the monitor mode.
In monitor mode, you can run commands that control the debug environment.
```

docs/guides/debugging.md
```diff
@@ -0,0 +1,112 @@
...
+
+For example, you can use `ls` to see the contents of the image:
+
+```console
+/ # ls
+bin    etc    lib    mnt    proc   run    srv    tmp    var
+dev    home   media  opt    root   sbin   sys    usr    work
+```
+
+## Using the monitor
```
dvdksn, line 52:
```suggestion
## Monitor mode
```

docs/guides/debugging.md
jedevc: Ah it's actually not - it's CTRL+A, and then C on the keyboard. https://github.com/docker/buildx/pull/1168#pullrequestreview-1006893745.

I *think* this is inspired by QEMU. CTRL+A enters a "command" view, and then next character indicates what to do, e.g. `c` might enter a console, `x` might kill the process, etc.

Good callout, I'm not sure how best to phrase this clearly :cry:

dvdksn: OK, got it. Sorry for the confusion. Not sure what the best way to put this is either. Vimmer in me wants to put something like:

the key combination `<C-a>c` (`Ctrl+A` followed by a `C`) opens the monitor mode

jedevc: > When you're in a debug shell, you can use the `Ctrl-a-c` key combination (press
`Ctrl`+`a` together, lift, then press `c`) to toggle between the debug shell
and the monitor mode. In monitor mode, you can run commands that control the
debug environment.

Hopefully that's descriptive enough :tada:

dvdksn (approved):
LGTM

crazy-max (approved):

PR #1786, opened 2023-05-10
debug: evaluate all refs to allow multi-platform debugging

jedevc:
This is a bit of a hack for now, but this allows us to debug multi-platform results.

On a success, the first entry in the result's platform metadata key is selected, on a failure, the platform which encountered an error is selected.


build/result.go
```diff
@@ -192,32 +216,25 @@ func (r *ResultContext) getProcessConfig(cfg *controllerapi.InvokeConfig, stdin
...
-	}
-	def, err := st.Marshal(ctx)
+
+	ps, err := exptypes.ParsePlatforms(res.Metadata)
 	if err != nil {
 		return nil, err
 	}
-	imgRef, err := c.Solve(ctx, gateway.SolveRequest{
-		Definition: def.ToPB(),
-	})
```
jedevc, line 211: @ktock I'm not quite sure if I'm missing something here - I don't think this second evaluation is necessary?

ktock: Yes, I think it's not necessary.

tonistiigi (approved):

crazy-max (approved):

PR #1787, opened 2023-05-10
ls: format opt

### LINKED ISSUE #325: Add --format support for command output

crazy-max:
fixes #325
follow-up https://github.com/docker/buildx/pull/830 with just the `format` opt.

```
$ docker buildx ls
NAME/NODE             DRIVER/ENDPOINT                                             STATUS     BUILDKIT   PLATFORMS
builder               docker-container
 \_ builder0           \_ unix:///var/run/docker.sock                             running    v0.11.6    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le
 \_ mac-mini-m1        \_ tcp://mac-mini-m1.home.foo.com:2376                     stopped               linux/arm64*
builder2*             docker-container
 \_ builder20          \_ unix:///var/run/docker.sock                             running    v0.11.6    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le
remote-builder        remote
 \_ remote-builder0    \_ docker-container://buildx_buildkit_dk-remote-builder0   inactive
 \_ aws_graviton2      \_ tcp://10.0.0.1:1234                                     running               darwin/arm64*, linux/arm64*, linux/arm/v5*, linux/arm/v6*, linux/arm/v7*, windows/arm64*
 \_ linuxone_s390x     \_ tcp://10.0.0.2:1234                                     running               linux/s390x*
default               docker
 \_ default            \_ default                                                 running    v0.8.2     linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386
desktop-linux         docker
 \_ desktop-linux      \_ desktop-linux                                           error
desktop-windows       docker
 \_ desktop-windows    \_ desktop-windows                                         error
docker2010            docker
 \_ docker2010         \_ docker2010                                              running    v0.8.2     linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386

Failed to get status for desktop-linux (desktop-linux): protocol not available
Failed to get status for desktop-windows (desktop-windows): protocol not available
```

```
$ docker buildx ls --format json
{"Name":"builder","Driver":"docker-container","Dynamic":false,"Nodes":[{"Name":"builder0","Endpoint":"unix:///var/run/docker.sock","Platforms":["linux/amd64","linux/amd64/v2","linux/amd64/v3","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/mips64le","linux/mips64","linux/arm/v7","linux/arm/v6"],"Flags":["--debug","--allow-insecure-entitlement","security.insecure","--allow-insecure-entitlement","network.host"],"DriverOpts":{"env.BUILDKIT_STEP_LOG_MAX_SIZE":"10485760","env.BUILDKIT_STEP_LOG_MAX_SPEED":"10485760","env.JAEGER_TRACE":"localhost:6831","image":"moby/buildkit:master","network":"host"},"Status":"running","Version":"v0.11.6"},{"Name":"mac-mini-m1","Endpoint":"tcp://mac-mini-m1.home.foo.com:2376","DriverOpts":{"image":"moby/buildkit:master"},"Status":"stopped"}]}
{"Name":"builder2","Driver":"docker-container","Dynamic":false,"Nodes":[{"Name":"builder20","Endpoint":"unix:///var/run/docker.sock","Platforms":["linux/amd64","linux/amd64/v2","linux/amd64/v3","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/mips64le","linux/mips64","linux/arm/v7","linux/arm/v6"],"Flags":["--debug","--allow-insecure-entitlement","security.insecure","--allow-insecure-entitlement","network.host"],"DriverOpts":{"env.BUILDKIT_STEP_LOG_MAX_SIZE":"10485760","env.BUILDKIT_STEP_LOG_MAX_SPEED":"10485760","env.JAEGER_TRACE":"localhost:6831","image":"moby/buildkit:master","network":"host"},"Status":"running","Version":"v0.11.6"}]}
{"Name":"remote-builder","Driver":"remote","Dynamic":false,"Nodes":[{"Name":"remote-builder0","Endpoint":"docker-container://buildx_buildkit_dk-remote-builder0","Status":"inactive"},{"Name":"aws_graviton2","Endpoint":"tcp://10.0.0.2:1234","Platforms":["linux/arm64","linux/arm/v7","linux/arm/v6"],"DriverOpts":{"cacert":"/home/crazy/.certs/aws_graviton2/ca.pem","cert":"/home/crazy/.certs/aws_graviton2/cert.pem","key":"/home/crazy/.certs/aws_graviton2/key.pem"},"Status":"running"},{"Name":"linuxone_s390x","Endpoint":"tcp://10.0.0.1:1234","Platforms":["linux/s390x"],"DriverOpts":{"cacert":"/home/crazy/.certs/linuxone_s390x/ca.pem","cert":"/home/crazy/.certs/linuxone_s390x/cert.pem","key":"/home/crazy/.certs/linuxone_s390x/key.pem"},"Status":"running"}]}
{"Name":"default","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"default","Endpoint":"default","Platforms":["linux/amd64","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/arm/v7","linux/arm/v6"],"Status":"running","Version":"v0.8.2"}]}
{"Name":"desktop-linux","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"desktop-linux","Endpoint":"desktop-linux","Status":"error","Err":"protocol not available"}]}
{"Name":"desktop-windows","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"desktop-windows","Endpoint":"desktop-windows","Status":"error","Err":"protocol not available"}]}
{"Name":"docker2010","Driver":"docker","Dynamic":false,"Nodes":[{"Name":"docker2010","Endpoint":"docker2010","Platforms":["linux/amd64","linux/arm64","linux/riscv64","linux/ppc64le","linux/s390x","linux/386","linux/arm/v7","linux/arm/v6"],"Status":"running","Version":"v0.8.2"}]}
```

```
$ docker buildx ls --format "{{.Builder.Name}}: {{range .Builder.Nodes}}{{ .Name }} {{end}}"
builder: builder0 
builder2: builder20 mac-mini-m1
remote-builder: remote-builder0 aws_graviton2 linuxone_s390x
default: default
desktop-linux: desktop-linux
desktop-windows: desktop-windows
docker2010: docker2010
```


crazy-max:
I was wondering if formatting should be moved to a `formatter` pkg and maybe then we should have a parent `cli` pkg.

thaJeztah:
I know there's been some shuffling around in the past between "a single formatter package, containing formatting for all (sub) commands / object-types", and "per-object-type formatters"

We need to have a look what works best; the "problem" I can see with putting all formatting in a single package, is that that package would now have to depend on all object-types (packages), whereas keeping it "per-object" could (potentially) allow for more granular imports (e.g. "I only need formatting for `<object-type X>`).

Of course, "common" / "generic" code should be put in a place that we can share.

commands/ls_print.go
```diff
@@ -0,0 +1,206 @@
...
+{{- end }}
+`
+)
+
+func lsPrint(dockerCli command.Cli, current *store.NodeGroup, builders []*builder.Builder, format string) (hasErrors bool, _ error) {
+	switch format {
+	case formatter.TableFormatKey:
+		format = lsDefaultTableFormat
+	case formatter.RawFormatKey:
+		format = lsDefaultRawFormat
```
thaJeztah, line 60: TBH, I'm always a bit on the fence on the `raw` format. I know we have it for (some?) object-types, but wonder how much it's used, and if there are specific use-cases for it, because we already have `json` for a "parseable" format, and the regular output and `table` for other formats.

commands/ls_print.go
```diff
@@ -0,0 +1,206 @@
...
+  error: {{.Error}}
+  {{- end }}
+{{- end }}
+`
+)
+
+func lsPrint(dockerCli command.Cli, current *store.NodeGroup, builders []*builder.Builder, format string) (hasErrors bool, _ error) {
+	switch format {
+	case formatter.TableFormatKey:
+		format = lsDefaultTableFormat
```
thaJeztah, line 58: We may have been inconsistent with this; slightly wondering if we should consider the default to be a `table` or not. (because of the nested presentation). We probably called it "table" for `docker service ps` (or which one was it?)

The default presentation _COULD_ be considered more of a `tree` format (https://github.com/moby/moby/issues/44582#issuecomment-1456938816), which could keep the road open to implement a "flat" table presentation.

(I haven't fully thought this through though!) /cc @jalonsogo

thaJeztah (commented):
Also quick check (ISTR we have some commands that do "one" and some that do "the other); for the JSON output, do we want to use "jsonlines" (https://jsonlines.org), or an array?

- I know `jq` is perfectly able to handle either
- Other tools may have more difficulty with line-delimited
- For some commands we used line-delimited, because we had to handle "streaming" responses (and didn't want to load _everything_ into memory before being able to display (👈 but this probably doesn't apply to this command (I _think_ we would have it all in memory, nothing streaming, right?)

docs/reference/buildx_ls.md
```diff
@@ -7,23 +7,66 @@ docker buildx ls
...
+|-------------------|---------------------------------------------|
+| `.NameNode`       | Name of the builder or node                 |
+| `.DriverEndpoint` | Driver (for builder) or Endpoint (for node) |
+| `.LastActivity`   | Builder last activity                       |
+| `.Status`         | Builder or node status                      |
+| `.Buildkit`       | BuildKit version of the node                |
+| `.Platforms`      | Available node's platforms                  |
+| `.Error`          | Error                                       |
+| `.IsNode`         | `true` if entry is a node                   |
+| `.IsCurrent`      | `true` if entry is the active builder       |
```
jedevc, line 56: Hm. I'm not sure I'm a fan of this approach for writing templates, I think we lose something from the structured data.

I think I'd rather have `.` be the builder, and have a `.Nodes` property contain a list of nodes that can be `range`d or `index`d to get the node details out.

jedevc:
> I was wondering if formatting should be moved to a formatter pkg and maybe then we should have a parent cli pkg with both commands and formatter. Thoughts?

IMO, we can keep it more specific for this PR, and as we grow the functionality we can pick the one that's most natural? I don't expect anyone outside of buildx to use that API (:pray: hopefully :pray:), so we should be fine to change it.

It also feels like we might want `buildx inspect` to support `--format` as well. It feels like they're quite complimentary commands, so depending on the amount of work involved, might be nice to grab that for v0.11 as well?

go.mod
```diff
@@ -120,6 +120,7 @@ require (
 	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0 // indirect
 	github.com/klauspost/compress v1.16.0 // indirect
 	github.com/mailru/easyjson v0.7.6 // indirect
+	github.com/mattn/go-runewidth v0.0.13 // indirect
```
thaJeztah, line 123: You _could_ decide to bump this to v0.0.14 already; see https://github.com/docker/cli/pull/4251 (v0.0.14 looks to have various performance improvements)

commands/ls_print.go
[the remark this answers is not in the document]
crazy-max: I'm fine to have it in `ls.go` in the meantime.

docs/reference/buildx_ls.md
crazy-max: Yeah me too but that's how the cli api does it today using the formatter context. I don't think there is something else than json-lines in the cli atm (cc @thaJeztah)

crazy-max: Ah I see @thaJeztah commented on this https://github.com/docker/buildx/pull/1787#pullrequestreview-1422131042. I would also prefer `json-array` tbh.

thaJeztah: @jedevc was that specific about the `.IsCurrent()` ? I know we added that because in the past we had `*` as part of the name, which was not ideal, and having a separate `.IsCurrent()` allow it to be "formatted how you'd like", but still keeping the raw name available to be used.

jedevc: Ah sorry my bad, comment on the wrong line :cry: I meant to comment on `IsNode`.

crazy-max: I was looking at https://docs.docker.com/engine/reference/commandline/service_ps/ which services and tasks (similar to us with builders and nodes) and it seems format only handles tasks and not services: https://docs.docker.com/engine/reference/commandline/service_ps/#format

commands/ls_print.go
crazy-max: That sounds like major change that should be first done upstream and then tackled in cli plugins no?

commands/ls_print.go
crazy-max: I see we are using the `raw` format for some commands already in cli: https://github.com/search?q=repo%3Adocker%2Fcli+path%3Acli%2Fcommand+RawFormatKey&type=code&ref=advsearch

thaJeztah: Yeah, I know it's there, but I _really_ doubt it's used much, and really doubt if anyone would miss it if it's not there (yet).

I can see it somewhat more useful for single objects, but in this case, it's neither structured, not (really) usable for automation or anything.

e.g., I  can't just grep for, well, anything really (because there's structured into, and I can't use it in anyway structured), and it's YAML(ish), but not usable as that either.

(fun fact: know that some people managed to parse the default `docker info` output as YAML (yes, apparently that worked!).

So if it's not targeting "parsing" / "scripting", then it should be targeting "humans", and (honestly), I don't think `raw` output is pretty / readable enough for Humans.

So "tempting" to consider "leave it out" (for now), as it'd be easy to add later if people _really_ _really_ miss it (whereas "remove later" is a lot more difficult)

crazy-max: Sure I'm fine removing it in the meantime :+1:

crazy-max:
Added `Builder` placeholder so we can access builder fields:

```
$ docker buildx ls --format "{{.Builder.Name}}: {{range .Builder.Nodes}}{{ .Name }} {{end}}"
builder: builder0 
builder2: builder20 mac-mini-m1
remote-builder: remote-builder0 aws_graviton2 linuxone_s390x
default: default
desktop-linux: desktop-linux
desktop-windows: desktop-windows
docker2010: docker2010
```

builder/builder.go
```diff
@@ -247,6 +250,28 @@ func (b *Builder) Factory(ctx context.Context, dialMeta map[string][]string) (_
 	return b.driverFactory.Factory, err
 }
 
+func (b *Builder) MarshalJSON() ([]byte, error) {
+	var berr string
+	if b.err != nil {
+		berr = strings.TrimSpace(b.err.Error())
+	}
+	return json.Marshal(struct {
+		Name         string
```
tonistiigi, line 259: Just confirming, these should not be lowercase?

docs/reference/buildx_ls.md
```diff
@@ -7,23 +7,75 @@ docker buildx ls
...
+### <a name="format"></a> Format the output (--format)
+
+The formatting options (`--format`) pretty-prints builder instances output
+using a Go template.
+
+Valid placeholders for the Go template are listed below:
+
+| Placeholder       | Description                                 |
+|-------------------|---------------------------------------------|
+| `.NameNode`       | Name of the builder or node                 |
```
tonistiigi, line 48: `Name` ?

builder/builder.go
crazy-max: Yes I don't think we should so we are aligned with the placeholders https://github.com/docker/buildx/pull/1787/files#diff-3d2c8ae27c3eb567b2564a6d5b460040e281261aa98ad464b86fc6a047892dcbR46

docs/reference/buildx_ls.md
crazy-max: Yes make sense. Updated to use `Name` as placeholder.

tonistiigi (approved):

PR #1804, opened 2023-05-16
Perform solve request in the main gateway call

jedevc:
:hammer_and_wrench: Replaces #1750. This is a lot neater than the original implementation and doesn't require that we go and change the _entire_ API.

Now, we always perform the full solve request in the main gateway call. This ensures that progress works properly (and doesn't stagger), and makes the lifetime semantics much clearer.

NewResultContext abstracts the details of a successful/failed build, to always return a single ResultContext, even though the details of how a gateway is created is different:

- For a failed build, we can just keep the gateway open.
- For a successful build, we immediately open another gateway and re-evaluate the build definition in that gateway. This should give an instant cache hit (since the build was just successful).

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -885,7 +885,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 
 					cc := c
 					var printRes map[string][]byte
-					rr, err := c.Build(ctx, so, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
+					buildFunc := func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
 						if opt.PrintFunc != nil {
 							if _, ok := req.FrontendOpt["frontend.caps"]; !ok {
 								req.FrontendOpt["frontend.caps"] = "moby.buildkit.frontend.subrequests+forward"
@@ -915,7 +915,6 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 							}
 
 							if fallback {
-								fmt.Println("falling back!")
 								req.FrontendOpt["build-arg:BUILDKIT_SYNTAX"] = printFallbackImage
 								res2, err2 := c.Solve(ctx, req)
 								if err2 != nil {
@@ -931,16 +930,16 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 						}
 
 						results.Set(resultKey(dp.driverIndex, k), res)
-						if resultHandleFunc != nil {
-							resultCtx, err := NewResultContext(ctx, cc, so, res)
-							if err == nil {
-								resultHandleFunc(dp.driverIndex, resultCtx)
-							} else {
-								logrus.Warnf("failed to record result: %s", err)
-							}
-						}
 						return res, nil
-					}, ch)
+					}
+					var rr *client.SolveResponse
+					if resultHandleFunc != nil {
+						var resultCtx *ResultContext
+						resultCtx, rr, err = NewResultContext(ctx, cc, so, "buildx", buildFunc, ch)
+						resultHandleFunc(dp.driverIndex, resultCtx)
+					} else {
+						rr, err = c.Build(ctx, so, "buildx", buildFunc, ch)
+					}
 					if err != nil {
 						return err
 					}
--- a/build/build.go
+++ b/build/build.go
@@ -886,59 +886,60 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 					cc := c
 					var printRes map[string][]byte
 					rr, err := c.Build(ctx, so, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
-						var isFallback bool
-						var origErr error
-						for {
-							if opt.PrintFunc != nil {
-								if _, ok := req.FrontendOpt["frontend.caps"]; !ok {
-									req.FrontendOpt["frontend.caps"] = "moby.buildkit.frontend.subrequests+forward"
-								} else {
-									req.FrontendOpt["frontend.caps"] += ",moby.buildkit.frontend.subrequests+forward"
-								}
-								req.FrontendOpt["requestid"] = "frontend." + opt.PrintFunc.Name
-								if isFallback {
-									req.FrontendOpt["build-arg:BUILDKIT_SYNTAX"] = printFallbackImage
-								}
+						if opt.PrintFunc != nil {
+							if _, ok := req.FrontendOpt["frontend.caps"]; !ok {
+								req.FrontendOpt["frontend.caps"] = "moby.buildkit.frontend.subrequests+forward"
+							} else {
+								req.FrontendOpt["frontend.caps"] += ",moby.buildkit.frontend.subrequests+forward"
 							}
-							res, err := c.Solve(ctx, req)
-							if err != nil {
-								if origErr != nil {
+							req.FrontendOpt["requestid"] = "frontend." + opt.PrintFunc.Name
+						}
+
+						res, err := c.Solve(ctx, req)
+						if err != nil {
+							fallback := false
+							var reqErr *errdefs.UnsupportedSubrequestError
+							if errors.As(err, &reqErr) {
+								switch reqErr.Name {
+								case "frontend.outline", "frontend.targets":
+									fallback = true
+								default:
 									return nil, err
 								}
-								var reqErr *errdefs.UnsupportedSubrequestError
-								if !isFallback {
-									if errors.As(err, &reqErr) {
-										switch reqErr.Name {
-										case "frontend.outline", "frontend.targets":
-											isFallback = true
-											origErr = err
-											continue
-										}
-										return nil, err
-									}
-									// buildkit v0.8 vendored in Docker 20.10 does not support typed errors
-									if strings.Contains(err.Error(), "unsupported request frontend.outline") || strings.Contains(err.Error(), "unsupported request frontend.targets") {
-										isFallback = true
-										origErr = err
-										continue
-									}
-								}
+							} else {
 								return nil, err
 							}
-							if opt.PrintFunc != nil {
-								printRes = res.Metadata
+							// buildkit v0.8 vendored in Docker 20.10 does not support typed errors
+							if strings.Contains(err.Error(), "unsupported request frontend.outline") || strings.Contains(err.Error(), "unsupported request frontend.targets") {
+								fallback = true
 							}
-							results.Set(resultKey(dp.driverIndex, k), res)
-							if resultHandleFunc != nil {
-								resultCtx, err := NewResultContext(ctx, cc, so, res)
-								if err == nil {
-									resultHandleFunc(dp.driverIndex, resultCtx)
-								} else {
-									logrus.Warnf("failed to record result: %s", err)
+
+							if fallback {
+								fmt.Println("falling back!")
+								req.FrontendOpt["build-arg:BUILDKIT_SYNTAX"] = printFallbackImage
+								res2, err2 := c.Solve(ctx, req)
+								if err2 != nil {
+									return nil, err
 								}
+								res = res2
+							} else {
+								return nil, err
 							}
-							return res, nil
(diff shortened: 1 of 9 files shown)


build/result.go
```diff
@@ -22,166 +20,258 @@ import (
...
+			case _, ok := <-done:
+				if !ok {
+					cancel()
+					return
+				}
+			default:
 			}
+			if err != nil {
+				resCtxCh <- nil
+				errCh <- err
```
ktock, line 170: At this point, both of `respCh` and `resCtxCh` have already been sent so the the for loop receiving `errCh`  isn't polling `errCh` anymore. So maybe this `errCh <- err` seems to block infinitly?

ktock (commented):
Thanks!

build/result.go
```diff
@@ -22,166 +20,258 @@ import (
...
+			res, err = buildFunc(ctx, c)
+
+			if res != nil && err == nil {
+				// force evaluation of all targets
+				def, err2 := getDefinition(ctx, res)
+				if err2 != nil {
+					return nil, err2
+				}
+				res2, err2 := evalDefinition(ctx, c, def)
+				if err != nil {
```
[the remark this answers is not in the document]
jedevc: Oops, this should actually not return here at all.

build/build.go
```diff
@@ -882,64 +882,65 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 
 					ch, done := progress.NewChannel(pw)
 					defer func() { <-done }()
+					_ = done
```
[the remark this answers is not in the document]
jedevc: :facepalm:

build/result.go
```diff
@@ -22,166 +20,258 @@ import (
...
-			res2, _ := result.ConvertResult(targets, func(def *pb.Definition) (gateway.Reference, error) {
-				if res, ok := results[def]; ok {
-					return res.Ref, nil
-				}
-				return nil, nil
-			})
+			return res, err
+		}, ch)
+		select {
+		case _, ok := <-done:
```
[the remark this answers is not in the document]
jedevc: Good call :+1:

build/result.go
jedevc: Good point, I think we can discard the error in this case.

ktock (approved):

build/build.go
```diff
@@ -885,61 +885,61 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
...
-							results.Set(resultKey(dp.driverIndex, k), res)
-							if resultHandleFunc != nil {
-								resultCtx, err := NewResultContext(ctx, cc, so, res)
-								if err == nil {
-									resultHandleFunc(dp.driverIndex, resultCtx)
-								} else {
-									logrus.Warnf("failed to record result: %s", err)
+
+							if fallback {
+								req.FrontendOpt["build-arg:BUILDKIT_SYNTAX"] = printFallbackImage
```
tonistiigi, line 918: btw, `printFallbackImage` value should be updated

build/result.go
```diff
@@ -22,166 +20,246 @@ import (
...
-		return resultCtx, nil
-	case err := <-errCh:
+func evalDefinition(ctx context.Context, c gateway.Client, defs *result.Result[*pb.Definition]) (*gateway.Result, error) {
+	// force evaluation of all targets in parallel
+	results := make(map[*pb.Definition]*gateway.Result)
+	resultsMu := sync.Mutex{}
+	eg, egCtx := errgroup.WithContext(ctx)
+	defs.EachRef(func(def *pb.Definition) error {
+		eg.Go(func() error {
+			res2, err := c.Solve(egCtx, gateway.SolveRequest{
```
tonistiigi, line 210: `res2` ?, same in 227

build/result.go
```diff
@@ -22,166 +20,246 @@ import (
...
-	defer cancel()
+// NewResultContext wraps a call to client.Build, additionally returning a
+// ResultContext alongside the standard response and error.
+//
+// This ResultContext can be used to execute build steps in the same context as
+// the build occurred, which can allow easy debugging of build failures and
+// successes.
+//
+// If the returned ResultContext is not nil, the caller must call Done() on it.
+func NewResultContext(ctx context.Context, cc *client.Client, opt client.SolveOpt, product string, buildFunc gateway.BuildFunc, ch chan *client.SolveStatus) (*ResultContext, *client.SolveResponse, error) {
```
tonistiigi, line 31: It's quite confusing that a function that actually runs the build is called `NewResultContext` (this can be follow-up)

build/result.go
```diff
@@ -22,166 +20,246 @@ import (
...
+		// status messages in the case that we get a solve error
+		baseCh := ch
+		ch := make(chan *client.SolveStatus)
+		go func() {
+			for {
+				s, ok := <-ch
+				if !ok {
+					return
+				}
+				baseCh <- s
```
tonistiigi, line 60: How is this safe/synchronized with closing of the channel.

build/result.go
```diff
@@ -22,166 +20,246 @@ import (
...
-				cancel(doneErr)
-				// wait for Build() completion(or timeout) to ensure the Build's finalizing and avoiding an error "context canceled"
+			// NOTE: ideally this second connection should be lazily opened
+			// NOTE: or... we could open this session *during* the other
+			//       session, which would let us grab the same vertices
+			//       immediately, mitigating the risk for "inconsistent solve
+			//       result"
+			opt := opt
+			opt.Ref = ""
+			cc.Build(ctx, opt, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
```
tonistiigi, line 133: No error?

build/result.go
```diff
@@ -22,166 +20,246 @@ import (
...
+// ResultContext alongside the standard response and error.
+//
+// This ResultContext can be used to execute build steps in the same context as
+// the build occurred, which can allow easy debugging of build failures and
+// successes.
+//
+// If the returned ResultContext is not nil, the caller must call Done() on it.
+func NewResultContext(ctx context.Context, cc *client.Client, opt client.SolveOpt, product string, buildFunc gateway.BuildFunc, ch chan *client.SolveStatus) (*ResultContext, *client.SolveResponse, error) {
+	resCtxCh := make(chan *ResultContext, 1)
+	respCh := make(chan *client.SolveResponse, 1)
```
tonistiigi, line 33: I don't really understand why these need to be channels if they are basically used to store a variable.

build/result.go
```diff
@@ -22,166 +20,246 @@ import (
...
+			}
 
-			// Record the client and ctx as well so that containers can be created from the SolveError.
-			resultCtx.res = res2
-			resultCtx.gwClient = c
-			resultCtx.gwCtx = ctx
-			resultCtx.gwDone = func() {
-				cancel(doneErr)
-				// wait for Build() completion(or timeout) to ensure the Build's finalizing and avoiding an error "context canceled"
+			// NOTE: ideally this second connection should be lazily opened
```
tonistiigi, line 126: The comments this function needs is an overall summary of what going on. A new reader has no idea why there would be a "second connection" with another `Build()` call.

build/build.go
jedevc: Have updated to `docker/dockerfile:1.5.2-labs@sha256:f2e91734a84c0922ff47aa4098ab775f1dfa932430d2888dd5cad5251fafdac4`.

build/result.go
jedevc: Have added *a lot* more comments, importantly, explaining the two different scenarios.

1. The build fails, from some solve error, and we preserve this gateway session.
2. The build succeeds, we close the gateway session, and open a new one (without this, the previous gateway session remains open, so we never run the exporter, which could easily have it's own collection of errors, and would require significant changes to the API)

build/result.go
tonistiigi: >  I really want to avoid another function that we call Build,

If that's what it does, then it should be called like it. We need to draw a chart of these new components/functions to see if the design makes sense and what is the the correct scope for each pkg/function.

This is not `Wrap`. Wrap is something that takes a build callback and returns a new build callback. This one takes a build callback and calls it.

build/result.go
```diff
@@ -22,166 +20,264 @@ import (
...
+//
+// This ResultContext can be used to execute additional build steps in the same
+// context as the build occurred, which can allow easy debugging of build
+// failures and successes.
+//
+// If the returned ResultContext is not nil, the caller must call Done() on it.
+func NewResultContext(ctx context.Context, cc *client.Client, opt client.SolveOpt, product string, buildFunc gateway.BuildFunc, ch chan *client.SolveStatus) (*ResultContext, *client.SolveResponse, error) {
+	// Create a new context to wrap the original, and cancel it when the
+	// caller-provided context is cancelled.
+	// However, we don't want to simply wrap the original context, as we want
```
tonistiigi, line 34: Iiuc the comment in here should be that we create a new context because we only want to forbid cancellation after a certain point(when <-done is closed).

build/result.go
```diff
@@ -22,166 +20,264 @@ import (
...
-				} else {
-					return nil, err
+					resCtx = &ResultContext{
+						done:     make(chan struct{}),
+						solveErr: se,
+						gwClient: c,
+						gwCtx:    ctx,
+					}
+					respErr = se
+					close(baseCh)
```
tonistiigi, line 110: I have trouble understanding this logic. Why does error close the status channel passed to this function. Status should close cleanly on build error, not dropped by the client.

build/result.go
```diff
@@ -22,166 +20,264 @@ import (
...
-			})
+			return res, err
+		}, ch)
+		select {
+		case <-done:
+			cancel(context.Canceled)
+			return
+		default:
+		}
+		close(baseCh)
```
tonistiigi, line 126: How is this not a close of already closed channel.

build/result.go
jedevc: Have renamed `ResultContext` to `ResultHandle` as suggested in slack, but will leave the name of `NewResultHandle` for a follow-up. I think we can do some other refactoring of the `Build` family of functions in Buildx that will make this a little easier.

build/result.go
jedevc: I've made this a little clearer in logic, by having this closed in a single place using a `defer`. I've also added some additional context of why we actually need to close this.

```go
	// Create a new channel to forward status messages to the original.
	//
	// We do this so that we can discard status messages after the main portion
	// of the build is complete. This is necessary for the solve error case,
	// where the original gateway is kept open until the ResultHandle is
	// closed - we don't want progress messages from operations in that
	// ResultHandle to display after this function exits.
	//
	// Additionally, callers should wait for the progress channel to be closed.
	// If we keep the session open and never close the progress channel, the
	// caller will likely hang.
```

build/result.go
jedevc: See above, hopefully the new comments and logic make a bit more sense?

jedevc:
A bit more tidying, I've removed one of the nested goroutines which wasn't necessary, and tidied up the calls to `cancel` and `close(baseCh)` by using a defer.

I think it *might* be possible to try and merge the `done` and solve status `baseCh`, but I've held off doing that, I think that might get a little confusing.

build/result.go
```diff
@@ -22,181 +20,278 @@ import (
...
-func getDefinition(ctx context.Context, res *gateway.Result) (*result.Result[*pb.Definition], error) {
-	return result.ConvertResult(res, func(ref gateway.Reference) (*pb.Definition, error) {
-		st, err := ref.ToState()
-		if err != nil {
-			return nil, err
-		}
-		def, err := st.Marshal(ctx)
-		if err != nil {
-			return nil, err
+// NewResultHandle wraps a call to client.Build, additionally returning a
```
tonistiigi, line 23: `ResultHandle` is not a "wrapper" but evaluated reference to a result. This function performs a build and returns an opaque result object that some other functions take as parameters.

tonistiigi (approved):
Let's get this in for the release 🤞 , but this area needs a follow-up refactoring. We need to take a step back, create a design issue that describes all the cases this code needs to solve and current code paths. Then put together a graph of components and function signatures with their scope description. If needed, we need to wrap or refactor the buildkit `client.Build()` signature if it is too complicated to use in these new cases.

PR #1656, opened 2023-03-01
monitor: breakpoint debugger on monitor and on IDEs (via DAP) 

ktock:
Related: #1104

- Depends on: #1640 and https://github.com/moby/buildkit/pull/3678
  - The first commit in this PR(f18bd40) is #1640

This commit adds an interactive breakpoint debugger on monitor.

This adds a `Continue` controller API for allowing requesting the specific vertex's result from the client.

Monitor walks the LLB graph with support for breakpoints and can exec into each step's container using Continue and Invoke API. Monitor provides DAP (Debug Adapter Protocol) so that IDE's debugger can use the breakpoints on their rich UI. DAP implementation is mostly inheited from buildg.

## Debugger on the monitor

`show` command shows the source file with breakpoint information.

```
(buildx) show
Filename: "Dockerfile"
      1| FROM busybox
      2| RUN echo hi > /hi
      3| RUN echo aaaa > /a
```

`continue` starts the build.
`=>` shows the position where the build is pausing.

```
(buildx) continue
...
Break at [stopOnEntry]
Filename: "Dockerfile"
 =>   1| FROM busybox
      2| RUN echo hi > /hi
      3| RUN echo aaaa > /a
```

`break` command sets a breakpoint
The position of the breakpoint is shown by `*` on the `show` command output.

```
(buildx) break 3
(buildx) show
Filename: "Dockerfile"
 =>   1| FROM busybox
      2| RUN echo hi > /hi
*     3| RUN echo aaaa > /a
```

`continue` resumes the build and stops when it encounters a breakpoint.
`C-a-c` switches IO to the shell.

```
(buildx) Switched IO

/ # cat /a
aaaa
```

- Implemented commands
  - `show`: shows the Dockerfile
  - `break`: set a breakpoint at the specified line
  - `breakpoints`: list key-value pairs of available breakpoints
  - `clear`: clear the breakpoint specified by the key
  - `clearall`: clear all breakpoints
  - `next`: proceed to the next line
  - `continue`: resume the build until the next breakpoint

## Debugger on IDEs

monitor provides DAP (Debug Adapter Protocol) and allows controlling the monitor on IDEs.
Editors can use buildx monitor via that standard protocol.

Please see `docs/guides/breakpoints.md` for details about configuration.

|VS Code|emacs|neovim|
|---|---|---|
|vscode-dap|emacs-dap|nvim-dap|

- VS Code
  - `vscode-buildx-debugger` extension can be used  https://github.com/ktock/vscode-buildx-debugger
[...]

PR #1807, opened 2023-05-17
monitor: Allow launching container from non-exec op with `on-error`

ktock:
Following-up https://github.com/docker/buildx/pull/1640#issuecomment-1512785173 :

*From @jedevc 's comment:*

> We should support breaking on other failed commands, like COPY - atm, on-error stops in the right place, but I can't spawn any session. Even if we just drop into a default shell, with no extra args, environment variables, that's alright for now. This will be necessary to provide a good experience with monitor: breakpoint debugger on monitor and on IDEs (via DAP)  #1656 as well.

Currently `on-error` launches container only from the errored ExecOp. This commit allows it from FileOp (COPY instruction) as well.

Unfortunately buildkitd doesn't seem to return the container execution information (`*pb.Meta`) on non-exec op failure, so we lanch the container from the result of the previous op of the failed one. Currently this is supported only for FileOp but we can expand this feature for other types of Ops if needed in the future.

In the following example Dockerfile, the build fails on `COPY --from=dev /none /` and the debugger container is created by the result of the input (`COPY --from=dev /hi /`).

```dockerfile
FROM ghcr.io/stargz-containers/ubuntu:22.04 AS dev
RUN echo hi > /hi

FROM ghcr.io/stargz-containers/ubuntu:22.04
ENV a=test
COPY --from=dev /hi /
COPY --from=dev /none /
```

```console
$ BUILDX_EXPERIMENTAL=1 /tmp/out/buildx build --detach=false --invoke sh /tmp/ctx4
[+] Building 1.7s (9/9) FINISHED                        docker-container:latest
 => [internal] connecting to local controller                              0.0s
 => => transferring dockerfile: 210B                                       0.0s
 => [internal] load metadata for ghcr.io/stargz-containers/ubuntu:22.04    1.3s
 => [auth] stargz-containers/ubuntu:pull token for ghcr.io                 0.0s
 => [internal] load .dockerignore                                          0.0s
 => => transferring context: 2B                                            0.0s
 => [dev 1/2] FROM ghcr.io/stargz-containers/ubuntu:22.04@sha256:20fa2d7b  0.0s
 => => resolve ghcr.io/stargz-containers/ubuntu:22.04@sha256:20fa2d7bb4de  0.0s
 => CACHED [stage-1 2/3] COPY --from=dev /hi /                             0.0s
 => CACHED [dev 2/2] RUN echo hi > /hi                                     0.0s
 => ERROR [stage-1 3/3] COPY --from=dev /none /                            0.0s
------
 > [stage-1 3/3] COPY --from=dev /none /:
------
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
Interactive container was restarted with process "g5ak43dqhulym5vnp9qs4h37x". Press Ctrl-a-c to switch to the new container
# echo $a
test
# cat hi
hi
# 
```

### DIFF
--- a/build/result.go
+++ b/build/result.go
@@ -11,11 +11,13 @@ import (
 
 	controllerapi "github.com/docker/buildx/controller/pb"
 	"github.com/moby/buildkit/client"
+	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/exporter/containerimage/exptypes"
 	gateway "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/solver/errdefs"
 	"github.com/moby/buildkit/solver/pb"
 	"github.com/moby/buildkit/solver/result"
+	"github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -101,11 +103,12 @@ func getResultAt(ctx context.Context, c *client.Client, solveOpt client.SolveOpt
 				})
 				return nil
 			})
-			resultCtx := ResultContext{}
+			resultCtx := ResultContext{def: targets}
 			if err := eg.Wait(); err != nil {
 				var se *errdefs.SolveError
 				if errors.As(err, &se) {
 					resultCtx.solveErr = se
+					resultCtx.rawErr = err
 				} else {
 					return nil, err
 				}
@@ -170,6 +173,9 @@ type ResultContext struct {
 
 	cleanups   []func()
 	cleanupsMu sync.Mutex
+
+	def    *result.Result[*pb.Definition]
+	rawErr error
 }
 
 func (r *ResultContext) Done() {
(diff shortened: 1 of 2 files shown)


build/result.go
```diff
@@ -196,10 +202,115 @@ func (r *ResultContext) build(buildFunc gateway.BuildFunc) (err error) {
...
+	if errors.As(r.rawErr, &ve) {
+		dgst = digest.Digest(ve.Digest)
+	} else {
+		return nil, errors.Errorf("unsupported vertex: cannot get parent of error")
+	}
+
+	var parentSelector func(inputs []llb.Output) llb.Output
+	switch r.solveErr.Solve.Op.GetOp().(type) {
+	case *pb.Op_File:
+		parentSelector = func(inputs []llb.Output) llb.Output { return inputs[0] } // TODO: allow user selecting one input?
```
ktock, line 279: An op can have multiple inputs and the current implementation always uses the first input (`input[0]`). But maybe we need to introduce a way to ensure the input is actually the previous step's rootfs. Or maybe we can introduce a way to let the user choose one of the inputs. 
Currently I don't have good idea for them so please let me know if there is a good way to solve it.

Issue #1522, opened 2023-01-16
Feature Request: `--load` equivalent for the OCI exporter


TBBle:
This use-case is driven by the Docker containerd image-store beta, see https://github.com/docker/roadmap/issues/371#issuecomment-1380326975 for the initial failure case, and then a bunch more messages after for detailed dives, if the below description is insufficient.

----

Currently `--load` is equivalent to `--output=type=docker,dest=`, which triggers a special code-path in buildx to receive the image back from BuildKit as a tar stream, and pass it into the moral equivalent of `docker load`.

The containerd image-store beta enables multi-platform builds, which produce a manifest list (or image index in OCI terms) ,which the `docker` exporter cannot handle.

[Some testing](https://github.com/docker/roadmap/issues/371#issuecomment-1384251028) shows that the OCI equivalent of `--load`, i.e. `--output=type=oci,dest=` fails with the following error:
```
ERROR: failed to solve: method /moby.filesync.v1.FileSend/diffcopy not supported by the client
```

As far as I can **guess** (which is why I'm posting it here) this is due to [the magic used for the `docker` exporter](https://github.com/docker/buildx/blob/5b3083e9e1bfd433280b0b09444bd39d21fa4bd2/build/build.go#L514-L519) not being hooked up for the `oci` exporter.

However, I haven't tested that this magic would be sufficient. I haven't actually worked out (from GitHub browsing) how adding that output writer leads to another protobuf method being registered on the client. And I'm also guessing that buildx is the client-in-question here.

I have noted that in the single-platform case, `--output=type=docker,dest=` works identically to `--load` (as documented, `--load` is a trivial alias), and `--output=type=oci,dest=` fails the same way, so this is not manifest-list specific.

I have also noted that `--output=type=oci,dest=somefile.tar` in both single-platform and multi-platform mode produces a file that is usable with `docker load -i`, which _suggests_ that everything is fine in both BuildKit and Docker for this data-flow. (`--output=type=oci` can be piped directly into `docker load` under WSL2, but on Windows native consoles (PowerShell and git bash) that's getting intermingled progress output for some reason. That's a different issue though.)

So there's really two requests here:
* Make `--output=type=oci,dest=` work like `--output=type=docker,dest=`, i.e. loading into a Docker instance.
* _Somehow_ make that shorter to type, e.g., `--load-oci` or something not-ugly. Or even make `--load` an alias for `--output=type=oci,dest=` instead, perhaps just in multi-platform mode?

[...]

PR #1813, opened 2023-05-18
`--load` with moby containerd store should use the `oci` exporter

jedevc:
Moby with the containerd store supports importing multi-platform OCI images, which means we *should* be able to add functionality to have `--load` work with multi-platform images.

Before:
```
$ buildx build . --platform linux/amd64,linux/arm64 --builder container --load
[+] Building 0.1s (1/1) FINISHED                                                                                         docker-container:container
 => [internal] connecting to local controller                                                                                                  0.0s
ERROR: failed to build: docker exporter does not currently support exporting manifest lists
```

After:
```
$ buildx build . --platform linux/amd64,linux/arm64 --builder container --load
[+] Building 4.0s (24/24) FINISHED                                                                                       docker-container:container
 => [internal] connecting to local controller                                                                                                  0.0s
 => [internal] load build definition from Dockerfile                                                                                           0.0s
 => => transferring dockerfile: 2.20kB                                                                                                         0.0s
...
 => exporting to docker image format                                                                                                              0.6s
 => => exporting layers                                                                                                                        0.0s
 => => exporting manifest sha256:c0dfaa2978d8ea698f0af7ce2a36373789ed20deffd837a4504891f04c28e0f4                                              0.0s
 => => exporting config sha256:6c835b10e81d195b8795ae810667bff2c0c6565430f16f3649f5e4d4f2b6a650                                                0.0s
 => => exporting attestation manifest sha256:b11b99e5db8890a7e66713e4476d20d2d6fa622519140830d3097ef3112d7ed3                                  0.0s
 => => exporting manifest sha256:ecae44f665a1727aea6c7a2adff17ebc2f954ae5aad957443e5aa35b24277983                                              0.0s
 => => exporting config sha256:dbca1a6373dd85c3bb3ea08a6d980fc7674976126ddcfdbcc2bd61c2b57dfebe                                                0.0s
 => => exporting manifest list sha256:c2b6927cbd1a651e05e462294d11e546d83a4b1c4146d0c8195c234d8f98d490                                         0.0s
 => => sending tarball                                                                                                                         0.6s
 => importing to docker  
```

This patch works by translating the `docker` exporter to an `oci` exporter (which supports multi-platform exports), just as we were previously translating the `docker` exporter to the `moby` exporter for the moby case.

I *think* this should just work? I've done some basic tests with some sample bake files, and build commands from my history, and get the expected results (and don't get any different behavior on non-containerd moby) :tada:

(I can add some integration tests once #1770 merges)

cc @tonistiigi @vvoland @rumpl

## Related issues

- https://github.com/docker/roadmap/issues/371
- Fixes https://github.com/docker/buildx/issues/1522 (cc @TBBle)

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -363,7 +363,7 @@ func toRepoOnly(in string) (string, error) {
 	return strings.Join(out, ","), nil
 }
 
-func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Options, bopts gateway.BuildOpts, configDir string, pw progress.Writer, dl dockerLoadCallback) (solveOpt *client.SolveOpt, release func(), err error) {
+func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Options, bopts gateway.BuildOpts, configDir string, pw progress.Writer, docker *dockerutil.Client) (solveOpt *client.SolveOpt, release func(), err error) {
 	nodeDriver := node.Driver
 	defers := make([]func(), 0, 2)
 	releaseF := func() {
@@ -531,14 +531,22 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 			return nil, nil, notSupported(nodeDriver, driver.OCIExporter)
 		}
 		if e.Type == "docker" {
-			if len(opt.Platforms) > 1 || len(attests) > 0 {
+			features := docker.Features(ctx, e.Attrs["context"])
+			if features[dockerutil.OCIImporter] && e.Output == nil {
+				// rely on oci importer if available (which supports
+				// multi-platform images), otherwise fall back to docker
+				opt.Exports[i].Type = "oci"
+			} else if len(opt.Platforms) > 1 || len(attests) > 0 {
+				if e.Output != nil {
+					return nil, nil, errors.Errorf("docker exporter does not support exporting manifest lists, use the oci exporter instead")
+				}
 				return nil, nil, errors.Errorf("docker exporter does not currently support exporting manifest lists")
 			}
 			if e.Output == nil {
 				if nodeDriver.IsMobyDriver() {
 					e.Type = "image"
 				} else {
-					w, cancel, err := dl(e.Attrs["context"])
+					w, cancel, err := docker.LoadImage(ctx, e.Attrs["context"], pw)
 					if err != nil {
 						return nil, nil, err
 					}
@@ -733,9 +741,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 				hasMobyDriver = true
 			}
 			opt.Platforms = np.platforms
-			so, release, err := toSolveOpt(ctx, node, multiDriver, opt, np.bopts, configDir, w, func(name string) (io.WriteCloser, func(), error) {
-				return docker.LoadImage(ctx, name, w)
-			})
+			so, release, err := toSolveOpt(ctx, node, multiDriver, opt, np.bopts, configDir, w, docker)
 			if err != nil {
 				return nil, err
 			}
@@ -1542,8 +1548,6 @@ func waitContextDeps(ctx context.Context, index int, results *waitmap.Map, so *c
 func notSupported(d driver.Driver, f driver.Feature) error {
 	return errors.Errorf("%s feature is currently not supported for %s driver. Please switch to a different driver (eg. \"docker buildx create --use\")", f, d.Factory().Name())
 }
-
-type dockerLoadCallback func(name string) (io.WriteCloser, func(), error)
 
 func noDefaultLoad() bool {
 	v, ok := os.LookupEnv("BUILDX_NO_DEFAULT_LOAD")
--- a/util/dockerutil/client.go
+++ b/util/dockerutil/client.go
@@ -63,6 +63,23 @@ func (c *Client) LoadImage(ctx context.Context, name string, status progress.Wri
 	}, nil
 }
 
+func (c *Client) Features(ctx context.Context, name string) map[Feature]bool {
+	features := make(map[Feature]bool)
+	if dapi, err := c.API(name); err == nil {
+		if info, err := dapi.Info(ctx); err == nil {
+			for _, v := range info.DriverStatus {
+				switch v[0] {
+				case "driver-type":
+					if v[1] == "io.containerd.snapshotter.v1" {
+						features[OCIImporter] = true
+					}
+				}
+			}
+		}
+	}
+	return features
+}
+
 type waitingWriter struct {
 	*io.PipeWriter
 	f      func()
--- a/util/dockerutil/features.go
+++ b/util/dockerutil/features.go
@@ -1,0 +1,5 @@
+package dockerutil
+
+type Feature string
+
+const OCIImporter Feature = "OCI importer"


tonistiigi:
> Allow --output type=oci to act as in --output type=docker when no dest argument is provided, thereby loading the result into moby.
> Change buildx build --load to default to using --output type=oci

I don't think this is necessarily what we want. I think we shouldn't change the behavior of `-o type=oci` or make `--load` dynamic. Instead, we should change `-o type=docker` to get the tarball in OCI format if the current Docker supports the OCI format. `-o type=docker` means "output to Docker". If Docker supports OCI then internally it should just output to Docker using OCI format internally.

TBBle:
I like the idea that when pushing to Docker, it automatically pushes in OCI format if supported. For `--load` (like the PR's approach) this "just works", which for me is _the_ UX goal. It also means that if people are writing `--load` longhand ([`--output=type=docker`](https://docs.docker.com/engine/reference/commandline/buildx_build/#docker)) then it "just works" with manifest lists. (I'm not as tied to that as a UX goal personally.)

However, this latter approach means `--output type=docker` and `--output type=docker,dest=- | docker load` behave differently, in that the latter still fails for an image index/manifest list.

> `-o type=docker` means "output to Docker".

I think this would be surprising to users, and isn't what the [`--output` documentation](https://docs.docker.com/engine/reference/commandline/buildx_build/#docker) emphasises, although the `--load` documentation implies this by being a fairly simple alias.

It's not _inconsistent_ with the docs, but if taking this latter approach, we might also want to rework the `--output` docs a little to make it clear that `type=docker` is focussed on _somehow_ getting it into a Docker instance, not on the Docker image tarball file format (which would be legacy under this construction). Right now, it is written as a direct contrast with the `oci` exporter, and is (apart from the note in `dest`) all about the tarballs.

tonistiigi:
I think the confusion may come from that in `buildx` and `buildctl` `-o type=docker` do not behave the same. `buildctl` does export a "docker-style" tar (that actually is OCI compatible as well) while buildx behavior is to perform a `docker load`. With my suggestion, no current commands should break, and `--load` works as expected. While in current form this PR breaks some current users of `type=oci`.  I agree that docs may need clarification.

tonistiigi:
> means --output type=docker and --output type=docker,dest=- | docker load behave differently, in that the latter still fails for an image index/manifest list.

First of all, there is no need for anyone to write `--output type=docker,dest=- | docker load` today. If anyone does, it would still work the same way as it works today. We could document that "dest" is an optional parameter, if provided, instead of loading tar, it will be exported in the path, with maximum compatibility with older Docker versions(1.10+). This would not upgrade automatically for a manifest list, but users would usually know if they export a manifest list and can easily change `type=oci` to strictly control the tar format if they need that.

Another thing is that you can imagine that there will be other ways to get build results into docker in the future, eg. https://github.com/moby/moby/issues/44369 . The behavior of the exporter should not be defined by a specific implementation but the expected end goal.

For bake, note that `load` is not bake property, it is only a shortcut override flag. So if exporters are not defined precisely the intended behavior of "load into docker" can't be defined in a file.

jedevc:
> I don't think this is necessarily what we want. I think we shouldn't change the behavior of -o type=oci or make --load dynamic. Instead, we should change -o type=docker to get the tarball in OCI format if the current Docker supports the OCI format. -o type=docker means "output to Docker".

Should be good now :tada:

However, we actually need to pass the `oci` exporter to buildkit, since the `docker` exporter explicitly *does not support multi-platform* images: https://github.com/moby/buildkit/blob/4e4b7a8ff8a5700b0c170185dce4ca86e3a4e33d/exporter/oci/export.go#L115-L117

I wonder if we should maybe change this requirement in the next version of buildkit? We have safeguards to prevent this in buildx now (since v0.6 https://github.com/docker/buildx/pull/582).

build/build.go
```diff
@@ -531,14 +531,22 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 			return nil, nil, notSupported(nodeDriver, driver.OCIExporter)
 		}
 		if e.Type == "docker" {
-			if len(opt.Platforms) > 1 || len(attests) > 0 {
+			features, err := docker.Features(ctx, e.Attrs["context"])
+			if err != nil {
+				return nil, nil, err
+			}
+			if features[dockerutil.OCIImporter] {
```
tonistiigi, line 538: This should probably be:
```
features[dockerutil.OCIImporter] && e.Output == nil
```

build/build.go
```diff
@@ -531,14 +531,22 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
...
+				// rely on oci importer if available (which supports
+				// multi-platform images), otherwise fall back to docker
+				opt.Exports[i].Type = "oci"
+			} else if len(opt.Platforms) > 1 || len(attests) > 0 {
 				return nil, nil, errors.Errorf("docker exporter does not currently support exporting manifest lists")
```
tonistiigi, line 543: We should turn this into two different error messages:

"does not currently support" - if daemon doesn't support OCI
"does not support, use OCI exporter instead" - if `e.Output != nil`

build/build.go
```diff
@@ -531,14 +531,22 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
...
 			}
 			if e.Output == nil {
 				if nodeDriver.IsMobyDriver() {
```
tonistiigi, line 546: Not changing in this PR but shouldn't this also have a "context" check?

tonistiigi (commented):
cc @rumpl

util/dockerutil/client.go
```diff
@@ -63,6 +63,28 @@ func (c *Client) LoadImage(ctx context.Context, name string, status progress.Wri
 	}, nil
 }
 
+func (c *Client) Features(ctx context.Context, name string) (map[Feature]bool, error) {
```
jedevc, line 66: In discussion with @crazy-max about https://github.com/docker/buildx/pull/1846, we noticed that this probably needs a similar cache to ensure that we don't create a new docker client in each loop of the `toSolveOpt` call.

jedevc: Can try and do this as a follow-up though, would be good to not block another RC on this.

jedevc:
Aha, reworked the `Features` call, we shouldn't allow errors from it to propagate to the user - if no docker daemon is available, we should just assume the `OCIImporter` isn't either. This got detected in the new tests we have :tada:

build/build.go
```diff
@@ -531,14 +531,27 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 			return nil, nil, notSupported(nodeDriver, driver.OCIExporter)
 		}
 		if e.Type == "docker" {
-			if len(opt.Platforms) > 1 || len(attests) > 0 {
-				return nil, nil, errors.Errorf("docker exporter does not currently support exporting manifest lists")
+			features := docker.Features(ctx, e.Attrs["context"])
+			if err != nil {
```
tonistiigi, line 535: bogus `err`

rumpl:
I have a couple of questions:

* Does the OCI exporter unpack the image? We use the containerimage exporter and force it to unpack the image, we also sanitize the image name and set a dangling prefix for the cases when the user doesn't set a tag https://github.com/moby/moby/blob/master/builder/builder-next/exporter/overrides/wrapper.go#L33-L41
* did you try and run the built images?

jedevc:
Could *potentially* be related to https://github.com/moby/buildkit/issues/3891?

tonistiigi:
This looks more like issue with the load command on dockerd with containerd enabled rather than build-side. We basically just issue `docker load`. All else is dockerd.

jedevc:
The case with `docker load` (using a non-docker driver) works fine - it's when we use the `type=image` exporter with buildkit (using the docker driver) that we have the issue.

rumpl:
> The case with `docker load` (using a non-docker driver) works fine - it's when we use the `type=image` exporter with buildkit (using the docker driver) that we have the issue.

Are we still talking about the `ERROR: failed to build: docker exporter does not currently support exporting manifest lists` issue or are you talking about something else?

tonistiigi:
> it's when we use the type=image exporter with buildkit (using the docker driver) that we have the issue. 

That case should be working since https://github.com/docker/buildx/pull/1262

jedevc:
Done some more investigation - I think this out of scope for this PR, so we probably shouldn't block on it.

I can reproduce the same issue with the `type=moby` exporter (equivalent to `type=image,unpack=true`) when building a multi-platform image, so something seems incorrect here - which is covered by https://github.com/moby/buildkit/issues/3891.

util/dockerutil/client.go
```diff
@@ -63,6 +63,23 @@ func (c *Client) LoadImage(ctx context.Context, name string, status progress.Wri
 	}, nil
 }
 
+func (c *Client) Features(ctx context.Context, name string) map[Feature]bool {
+	features := make(map[Feature]bool)
+	if dapi, err := c.API(name); err == nil {
```
tonistiigi, line 68: In a follow-up this function could do caching these return values for `bake` and if we add any more `Feature` values we probably would need to ask for a specific one to make sure expensive call like `info` is not done needlessly.

tonistiigi (approved):

rumpl (approved):
Just tested it, seems to work great, awesome work @jedevc, thanks!

Felixoid:
Hello, is there a way to use OCI exported in case of `--push`? Our command looks like `docker buildx build --push --output=type=image,push-by-digest=true`, but it's very slow for low-power instances.

jedevc:
@Felixoid this doesn't seem related to this PR, which is about `--load` - if you have other questions around it, can you open a separate issue?

In any case, if looks like you might want `--output type=image,push-by-digest=true,oci-mediatypes=true`.

Felixoid:
Thank you, I'll try the parameter `oci-mediatypes=true`. It may solve our issue!

PR #1815, opened 2023-05-19
vendor: update buildkit to v0.11@348e79dfed17

jedevc:
Significant changes:

- Includes session sync fix https://github.com/moby/buildkit/pull/3881
- Includes `NO_COLOR` fix from https://github.com/moby/buildkit/pull/3556 (we should have picked this across earlier :scream:)

We may want to wait to do a BuildKit v0.11.7 release if we want to don't want to vendor the pseudo-version.


tonistiigi:
Unless there are any other bugfixes here, I would rather just pick the patches with a replace rule for a quick release. We are about to release v0.11 in a couple of days anyway and this will be EOL. The other updates in buildkit seem harmless, but the containerd has actual logic changes.

jedevc:
So a separate branch in buildkit? Any preferences on name?

tonistiigi:
`fix-session-close-race`, can be from your fork.

jedevc:
Going through the containerd changes, I actually think there's only a single logic change that should affect us: https://github.com/jedevc/buildx/blob/v0.10-vendor-buildkit/vendor/github.com/containerd/ttrpc/server.go. The others are doc comments, formatting, etc. OCI image-spec just adds a few new struct fields that we don't use really.

Aside from that, we just pull in our own buildkit changes, which I'm tempted to trust :tada: Nothing there looks particularly high risk.

I think we should just pull all the changes in.

tonistiigi (approved):

PR #1816, opened 2023-05-19
Fix AWS Authentication when mixing static creds and IAM profile

bpaquet:
When the user supply static creds, we must not enrich them with a session token which is unrelated.


jedevc (approved):

jedevc (approved):

jedevc:
@bpaquet just pushed a fixup to resolve the linter issue :heart:

PR #1818, opened 2023-05-19
test: relax validation of TEST_BUILDKIT_IMAGE to reduce direct dependencies

thaJeztah:
commit 2d124e0ce954727d640fb3de81d8b36615e7b84f (https://github.com/docker/buildx/pull/1770) added some integration-tests, and as part of those tests added github.com/distribution/distribution/v3 as a direct dependency.

github.com/distribution/distribution/v3 is unreleased, and is still receiving backward breaking changes. While this dependency is already used elsewhere, it looks like the only use in buildx was for a sanity check of the custom TEST_BUILDKIT_IMAGE; we can probably relax that check to avoid adding the direct dependency.

### DIFF
--- a/tests/integration_test.go
+++ b/tests/integration_test.go
@@ -4,7 +4,6 @@ import (
 	"os"
 	"testing"
 
-	"github.com/distribution/distribution/v3/reference"
 	"github.com/docker/buildx/tests/workers"
 	"github.com/moby/buildkit/util/testutil/integration"
 )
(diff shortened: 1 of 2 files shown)


tests/integration_test.go
```diff
@@ -30,11 +29,9 @@ func testIntegration(t *testing.T, funcs ...func(t *testing.T, sb integration.Sa
 	mirroredImages := integration.OfficialImages("busybox:latest", "alpine:latest")
 	buildkitImage := "docker.io/moby/buildkit:buildx-stable-1"
 	if integration.IsTestDockerd() {
-		if img, ok := os.LookupEnv("TEST_BUILDKIT_IMAGE"); ok {
-			ref, err := reference.ParseNormalizedNamed(img)
-			if err == nil {
-				buildkitImage = ref.String()
-			}
+		if img, ok := os.LookupEnv("TEST_BUILDKIT_IMAGE"); ok && img != "" {
+			t.Log("using TEST_BUILDKIT_IMAGE", img)
```
thaJeztah, line 33: Not sure if needed, but thought to make it more visible if a custom image is used, in that odd case an invalid image reference is passed.

thaJeztah:
@jedevc @tonistiigi PTAL (let me know if you feel strongly about this; the alternative was to use the same version as BuildKit (v2.8.2), but when looking what it was used for, it seemed a very minimal "added value"

tonistiigi:
This looks fine but how does it matter if it is a direct or indirect dependency? Looks like this came from latest compose update.

We can also just use the regular `reference` pkg that is already used in many places.

PR #1820, opened 2023-05-22
vendor: use docker/distribution instead of distribution/distribution

jedevc:
:crossed_swords: Replaces https://github.com/docker/buildx/pull/1818 (cc @thaJeztah)

This doesn't require pulling in a new direct dependency, we already use docker/distribution throughout imagetools and build.


crazy-max (approved):

thaJeztah (approved):
LGTM

PR #1821, opened 2023-05-22
commands: consume DEBUG environment variable

jedevc:
When running in standalone mode, the --debug flag passed to docker cannot be passed. The docker cli also supports a DEBUG env var, however, in standalone mode this won't be consumed.

This patch reads the contents of the DEBUG environment variable and enables debugging logs when it's been set.


crazy-max (approved):

thaJeztah (approved):
SGTM

Issue #1704, opened 2023-03-29
Azure/GCP/Openstack plugins removed in latest release of client-go


jedevc:
See https://github.com/kubernetes/client-go/tree/master/plugin/pkg/client/auth - until we resolve this, we can't bump the versions of `client-go` (which it tries to do if revendoring buildkit, as in https://github.com/docker/buildx/pull/1703).

Both the Azure and GCP plugins have been deprecated, while the Openstack plugin has been completely removed, and all moved into client credential plugins: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins.

I think we need to update the buildx code to use the [k8s.io/client-go/plugin/pkg/client/auth/exec](https://github.com/kubernetes/client-go/tree/master/plugin/pkg/client/auth/exec) package? @AkihiroSuda do you know what the right thing to do here is?

AkihiroSuda:
I guess we can just drop these lines?
https://github.com/docker/buildx/blob/f360088ae73988ef478ecddb2cbc0c882bedf392/cmd/buildx/main.go#L19-L22

jedevc:
I would guess so? I'm not 100% sure, I don't know what we used them for in the first place :eyes:

PR #1824, opened 2023-05-22
cmd/buildx: non-functional k8s.io/client-go/plugin/pkg/client/auth/openstack

thaJeztah:
- fixes https://github.com/docker/buildx/issues/1704

Kubernetes v1.26 removed the in-tree openstack cloud provider. While buildx has a replace rule to pin the k8s dependencies to an older version, such replace rules are non-transitive, which means that consumers of buildx as a module could potentially end up with non-resolvable dependencies.

The used code was already non-functional since c41b006be1078aa358721b9a92e17eab9a51a958 (buildx v0.5.0-rc1);

https://github.com/docker/buildx/blob/a6b0959276cd4494eed8caabbc2656f9425f2c77/vendor/k8s.io/client-go/plugin/pkg/client/auth/openstack/openstack_stub.go#L26-L36

See:

- https://github.com/kubernetes/kubernetes/commit/9bbf01bae93600d3314bce0acbff1f3a0a42e74b / https://github.com/kubernetes/kubernetes/pull/67782
- https://github.com/kubernetes/client-go/commit/98e81a77843e9ef2ad5c98a6919e450028610631


jedevc (approved):

thaJeztah:
@crazy-max PTAL 🤗

crazy-max (approved):

PR #1825, opened 2023-05-22
go.mod: update k8s deps to v0.26.2 (remove "replace" rule)

thaJeztah:
- [x] follow-up to https://github.com/docker/buildx/pull/1824

Replace rules are not inherited by consumers of buildx as a module, and as
such would default to use the v0.26.2 version. Removing the replace rules
also removes various (indirect) dependencies (although brings in some new
packages from k8s itself).


thaJeztah:
- Perhaps after this, we can also remove the replace rule in compose 🤔 https://github.com/docker/compose/pull/10592

thaJeztah:
Ah, some deprecations;

```
 > [stage-0 5/5] RUN --mount=target=/go/src/github.com/docker/buildx --mount=target=/root/.cache,type=cache   golangci-lint run:
#13 185.9 driver/kubernetes/execconn/execconn.go:45:11: SA1019: exec.Stream is deprecated: use StreamWithContext instead to avoid possible resource leaks. See https://github.com/kubernetes/kubernetes/pull/103177 for details. (staticcheck)
#13 185.9 		serr := exec.Stream(remotecommand.StreamOptions{
#13 185.9 		        ^
```

driver/kubernetes/execconn/execconn.go
```diff
@@ -42,7 +43,8 @@ func ExecConn(restClient rest.Interface, restConfig *rest.Config, namespace, pod
 		remoteAddr: dummyAddr{network: "dummy", s: "dummy-1"},
 	}
 	go func() {
-		serr := exec.Stream(remotecommand.StreamOptions{
+		// FIXME(thaJeztah): should ExecConn get passed a context here? See https://github.com/kubernetes/kubernetes/pull/103177
+		serr := exec.StreamWithContext(context.TODO(), remotecommand.StreamOptions{
```
thaJeztah, line 47: Should `ExecConn` get passed the context, and punched through here? (Not too familiar with this code, and I'm horrible at contexts 😂)

vendor/k8s.io/client-go/plugin/pkg/client/auth/gcp/gcp_stub.go
```diff
@@ -0,0 +1,36 @@
...
+func init() {
+	if err := rest.RegisterAuthProviderPlugin("gcp", newGCPAuthProvider); err != nil {
+		klog.Fatalf("Failed to register gcp auth plugin: %v", err)
+	}
+}
+
+func newGCPAuthProvider(_ string, _ map[string]string, _ rest.AuthProviderConfigPersister) (rest.AuthProvider, error) {
+	return nil, errors.New(`The gcp auth plugin has been removed.
+Please use the "gke-gcloud-auth-plugin" kubectl/client-go credential plugin instead.
+See https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke for further details`)
```
crazy-max, line 35: Same for GCP

driver/kubernetes/execconn/execconn.go
AkihiroSuda: Yes, that's probably better

vendor/k8s.io/client-go/plugin/pkg/client/auth/gcp/gcp_stub.go
thaJeztah: oh! good call; they're stubs now as well

AkihiroSuda (approved):

AkihiroSuda (approved):

PR #1830, opened 2023-05-23
build: fix missing "docker" driver name in build progress

jedevc:
This was missing, since the driver property can only be fully populated after loading nodes from disk. So we add logic to load the nodes, and check for an error, which ensures that the "docker" driver is always correctly present in the progress description.

I think this was introduced in #1737, when I refactored the progress printing.

Before:

```
$ buildx build . --builder=desktop-linux
[+] Building 2.2s (10/10) FINISHED                                                                                           :desktop-linux
```

After:

```
$ buildx build . --builder=desktop-linux
[+] Building 0.9s (6/6) FINISHED                                                                                       docker:desktop-linux
```

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -233,6 +233,10 @@ func runBuild(dockerCli command.Cli, options buildOptions) (err error) {
 	if err != nil {
 		return err
 	}
+	_, err = b.LoadNodes(ctx, false)
+	if err != nil {
+		return err
+	}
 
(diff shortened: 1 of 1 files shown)


crazy-max (approved):
LGTM

Looking at https://github.com/docker/buildx/blob/69a9c6609a8de43aa0cd996526477f51f01e87c9/controller/build/build.go#L160-L174 it seems we instantiate the builder and load nodes twice now. Might look at this in follow-up.

PR #1838, opened 2023-05-24
command(bake): Specify local and remote bake files

c-ameron:
As I mentioned on Slack. This adds the ability to source additional local build definition files when sourcing Bake files via a remote url.
Prefixing a file with `cwd://` will source the bake file from the local machine, instead of the remote location.
Local files will be read/have precedence before remote files.

Usage:
```
docker buildx bake https://github.com/example/upstream.git --file cwd://docker-bake.override.hcl --print
```
This will source a default file from the example/upstream repository, and also source a build definition from the local machine.

I'm a Go novice so any suggestions are welcome 😄 

Thanks!

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
+	"strings"
 
 	"github.com/containerd/console"
 	"github.com/containerd/containerd/platforms"
@@ -98,8 +100,6 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
 	defer cancel()
 
 	var nodes []builder.Node
-	var files []bake.File
-	var inp *bake.Input
 	var progressConsoleDesc, progressTextDesc string
 
 	// instance only needed for reading remote bake files or building
@@ -147,14 +147,7 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
 		}
 	}()
 
-	if url != "" {
-		files, inp, err = bake.ReadRemoteFiles(ctx, nodes, url, in.files, printer)
-	} else {
-		progress.Wrap("[internal] load local bake definitions", printer.Write, func(sub progress.SubLogger) error {
-			files, err = bake.ReadLocalFiles(in.files, dockerCli.In(), sub)
-			return nil
-		})
-	}
+	files, inp, err := readBakeFiles(ctx, nodes, url, in.files, dockerCli.In(), printer)
 	if err != nil {
 		return err
 	}
@@ -295,4 +288,43 @@ func saveLocalStateGroup(dockerCli command.Cli, ref string, lsg localstate.State
 		return err
 	}
 	return l.SaveGroup(ref, lsg)
+}
+
+func readBakeFiles(ctx context.Context, nodes []builder.Node, url string, names []string, stdin io.Reader, pw progress.Writer) (files []bake.File, inp *bake.Input, err error) {
+	var lnames []string
+	var rnames []string
+	for _, v := range names {
+		if strings.HasPrefix(v, "cwd://") {
+			lnames = append(lnames, strings.TrimPrefix(v, "cwd://"))
+		} else {
+			rnames = append(rnames, v)
+		}
+	}
+
+	if url != "" {
+		var rfiles []bake.File
+		rfiles, inp, err = bake.ReadRemoteFiles(ctx, nodes, url, rnames, pw)
+		if err != nil {
+			return nil, nil, err
+		}
+		files = append(files, rfiles...)
+	}
+
+	if len(lnames) > 0 || url == "" {
+		var lfiles []bake.File
+		progress.Wrap("[internal] load local bake definitions", pw.Write, func(sub progress.SubLogger) error {
+			if url != "" {
+				lfiles, err = bake.ReadLocalFiles(lnames, stdin, sub)
+			} else {
+				lfiles, err = bake.ReadLocalFiles(append(lnames, rnames...), stdin, sub)
+			}
+			return nil
+		})
+		if err != nil {
+			return nil, nil, err
+		}
+		files = append(files, lfiles...)
+	}
+
+	return
 }
--- a/tests/bake.go
+++ b/tests/bake.go
@@ -23,6 +23,7 @@ var bakeTests = []func(t *testing.T, sb integration.Sandbox){
 	testBakeLocalMulti,
 	testBakeRemote,
 	testBakeRemoteCmdContext,
+	testBakeRemoteLocalOverride,
 	testBakeRemoteCmdContextOverride,
 	testBakeRemoteContextSubdir,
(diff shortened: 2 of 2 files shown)


commands/bake.go
```diff
@@ -247,3 +259,32 @@ func bakeCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 	return cmd
 }
+
+func containsLocalFiles(s []string) bool {
+	for _, v := range s {
+		if strings.Contains(v, "local://") {
```
tonistiigi, line 265: `HasPrefix`

Same for rest.

commands/bake.go
```diff
@@ -134,7 +135,18 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
...
+			var remoteFiles []bake.File
+			remoteFiles, inp, err = bake.ReadRemoteFiles(ctx, nodes, url, getRemoteFiles(in.files), printer)
+			if err != nil {
+				return err
+			}
+			files, err = bake.ReadLocalFiles(getLocalFiles(in.files))
+			files = append(files, remoteFiles...)
+
+		} else {
+			files, inp, err = bake.ReadRemoteFiles(ctx, nodes, url, in.files, printer)
```
tonistiigi, line 148: Avoid duplicating this call. Instead split the files list into remote and local files with a single function and always call `ReadLocalFiles/ReadRemoteFiles` if array has items.

tonistiigi (commented):
The "local://" prefix is not ideal but don't have better ideas either.

Ideally, this could have an integration test under `tests` but let us know if you need help with that.

jedevc:
Fun tangent: was looking into this PR, only to discover that `BAKE_CMD_CONTEXT` doesn't work correctly anymore: https://github.com/docker/buildx/pull/1840.

> The "local://" prefix is not ideal but don't have better ideas either.

This seems kinda similar to the `BAKE_CMD_CONTEXT` var actually. Internally, we have this work by doing `cwd://` - we could re-use this instead of `local://` which should also let us share some code?

tonistiigi (changes requested):
^ PTAL comment from @jedevc

c-ameron:
Apologies for the late reply, I was away on holiday so will get back to this PR at the end of the week.

> Our integration tests are quite new here, so any help you can give to help us build those out is very appreciated 

Could you please point me in the right direction where to do this? I should do it like the [TestReadLocalFilesDefault](https://github.com/docker/buildx/blob/master/bake/bake_test.go#L1369) function right?

> This seems kinda similar to the BAKE_CMD_CONTEXT var actually. Internally, we have this work by doing cwd:// - we could re-use this instead of local:// which should also let us share some code?

Ok thanks - so I should use the `cwd://` prefix instead? I had a look through [toBuildOpt](https://github.com/docker/buildx/blob/master/bake/bake.go#LL1022C6-L1022C16) but I don't see any reusable methods?

jedevc:
> Could you please point me in the right direction where to do this? I should do it like the [TestReadLocalFilesDefault](https://github.com/docker/buildx/blob/master/bake/bake_test.go?rgh-link-date=2023-06-21T15%3A42%3A09Z#L1369) function right?

We have our integration tests in https://github.com/docker/buildx/blob/master/tests/bake.go, and our bake unit tests in https://github.com/docker/buildx/blob/master/bake/bake_test.go. Ideally - we'd be able to have both, though if the logic is only defined in the `commands` package, we can't have a unit test under our current structure really (which is absolutely fine).

> Ok thanks - so I should use the cwd:// prefix instead? I had a look through [toBuildOpt](https://github.com/docker/buildx/blob/master/bake/bake.go?rgh-link-date=2023-06-21T15%3A42%3A09Z#LL1022C6-L1022C16) but I don't see any reusable methods?

Yeah, `cwd://` would be preferable I think, just so we don't end up having multiple different prefixes for the same thing, referencing local directories.

It's fine if the code doesn't share nicely, this is for different purposes :tada: Maybe it would be good to try and define the `cwd://` constant somewhere though, so at least when manipulating the strings we're using common code (having different code for manipulating URLs has gotten us before :cry:).

commands/bake.go
```diff
@@ -144,7 +145,13 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
...
-		files, inp, err = bake.ReadRemoteFiles(ctx, nodes, url, in.files, printer)
+		localFilesArg, remoteFilesArg := getFilesArgs(in.files)
+		files, inp, err = bake.ReadRemoteFiles(ctx, nodes, url, remoteFilesArg, printer)
+		if len(localFilesArg) > 0 {
+			var localFiles []bake.File
+			localFiles, err = bake.ReadLocalFiles(localFilesArg, dockerCli.In())
+			files = append(localFiles, files...)
+		}
 	} else {
 		files, err = bake.ReadLocalFiles(in.files, dockerCli.In())
```
jedevc, line 156: We should make sure that `cwd://` is still valid for local files - even though it'll just be stripped.

We can move the `getFilesArgs` call out of the branch loop and just call `ReadLocalFiles` on both local and remote splits.

jedevc:
Also, can you squash commits, and add a sign-off (see https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md#sign-your-work)

commands/bake.go
c-ameron: >just call ReadLocalFiles on both local and remote splits.

@jedevc are you sure about that? don't we need 2 different calls do either bake.ReadRemoteFiles or bake.ReadLocalFiles?

c-ameron:
Hi, apologies for the delay in getting this done. I've added an integration test as well. Let me know if you require anything else. Thanks for your time 😃

commands/bake.go
jedevc: Aha, yeah you got the gist of what I meant, thanks :)

jedevc (approved):

jedevc:
@c-ameron I realized that https://github.com/docker/buildx/commit/56389195c506d356d4f9d44db8b5be4c678fd206#diff-3f9c21dce82a82fae6d5ca096e77e5ddcf0eccd7051f815b452ec26fbfeb314bR284 was incorrect, which was causing the e2e tests to fail.

I pushed a commit that should fix that up, and simplifies the splitting logic, so we don't need the `isRemote` argument - hopefully looks good to you.

c-ameron:
Cool! Thanks for your help @jedevc ! 😃

c-ameron:
Is this ok to be approved/merged @tonistiigi ? Thanks in advance! 😃

c-ameron:
I've rebased the commit onto master as there was a conflict in `commands/bake.go` :)

crazy-max:
> Although I'm not sure why this test is now failing https://github.com/docker/buildx/actions/runs/6408757454/job/17398732309?pr=1838

@c-ameron Sorry for the delay. I have rebased to fix this ci issue.

I pushed an extra commit to fix some error handling when reading local/remote files and consolidate the logic in a single func as well. Feel free to squash if that looks good to you.

c-ameron:
No worries, thanks @crazy-max . I've squashed and added you as a co-author too.

PR #1841, opened 2023-05-25
build: display build details link

crazy-max:
https://github.com/docker/buildx/assets/1951866/41a8935a-438f-48f4-be2c-957a56e80f8f

Display build details links if Docker Desktop build backend is enabled.

I'm using an OSC 8 escape sequence with the target links to be displayed in the terminal. Most of the terminal emulators auto-detect when a URL appears onscreen and allow to conveniently open them (e.g. via Ctrl+click or Cmd+click, or the right click menu). A hyperlink is opened upon encountering an OSC 8 escape sequence with the link. The list of supporting terminal emulators and terminal-based apps can be seen here: https://github.com/Alhadis/OSC8-Adoption/.

Atm it only displays links when build is completed and not for active builds. For active builds we need to put links right before build starts so user can just follow build status in desktop instead of his terminal. Would also need a callback to retrieve the ref when attributes are converted to solve opts.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -26,6 +26,7 @@ import (
 	"github.com/docker/buildx/builder"
 	"github.com/docker/buildx/driver"
 	"github.com/docker/buildx/localstate"
+	"github.com/docker/buildx/util/desktop"
 	"github.com/docker/buildx/util/dockerutil"
 	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/buildx/util/progress"
@@ -822,6 +823,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 
 			for i, dp := range dps {
 				i, dp, so := i, dp, *dp.so
+				node := nodes[dp.driverIndex]
 				if multiDriver {
 					for i, e := range so.Exports {
 						switch e.Type {
@@ -939,6 +941,16 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 						resultHandleFunc(dp.driverIndex, resultHandle)
 					} else {
 						rr, err = c.Build(ctx, so, "buildx", buildFunc, ch)
+					}
+					if node.Driver.Features(ctx)[driver.HistoryAPI] && desktop.BuildBackendEnabled() {
+						buildRef := fmt.Sprintf("%s/%s/%s", node.Builder, node.Name, so.Ref)
+						if err != nil {
+							return &desktop.ErrorWithBuildRef{
+								Ref: buildRef,
+								Err: err,
+							}
+						}
+						progress.WriteBuildRef(w, k, buildRef)
 					}
 					if err != nil {
 						return err
--- a/commands/build.go
+++ b/commands/build.go
@@ -26,6 +26,7 @@ import (
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/store/storeutil"
 	"github.com/docker/buildx/util/buildflags"
+	"github.com/docker/buildx/util/desktop"
 	"github.com/docker/buildx/util/ioset"
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/buildx/util/tracing"
@@ -238,6 +239,11 @@ func runBuild(dockerCli command.Cli, options buildOptions) (err error) {
 		return err
 	}
 
+	var term bool
+	if _, err := console.ConsoleFromFile(os.Stderr); err == nil {
+		term = true
+	}
+
 	ctx2, cancel := context.WithCancel(context.TODO())
 	defer cancel()
 	progressMode, err := options.toProgress()
@@ -273,7 +279,9 @@ func runBuild(dockerCli command.Cli, options buildOptions) (err error) {
 		return retErr
 	}
 
-	if progressMode == progress.PrinterModeQuiet {
+	if progressMode != progress.PrinterModeQuiet {
+		desktop.PrintBuildDetails(os.Stderr, printer.BuildRefs(), term)
+	} else {
 		fmt.Println(getImageID(resp.ExporterResponse))
 	}
 	if options.imageIDFile != "" {
--- a/commands/build.go
+++ b/commands/build.go
@@ -273,7 +273,7 @@ func runBuild(dockerCli command.Cli, options buildOptions) (err error) {
 		return retErr
 	}
 
-	if options.quiet {
+	if progressMode == progress.PrinterModeQuiet {
 		fmt.Println(getImageID(resp.ExporterResponse))
 	}
 	if options.imageIDFile != "" {
--- a/util/desktop/desktop.go
+++ b/util/desktop/desktop.go
@@ -1,0 +1,86 @@
+package desktop
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/containerd/console"
+)
+
+var (
+	bbEnabledOnce sync.Once
+	bbEnabled     bool
+)
+
+func BuildBackendEnabled() bool {
+	bbEnabledOnce.Do(func() {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return
+		}
+		_, err = os.Stat(filepath.Join(home, ".docker", "desktop-build", ".lastaccess"))
+		bbEnabled = err == nil
+	})
+	return bbEnabled
+}
+
+func BuildDetailsOutput(refs map[string]string, term bool) string {
+	if len(refs) == 0 {
+		return ""
+	}
+	refURL := func(ref string) string {
+		return fmt.Sprintf("docker-desktop://dashboard/build/%s", ref)
+	}
+	var out bytes.Buffer
+	out.WriteString("View build details: ")
+	multiTargets := len(refs) > 1
+	for target, ref := range refs {
+		if multiTargets {
+			out.WriteString(fmt.Sprintf("\n  %s: ", target))
+		}
+		if term {
+			out.WriteString(hyperlink(refURL(ref)))
+		} else {
+			out.WriteString(refURL(ref))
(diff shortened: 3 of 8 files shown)


go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	github.com/moby/sys/mountinfo v0.6.2
 	github.com/moby/sys/signal v0.7.0
 	github.com/morikuni/aec v1.0.0
+	github.com/muesli/termenv v0.15.1
```
crazy-max, line 29: I could get rid of this lib and just handle the escape sequence in an util package.

build/build.go
```diff
@@ -769,6 +777,9 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 		}
 	}
 
+	// print build details links
+	logrus.Printf(desktop.BuildDetailsOutput(buildDetailsOpts))
```
crazy-max, line 781: As discussed, we should probably output to progress instead of logrus.

go.mod
crazy-max: Done

util/desktop/desktop.go
```diff
@@ -0,0 +1,54 @@
...
+	bbEnabled     bool
+)
+
+func BuildBackendEnabled() bool {
+	bbEnabledOnce.Do(func() {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return
+		}
+		_, err = os.Stat(filepath.Join(home, ".docker", "desktop-build", ".lock"))
```
jedevc, line 22: Think we should be able to get the right filename using https://github.com/docker/buildx/blob/341fb65f6f568696ed5e2585eb747aedaed7ee1a/util/confutil/config.go#L23 the docker cli?

Or is that inaccurate for some reason?

util/desktop/desktop.go
```diff
@@ -0,0 +1,54 @@
...
+	if multiTargets {
+		out.WriteString("\n")
+	}
+	for target, opt := range opts {
+		url := fmt.Sprintf("docker-desktop://dashboard/build/%s/%s/%s", opt.Builder, opt.Node, opt.Ref)
+		if multiTargets {
+			out.WriteString(fmt.Sprintf("  %s: ", target))
+		}
+		// create an escape sequence using the OSC 8 format: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
+		out.WriteString(fmt.Sprintf("\033]8;;%s\033\\%s\033]8;;\033\\\n", url, url))
```
jedevc, line 51: We can split this our to a separate function in this file.

util/desktop/desktop.go
crazy-max: The desktop build daemon sets the data folder to `$HOME/.docker/desktop-build` and doesn't take into account `DOCKER_CONFIG`.

util/desktop/desktop.go
```diff
@@ -0,0 +1,54 @@
...
+		if multiTargets {
+			out.WriteString(fmt.Sprintf("  %s: ", target))
+		}
+		out.WriteString(hyperlink(url, url))
+	}
+	return out.String()
+}
+
+func hyperlink(url, text string) string {
+	// create an escape sequence using the OSC 8 format: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
```
tonistiigi, line 52: Is the hyperlink escape doing anything in here? Iterm works same way without any escape sequences if you just print a link.

If you print escape sequences then you also need to detect that you are actually writing to a terminal, not to a file etc.

build/build.go
tonistiigi: Progress output should be in beginning (and needs more testing with bake). After build this could be a "build summary" section. I think logrus is fine for now and we can turn it into proper summary in next iteration.

tonistiigi: I think it is better if this logging is in `cmd` packages.

build/build.go
```diff
@@ -736,6 +737,13 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 			so, release, err := toSolveOpt(ctx, node, multiDriver, opt, np.bopts, configDir, w, func(name string) (io.WriteCloser, func(), error) {
 				return docker.LoadImage(ctx, name, w)
 			})
+			if desktop.BuildBackendEnabled() {
+				buildDetailsOpts[k] = desktop.BuildDetailsOpt{
+					Builder: node.Builder,
+					Node:    node.Name,
+					Ref:     so.Ref,
```
tonistiigi, line 744: We should only print if `Build()` ran and either was successful or we got error from BuildKit. Otherwise BuildKit has not registered the ref yet and it is useless.

Some options:
- Add link details based on status stream. If we get something on status stream then we know that buildkit has registered ref. This is also the time where we would write the link to progress in the future.
- In the `client.Build` error handler wrap the error with a custom error type. Eg. `RemoteBuildError` that does not change the `Error()` method but then caller can check which component returned the error with `errors.Is`.

util/desktop/desktop.go
```diff
@@ -0,0 +1,54 @@
...
+
+type BuildDetailsOpt struct {
+	Builder string
+	Node    string
+	Ref     string
+}
+
+func BuildDetailsOutput(opts map[string]BuildDetailsOpt) string {
+	var out bytes.Buffer
+	out.WriteString("Open build details on Docker Desktop: ")
```
tonistiigi, line 36: `s/on/in/`

Maybe `s/Open/Review/` . The shorter we can make this prefix the better.

util/desktop/desktop.go
crazy-max: Some terminal like Windows Terminal need the escape sequence as it doesn't solve on its own unknown schemes like `docker-desktop://`

build/build.go
```diff
@@ -929,6 +933,13 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
...
 							results.Set(resultKey(dp.driverIndex, k), res)
+							if node.Driver.Features()[driver.HistoryAPI] && desktop.BuildBackendEnabled() {
+								buildDetailsOpts[k] = desktop.BuildDetailsOpt{
+									Builder: node.Builder,
+									Node:    node.Name,
+									Ref:     so.Ref,
+								}
+							}
 							if resultHandleFunc != nil {
 								resultCtx, err := NewResultContext(ctx, cc, so, res)
```
crazy-max, line 944: @jedevc If we want to print in commands pkg I was wondering if we could use the result context handler to set the build details opts?

jedevc: `ResultContext` doesn't leave outside the controller boundary, so that would likely only work for the `runBasicBuild` case I think :cry: Although, we do record the build options, so `controller.Inspect` could work for this?

Though that's only the controller options, not the full solve request.

crazy-max: Ah right thanks for the feedback, I will take a closer look.

crazy-max: Builder name, node name when driver pairs are computed during build are not inputs so doesn't look good to have it in build options. Same for the `refID` computed for each solve opt.

tonistiigi:
@crazy-max Looks like multiple CI issues.

tonistiigi:
These comments have not been resolved:


> I think it is better if this logging is in cmd packages.

> If you print escape sequences then you also need to detect that you are actually writing to a terminal, not to a file etc. (see full previous comment for alternatives)

> We should only print if Build() ran and either was successful or we got error from BuildKit. Otherwise BuildKit has not registered the ref yet and it is useless.

The link is not shown at all on error anymore iiuc.

I'm not sure if we should show the link with `--print`. If not then these builds should be marked as internal so they don't get registered by history API.

crazy-max:
> I think it is better if this logging is in cmd packages.

Moved to commands package

> If you print escape sequences then you also need to detect that you are actually writing to a terminal, not to a file etc. (see full previous comment for alternatives)

Added detection using `containerd/console`.

> We should only print if Build() ran and either was successful or we got error from BuildKit. Otherwise BuildKit has not registered the ref yet and it is useless.

It now prints on error as well. I was looking at using `ResultContext` and it works fine with `build` command but we don't have any implementation yet for `bake` so in the meantime I return an extra arg that contains the build details.

build/build.go
```diff
@@ -663,18 +664,18 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	return &so, releaseF, nil
 }
 
-func Build(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer) (resp map[string]*client.SolveResponse, err error) {
+func Build(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer) (resp map[string]*client.SolveResponse, bd map[string]desktop.BuildDetails, err error) {
```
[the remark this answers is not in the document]
crazy-max: Now using progress to set build detail and print `onclose` in commands package for build and bake.

build/build.go
crazy-max: Opened https://github.com/docker/buildx/pull/1861 to carry changes from #1680. Will revisit changes here once switching bake to controller for building.

commands/build.go
```diff
@@ -614,10 +620,15 @@ func decodeExporterResponse(exporterResponse map[string]string) map[string]inter
 	return out
 }
 
-func wrapBuildError(err error, bake bool) error {
+func wrapBuildError(err error, term bool, bake bool) error {
 	if err == nil {
 		return nil
 	}
+	if ebr, ok := err.(*build.ErrorWithBuildRef); ok {
+		return errors.Errorf("%v\n\n%s", ebr.Err, buildDetailsOutput(map[string]string{
```
tonistiigi, line 628: wrap function should not flatten the typed error into a string. This should happen when printing the error, currently handled in main, eg. same as the `errdefs.Sources()` handling.

tonistiigi (approved):

PR #1846, opened 2023-05-26
driver: check history capability

crazy-max:
related to #1841 

Adds history API feature check for each driver. Also adds cache for driver features.

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -387,9 +387,8 @@ func (d *Driver) Factory() driver.Factory {
 	return d.factory
 }
 
-func (d *Driver) Features() map[driver.Feature]bool {
+func (d *Driver) Features(ctx context.Context) map[driver.Feature]bool {
 	var historyAPI bool
-	ctx := context.Background()
 	c, err := d.Client(ctx)
 	if err == nil {
 		historyAPI = driver.HistoryAPISupported(ctx, c)
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -388,12 +388,19 @@ func (d *Driver) Factory() driver.Factory {
 }
 
 func (d *Driver) Features() map[driver.Feature]bool {
+	var historyAPI bool
+	ctx := context.Background()
+	c, err := d.Client(ctx)
+	if err == nil {
+		historyAPI = driver.HistoryAPISupported(ctx, c)
+		c.Close()
+	}
 	return map[driver.Feature]bool{
 		driver.OCIExporter:    true,
 		driver.DockerExporter: true,
-
-		driver.CacheExport:   true,
-		driver.MultiPlatform: true,
+		driver.CacheExport:    true,
+		driver.MultiPlatform:  true,
+		driver.HistoryAPI:     historyAPI,
 	}
 }
 
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -6,8 +6,11 @@ import (
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/util/progress"
 	clitypes "github.com/docker/cli/cli/config/types"
+	controlapi "github.com/moby/buildkit/api/services/control"
 	"github.com/moby/buildkit/client"
+	"github.com/moby/buildkit/util/grpcerrors"
 	"github.com/pkg/errors"
+	"google.golang.org/grpc/codes"
 )
 
 var ErrNotRunning = errors.Errorf("driver not running")
@@ -88,4 +91,27 @@ func Boot(ctx, clientContext context.Context, d Driver, pw progress.Writer) (*cl
 		}
 		return c, nil
 	}
+}
+
+func HistoryAPISupported(ctx context.Context, c *client.Client) (res bool) {
+	res = true
+	checkErrF := func(err error) {
+		if s, ok := grpcerrors.AsGRPCStatus(err); ok {
+			if s.Code() == codes.Unimplemented {
+				res = false
+			}
+		}
+	}
+	cl, err := c.ControlClient().ListenBuildHistory(ctx, &controlapi.BuildHistoryRequest{
+		ActiveOnly: true,
+		Ref:        "buildx-dummy-ref", // dummy ref to check if the server supports the API
+		EarlyExit:  true,
+	})
+	if err != nil {
+		checkErrF(err)
+		return
+	}
+	_, err = cl.Recv()
+	checkErrF(err)
+	return
 }
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -60,7 +60,7 @@ type Driver interface {
 	Stop(ctx context.Context, force bool) error
 	Rm(ctx context.Context, force, rmVolume, rmDaemon bool) error
 	Client(ctx context.Context) (*client.Client, error)
-	Features() map[Feature]bool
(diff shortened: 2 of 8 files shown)


driver/kubernetes/driver.go
```diff
@@ -229,11 +229,18 @@ func (d *Driver) Factory() driver.Factory {
...
+		c.Close()
+	}
 	return map[driver.Feature]bool{
 		driver.OCIExporter:    true,
 		driver.DockerExporter: d.DockerAPI != nil,
-
-		driver.CacheExport:   true,
-		driver.MultiPlatform: true, // Untested (needs multiple Driver instances)
+		driver.CacheExport:    true,
+		driver.MultiPlatform:  true, // Untested (needs multiple Driver instances)
```
jedevc, line 243: Unrelated, but I think this comment is out-of-date now?

I think this is tested now, at least we have docs for this scenario?

driver/remote/driver.go
```diff
@@ -88,11 +88,19 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
 }
 
 func (d *Driver) Features() map[driver.Feature]bool {
+	var historyAPI bool
+	ctx := context.Background()
```
jedevc, line 92: Hm, since `Features` now will pretty much *always* make external requests, it feels like we should probably just modify the interface to take a `Context` instead of using the background? It's not great that this operation wouldn't cancel properly, we also do it from inside `toSolveOpt`, so we've always got a context there.

jedevc (dismissed):

driver/kubernetes/driver.go
crazy-max: @AkihiroSuda @morlay

driver/remote/driver.go
jedevc: Ignore me. That's a terrible idea - if the call fails, the feature should just be marked as false.

driver/driver.go
```diff
@@ -89,3 +92,15 @@ func Boot(ctx, clientContext context.Context, d Driver, pw progress.Writer) (*cl
 		return c, nil
 	}
 }
+
+func HistoryAPISupported(ctx context.Context, c *client.Client) bool {
+	res := true
+	if _, err := c.ControlClient().UpdateBuildHistory(ctx, &controlapi.UpdateBuildHistoryRequest{}); err != nil {
```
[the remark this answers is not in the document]
tonistiigi: I think dockerd doesn't enable reflection. 😢 

What should work would be `ListenBuildHistory` with a dummy ref, activeonly=true and earlyExit=true. That should basically also be a no-op.

Still error log on old servers, but that is not that big of a deal.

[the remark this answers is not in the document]
crazy-max: I guess we need to enable reflection in client when creating the gRPC connection: https://github.com/moby/buildkit/blob/61e6f28bcf3e19f4f4f5d90761624d7478b42da4/client/client.go#L149-L152

```go
conn, err := grpc.DialContext(ctx, address, gopts...)
if err != nil {
	return nil, errors.Wrapf(err, "failed to dial %q . make sure buildkitd is running", address)
}
reflection.Register(conn)
```

and make it optional in `client.ClientOpt` with smth like `WithReflection`.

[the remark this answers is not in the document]
tonistiigi: No, on client side you would just need to create `NewServerReflectionClient`. But see my comment above on why the reflection approach doesn't work.

driver/driver.go
```diff
@@ -89,3 +92,26 @@ func Boot(ctx, clientContext context.Context, d Driver, pw progress.Writer) (*cl
...
+	checkErrF := func(err error) {
+		if s, ok := grpcerrors.AsGRPCStatus(err); ok {
+			if s.Code() == codes.Unimplemented {
+				res = false
+			}
+		}
+	}
+	cl, err := c.ControlClient().ListenBuildHistory(ctx, &controlapi.BuildHistoryRequest{
+		ActiveOnly: true,
+		Ref:        "0123456789-abcdefghijklmn", // dummy ref to check if the server supports the API
```
jedevc, line 107: *super nitpick*: can we set something that contains the string `buildx-dummy-ref` or similar, so it's ultra-clear if this ends up in a buildx/buildkit error report?

jedevc (approved):

driver/driver.go
```diff
@@ -89,3 +92,26 @@ func Boot(ctx, clientContext context.Context, d Driver, pw progress.Writer) (*cl
...
+	checkErrF := func(err error) {
+		if s, ok := grpcerrors.AsGRPCStatus(err); ok {
+			if s.Code() == codes.Unimplemented {
+				res = false
+			}
+		}
+	}
+	cl, err := c.ControlClient().ListenBuildHistory(ctx, &controlapi.BuildHistoryRequest{
+		ActiveOnly: true,
+		Ref:        "buildx-dummy-ref", // dummy ref to check if the server supports the API
```
tonistiigi, line 107: "buildx-test-history-api-feature" so that if you look at it from server side you understand where the request is coming from.

driver/driver.go
```diff
@@ -89,3 +92,26 @@ func Boot(ctx, clientContext context.Context, d Driver, pw progress.Writer) (*cl
 		return c, nil
 	}
 }
+
+func HistoryAPISupported(ctx context.Context, c *client.Client) (res bool) {
+	res = true
+	checkErrF := func(err error) {
+		if s, ok := grpcerrors.AsGRPCStatus(err); ok {
+			if s.Code() == codes.Unimplemented {
```
tonistiigi, line 100: What error except for EOF in here means there is support?

driver/driver.go
```diff
@@ -89,3 +92,26 @@ func Boot(ctx, clientContext context.Context, d Driver, pw progress.Writer) (*cl
...
+	cl, err := c.ControlClient().ListenBuildHistory(ctx, &controlapi.BuildHistoryRequest{
+		ActiveOnly: true,
+		Ref:        "buildx-dummy-ref", // dummy ref to check if the server supports the API
+		EarlyExit:  true,
+	})
+	if err != nil {
+		checkErrF(err)
+		return
+	}
+	_, err = cl.Recv()
```
tonistiigi, line 114: Client should be read until it is empty. It can be closed/canceled if there is an issue.

driver/docker-container/driver.go
```diff
@@ -387,13 +387,19 @@ func (d *Driver) Factory() driver.Factory {
 	return d.factory
 }
 
-func (d *Driver) Features() map[driver.Feature]bool {
+func (d *Driver) Features(ctx context.Context) map[driver.Feature]bool {
+	var historyAPI bool
+	c, err := d.Client(ctx)
```
tonistiigi, line 392: This call is expensive. It makes an new `docker exec`.

crazy-max: The client is cached https://github.com/docker/buildx/blob/a906149930ebe4dbd33d96bd42e157a661956d4c/driver/manager.go#L157-L162 so we are probably fine here no?

tonistiigi:
That is not the struct that gets called in line 392.

driver/docker-container/driver.go
tonistiigi: That is not the struct that's get called in the line 392.

Issue #1847, opened 2023-05-26
opentelemetry support does not send buildx spans


nicks:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

I followed the instructions here:
https://docs.docker.com/build/building/opentelemetry/
but did not get any buildx spans.

### Expected behaviour

Both buildx and buildkitd spans

### Actual behaviour

Only buildkitd spans

### Buildx version

HEAD

### Docker info

```text
n/a
```

### Builders list

```text
n/a
```

### Configuration

n/a

### Additional info

we forgot to import the exporter :sob:

jedevc:
Out of curiosity, which version of buildx was this on? Is it just on master, or did you see this in the 0.10 release?

PR #1848, opened 2023-05-26
tracing: add the jaeger exporter

nicks:
once this env variable is set, the
the docs in https://docs.docker.com/build/building/opentelemetry/ will work

fixes https://github.com/docker/buildx/issues/1847


nicks:
actually poking around a bit more at the delegated tracer, i don't think this is right, i think the tracer delegation isn't work right...still digging...

nicks:
ok, i see the real problem, closing this PR

PR #1850, opened 2023-05-29
monitor: add long help for commands

ktock:
This commit adds long help messages about the usage of  the monitor commands .

example:
```console
(buildx) help attach
attach to a buildx server or a process in the container

Usage:
  attach ID

ID is for a session (visible via list command) or a process (visible via ps command).
If you attached to a process, use Ctrl-c-a for swithcing the monitor to that process's STDIO.
```

### DIFF
--- a/monitor/commands/kill.go
+++ b/monitor/commands/kill.go
@@ -16,7 +16,16 @@ func NewKillCmd(m types.Monitor) types.Command {
 }
 
 func (cm *KillCmd) Info() types.CommandInfo {
-	return types.CommandInfo{HelpMessage: "kill buildx server"}
+	return types.CommandInfo{
+		Name:        "kill",
+		HelpMessage: "kill buildx server",
+		HelpMessageLong: `
+Usage:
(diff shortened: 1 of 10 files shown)


monitor/commands/kill.go
```diff
@@ -16,7 +16,16 @@ func NewKillCmd(m types.Monitor) types.Command {
...
 func (cm *KillCmd) Info() types.CommandInfo {
-	return types.CommandInfo{HelpMessage: "kill buildx server"}
+	return types.CommandInfo{
+		Name:        "kill",
+		HelpMessage: "kill buildx server",
+		HelpMessageLong: `
+Usage:
+  kill
+
+Kills the currently connecting buildx server process.
```
tonistiigi, line 26: Not a docs comment but not sure if "kill" is the best name in here. Kill should be "send a signal". Otherwise, it should be "stop" or "shutdown".

tonistiigi (approved):

PR #1851, opened 2023-05-29
test: update workflow

crazy-max:
See individual commits.

I keep the `test` bake target for now. I think it's still useful to run unit tests without make or shell script.


hack/test
```diff
@@ -3,60 +3,32 @@
...
 fi
-
-if [ "$TEST_INTEGRATION" == 1 ]; then
-  cid=$(docker create --rm -v /tmp $testReportsVol --volumes-from=$cacheVolume -e GITHUB_REF -e TEST_DOCKERD -e TEST_BUILDKIT_IMAGE -e SKIP_INTEGRATION_TESTS -e GOTESTSUM_FORMAT ${BUILDKIT_INTEGRATION_SNAPSHOTTER:+"-eBUILDKIT_INTEGRATION_SNAPSHOTTER"} -e BUILDKIT_REGISTRY_MIRROR_DIR=/root/.cache/registry --privileged $iid gotestsum $gotestsumArgs --packages="${TESTPKGS:-./...}" -- $gotestArgs ${TESTFLAGS:--v})
-  docker start -a -i $cid
-fi
-
 if [ "$TEST_KEEP_CACHE" != "1" ]; then
-  docker rm -v $cacheVolume
+  trap 'docker rm -v $cacheVolume' EXIT
```
jedevc, line 30: FYI - this was pulled from buildkit's `./hack/test`, so maybe we need an update there as well?

crazy-max: Yes indeed I was looking at making some changes in BuildKit related to this :+1:

hack/test
```diff
@@ -23,20 +22,6 @@ if [ -n "$CACHE_TO" ]; then
...
-  case "$1" in
-    integration)
-      TEST_INTEGRATION=1
-      ;;
-    *)
-      echo "unknown arg $1"
-      ;;
-  esac
-  shift
-done
```
jedevc: I think I'd like to keep this if possible, we still default to `integration` if nothing else is specified.

I think we're going to need to create another suite of integration tests at some point to test the complexities of driver creation, e.g. there are a lot of options for setting up the kubernetes driver. These tests don't test the `commands` package directly, so it would make more sense to split it out, so we can run them in parallel etc.

hack/test
```diff
@@ -3,60 +3,32 @@
...
```
jedevc, line 38: I think I'd like to keep this if possible, we still default to integration if nothing else is specified.

I think we're going to need to create another suite of integration tests at some point to test the complexities of driver creation, e.g. there are a lot of options for setting up the kubernetes driver. These tests don't test the commands package directly, so it would make more sense to split it out, so we can run them in parallel etc.

crazy-max: > I think I'd like to keep this if possible, we still default to integration if nothing else is specified.

On BuildKit we needed this special case because we got to test the frontend specifically and also the gateway and I don't think we need to have this kind of logic in buildx.

> I think we're going to need to create another suite of integration tests at some point to test the complexities of driver creation, e.g. there are a lot of options for setting up the kubernetes driver. These tests don't test the commands package directly, so it would make more sense to split it out, so we can run them in parallel etc.

Right we just invoke buildx binary for the inspect command integration test atm so we don't cover the `commands` package but don't think we need other kind of invocation of the integrations tests container like we do in BuildKit.

jedevc: > On BuildKit we needed this special case because we got to test the frontend specifically and also the gateway and I don't think we need to have this kind of logic in buildx.

I think our special case will come from wanting to test the commands specifically, and then also test the driver creation specifically. We'll probably want different types of environment for both (e.g. the driver creation tests will probably need to configure buildkit servers with different combinations of tls options, or spin up unique k8s clusters per test).

We might not do that though, so I think it's ok to simplify this for now, and we can always add it back in later.

jedevc (approved):

PR #1858, opened 2023-05-31
Fix typo "Shortand" -> "Shorthand"

Potherca:
This typo was present in two files.

This MR fixes both.


tonistiigi:
The part of docs is generated and should not be updated directly. Run `make docs` and check if CI goes green.

tonistiigi (approved):

PR #1861, opened 2023-06-01
bake: switch to controller

crazy-max:
Carry https://github.com/docker/buildx/pull/1680 to handle controller options with bake and also use controller build. This will be used in #1841 so we can pass the build reference to the result context for both build and bake efficiently.


controller/build/build.go
```diff
@@ -23,43 +23,137 @@ import (
...
+	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/util/grpcerrors"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 )
 
-const defaultTargetName = "default"
+const DefaultTargetName = "default"
```
ktock, line 34: We don't need to make it public.

controller/build/build.go
```diff
@@ -23,43 +23,137 @@ import (
...
 
 // RunBuild runs the specified build and returns the result.
 //
 // NOTE: When an error happens during the build and this function acquires the debuggable *build.ResultContext,
 // this function returns it in addition to the error (i.e. it does "return nil, res, err"). The caller can
 // inspect the result and debug the cause of that error.
-func RunBuild(ctx context.Context, dockerCli command.Cli, in controllerapi.BuildOptions, inStream io.Reader, progress progress.Writer, generateResult bool) (*client.SolveResponse, *build.ResultContext, error) {
-	if in.NoCache && len(in.NoCacheFilter) > 0 {
-		return nil, nil, errors.Errorf("--no-cache and --no-cache-filter cannot currently be used together")
+func RunBuild(ctx context.Context, dockerCli command.Cli, builderName string, in *controllerapi.BuildOptions, inStream io.Reader, progress progress.Writer, generateResult bool) (*client.SolveResponse, *build.ResultContext, error) {
```
ktock, line 41: Why do we need to have a separated `builderName` option? This looks already included in the `controllerapi.BuildOptions`.

controller/build/build.go
crazy-max: Yes indeed, this is an oversight related to previous changes.

controller/build/build.go
crazy-max: I was looking at passing just the list of nodes necessary to build so we don't create another builder instance. But when building multiple targets controller.BuildOptions is a map and I don't think having Builder there is right: https://github.com/docker/buildx/blob/0758a9b75dcb7de2cb89944f56afe1151e06f176/controller/pb/controller.proto#L74
I will take another look to improve this.

crazy-max:
@ktock @jedevc Bake does not handle the context path hash like build (we just take the current work dir atm). This is also just specific to the k8s driver and should move there. I think we already discuss about it but can't put my finger on it.

PR #1865, opened 2023-06-06
vendor: github.com/Masterminds/semver/v3 v3.2.1

crazy-max:
https://github.com/Masterminds/semver/compare/v3.2.0...v3.2.1


thaJeztah:
Orthogonal to this PR, but I guess our hope is to get rid of this dependency eventually (as this is only to fill-in versions for old / legacy versions of the Docker Engine).

Looking at https://github.com/docker/buildx/blob/ea06685c113ce4a0b182c6932b43c36297f11061/driver/docker/version.go#L169-L174

And the list of versions we check against;
https://github.com/docker/buildx/blob/ea06685c113ce4a0b182c6932b43c36297f11061/driver/docker/version.go#L14-L26


Can we;

- reverse the order of versions in that list, so that we (assuming most people run a "recent version" of the engine), can return early?
- should we consider removing the very old (and EOL) versions of the engine? BuildKit was still _experimental _in 18.06 (and that version of the Engine has been EOL for 5 years)
- does the API itself contain a version? Is this also connecting to the Engine API (and if so, has a `/_ping` endpoint)? Because in that case, we can use the ping-result to skip this version check altogether if the API version is v1.43 (v1.44?) or up. For those versions we already set the correct version, so there's no need to do a mapping.

thaJeztah (approved):
LGTM

thaJeztah:
> should we consider removing the very old (and EOL) versions of the engine?

We should at least consider removing all the old `-beta` and `-rc` versions from there.

crazy-max:
> reverse the order of versions in that list, so that we (assuming most people run a "recent version" of the engine), can return early?

`resolveBuildKitVersion` is not called for Docker > 23. It will just use the API (available since BuildKit 0.11 and Docker 24)

> We should at least consider removing all the old `-beta` and `-rc` versions from there.

I don't mind, it was just for completeness.

crazy-max:
@thaJeztah Looking at the returned format here like https://github.com/docker/buildx/blob/ea06685c113ce4a0b182c6932b43c36297f11061/driver/docker/version.go#L161

It would have been nice to have the same in moby: https://github.com/moby/moby/blob/8e70a1b23e965d86ec8c2feb77605196ae124630/builder/builder-next/worker/worker.go#L53

When doing `docker buildx ls`, it's quite huge:

```
NAME/NODE            DRIVER/ENDPOINT                                       STATUS   BUILDKIT                              PLATFORMS
builder              docker-container
  builder0           default                                               inactive
builder2             docker-container
  builder20          unix:///var/run/docker.sock                           stopped
buildkit0106         docker-container
  buildkit01060      unix:///var/run/docker.sock                           inactive
buildkit093          docker-container
  buildkit0930       unix:///var/run/docker.sock                           inactive
ctn-builder          docker-container
  ctn-builder0       unix:///var/run/docker.sock                           inactive
dk-remote-builder    docker-container
  dk-remote-builder0 unix:///var/run/docker.sock                           inactive
default *            docker
  default            default                                               running  v0.11.7-0.20230525183624-798ad6b0ce9f linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
desktop-linux                                                              error
docker2010           docker
  docker2010         docker2010                                            running  20.10.23                              linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```

thaJeztah:
Ah; I found that `Version()` is called "conditionally" indeed https://github.com/docker/buildx/blob/ea06685c113ce4a0b182c6932b43c36297f11061/builder/node.go#L200


.... that's slightly confusing;

- `Version()` is part of the interface for drivers https://github.com/docker/buildx/blob/ea06685c113ce4a0b182c6932b43c36297f11061/driver/driver.go#L58
- ... but the _actual_ version used is `node.Version` (not `node.Driver.Version`) https://github.com/docker/buildx/blob/ea06685c113ce4a0b182c6932b43c36297f11061/builder/node.go#L207
- ... and there's also a `Driver.Info()` which returns "some info", but later on we manually collect "some other info as well" https://github.com/docker/buildx/blob/ea06685c113ce4a0b182c6932b43c36297f11061/builder/node.go#L186-L196

So it feels like `Driver.Version()` is effectively "deprecated" (but the thing described in the interface). Wondering if we can somehow;

- Make `Version()` do what it needs to do (if the driver can provide `inf.BuildkitVersion.Version`, then make it return that as version)
- Roll-up the custom collecting of extra (worker) properties as part of `Driver.Info()` (perhaps this needs (optional) arguments to be added to `Info()` to collect extra data if it's not "without cost")

thaJeztah:
> When doing docker buildx ls, it's quite huge:

Well, for v23.0 the fix would be easy; BuildKit to tag releases for v0.10 😉 

@crazy-max As to the output format, I guess that relates to https://github.com/moby/moby/pull/45582#discussion_r1198853211

I picked the version format that Go modules use; wouldn't that be the format that's set with https://github.com/moby/moby/pull/45304 as well?

crazy-max:
> wouldn't that be the format that's set with [moby/moby#45304](https://github.com/moby/moby/pull/45304) as well?

It slightly differs, see https://github.com/moby/moby/pull/45304/files#diff-efb84f7c8d09340a77b30fffc9da9f30fd2f7e423afdb9d14f2ab2848b8d18e9R48

> As to the output format, I guess that relates to [moby/moby#45582 (comment)](https://github.com/moby/moby/pull/45582#discussion_r1198853211)

Ah yes we have not backported this to 24. Forgot about it.

thaJeztah:
Oh! I thought it would use the default module version format; which is why I used that (didn't like it, but thought that was what the new version would look like); feel free to open a PR to update it to look the same (perhaps a comment added that describes how to manually construct the format)

PR #1869, opened 2023-06-07
builder: return error if all nodes fail to boot

crazy-max:
fixes https://github.com/docker/setup-buildx-action/issues/228#issuecomment-1510707981

Before:

```
#1 [internal] booting buildkit
#1 pulling image moby/buildkit:buildx-stable-1
#1 pulling image moby/buildkit:buildx-stable-1 1.6s done
#1 creating container buildx_buildkit_builder-cc8757dc-a94d-4ea0-ac4a-592280318e710 done
#1 ERROR: Error response from daemon: Windows does not support privileged mode
------
 > [internal] booting buildkit:
------
Name:          builder-cc8757dc-a94d-4ea0-ac4a-592280318e71
Last Activity: 2023-10-23 12:13:15 +0000 UTC

Nodes:
Name:      builder-cc8757dc-a94d-4ea0-ac4a-592280318e710
Endpoint:  npipe:////./pipe/docker_engine
Status:    inactive
Flags:     --allow-insecure-entitlement security.insecure --allow-insecure-entitlement network.host
Platforms:
```

Now:

```
#1 [internal] booting buildkit
pedantic_hopper
#1 pulling image moby/buildkit:buildx-stable-1
#1 pulling image moby/buildkit:buildx-stable-1 1.4s done
#1 creating container buildx_buildkit_pedantic_hopper0 0.0s done
#1 ERROR: Error response from daemon: Windows does not support privileged mode
------
 > [internal] booting buildkit:
------
ERROR: Error response from daemon: Windows does not support privileged mode
```

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -165,6 +165,7 @@ func (b *Builder) Boot(ctx context.Context) (bool, error) {
 
 	baseCtx := ctx
 	eg, _ := errgroup.WithContext(ctx)
+	errCh := make(chan error, len(toBoot))
 	for _, idx := range toBoot {
 		func(idx int) {
 			eg.Go(func() error {
@@ -172,6 +173,7 @@ func (b *Builder) Boot(ctx context.Context) (bool, error) {
 				_, err := driver.Boot(ctx, baseCtx, b.nodes[idx].Driver, pw)
 				if err != nil {
 					b.nodes[idx].Err = err
+					errCh <- err
 				}
 				return nil
 			})
@@ -179,11 +181,15 @@ func (b *Builder) Boot(ctx context.Context) (bool, error) {
 	}
 
 	err = eg.Wait()
+	close(errCh)
 	err1 := printer.Wait()
 	if err == nil {
 		err = err1
 	}
 
+	if err == nil && len(errCh) == len(toBoot) {
+		return false, <-errCh
+	}
 	return true, err
 }
 


builder/builder.go
```diff
@@ -165,25 +165,34 @@ func (b *Builder) Boot(ctx context.Context) (bool, error) {
...
 	for _, idx := range toBoot {
 		func(idx int) {
 			eg.Go(func() error {
 				pw := progress.WithPrefix(printer, b.NodeGroup.Nodes[idx].Name, len(toBoot) > 1)
 				_, err := driver.Boot(ctx, baseCtx, b.nodes[idx].Driver, pw)
 				if err != nil {
 					b.nodes[idx].Err = err
+					mu.Lock()
+					errCh <- err
+					mu.Unlock()
```
jsternberg, line 179: Is the mutex here needed? I don't think this requires a mutex because `errCh` is written to outside of the area where it's used and it doesn't seem to be changed or modified until after the error group is finished.

Also `errCh` is initialized with a buffer the size of the number of nodes to boot so this should be non-blocking and a-ok without any other code.

jsternberg (changes requested):

builder/builder.go
crazy-max: Ah my bad yes there should be no contention, and it won't block until the buffer is full.

jsternberg (approved):

PR #1876, opened 2023-06-13
avoid extra client for history API detection

tonistiigi:
This makes sure there is no extra client initialization/request unless it is actually needed to detect.

We could also return the history API indicator with the build result or add a cap for it. Actually, if we want to be really sneaky we could use another cap like `source.policy` that was also added in v0.11 to detect this.


crazy-max (approved):
PTAL @jedevc

jedevc (approved):

PR #1879, opened 2023-06-14
builder: skip name validation for docker context

crazy-max:
related to https://github.com/docker/for-win/issues/13543

Although a builder from the store cannot be created unless it has a valid name, this is not the case for a Docker context.

We should skip name validation when checking a node from the store and fall back to finding one from Docker context instead.


store/store_test.go
```diff
@@ -235,3 +235,19 @@ func TestNodeManagement(t *testing.T) {
...
+	s, err := New(tmpdir)
+	require.NoError(t, err)
+
+	txn, release, err := s.Txn()
+	require.NoError(t, err)
+	defer release()
+
+	_, err = txn.NodeGroupByName("123builder")
+	require.Error(t, err)
+	require.True(t, IsErrInvalidName(err))
```
jedevc, line 252: Could we add a similar test case for `txn.Save`?

store/storeutil/storeutil.go
```diff
@@ -74,7 +74,7 @@ func GetCurrentInstance(txn *store.Txn, dockerCli command.Cli) (*store.NodeGroup
 func GetNodeGroup(txn *store.Txn, dockerCli command.Cli, name string) (*store.NodeGroup, error) {
 	ng, err := txn.NodeGroupByName(name)
 	if err != nil {
-		if !os.IsNotExist(errors.Cause(err)) {
+		if !os.IsNotExist(errors.Cause(err)) && !store.IsErrInvalidName(err) {
```
jedevc, line 77: `errors.Is`?

crazy-max: Would need to make the interface public and also error contains a dynamic value.

jedevc: Yeah good point, that sounds alright for now. We're not consistent on styles in buildx anyways, so :shrug:

crazy-max: Agree we should improve this in our code base with well defined error types.

jedevc (approved):

thaJeztah:
Interesting; it looks like context-store in docker/cli _does_ have a validation function; https://github.com/docker/buildx/blob/bd672eaf5bf6d7c8660d4d469f43770085aa5963/vendor/github.com/docker/cli/cli/context/store/store.go#L214-L226

https://github.com/docker/buildx/blob/bd672eaf5bf6d7c8660d4d469f43770085aa5963/vendor/github.com/docker/cli/cli/context/store/store.go#L22-L24


But I was actually looking at other parts of the code, and it looks like validation is done externally 😞 https://github.com/docker/cli/blob/085d5c281612298583d19806aa8ec637b572e4cc/cli/command/context/create.go#L124-L135

```go
func checkContextNameForCreation(s store.Reader, name string) error {
	if err := store.ValidateContextName(name); err != nil {
		return err
	}
	if _, err := s.GetMetadata(name); !errdefs.IsNotFound(err) {
		if err != nil {
			return errors.Wrap(err, "error while getting existing contexts")
		}
		return errors.Errorf("context %q already exists", name)
	}
	return nil
}
```

https://github.com/docker/cli/blob/085d5c281612298583d19806aa8ec637b572e4cc/cli/command/context/create.go#L71-L76

```go
// RunCreate creates a Docker context
func RunCreate(cli command.Cli, o *CreateOptions) error {
	s := cli.ContextStore()
	err := checkContextNameForCreation(s, o.Name)
	if err != nil {
		return err
	}
	swit
```

So I wonder if there's a bug at hand here, where validation was missed?

crazy-max:
Oh good catch @thaJeztah. Yeah there's smth fishy here :thinking:

thaJeztah:
Yeah, was looking "why isn't the validation done by the context _store_ code?", and then saw this PR fly by in my notifications 😂 

Is the case here that BuildKit / buildx can use _either_ hostname or context-name, and doesn't know which is the case?

Issue #1627, opened 2023-02-14
[BUG] Bake build fails when using BAKE_CMD_CONTEXT locally


nVlast:
I noticed that when I use the `BAKE_CMD_CONTEXT` built-in variable in a local HCL bake file, the build fails.

**Steps to reproduce:**
1. Create a Dockerfile (named "Dockerfile"):
  ```Dockerfile
  FROM alpine
  CMD ["sleep", "infinity"]
  ```
2. Create a bake file (named "docker-bake.hcl"):
  ```HCL
  target "default" {
    dockerfile = "Dockerfile"
    context = BAKE_CMD_CONTEXT
    tags = ["myimage:myversion"]
  }
  ```
3. Run `docker buildx bake`

**Expected outcome:**
The docker image specified should be built

**Actual outcome:**
```
ERROR: could not find cwd:: stat cwd:: no such file or directory
```

---

I looked into it a bit further, and it seems that when the aforementioned built-in variable was introduced, while the context [was arranged to be "cleaned"](https://github.com/docker/buildx/pull/671/files#diff-fbe2fa23203f2ff05f06b6c0731ea90529f90906807483db76a79f6018781465R581-R585) from the [cwd protocol](https://github.com/docker/buildx/pull/671/files#r668896194), the dockerfilePath is left with a cwd reference in it. Perhaps something as simple as adding a `!strings.HasPrefix(contextPath, "cwd://")` check [here](https://github.com/docker/buildx/pull/671/files#diff-fbe2fa23203f2ff05f06b6c0731ea90529f90906807483db76a79f6018781465R560) would be sufficient but take that with a pinch of salt as I'm not all that familiar with the codebase.

Also, if this variable was never meant to work in a local setting, please close this issue.

crazy-max:
@nVlast Yes `BAKE_CMD_CONTEXT` has to be used with a remote definition as shown here: https://docs.docker.com/build/bake/file-definition/#remote-definition but it should not fail for local builds.

I'm able to repro locally but `--print` gives the expected value for `context`:

```
$ docker buildx bake --print
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "myimage:myversion"
      ]
    }
  }
}
```

Will take a look, thanks for your feedback.

jedevc:
[...]

PR #1880, opened 2023-06-14
bake: fix incorrect dockerfile resolution against `cwd://` context

jedevc:
:hammer_and_wrench: Fixes https://github.com/docker/buildx/issues/1627

We need to resolve the strip the `cwd://` prefix before attempting to resolve the dockerfile. Otherwise, we'll get the `cwd://` prefix in the dockerfile name, which isn't stripped out later.

Also added this to the test case, where previously we would get a `cwd://` in the dockerfile name, now we correctly resolve it.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -1048,11 +1048,11 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 		bi.DockerfileInline = *t.DockerfileInline
 	}
 	updateContext(&bi, inp)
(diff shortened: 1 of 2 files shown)


crazy-max (approved):

tonistiigi:
I think the old behavior is intentional. It is for the use case where you have docker build logic only in the remote repository and source files in local dir. In that case, the Dockerfile and Bake file are stored together (similar to using `dockerfile_inline` in bake). If a remote Bake file also wants to load Dockerfile from local cmd dir it should use `cwd://` there as well.

If someone can prove this is a regression that would change things.

Issue #1175, opened 2022-06-21
imagetools inspect `--format` ignores formatting unless `json` is used


thaJeztah:
Was writing a reply to https://github.com/docker/hub-feedback/issues/2043, and noticed this.

Reproduced on;

```bash
docker buildx version
github.com/docker/buildx v0.8.2-docker 6224def4dd2c3d347eee19db595348c50d7cb491
```

When using the `-f` / `--format` option, the template is ignored, unless it's using the `json` function;

```bash
docker buildx imagetools inspect --format '{{.Manifest.Digest}}' tensorflow/tensorflow:1.13.1-gpu-py3
Name:      docker.io/tensorflow/tensorflow:1.13.1-gpu-py3
MediaType: application/vnd.docker.distribution.manifest.v2+json
Digest:    sha256:0f949ccc690d9c50e9b46b16d9030c2c6845af621c2e7e82c4bf59803edc69b5
```

```bash
docker buildx imagetools inspect --format '{{json .Manifest.Digest}}' tensorflow/tensorflow:1.13.1-gpu-py3
"sha256:0f949ccc690d9c50e9b46b16d9030c2c6845af621c2e7e82c4bf59803edc69b5"
```

This is inconsistent with other commands, for example, here's `docker image inspect` (difference between using `json` and without is that the `json` output uses JSON quoting for the output);

```bash
docker image inspect -f '{{.ID}}' busybox
sha256:62aedd01bd8520c43d06b09f7a0f67ba9720bdc04631a8242c65ea995f3ecac8

docker image inspect -f '{{json .ID}}' busybox
"sha256:62aedd01bd8520c43d06b09f7a0f67ba9720bdc04631a8242c65ea995f3ecac8"
```

Also note that the `json` output does not print a newline at the end, which can make it a bit cumbersome to use; here's the full output (notice that the prompt is appended to the output, because there's no trailing newline)

```bash
root@swarm-test-01:~# docker buildx imagetools inspect --format '{{json .Manifest.Digest}}' tensorflow/tensorflow:1.13.1-gpu-py3
"sha256:0f949ccc690d9c50e9b46b16d9030c2c6845af621c2e7e82c4bf59803edc69b5"root@swarm-test-01:~#
```

crazy-max:
@thaJeztah Since https://github.com/docker/cli/pull/2936 has been merged I will revisit the format output to be aligned.

thaJeztah:
[...]

PR #1882, opened 2023-06-14
imagetools: use docker/cli formatting

uhthomas:
Use the formatting packages provided by docker/cli to comply with the official formatting documentation.

https://docs.docker.com/config/formatting/

The use of the official formatting packages also fixes the format templating.

Fixes: #1175


uhthomas:
@thaJeztah, this isn't finished but I thought it would be good to create a draft PR for some early feedback. I was initially just going to fix the formatting directive, but thought it would be nicer to make use of [github.com/docker/cli/cli/command/inspect](https://pkg.go.dev/github.com/docker/cli/cli/command/inspect) and [github.com/docker/cli/cli/command/formatter](https://pkg.go.dev/github.com/docker/cli/cli/command/formatter). Using the formatting package should result in a lot less duplicate code, and would be consistent with `docker/cli`.

What are your thoughts?

PR #1886, opened 2023-06-15
build: prefer local image resolution for docker driver

crazy-max:
If pull option is not set we should prefer local image resolution for docker driver. In follow-up we should look at making the `--pull` flag accepting a specific state like the `docker run` command does: https://docs.docker.com/engine/reference/commandline/run/#pull

### DIFF
--- a/tests/build.go
+++ b/tests/build.go
@@ -32,6 +32,7 @@ var buildTests = []func(t *testing.T, sb integration.Sandbox){
 	testBuildLocalExport,
 	testBuildRegistryExport,
 	testBuildTarExport,
+	testBuildMobyFromLocalImage,
 }
 
 func testBuild(t *testing.T, sb integration.Sandbox) {
@@ -137,6 +138,53 @@ func testImageIDOutput(t *testing.T, sb integration.Sandbox) {
 
 	require.NotEmpty(t, md.ConfigDigest)
 	require.Equal(t, dgst, digest.Digest(md.ConfigDigest))
+}
+
+func testBuildMobyFromLocalImage(t *testing.T, sb integration.Sandbox) {
+	if !isDockerWorker(sb) {
+		t.Skip("skipping test for non-docker workers")
+	}
+
+	// pull image
+	cmd := dockerCmd(sb, withArgs("pull", "-q", "busybox:latest"))
+	stdout := bytes.NewBuffer(nil)
+	cmd.Stdout = stdout
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
+	require.Equal(t, "docker.io/library/busybox:latest", strings.TrimSpace(stdout.String()))
+
+	// create local tag
+	cmd = dockerCmd(sb, withArgs("tag", "busybox:latest", "buildx-test:busybox"))
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
+
+	// build image
+	dockerfile := []byte(`FROM buildx-test:busybox`)
+	dir := tmpdir(t, fstest.CreateFile("Dockerfile", dockerfile, 0600))
+	cmd = buildxCmd(
+		sb,
+		withArgs("build", "-q", "--output=type=cacheonly", dir),
+	)
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
+
+	// create local tag matching a remote one
+	cmd = dockerCmd(sb, withArgs("tag", "busybox:latest", "busybox:1.36"))
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
+
+	// build image and check that it uses the local tag
+	dockerfile = []byte(`
+FROM busybox:1.36
+RUN busybox | head -1 | grep v1.35.0
+`)
+	dir = tmpdir(t, fstest.CreateFile("Dockerfile", dockerfile, 0600))
+	cmd = buildxCmd(
+		sb,
+		withArgs("build", "-q", "--output=type=cacheonly", dir),
+	)
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
 }
 
 func createTestProject(t *testing.T) string {
--- a/build/build.go
+++ b/build/build.go
@@ -593,7 +593,10 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	}
 
 	if opt.Pull {
-		so.FrontendAttrs["image-resolve-mode"] = "pull"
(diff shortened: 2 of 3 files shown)


tonistiigi (approved):

tests/build.go
```diff
@@ -139,6 +140,35 @@ func testImageIDOutput(t *testing.T, sb integration.Sandbox) {
 	require.Equal(t, dgst, digest.Digest(md.ConfigDigest))
 }
 
+func testBuildMobyFromLocalImage(t *testing.T, sb integration.Sandbox) {
+	if sb.Name() != "docker" {
+		t.Skip("skipping test for non-docker workers")
+	}
+
+	// pull image
+	cmd := dockerCmd(sb, withArgs("pull", "-q", "busybox:latest"))
```
tonistiigi, line 149: This will probably step over the local mirrors support in tests that avoid new pulls for every test. We could either connect the local mirrors to dockerd conf as well or  maybe rewrite the test so it doesn't require pull and instead does a "build + tag" as a setup.

thaJeztah:
>  In follow-up we should look at making the --pull flag accepting a specific state like the docker run command does: https://docs.docker.com/engine/reference/commandline/run/#pull


- @crazy-max should I transfer https://github.com/docker/cli/issues/2253 to this repository?

tests/build.go
crazy-max: > We could either connect the local mirrors to dockerd conf

It looks already handled afaik: https://github.com/moby/buildkit/blob/cb86cc619256af8b3e6c9191b094d280fd804f86/util/testutil/integration/dockerd.go#L88-L92

tests/build.go
```diff
@@ -139,6 +140,35 @@ func testImageIDOutput(t *testing.T, sb integration.Sandbox) {
 	require.Equal(t, dgst, digest.Digest(md.ConfigDigest))
 }
 
+func testBuildMobyFromLocalImage(t *testing.T, sb integration.Sandbox) {
```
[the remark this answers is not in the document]
crazy-max: > Do we have the capability to test docker driver in both storage modes?

Not currently, will look into it.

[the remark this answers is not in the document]
crazy-max: > Does this test fail in the current master? I thought this was only broken for containerd storage?

Indeed it works fine with graphdriver: https://github.com/crazy-max/buildx/actions/runs/5308541272/jobs/9608171567#step:6:88

tests/build.go
```diff
@@ -106,7 +107,7 @@ func testImageIDOutput(t *testing.T, sb integration.Sandbox) {
 
 	cmd := buildxCmd(
 		sb,
-		withArgs("build", "-q", outFlag, "--iidfile", filepath.Join(targetDir, "iid.txt"), "--metadata-file", filepath.Join(targetDir, "md.json"), dir),
+		withArgs("build", "-q", "--provenance", "false", outFlag, "--iidfile", filepath.Join(targetDir, "iid.txt"), "--metadata-file", filepath.Join(targetDir, "md.json"), dir),
```
crazy-max, line 110: @tonistiigi Needs to disable provenance for this test with containerd storage as it would export an OCI image:

```json
{
  "containerimage.buildinfo": {
    "frontend": "dockerfile.v0",
    "attrs": {
      "filename": "Dockerfile"
    },
    "sources": [
      {
        "type": "docker-image",
        "ref": "docker.io/library/busybox:latest",
        "pin": "sha256:0d5a701f0ca53f38723108687add000e1922f812d4187dea7feaee85d2f5a6c5"
      }
    ]
  },
  "containerimage.descriptor": {
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "digest": "sha256:ce51e4ac359feabf7a34388bb599e84873ba9299ce8b863ead9a093a16ffc875",
    "size": 855
  },
  "containerimage.digest": "sha256:ce51e4ac359feabf7a34388bb599e84873ba9299ce8b863ead9a093a16ffc875",
  "image.name": "moby-dangling@sha256:ce51e4ac359feabf7a34388bb599e84873ba9299ce8b863ead9a093a16ffc875"
}
```

This is a test case we can add in follow-up. I guess we need to read `"containerimage.descriptor".annotations."config.digest"`.

.github/workflows/build.yml
```diff
@@ -66,6 +66,7 @@ jobs:
       matrix:
         worker:
           - docker
+          - docker-containerd  # same as docker, but with containerd snapshotter
```
jedevc, line 69: Could we maybe call this worker something different? Looks like it clashes with `docker-container`.

If we could use `+` in the worker name, we could do a variant style like `docker+containerd`?

Dockerfile
```diff
@@ -25,6 +24,19 @@ FROM registry:$REGISTRY_VERSION AS registry
...
 FROM moby/buildkit:$BUILDKIT_VERSION AS buildkit
 
+FROM gobase AS docker
+ARG TARGETPLATFORM
+ARG DOCKER_VERSION
+WORKDIR /opt/docker
+RUN DOCKER_ARCH=$(case ${TARGETPLATFORM:-linux/amd64} in \
+    "linux/amd64")   echo "x86_64"  ;; \
+    "linux/arm/v7")  echo "armhf"   ;; \
+    "linux/arm64")   echo "aarch64" ;; \
```
jedevc, line 34: Should we include all the platforms we release docker for here?

Dockerfile
```diff
@@ -25,6 +24,19 @@ FROM registry:$REGISTRY_VERSION AS registry
...
+    *)               echo ""        ;; esac) \
+  && echo "DOCKER_ARCH=$DOCKER_ARCH" \
+  && wget -qO- "https://download.docker.com/linux/static/stable/${DOCKER_ARCH}/docker-${DOCKER_VERSION}.tgz" | tar xvz --strip 1
+RUN ./dockerd --version; ./containerd --version; ./ctr --version; ./runc --version
```
jedevc, line 38:
```suggestion
RUN ./dockerd --version && ./containerd --version && ./ctr --version && ./runc --version
```

We should fail if any of the above fail.

tests/build.go
```diff
@@ -139,6 +140,53 @@ func testImageIDOutput(t *testing.T, sb integration.Sandbox) {
 	require.Equal(t, dgst, digest.Digest(md.ConfigDigest))
 }
 
+func testBuildMobyFromLocalImage(t *testing.T, sb integration.Sandbox) {
+	if sb.Name() != "docker" && sb.Name() != "docker+containerd" {
```
jedevc, line 144: Can we do the same `strings.Cut` trick here, so we don't need to hardcode the `containerd` part of the string?

jedevc (approved):

tests/build.go
jedevc: Perfect, LGTM :tada:

PR #1890, opened 2023-06-16
tests: share single docker between docker-container backends

jedevc:
~~:warning: Requires https://github.com/moby/buildkit/pull/3954.~~

This means that we can run our docker-container tests in parallel again, which can help speed up our test runs by a *significant* amount.


crazy-max (approved):
LGTM :tada:

PR #1894, opened 2023-06-19
build: fix host-gateway handling

### LINKED ISSUE #1832: host-gateway no longer supported in docker build
### Description

Hi all, I've noticed that once I upgraded my docker to the new version I'm unable to use `host-gateway` during the build process.

I get the following error when using the `--add-host=host.docker.internal:host-gateway` argument for `docker build`.

```
[+] Building 0.0s (0/0)
 ERROR: failed to solve: failed to parse additional hosts: failed to parse IP host-gateway
```

This used to work before the upgrade and what's interesting is that the `docker run` command and even `docker comnpose` still support the argument.

### Reproduce

1. Run a docker build command with the `--add-host=host.docker.internal:host-gateway`

For example

```
docker build --add-host=host.docker.internal:host-gateway .
```

### Expected behavior

`docker build` is able to use the `host-gateway` special string to bridge the host network with the containers network.

### docker version

```bash
  Version:           23.0.0
  API version:       1.42
  Go version:        go1.19.5
  Git commit:        e92dd87
  Built:             Wed Feb  1 17:49:08 2023
  OS/Arch:           linux/amd64
  Context:           default
 
 Server: Docker Engine - Community
  Engine:
   Version:          23.0.0
   API version:      1.42 (minimum version 1.12)
   Go version:       go1.19.5
   Git commit:       d7573ab
   Built:            Wed Feb  1 17:49:08 2023
   OS/Arch:          linux/amd64
   Experimental:     false
  containerd:
   Version:          1.6.16
   GitCommit:        31aa4358a36870b21a992d3ad2bef29e1d693bec
  runc:
   Version:          1.1.4
   GitCommit:        v1.1.4-0-g5fd4c4d
  docker-init:
   Version:          0.19.0
   GitCommit:        de40ad0
```

### docker info

```bash
Client:
  Context:    default
  Debug Mode: false
  Plugins:
     Version:  v0.10.2
     Version:  v2.15.1
   scan: Docker Scan (Docker Inc.)
     Version:  v0.23.0
     Path:     /usr/libexec/docker/cli-plugins/docker-scan
 
 Server:
  Containers: 9
   Running: 7
   Paused: 0
   Stopped: 2
  Images: 115
  Server Version: 23.0.0
  Storage Driver: overlay2
   Backing Filesystem: extfs
   Supports d_type: true
   Using metacopy: false
   Native Overlay Diff: true
   userxattr: false
  Logging Driver: json-file
  Cgroup Driver: cgroupfs
  Cgroup Version: 1
  Plugins:
   Volume: local
  Swarm: inactive
  Default Runtime: runc
  Init Binary: docker-init
  containerd version: 31aa4358a36870b21a992d3ad2bef29e1d693bec
  runc version: v1.1.4-0-g5fd4c4d
  init version: de40ad0
  Security Options:
   apparmor
   seccomp
    Profile: builtin
  Kernel Version: 5.4.0-121-generic
  Operating System: Ubuntu 20.04.5 LTS
  OSType: linux
  Architecture: x86_64
  CPUs: 4
  Total Memory: 3.666GiB
  Name: dev-testing-amlesh
  ID: HYQS:EF3Y:2IJR:XJHT:IPNI:2M3H:SPSX:3UBG:7OFQ:SE77:7U7J:WHYU
  Debug Mode: false
  Registry: https://index.docker.io/v1/
  Experimental: false
  Insecure Registries:
   127.0.0.0/8
  Live Restore Enabled: false
 
 WARNING: No swap limit support
```

### Additional Info

_No response_

crazy-max:
fixes #1832 
related to https://github.com/moby/moby/pull/45767

### DIFF
--- a/build/utils.go
+++ b/build/utils.go
@@ -3,10 +3,12 @@ package build
 import (
 	"archive/tar"
 	"bytes"
+	"context"
 	"net"
 	"os"
 	"strings"
 
+	"github.com/docker/buildx/driver"
 	"github.com/docker/cli/opts"
 	"github.com/docker/docker/builder/remotecontext/urlutil"
 	"github.com/moby/buildkit/util/gitutil"
@@ -57,7 +59,7 @@ func isArchive(header []byte) bool {
 }
 
 // toBuildkitExtraHosts converts hosts from docker key:value format to buildkit's csv format
-func toBuildkitExtraHosts(inp []string, mobyDriver bool) (string, error) {
+func toBuildkitExtraHosts(ctx context.Context, inp []string, nodeDriver *driver.DriverHandle) (string, error) {
 	if len(inp) == 0 {
 		return "", nil
 	}
@@ -67,11 +69,16 @@ func toBuildkitExtraHosts(inp []string, mobyDriver bool) (string, error) {
 		if !ok || host == "" || ip == "" {
 			return "", errors.Errorf("invalid host %s", h)
 		}
-		// Skip IP address validation for "host-gateway" string with moby driver
-		if !mobyDriver || ip != mobyHostGatewayName {
-			if net.ParseIP(ip) == nil {
-				return "", errors.Errorf("invalid host %s", h)
+		// If the IP Address is a "host-gateway", replace this value with the
+		// IP address provided by the worker's label.
+		if ip == mobyHostGatewayName {
+			hgip, err := nodeDriver.HostGatewayIP(ctx)
+			if err != nil {
+				return "", errors.Wrap(err, "unable to derive the IP value for host-gateway")
 			}
+			ip = hgip.String()
+		} else if net.ParseIP(ip) == nil {
+			return "", errors.Errorf("invalid host %s", h)
 		}
 		hosts = append(hosts, host+"="+ip)
 	}
--- a/driver/manager.go
+++ b/driver/manager.go
@@ -2,17 +2,17 @@ package driver
 
 import (
 	"context"
+	"net"
 	"os"
 	"sort"
 	"strings"
 	"sync"
 
-	"k8s.io/client-go/rest"
-
 	dockerclient "github.com/docker/docker/client"
 	"github.com/moby/buildkit/client"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	"k8s.io/client-go/rest"
 )
 
 type Factory interface {
@@ -154,6 +154,9 @@ type DriverHandle struct {
 	features                map[Feature]bool
 	historyAPISupportedOnce sync.Once
 	historyAPISupported     bool
+	hostGatewayIPOnce       sync.Once
+	hostGatewayIP           net.IP
+	hostGatewayIPErr        error
 }
 
 func (d *DriverHandle) Client(ctx context.Context) (*client.Client, error) {
@@ -177,4 +180,37 @@ func (d *DriverHandle) HistoryAPISupported(ctx context.Context) bool {
 		}
 	})
 	return d.historyAPISupported
+}
+
+func (d *DriverHandle) HostGatewayIP(ctx context.Context) (net.IP, error) {
+	d.hostGatewayIPOnce.Do(func() {
+		if !d.Driver.IsMobyDriver() {
+			d.hostGatewayIPErr = errors.New("host-gateway is only supported with the docker driver")
+			return
+		}
+		c, err := d.Client(ctx)
+		if err != nil {
+			d.hostGatewayIPErr = err
+			return
+		}
+		workers, err := c.ListWorkers(ctx)
+		if err != nil {
+			d.hostGatewayIPErr = errors.Wrap(err, "listing workers")
+			return
+		}
+		for _, w := range workers {
+			// should match github.com/docker/docker/builder/builder-next/worker/label.HostGatewayIP const
+			if v, ok := w.Labels["org.mobyproject.buildkit.worker.moby.host-gateway-ip"]; ok && v != "" {
+				ip := net.ParseIP(v)
+				if ip == nil {
+					d.hostGatewayIPErr = errors.Errorf("failed to parse host-gateway IP: %s", v)
+					return
+				}
+				d.hostGatewayIP = ip
+				return
+			}
+		}
+		d.hostGatewayIPErr = errors.New("host-gateway IP not found")
+	})
+	return d.hostGatewayIP, d.hostGatewayIPErr
 }
--- a/build/build.go
+++ b/build/build.go
@@ -645,7 +645,7 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	}
 
 	// setup extrahosts
-	extraHosts, err := toBuildkitExtraHosts(opt.ExtraHosts, nodeDriver.IsMobyDriver())
+	extraHosts, err := toBuildkitExtraHosts(ctx, opt.ExtraHosts, nodeDriver)
 	if err != nil {
 		return nil, nil, err
 	}


crazy-max:
PTAL @tonistiigi @jedevc 

Will add integration test in follow-up when https://github.com/moby/moby/pull/45767 is merged as discussed.

build/utils.go
```diff
@@ -67,11 +67,16 @@ func toBuildkitExtraHosts(inp []string, mobyDriver bool) (string, error) {
...
 			return "", errors.Errorf("invalid host %s", h)
 		}
-		// Skip IP address validation for "host-gateway" string with moby driver
-		if !mobyDriver || ip != mobyHostGatewayName {
-			if net.ParseIP(ip) == nil {
-				return "", errors.Errorf("invalid host %s", h)
+		// If the IP Address is a "host-gateway", replace this value with the
+		// IP address provided by the worker's label.
+		if ip == mobyHostGatewayName {
+			if hgip == nil {
```
thaJeztah, line 73: Would it make  sense to just pass a reference to the `nodeDriver` as argument, and do the lookup here? (Could even be more generic and a `nodeDriver.LookupIP(name)` perhaps, but maybe that's not needed). And we could return the error it returned here. (Function could return a `NotFound()` error if not supported, but not sure about that one)

build/utils.go
```diff
@@ -67,11 +69,19 @@ func toBuildkitExtraHosts(inp []string, mobyDriver bool) (string, error) {
...
-				return "", errors.Errorf("invalid host %s", h)
+		// If the IP Address is a "host-gateway", replace this value with the
+		// IP address provided by the worker's label.
+		if ip == mobyHostGatewayName {
+			hgip, err := nodeDriver.HostGatewayIP(ctx)
+			if err != nil {
+				return "", err
+			} else if hgip == nil {
+				return "", errors.Errorf("unable to derive the IP value for host-gateway")
 			}
```
thaJeztah, line 80: nit: don't think we need an "else" here (first one already returns). Also can use `errors.New()` here, because there's no formatting 😅 

```suggestion
			if err != nil {
				return "", err
			}
			if hgip == nil {
				return "", errors.New("unable to derive the IP value for host-gateway")
			}
```

driver/manager.go
```diff
@@ -178,3 +181,34 @@ func (d *DriverHandle) HistoryAPISupported(ctx context.Context) bool {
...
+			return
+		}
+		if len(workers) == 0 {
+			return
+		}
+		// TODO: Use HostGatewayIP const from github.com/docker/docker/builder/builder-next/worker/label pkg
+		if v, ok := workers[0].Labels["org.mobyproject.buildkit.worker.moby.host-gateway-ip"]; ok && v != "" {
+			ip := net.ParseIP(v)
+			if ip == nil {
+				d.hostGatewayIPErr = errors.Wrapf(err, "failed to parse host gateway IP: %s", v)
```
thaJeztah, line 207: This will always be `nil`, because `errors.Wrap()` returns `nil` if `err` is `nil`

driver/manager.go
```diff
@@ -178,3 +181,34 @@ func (d *DriverHandle) HistoryAPISupported(ctx context.Context) bool {
 	})
 	return d.historyAPISupported
 }
+
+func (d *DriverHandle) HostGatewayIP(ctx context.Context) (net.IP, error) {
+	d.hostGatewayIPOnce.Do(func() {
+		if !d.Driver.IsMobyDriver() {
+			return
```
thaJeztah, line 188: Instead of returning `nil, nil`, should we return an error (host gateway IP not supported by driver)?

driver/manager.go
```diff
@@ -178,3 +181,34 @@ func (d *DriverHandle) HistoryAPISupported(ctx context.Context) bool {
...
+			d.hostGatewayIPErr = err
+			return
+		}
+		workers, err := c.ListWorkers(ctx)
+		if err != nil {
+			d.hostGatewayIPErr = err
+			return
+		}
+		if len(workers) == 0 {
+			return
```
thaJeztah, line 201: Wondering the same here; would this be an error condition? What does the rest of the build do if there's no workers? (somewhat assuming that would also produce an error somewhere).

Somewhat unrelated question: can this code be used in other situationsions than the `buildx` _CLI_ ? If this code is used in a long lived process, could the state of the world change (as in: `ListWorkers()` later now having actual workers).

Because if that's the case, I don't think we can use a `sync.Once()` for this code ⚠️

build/utils.go
thaJeztah: Also left a comment on the `hgip == nil` case (ideally it would _either_ return an error _or_ return `nil`, not both (I ... think?).

build/utils.go
```diff
@@ -67,11 +69,19 @@ func toBuildkitExtraHosts(inp []string, mobyDriver bool) (string, error) {
...
+			ip = hgip.String()
+		}
+		if net.ParseIP(ip) == nil {
```
thaJeztah, line 83: _This one_ should actually be an `else if` (or at least, I would not expect `hgip.String()` to be producing an invalid IP; if that's the case, that would be a bug in Go's stdlib 😅)

 ```suggestion
		} else if net.ParseIP(ip) == nil {
```

**Review by thaJeztah — CHANGES_REQUESTED** [review]
left some comments / thoughts (but could use verification on some of them)

File: build/utils.go (hunk 2 above)
Comment (crazy-max) [reply build/utils.go]: Yeah returning an error is better you right, don't need an extra cond on this.

File: driver/manager.go (hunk 3 above)
Comment (crazy-max) [reply driver/manager.go]: :man_facepalming:

File: driver/manager.go (hunk 5 above)
Comment (crazy-max) [reply driver/manager.go:199]: Workers are spawned on startup when instantiating a new controller such as https://github.com/moby/moby/blob/8d070e30f5a7e8aaa3dd3dd431f98b8c8909fbe2/builder/builder-next/controller.go#L114 and BuildKit would fail to start if there is none available afaik: https://github.com/moby/buildkit/blob/23d38a7220bff812b8a1de07c1f02b4e46cacd7c/cmd/buildkitd/main.go#L752-L754

We also don't support multi workers when solving a request atm iirc: https://github.com/moby/buildkit/blob/23d38a7220bff812b8a1de07c1f02b4e46cacd7c/control/control.go#L326

I changed the logic for this case and takes the very first one that has the label set.

File: driver/manager.go [hunk 7]
```diff
@@ -178,3 +181,36 @@ func (d *DriverHandle) HistoryAPISupported(ctx context.Context) bool {
...
+			d.hostGatewayIPErr = err
+			return
+		}
+		workers, err := c.ListWorkers(ctx)
+		if err != nil {
+			d.hostGatewayIPErr = errors.Wrap(err, "listing workers")
+			return
+		}
+		for _, w := range workers {
+			// TODO: Use HostGatewayIP const from github.com/docker/docker/builder/builder-next/worker/label pkg
```
thaJeztah, line 202: I should add here that if using the const brings in a lot of complexity (dependency-wise) that it's probably okay to keep a local const, and call it a day (but with a mention of "this is what it must match")

build/utils.go
thaJeztah: I think you missed this one 😅

thaJeztah (commented):
I think you missed one comment, but LGTM otherwise :)

thaJeztah (approved):
LGTM

jedevc (approved):

thaJeztah:
@jedevc good to go?

dvdksn (approved):
I'm documenting the special `host-gateway` value now, and it looks like this is pretty close to getting merged, so I will not mention that `host-gateway` is not supported for `docker build`

zamlz:
@dvdksn why is host-gateway no longer supported in `docker build`? It used to be an is a necessary feature for my team's build system with docker.

zamlz:
please disregard my comment lol

PR #1900, opened 2023-06-21
build: add GO_EXTRA_FLAGS argument

cyphar:
This is useful for setting things like -buildmode=pie when packaging docker-buildx for distributions.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>


cyphar:
We only really need this for `-buildmode=pie` so if you'd prefer that `-buildmode=pie` is auto-added, I can write a patch to do that instead.

jedevc (approved):

Issue #1902, opened 2023-06-22
v0.11.x breaking with docker compose profiles


ClayBenson94:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [ ] ... the documentation does not mention anything about my problem
- [ ] ... there are no open or closed issues that are related to my problem

### Description

When building a simple docker-compose file that has a `profile`, buildx returns an error `failed to find target` on v0.11.0, but this isn't the case on v0.10.4

### Expected behaviour

A successful build of the target service

### Actual behaviour

I receive an error with `failed to find target` - details in the logs below

### Buildx version

github.com/docker/buildx v0.11.0 687feca9e8dcd1534ac4c026bc4db5a49de0dd6e

### Docker info

```text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.4
    Path:     /Users/clay/.docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.17.3
    Path:     /Users/clay/.docker/cli-plugins/docker-compose
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /Users/clay/.docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.19
    Path:     /Users/clay/.docker/cli-plugins/docker-extension
  init: Creates Docker-related starter files for your project (Docker Inc.)
    Version:  v0.1.0-beta.4
    Path:     /Users/clay/.docker/cli-plugins/docker-init
  sbom: View the packaged-based Software Bill Of Materials (SBOM) for an image (Anchore Inc.)
    Version:  0.6.0
    Path:     /Users/clay/.docker/cli-plugins/docker-sbom
  scan: Docker Scan (Docker Inc.)
    Version:  v0.26.0
    Path:     /Users/clay/.docker/cli-plugins/docker-scan
  scout: Command line tool for Docker Scout (Docker Inc.)
    Version:  v0.10.0
    Path:     /Users/clay/.docker/cli-plugins/docker-scout

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 23
 Server Version: 23.0.5
 Storage Driver: overlay2
[...]

PR #1903, opened 2023-06-22
bake: ignore profiles in compose definitions

crazy-max:
fixes https://github.com/docker/buildx/issues/1902

If a profile is being set in the compose file, build will not run against this service unless this profile is enabled: https://docs.docker.com/compose/compose-file/03-compose-file/#profiles

For compose it makes sense but bake doesn't have any support for profiles so I think we should just ignore them: https://github.com/compose-spec/compose-go/blob/532cd92682a3f00c22c1286f7eedccab2a232baa/types/project.go#L258


milas (approved):
SGTM

glours (approved):
LGTM

jedevc (approved):

PR #1905, opened 2023-06-23
update flag-description for --cgroup-parent

thaJeztah:
- relates to https://github.com/moby/moby/issues/34469#issuecomment-362042644

This attempts to make it clearer that the --cgroup-parent option is only used for the containers used during build. Instead of mentioning "build container", I opted for using "RUN instructions" (to match the --network description), although this may not be ideal (as it assumes the "Dockerfile" front-end, which of course may not be the case).


commands/build.go
```diff
@@ -464,7 +464,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 	flags.StringArrayVar(&options.cacheTo, "cache-to", []string{}, `Cache export destinations (e.g., "user/app:cache", "type=local,dest=path/to/dir")`)
 
-	flags.StringVar(&options.cgroupParent, "cgroup-parent", "", "Optional parent cgroup for the container")
+	flags.StringVar(&options.cgroupParent, "cgroup-parent", "", `Set the parent cgroup for the "RUN" instructions during build`)
```
thaJeztah, line 467: I'm definitely open to suggestions for the description (also see PR description above). Not sure what the least ambiguous description would be for this.

dvdksn (approved):
LGTM!

dvdksn:
I think it's safe to assume that "RUN" instruction is the most approachable way to convey this at the moment. IMO there's no better way to refer to "build container exec" that's also a one-liner.

thaJeztah:
I'm considering; perhaps we should put the options in groups.

- options used during build (during "RUN")
- options for the output
- options for ....



I added some "ad-hoc" grouping for _commands_ in the CLI at some point

- https://github.com/docker/cli/pull/3506

But that one is really tacky, and really for "our specific purpose". ISTR though that Cobra also added some options to group things, so perhaps we need to look into that.

PR #1907, opened 2023-06-23
build: toControllerOptions: micro-optimization

thaJeztah:
Swapping the order of these checks, which I guess is the most minimal optimization possible

    BenchmarkGetEnv-10    68764720    16.82  ns/op   0 B/op   0 allocs/op
    BenchmarkMap-10      454135184     2.635 ns/op   0 B/op   0 allocs/op


crazy-max (approved):

thaJeztah:
opened https://github.com/docker/buildx/pull/1908 as alternative

PR #1908, opened 2023-06-23
use containerd's epoch package for handling SOURCE_DATE_EPOCH

thaJeztah:
- alternative to / closes https://github.com/docker/buildx/pull/1907

This allows us to validate the value before we're using it, and makes sure we handle things in the same way.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 
 	"github.com/containerd/console"
+	"github.com/containerd/containerd/pkg/epoch"
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/builder"
 	"github.com/docker/buildx/controller"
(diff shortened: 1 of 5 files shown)


commands/build.go
```diff
@@ -120,10 +121,13 @@ func (o *buildOptions) toControllerOptions() (*controllerapi.BuildOptions, error
...
-	if v := os.Getenv("SOURCE_DATE_EPOCH"); v != "" {
-		if _, ok := opts.BuildArgs["SOURCE_DATE_EPOCH"]; !ok {
-			opts.BuildArgs["SOURCE_DATE_EPOCH"] = v
+	if _, ok := opts.BuildArgs[epoch.SourceDateEpochEnv]; !ok {
+		v, err := epoch.SourceDateEpoch()
+		if err != nil {
+			return nil, err
+		}
+		if v != nil {
+			opts.BuildArgs[epoch.SourceDateEpochEnv] = strconv.FormatInt(v.Unix(), 10)
```
crazy-max, line 130: I'm not sure parsing string to time and back to string for this purpose is ideal.

thaJeztah: I could still use the parsing, and just get the env-var separately; let me check for a bit

thaJeztah: hm.. no no utility on its own, so if could be;

```go
opts.BuildArgs[epoch.SourceDateEpochEnv] = os.Getenv(epoch.SourceDateEpochEnv)
```

which also looks a bit odd, as we're setting a value that's not the `v != nil` that we check 😞

tonistiigi:
@thaJeztah What's the status of this?

PR #1913, opened 2023-06-27
Controller: Include CgroupParent in build.Options

yastanotheruser:
The `--cgroup-parent` option had no effect in build containers due to the CgroupParent field in build.Options not being set. Related: moby/moby#45796.


jedevc (approved):

jedevc:
Ugh, that one's my oversight.

Thanks so much for catching that @yastanotheruser :tada: :tada:

PR #1897, opened 2023-06-20
test: register docker worker with containerd snapshotter

crazy-max:
follow-up https://github.com/docker/buildx/pull/1886#discussion_r1233562589

Adds `docker+containerd` worker to our matrix to be able to test against dockerd with containerd storage enabled.

PR #1931, opened 2023-07-07
dockerfile: use xx-info march when downloading dockerd

crazy-max:
follow-up https://github.com/docker/buildx/pull/1897#discussion_r1254948724

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,16 +28,7 @@ FROM gobase AS docker
 ARG TARGETPLATFORM
 ARG DOCKER_VERSION
 WORKDIR /opt/docker
-RUN DOCKER_ARCH=$(case ${TARGETPLATFORM:-linux/amd64} in \
-    "linux/amd64")   echo "x86_64"  ;; \
-    "linux/arm/v6")  echo "armel"   ;; \
-    "linux/arm/v7")  echo "armhf"   ;; \
-    "linux/arm64")   echo "aarch64" ;; \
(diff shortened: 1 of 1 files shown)


tonistiigi (approved):

Dockerfile
```diff
@@ -28,16 +28,7 @@ FROM gobase AS docker
 ARG TARGETPLATFORM
 ARG DOCKER_VERSION
 WORKDIR /opt/docker
-RUN DOCKER_ARCH=$(case ${TARGETPLATFORM:-linux/amd64} in \
-    "linux/amd64")   echo "x86_64"  ;; \
-    "linux/arm/v6")  echo "armel"   ;; \
```
tonistiigi, line 33: Actually this doesn't look correct for arm/vX . https://github.com/tonistiigi/xx/blob/master/base/xx-info#L211 . I'm not sure if this is based on anything? Arm is like debian but aarch64/ppc64le are not debian 😕

crazy-max: Yes indeed, I recall we had a discussion with @tianon and @thaJeztah to align them in the future.

tonistiigi: In that case, we could still do a `case` in `xx-info` with fewer checks than the original, or if you think it's not worth it then we can close this and keep as is.

crazy-max: Yes let's keep it as is. I'm closing this one and mark it as needs/follow-up.

PR #1937, opened 2023-07-12
build: assume '.' if no args are passed

imjasonh:
Before:

```shell
$ docker build -t foo
ERROR: "docker buildx build" requires exactly 1 argument.
# swearing omitted
$ docker build -t foo .
[+] Building 0.4s (3/3) FINISHED
 => [internal] load .dockerignore                                                                                                                                                                            
  ...etc...
```

After:

```shell
$ docker build -t foo
[+] Building 0.4s (3/3) FINISHED
 => [internal] load .dockerignore                                                                                                                                                                            
  ...etc...
```

If this seems fine to you I can make a [similar change in docker/cli](https://github.com/docker/cli/compare/master...imjasonh:cli-1:assume-dot)

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -415,9 +415,13 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 		Use:     "build [OPTIONS] PATH | URL | -",
 		Aliases: []string{"b"},
 		Short:   "Start a build",
-		Args:    cli.ExactArgs(1),
+		Args:    cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			options.contextPath = args[0]
+			if len(args) == 0 {
+				options.contextPath = "."
+			} else {
+				options.contextPath = args[0]
+			}
 			options.builder = rootOpts.builder
 			options.metadataFile = cFlags.metadataFile
 			options.noCache = false


thaJeztah (changes requested):
I need to find related tickets about this (as there was prior discussion with more details), but I don't think we should do this as we're considering keeping this option open for future features (interactive builds). Setting this default means that this will no longer be an option in future.

imjasonh:
> I need to find related tickets about this (as there was prior discussion with more details), but I don't think we should do this as we're considering keeping this option open for future features (interactive builds). Setting this default means that this will no longer be an option in future.

I totally understand, and that's absolutely your call to make. This CLI is invoked probably bajillions of times each day, and I was pretty sure I wasn't the first person to think of this 😆. Any previous conversation/context about it you could share would be helpful.

Speaking only for myself, I stumble on this embarrassingly often, and I have since I started using Docker years ago (embarrassing, as I said). I imagine others stumble on it too, some small percentage of those bajillions of times per day. It adds up.

I don't know what you plan to do for interactive builds (it sounds cool!) but personally I'd probably get more value out of `docker build -t foo` doing what I've expected it to do for years, rather than start dropping me into some new interactive build experience, just because I forgot the `.` again.

tonistiigi:
My personal preference would be that no-args could use the current Git state as main context, as this is a much better default because all the sources are immutable and properly tracked.

I don't think this is a quick merge, but we can definitely have an open issue about what the default behavior could be and feedback for various options.

tonistiigi:
Closing this for now, but we can continue tracking this in issue.

PR #1938, opened 2023-07-12
kubernetes driver // allow custom annotations and labels

jonapich:
This implements custom annotations and labels to be included in kubernetes manifests. Currently, the user has to patch them in a subsequent call.

Motivation: We provide a tool for developers to create/remove buildx builders on a shared k8s cluster. To prevent wasting resources, the buildx deployments are patched with a label (for filtering) and an annotation (an `expire-after` timestamp). A cron janitor eventually finds and removes them. The user may also refresh the timestamp in order to retain their cache for longer if they desire.

Issue: Because creating the builder and patching the annotations requires two calls, I need to handle the cases when the `create` call fails. This isn't always possible in an exception handler, for instance when this happens on a CI server that gets killed abruptly. This forces the janitor to include additional heuristics to find and remove the leaked deployments. It would be simpler if these annotations would be included from the start.

### DIFF
--- a/docs/reference/buildx_create.md
+++ b/docs/reference/buildx_create.md
@@ -141,7 +141,7 @@ to achieve that.
 
 Passes additional driver-specific options.
 
-Note: When using quoted values for example for the `nodeselector` or
+Note: When using quoted values for the `nodeselector`, `annotations`, `labels` or
 `tolerations` options, ensure that quotes are escaped correctly for your shell.
 
 #### `docker` driver
@@ -165,6 +165,8 @@ No driver options.
 - `limits.memory` - Sets the limit memory value specified in bytes or with a valid suffix. Example `limits.memory=500Mi`, `limits.memory=4G`
 - `serviceaccount` - Sets the created pod's service account. Example `serviceaccount=example-sa`
 - `"nodeselector=label1=value1,label2=value2"` - Sets the kv of `Pod` nodeSelector. No Defaults. Example `nodeselector=kubernetes.io/arch=arm64`
+- `"annotations=domain/thing1=value1,domain/thing2=value2"` - Sets additional annotations on the deployments and pods. No Defaults. Example `annotations=example.com/owner=sarah`
+- `"labels=domain/thing1=value1,domain/thing2=value2"` - Sets additional labels on the deployments and pods. No Defaults. Example `labels=example.com/team=rd`
 - `"tolerations=key=foo,value=bar;key=foo2,operator=exists;key=foo3,effect=NoSchedule"` - Sets the `Pod` tolerations. Accepts the same values as the kube manifest tolera>tions. Key-value pairs are separated by `,`, tolerations are separated by `;`. No Defaults. Example `tolerations=operator=exists`
 - `rootless=(true|false)` - Run the container as a non-root user without `securityContext.privileged`. Needs Kubernetes 1.19 or later. [Using Ubuntu host kernel is recommended](https://github.com/moby/buildkit/blob/master/docs/rootless.md). Defaults to false.
 - `loadbalance=(sticky|random)` - Load-balancing strategy. If set to "sticky", the pod is chosen using the hash of the context path. Defaults to "sticky"
--- a/driver/kubernetes/factory.go
+++ b/driver/kubernetes/factory.go
@@ -148,15 +148,20 @@ func (f *factory) processDriverOpts(deploymentName string, namespace string, cfg
 		case "serviceaccount":
 			deploymentOpt.ServiceAccountName = v
 		case "nodeselector":
-			kvs := strings.Split(strings.Trim(v, `"`), ",")
-			s := map[string]string{}
-			for i := range kvs {
-				kv := strings.Split(kvs[i], "=")
-				if len(kv) == 2 {
-					s[kv[0]] = kv[1]
-				}
+			deploymentOpt.NodeSelector, err = splitMultiValues(v, ",", "=")
+			if err != nil {
+				return nil, "", "", errors.Wrap(err, "cannot parse node selector")
 			}
-			deploymentOpt.NodeSelector = s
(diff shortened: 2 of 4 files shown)


docs/reference/buildx_create.md
```diff
@@ -165,6 +165,8 @@ No driver options.
 - `limits.memory` - Sets the limit memory value specified in bytes or with a valid suffix. Example `limits.memory=500Mi`, `limits.memory=4G`
 - `serviceaccount` - Sets the created pod's service account. Example `serviceaccount=example-sa`
 - `"nodeselector=label1=value1,label2=value2"` - Sets the kv of `Pod` nodeSelector. No Defaults. Example `nodeselector=kubernetes.io/arch=arm64`
+- `"annotations=domain/thing1=value1,domain/thing2=value2"` - Sets additional annotations on the deployments and pods. No Defaults. Example `annotations=mycorp.io/owner=sarah`
```
AkihiroSuda, line 168:
```suggestion
- `"annotations=domain/thing1=value1,domain/thing2=value2"` - Sets additional annotations on the deployments and pods. No Defaults. Example `annotations=example.com/owner=sarah`
```

docs/reference/buildx_create.md
```diff
@@ -165,6 +165,8 @@ No driver options.
 - `limits.memory` - Sets the limit memory value specified in bytes or with a valid suffix. Example `limits.memory=500Mi`, `limits.memory=4G`
 - `serviceaccount` - Sets the created pod's service account. Example `serviceaccount=example-sa`
 - `"nodeselector=label1=value1,label2=value2"` - Sets the kv of `Pod` nodeSelector. No Defaults. Example `nodeselector=kubernetes.io/arch=arm64`
+- `"annotations=domain/thing1=value1,domain/thing2=value2"` - Sets additional annotations on the deployments and pods. No Defaults. Example `annotations=example.com/owner=sarah`
+- `"labels=domain/thing1=value1,domain/thing2=value2"` - Sets additional labels on the deployments and pods. No Defaults. Example `labels=domain.com/team=rd`
```
AkihiroSuda, line 169:
```suggestion
- `"labels=domain/thing1=value1,domain/thing2=value2"` - Sets additional labels on the deployments and pods. No Defaults. Example `labels=example.com/team=rd`
```

driver/kubernetes/factory_test.go
```diff
@@ -47,6 +47,8 @@ func TestFactory_processDriverOpts(t *testing.T) {
 				"rootless":        "true",
 				"nodeselector":    "selector1=value1,selector2=value2",
 				"tolerations":     "key=tolerationKey1,value=tolerationValue1,operator=Equal,effect=NoSchedule,tolerationSeconds=60;key=tolerationKey2,operator=Exists",
+				"annotations":     "my.company.io/expires-after=annotation1,my.company.io/other=annotation2",
```
AkihiroSuda, line 50:
```suggestion
				"annotations":     "example.com/expires-after=annotation1,my.company.io/other=annotation2",
```

AkihiroSuda (approved):

docs/reference/buildx_create.md
```diff
@@ -141,7 +141,7 @@ to achieve that.
 
 Passes additional driver-specific options.
 
-Note: When using quoted values for example for the `nodeselector` or
+Note: When using quoted values for example for the `nodeselector`, `annotations` or
```
crazy-max, line 144:
```suggestion
Note: When using quoted values for example for the `nodeselector`, `annotations`, `labels` or
```

driver/kubernetes/factory.go
```diff
@@ -148,15 +149,20 @@ func (f *factory) processDriverOpts(deploymentName string, namespace string, cfg
...
-			kvs := strings.Split(strings.Trim(v, `"`), ",")
-			s := map[string]string{}
-			for i := range kvs {
-				kv := strings.Split(kvs[i], "=")
-				if len(kv) == 2 {
-					s[kv[0]] = kv[1]
-				}
+			deploymentOpt.NodeSelector, err = splitMultiValues(v, ",", "=")
+			if err != nil {
+				return nil, "", "", fmt.Errorf("cannot parse node selector: %w", err)
```
crazy-max, line 154:
```suggestion
				return nil, "", "", errors.Wrap(err, "cannot parse node selector")
```

driver/kubernetes/factory_test.go
```diff
@@ -192,6 +210,30 @@ func TestFactory_processDriverOpts(t *testing.T) {
...
 	)
 
+	t.Run(
+		"InvalidCustomAnnotation", func(t *testing.T) {
+			cfg.DriverOpts = map[string]string{
+				"annotations": "key,value",
+			}
+
+			_, _, _, err := f.processDriverOpts(cfg.Name, "test", cfg)
+
```
crazy-max, line 220: nit: remove extra lines

docs/reference/buildx_create.md
jonapich: I was expecting this comment :)

Should I remove `for example` from the sentence, since it's supposed to be all-inclusive?

```suggestion
Note: When using quoted values for the `nodeselector`, `annotations`, `labels` or
```

driver/kubernetes/factory_test.go
jonapich: I applied this comment on the other identical surrounding tests as well. I also noticed an extra `ns := "test"` at the beginning that was not necessary since it was rewritten in the next call.

driver/kubernetes/factory.go
jonapich: TIL

docs/reference/buildx_create.md
jonapich: (note: I applied my suggestion)

crazy-max (approved):
LGTM thanks!

PR #1945, opened 2023-07-17
vendor: github.com/docker/docker@24.0 36e9e79

crazy-max:
relates to
* https://github.com/moby/moby/pull/45942
* https://github.com/moby/moby/issues/45935

closes #1944 

client: define a "dummy" hostname to use for local connections fixes "http: invalid Host header" errors when compiling with go1.20.6 or go1.19.11

full diff: https://github.com/docker/docker/compare/v24.0.2...36e9e796c6fc

PR #1946, opened 2023-07-17
update go to 1.20.6

crazy-max:
* follow-up #1945

### DIFF
--- a/hack/dockerfiles/generated-files.Dockerfile
+++ b/hack/dockerfiles/generated-files.Dockerfile
@@ -9,7 +9,7 @@ ARG GO_VERSION="1.20.6"
 ARG PROTOC_VERSION="3.11.4"
 
 # protoc is dynamically linked to glibc so can't use alpine base
-FROM golang:${GO_VERSION}-buster AS base
+FROM golang:${GO_VERSION}-bookworm AS base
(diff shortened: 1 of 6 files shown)


crazy-max:
`generated-files.Dockerfile` needs to be updated:

```
#7 [internal] load metadata for docker.io/library/golang:1.20.6-buster
#7 ERROR: docker.io/library/golang:1.20.6-buster: not found
```

hack/dockerfiles/generated-files.Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG GO_VERSION="1.20.6"
 ARG PROTOC_VERSION="3.11.4"
 
 # protoc is dynamically linked to glibc so can't use alpine base
-FROM golang:${GO_VERSION}-buster AS base
+FROM golang:${GO_VERSION} AS base
```
jedevc, line 12: I think it makes more sense to bump this to `bullseye` instead? Since we do apt commands below, it would be good to keep the debian base clear.

Also, there may be subtle versions between debian versions, that would be good to explicitly opt into.

crazy-max: Sure, will use `bookworm` as it's the latest stable.

jedevc (approved):

Issue #1171, opened 2022-06-16
Proposal: annotations helpers


jedevc:
With https://github.com/moby/buildkit/pull/2879 close to merging, I thought I'd propose some new syntax here :tada:

When merged, it'll be possible to use annotations as follows (with no changes to buildx):
```
docker buildx build . --output type=image,name=example/foo,push=true,annotation.x=1 --platform=linux/arm64,linux/amd64
```

However, this isn't ideal, and it would be nice to provide a shorthand, similar to how `--push` appends `push=true`.

## Buildx

For buildx, we can provide the `--annotation` flag:

```
docker buildx build . --output type=image,name=example/foo,push=true --annotation x=1 --platform=linux/arm64,linux/amd64
```

For platform-specific annotations, we can allow prefixing with the platform name:

```
docker buildx build . --output type=image,name=example/foo,push=true --annotation linux/amd64:x=1 --platform=linux/arm64,linux/amd64
```

Not super sure about the exact syntax, but something along those lines should work.

I think we should explicitly avoid allowing using anything other than the `manifest` annotation type, directing users to manually add the output attribute if they want more advanced behavior.

## Bake

For bake, we can add an `annotations` key:

```hcl
target "x" {
    ...
    annotations = {
        "linux/amd64:x" = "1"
    }
    ...
}
```

Comments appreciated :tada:

johnsonshi:
This is nice work @jedevc! I think the experience of allowing multiple `--annotation` flags (like the example below) would allow image builders to support [OCI Annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys).

```bash
docker buildx build \
  --annotation 'org.opencontainers.image.authors=authorName' \
  --annotation 'org.opencontainers.image.licenses=Apache' \
  -f Dockerfile -t myimage:latest .
```

I suppose the annotations should be added at the `manifest` type.

---

For multi-arch builds, would the annotations be only in the `manifestList`, or should they also be in each `manifest` within the multi-arch `manifestList`?


```bash
docker buildx build \
  --annotation 'org.opencontainers.image.authors=authorName' \
  --annotation 'org.opencontainers.image.licenses=Apache' \
  --platform=linux/arm64,linux/amd64 \
  -f Dockerfile -t myimage:latest .
```

johnsonshi:
> I think we should explicitly avoid allowing using anything other than the manifest annotation type, directing users to manually add the output attribute if they want more advanced behavior.

[...]

PR #1965, opened 2023-07-20
imagetools: Allow annotations for OCI image index

mqasimsarfraz:
Currently there isn't any way to add annotations to OCI image index for [multi-arch](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners). The use case is to have a description in ["About this version"](https://github.com/inspektor-gadget/inspektor-gadget/pkgs/container/inspektor-gadget/106798013?tag=v0.18.1) in GitHub. It will introduce an optional flag of `--annotation` to `imagetools create` to allow adding OCI annotations.

Related: #1171 

#### OCI Image Index (annotation added)

```
$ ./bin/build/buildx  imagetools create  -t ghcr.io/inspektor-gadget/inspektor-gadget:v0.18.1 ghcr.io/inspektor-gadget/inspektor-gadget:v0.18.1@sha256:c3780808973801b24c80f8b46835b882fe0d69c08a4460333f28143569665861 ghcr.io/inspektor-gadget/inspektor-gadget:v0.18.1@sha256:d6d5661b02002aa8035e035e9210e3476e0401a948a86fabf104f71c2cbe0efe  --dry-run --annotation index:org.opencontainers.image.description="I am a description"
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:c3780808973801b24c80f8b46835b882fe0d69c08a4460333f28143569665861",
      "size": 2963,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:d6d5661b02002aa8035e035e9210e3476e0401a948a86fabf104f71c2cbe0efe",
      "size": 2962,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ],
  "annotations": {
    "org.opencontainers.image.description": "I am a description"
  }
}
```

#### Manifest Descriptor

```
$ ./bin/build/buildx  imagetools create  -t ghcr.io/inspektor-gadget/inspektor-gadget:v0.18.1 ghcr.io/inspektor-gadget/inspektor-gadget:v0.18.1@sha256:c3780808973801b24c80f8b46835b882fe0d69c08a4460333f28143569665861 ghcr.io/inspektor-gadget/inspektor-gadget:v0.18.1@sha256:d6d5661b02002aa8035e035e9210e3476e0401a948a86fabf104f71c2cbe0efe  --dry-run --annotation manifest-descriptor:org.opencontainers.image.description="I am a description"
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:c3780808973801b24c80f8b46835b882fe0d69c08a4460333f28143569665861",
      "size": 2963,
      "annotations": {
        "org.opencontainers.image.description": "I am a description"
      },
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:d6d5661b02002aa8035e035e9210e3476e0401a948a86fabf104f71c2cbe0efe",
      "size": 2962,
      "annotations": {
        "org.opencontainers.image.description": "I am a description"
      },
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}

```

#### Docker manifest list (noop)

```
$ /bin/build/buildx  imagetools create  -t hello-world hello-world hello-world  --dry-run --annotation org.opencontainers.image.description="I am a description"
...
```

### DIFF
--- a/tests/imagetools.go
+++ b/tests/imagetools.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"encoding/json"
+	"os/exec"
 	"testing"
 
 	"github.com/containerd/containerd/platforms"
@@ -14,6 +15,7 @@ import (
 
 var imagetoolsTests = []func(t *testing.T, sb integration.Sandbox){
 	testImagetoolsInspectAndFilter,
+	testImagetoolsAnnotation,
 }
 
 func testImagetoolsInspectAndFilter(t *testing.T, sb integration.Sandbox) {
@@ -21,18 +23,7 @@ func testImagetoolsInspectAndFilter(t *testing.T, sb integration.Sandbox) {
 		t.Skip("imagetools tests are not driver specific and only run on docker-container")
 	}
 
-	dockerfile := []byte(`
-	FROM scratch
-	ARG TARGETARCH
-	COPY foo-${TARGETARCH} /foo
-	`)
-	dir := tmpdir(
-		t,
-		fstest.CreateFile("Dockerfile", dockerfile, 0600),
-		fstest.CreateFile("foo-amd64", []byte("foo-amd64"), 0600),
-		fstest.CreateFile("foo-arm64", []byte("foo-arm64"), 0600),
-	)
-
+	dir := createDockerfile(t)
 	registry, err := sb.NewRegistry()
 	if errors.Is(err, integration.ErrRequirements) {
 		t.Skip(err.Error())
@@ -76,4 +67,90 @@ func testImagetoolsInspectAndFilter(t *testing.T, sb integration.Sandbox) {
 
 	require.Equal(t, idx.Manifests[1].Digest, idx2.Manifests[0].Digest)
 	require.Equal(t, platforms.Format(*idx.Manifests[1].Platform), platforms.Format(*idx2.Manifests[0].Platform))
+}
+
+func testImagetoolsAnnotation(t *testing.T, sb integration.Sandbox) {
+	if sb.Name() != "docker-container" {
+		t.Skip("imagetools tests are not driver specific and only run on docker-container")
+	}
+
+	dir := createDockerfile(t)
+	registry, err := sb.NewRegistry()
+	if errors.Is(err, integration.ErrRequirements) {
+		t.Skip(err.Error())
+	}
+	require.NoError(t, err)
+	target := registry + "/buildx/imtools:latest"
+
+	out, err := buildCmd(sb, withArgs("--output", "type=registry,oci-mediatypes=true,name="+target, "--platform=linux/amd64,linux/arm64", "--provenance=false", dir))
+	require.NoError(t, err, string(out))
+
+	cmd := buildxCmd(sb, withArgs("imagetools", "inspect", target, "--raw"))
+	dt, err := cmd.CombinedOutput()
+	require.NoError(t, err, string(dt))
+
+	var idx ocispecs.Index
+	err = json.Unmarshal(dt, &idx)
+	require.NoError(t, err)
+	require.Empty(t, idx.Annotations)
+
+	imagetoolsCmd := func(source []string) *exec.Cmd {
+		args := []string{"imagetools", "create", "-t", target, "--annotation", "index:foo=bar", "--annotation", "index:bar=baz",
+			"--annotation", "manifest-descriptor:foo=bar", "--annotation", "manifest-descriptor[linux/amd64]:bar=baz"}
+		args = append(args, source...)
+		return buildxCmd(sb, withArgs(args...))
+	}
+	sources := [][]string{
+		{
+			target,
+		},
+		{
+			target + "@" + string(idx.Manifests[0].Digest),
+			target + "@" + string(idx.Manifests[1].Digest),
+		},
+	}
+	for _, source := range sources {
+		cmd = imagetoolsCmd(source)
+		dt, err = cmd.CombinedOutput()
+		require.NoError(t, err, string(dt))
+
+		newTarget := registry + "/buildx/imtools:annotations"
+		cmd = buildxCmd(sb, withArgs("imagetools", "create", "-t", newTarget, target))
+		dt, err = cmd.CombinedOutput()
+		require.NoError(t, err, string(dt))
+
+		cmd = buildxCmd(sb, withArgs("imagetools", "inspect", newTarget, "--raw"))
+		dt, err = cmd.CombinedOutput()
+		require.NoError(t, err, string(dt))
+
+		err = json.Unmarshal(dt, &idx)
+		require.NoError(t, err)
+		require.Len(t, idx.Annotations, 2)
+		require.Equal(t, "bar", idx.Annotations["foo"])
+		require.Equal(t, "baz", idx.Annotations["bar"])
+		require.Len(t, idx.Manifests, 2)
+		for _, mfst := range idx.Manifests {
+			require.Equal(t, "bar", mfst.Annotations["foo"])
+			if platforms.Format(*mfst.Platform) == "linux/amd64" {
+				require.Equal(t, "baz", mfst.Annotations["bar"])
+			} else {
+				require.Empty(t, mfst.Annotations["bar"])
+			}
+		}
+	}
+}
+
+func createDockerfile(t *testing.T) string {
+	dockerfile := []byte(`
+	FROM scratch
+	ARG TARGETARCH
+	COPY foo-${TARGETARCH} /foo
+	`)
+	dir := tmpdir(
+		t,
+		fstest.CreateFile("Dockerfile", dockerfile, 0600),
+		fstest.CreateFile("foo-amd64", []byte("foo-amd64"), 0600),
+		fstest.CreateFile("foo-arm64", []byte("foo-arm64"), 0600),
+	)
+	return dir
 }
--- a/util/imagetools/create.go
+++ b/util/imagetools/create.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/url"
+	"regexp"
 	"strings"
 
 	"github.com/containerd/containerd/content"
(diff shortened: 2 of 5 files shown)


util/imagetools/create.go
```diff
@@ -138,12 +138,19 @@ func (r *Resolver) Combine(ctx context.Context, srcs []*Source) ([]byte, ocispec
 		mt = ocispec.MediaTypeImageIndex
 	}
 
+	// annotations are only allowed on OCI indexes
+	annotations := make(map[string]string)
+	if mt == ocispec.MediaTypeImageIndex && len(ann) > 0 {
```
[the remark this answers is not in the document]
mqasimsarfraz: Good Catch! It can be removed!

mqasimsarfraz:
@crazy-max @jedevc any thoughts on this? :)

jedevc:
Hm, in general I think the idea for this is alright :tada: 

However, I think we'd probably want to retain consistency with how buildkit sets annotations: https://github.com/moby/buildkit/blob/master/docs/annotations.md.

There's not really any strong consensus on how we'd expose this in buildx build, but imagetools should probably follow an identical setup to avoid confusion.

https://github.com/docker/buildx/issues/1171#issuecomment-1499667083 gets pretty close to an ideal syntax imo. With, that to annotate your multi-arch image, you'd instead have:

```
$ buildx imagetools create -t hello-world hello-world hello-world --dry-run --annotation index:org.opencontainers.image.description="I am a description"
```

Would this kind of syntax work for you? I think we need consistency through buildx, so maybe good to continue the discussion there.

crazy-max:
For the sake of parity we should consider supporting adding labels as well in follow-up.

util/imagetools/create.go
```diff
@@ -138,12 +138,25 @@ func (r *Resolver) Combine(ctx context.Context, srcs []*Source) ([]byte, ocispec
...
 
+	// annotations are only allowed on OCI indexes
+	annotations := make(map[string]string)
+	if mt == ocispec.MediaTypeImageIndex {
+		for k, v := range ann {
+			if !strings.HasPrefix(k, "index:") {
+				return nil, ocispec.Descriptor{}, errors.Errorf("invalid annotation key %q, expected it to have 'index:' prefix", k)
+			}
+			k = strings.TrimPrefix(k, "index:")
+			annotations[k] = v
```
jedevc, line 149: We should also support `manifest-descriptor` here - this would need to modify `newDescs`, but shouldn't be too tricky?

We don't have to do `manifest` right now, since that involves mutating the manifest, so we can leave that as a future extension.

util/imagetools/create.go
```diff
@@ -138,12 +138,25 @@ func (r *Resolver) Combine(ctx context.Context, srcs []*Source) ([]byte, ocispec
 		mt = ocispec.MediaTypeImageIndex
 	}
 
+	// annotations are only allowed on OCI indexes
+	annotations := make(map[string]string)
+	if mt == ocispec.MediaTypeImageIndex {
+		for k, v := range ann {
+			if !strings.HasPrefix(k, "index:") {
+				return nil, ocispec.Descriptor{}, errors.Errorf("invalid annotation key %q, expected it to have 'index:' prefix", k)
+			}
```
jedevc, line 147: Suggestion: let's change this to a switch statement, checking for all the variants: `index` is supported, `index-descriptor` doesn't make sense in this context (so we should return a suitable error message), `manifest-descriptor` and `manifest` should error as not currently supported, all others should error with an "unknown"-like error.

jedevc (changes requested):

tests/imagetools.go
```diff
@@ -77,3 +78,56 @@ func testImagetoolsInspectAndFilter(t *testing.T, sb integration.Sandbox) {
 	require.Equal(t, idx.Manifests[1].Digest, idx2.Manifests[0].Digest)
 	require.Equal(t, platforms.Format(*idx.Manifests[1].Platform), platforms.Format(*idx2.Manifests[0].Platform))
 }
+
+func testImagetoolsAnnotation(t *testing.T, sb integration.Sandbox) {
+	if sb.Name() != "docker-container" {
+		t.Skip("imagetools tests are not driver specific and only run on docker-container")
+	}
+
+	dockerfile := []byte(`
```
tonistiigi, line 87: Please move lines 87-97 to a private helper function so it is shared with another test.

tests/imagetools.go
```diff
@@ -77,3 +78,56 @@ func testImagetoolsInspectAndFilter(t *testing.T, sb integration.Sandbox) {
...
+	dt, err := cmd.CombinedOutput()
+	require.NoError(t, err, string(dt))
+
+	var idx ocispecs.Index
+	err = json.Unmarshal(dt, &idx)
+	require.NoError(t, err)
+	require.Empty(t, idx.Annotations)
+
+	// update the index with annotations
+	cmd = buildxCmd(sb, withArgs("imagetools", "create", "-t", target, target+"@"+string(idx.Manifests[0].Digest),
```
tonistiigi, line 119: I believe for this case, you don't need to define manifests separately. Can just be `-t newtarget target`.

We could also test both index input and list of separate manifests input.

mqasimsarfraz: Done. Actually the test started failing with single source afterwards I realized annotations were [skipped because of early return](https://github.com/docker/buildx/pull/1965/files#diff-273961e2b2f5a64c8ad90d15663a519f07200712496ae01f0b8b8d989a0d95aaR80). I fixed it now. Thanks!

util/imagetools/create.go
mqasimsarfraz: Done.

tests/imagetools.go
```diff
@@ -14,25 +15,15 @@ import (
...
-	COPY foo-${TARGETARCH} /foo
-	`)
-	dir := tmpdir(
-		t,
-		fstest.CreateFile("Dockerfile", dockerfile, 0600),
-		fstest.CreateFile("foo-amd64", []byte("foo-amd64"), 0600),
-		fstest.CreateFile("foo-arm64", []byte("foo-arm64"), 0600),
-	)
-
+	dir := createDockerfile(t)
```
jedevc, line 26: We shouldn't split this out to a separate function, this is specific to doing multi-arch filtering stuff. We can define a much simpler dockerfile for the annotations tests IMO.

util/imagetools/create.go
```diff
@@ -266,3 +291,32 @@ func detectMediaType(dt []byte) (string, error) {
 
 	return images.MediaTypeDockerSchema2ManifestList, nil
 }
+
+func parseAnnotations(ann map[string]string) (map[string]map[string]string, error) {
+	annotationRegexp := regexp.MustCompile(`^([a-z-]+):(\S+)$`)
+	indexAnnotations := make(map[string]string)
+	manifestDescriptorAnnotations := make(map[string]string)
+	for k, v := range ann {
+		groups := annotationRegexp.FindStringSubmatch(k)
```
jedevc, line 300: Instead of using a regexp here, wdyt about using https://github.com/moby/buildkit/blob/dd0053cdce470b1355fdb0bd5a8f2b0fc506d842/exporter/containerimage/exptypes/annotations.go#L85 instead?

Obviously, it's not perfect, since we'd need to add the `annotation-` prefix here, which might make the error message a bit odd, but we could always upstream a buildkit fix later that would rework the function to have the caller remove the prefix there (so we wouldn't have to do that).

mqasimsarfraz: > Instead of using a regexp here, wdyt about using https://github.com/moby/buildkit/blob/dd0053cdce470b1355fdb0bd5a8f2b0fc506d842/exporter/containerimage/exptypes/annotations.go#L85 instead?

Initially I had the same idea but we are using `:` as a separator for `type` and annotation key compared to `.` expected [here](https://github.com/moby/buildkit/blob/dd0053cdce470b1355fdb0bd5a8f2b0fc506d842/exporter/containerimage/exptypes/annotations.go#L20). So it needs more work than just appending `annotation-` prefix.

> but we could always upstream a buildkit fix later that would rework the function

Does it make sense to make buildkit parser to also handle `:` separator. It sounded specific to buildx so I feel we should keep it here. wdyt?

tests/imagetools.go
mqasimsarfraz: I don't have a strong opinion on this so would be fine going to previous/different implementation. But it has been suggested [here](https://github.com/docker/buildx/pull/1965#discussion_r1281631431).

jedevc: Ah fair, I missed that, sorry!

util/imagetools/create.go
jedevc: Personally, I'd prefer the conversion to buildkit's form for now, so that we only have one source of truth for parsing these keys - I'm happy to follow up in buildkit to rework the logic to be a bit more reusable for this case.

I also just noticed (sorry), that we aren't handling platforms? `manifest-descriptor` supports a platform, and should only be attached to the descriptor for those platforms.

mqasimsarfraz: > I'm happy to follow up in buildkit to rework the logic to be a bit more reusable for this case.

Agreed. That should be the way moving forward. I added a comment regrading using buildkit once it supports our use-case here.

> I also just noticed (sorry), that we aren't handling platforms?

Great Catch. Done.

jedevc (approved):
LGTM :tada: Thanks @mqasimsarfraz!

mqasimsarfraz:
Can you please merge this ? or do we have any open items against it? :)

PR #1966, opened 2023-07-24
build: rework node resolution

jedevc:
This patch reworks and updates the node resolution logic for selecting a node from a builder.

The new implementation reworks the logic to make use of containerd's `platforms.MatchComparer` interface instead of manually associated strings, and additionally provides a few behavioural changes over the original implementation:

- Platforms can be matched with non-strict semantics. e.g. `i386` builds can be scheduled on an `amd64` node, `arm/v7` builds can be scheduled on an `arm/v8` node.

We also add a new collection of tests for tracking regressions and making the intended behaviour clearer.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -135,218 +135,6 @@ func filterAvailableNodes(nodes []builder.Node) ([]builder.Node, error) {
 	return nil, err
 }
 
-type driverPair struct {
-	driverIndex int
-	platforms   []specs.Platform
-	so          *client.SolveOpt
-	bopts       gateway.BuildOpts
-}
-
-func driverIndexes(m map[string][]driverPair) []int {
-	out := make([]int, 0, len(m))
-	visited := map[int]struct{}{}
-	for _, dp := range m {
-		for _, d := range dp {
-			if _, ok := visited[d.driverIndex]; ok {
-				continue
-			}
-			visited[d.driverIndex] = struct{}{}
-			out = append(out, d.driverIndex)
-		}
-	}
-	return out
-}
-
-func allIndexes(l int) []int {
-	out := make([]int, 0, l)
-	for i := 0; i < l; i++ {
-		out = append(out, i)
-	}
-	return out
-}
-
-func ensureBooted(ctx context.Context, nodes []builder.Node, idxs []int, pw progress.Writer) ([]*client.Client, error) {
-	clients := make([]*client.Client, len(nodes))
-
-	baseCtx := ctx
-	eg, ctx := errgroup.WithContext(ctx)
-
-	for _, i := range idxs {
-		func(i int) {
-			eg.Go(func() error {
-				c, err := driver.Boot(ctx, baseCtx, nodes[i].Driver, pw)
-				if err != nil {
-					return err
-				}
-				clients[i] = c
-				return nil
-			})
-		}(i)
-	}
-
-	if err := eg.Wait(); err != nil {
-		return nil, err
-	}
-
-	return clients, nil
-}
-
-func splitToDriverPairs(availablePlatforms map[string]int, opt map[string]Options) map[string][]driverPair {
-	m := map[string][]driverPair{}
-	for k, opt := range opt {
-		mm := map[int][]specs.Platform{}
-		for _, p := range opt.Platforms {
-			k := platforms.Format(p)
-			idx := availablePlatforms[k] // default 0
-			pp := mm[idx]
-			pp = append(pp, p)
-			mm[idx] = pp
-		}
-		// if no platform is specified, use first driver
-		if len(mm) == 0 {
-			mm[0] = nil
-		}
-		dps := make([]driverPair, 0, 2)
-		for idx, pp := range mm {
-			dps = append(dps, driverPair{driverIndex: idx, platforms: pp})
-		}
-		m[k] = dps
-	}
-	return m
-}
-
-func resolveDrivers(ctx context.Context, nodes []builder.Node, opt map[string]Options, pw progress.Writer) (map[string][]driverPair, []*client.Client, error) {
-	dps, clients, err := resolveDriversBase(ctx, nodes, opt, pw)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	bopts := make([]gateway.BuildOpts, len(clients))
-
-	span, ctx := tracing.StartSpan(ctx, "load buildkit capabilities", trace.WithSpanKind(trace.SpanKindInternal))
-
-	eg, ctx := errgroup.WithContext(ctx)
-	for i, c := range clients {
-		if c == nil {
-			continue
-		}
-
-		func(i int, c *client.Client) {
-			eg.Go(func() error {
-				clients[i].Build(ctx, client.SolveOpt{
-					Internal: true,
-				}, "buildx", func(ctx context.Context, c gateway.Client) (*gateway.Result, error) {
-					bopts[i] = c.BuildOpts()
-					return nil, nil
-				}, nil)
-				return nil
-			})
-		}(i, c)
-	}
-
-	err = eg.Wait()
-	tracing.FinishWithError(span, err)
-	if err != nil {
-		return nil, nil, err
-	}
-	for key := range dps {
-		for i, dp := range dps[key] {
-			dps[key][i].bopts = bopts[dp.driverIndex]
-		}
-	}
-
-	return dps, clients, nil
-}
-
-func resolveDriversBase(ctx context.Context, nodes []builder.Node, opt map[string]Options, pw progress.Writer) (map[string][]driverPair, []*client.Client, error) {
-	availablePlatforms := map[string]int{}
-	for i, node := range nodes {
-		for _, p := range node.Platforms {
-			availablePlatforms[platforms.Format(p)] = i
-		}
-	}
-
-	undetectedPlatform := false
-	allPlatforms := map[string]struct{}{}
-	for _, opt := range opt {
-		for _, p := range opt.Platforms {
-			k := platforms.Format(p)
-			allPlatforms[k] = struct{}{}
-			if _, ok := availablePlatforms[k]; !ok {
-				undetectedPlatform = true
-			}
-		}
-	}
-
-	// fast path
-	if len(nodes) == 1 || len(allPlatforms) == 0 {
-		m := map[string][]driverPair{}
-		for k, opt := range opt {
-			m[k] = []driverPair{{driverIndex: 0, platforms: opt.Platforms}}
-		}
-		clients, err := ensureBooted(ctx, nodes, driverIndexes(m), pw)
-		if err != nil {
-			return nil, nil, err
-		}
-		return m, clients, nil
-	}
-
-	// map based on existing platforms
-	if !undetectedPlatform {
-		m := splitToDriverPairs(availablePlatforms, opt)
-		clients, err := ensureBooted(ctx, nodes, driverIndexes(m), pw)
-		if err != nil {
-			return nil, nil, err
-		}
-		return m, clients, nil
-	}
-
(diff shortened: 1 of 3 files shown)


build/resolver.go
```diff
@@ -0,0 +1,365 @@
...
+		// We want to match against the strictest possible platform first,
+		// descending to the least strict. If the match is not perfect, we
+		// want to boot all the nodes to get the full list of available
+		// platforms and retry all the perfect matchers before trying the
+		// inperfect ones.
+		var matchers []matcher
+		matchers = append(matchers, matcher{platforms.OnlyStrict(p), true})
+		matchers = append(matchers, matcher{platforms.Only(p), true})
+		matchers = append(matchers, matcher{newArchMatcherStrict(p), false})
+		matchers = append(matchers, matcher{newArchMatcher(p), false})
```
crazy-max, line 79: I think we could add a specific case where `arm` arch would match `arm64`.

crazy-max: Ah damn forgot it's already the case nevermind https://github.com/docker/buildx/blob/master/vendor/github.com/containerd/containerd/platforms/compare.go#L70-L82

build/resolver_test.go
```diff
@@ -0,0 +1,320 @@
...
+	require.True(t, perfect)
+	require.Len(t, res, 1)
+	require.Equal(t, "bbb", res[0].Node().Builder)
+
+	// arm/v7 can be built on arm/v8
+	res, perfect, err = r.resolve(context.TODO(), []specs.Platform{platforms.MustParse("linux/arm/v7")}, nil, nil)
+	require.NoError(t, err)
+	require.True(t, perfect)
+	require.Len(t, res, 1)
+	require.Equal(t, "bbb", res[0].Node().Builder)
```
crazy-max, line 148: `arm` can also be built on `arm64`, we should have a case for this.

crazy-max (commented):
It's a bit out of scope but it would be nice to check if supported platforms reported by BuildKit are emulated or not and if so, native ones should take priority.

build/resolver_test.go
```diff
@@ -0,0 +1,320 @@
...
+
+	// arm64 is an alias for arm/v8
+	res, perfect, err = r.resolve(context.TODO(), []specs.Platform{platforms.MustParse("linux/arm/v8")}, nil, nil)
+	require.NoError(t, err)
+	require.True(t, perfect)
+	require.Len(t, res, 1)
+	require.Equal(t, "bbb", res[0].Node().Builder)
+
+	// arm/v7 can be built on arm/v8
```
crazy-max, line 143: Actually I misread in https://github.com/docker/buildx/pull/1966/files#r1272167958 and it works as expected but comment on this case should be:

```suggestion
	// arm/v7 can be built on arm64
```

build/resolver_test.go
```diff
@@ -0,0 +1,320 @@
...
+	})
+
+	// v8 can't be built on v7 (so we should select the default)...
+	res, perfect, err := r.resolve(context.TODO(), []specs.Platform{platforms.MustParse("linux/arm/v8")}, nil, nil)
+	require.NoError(t, err)
+	require.False(t, perfect)
+	require.Len(t, res, 1)
+	require.Equal(t, "aaa", res[0].Node().Builder)
+
+	// ...but v6 can be built on v8
```
crazy-max, line 164:
```suggestion
	// ...but v6 can be built on v7
```

build/resolver_test.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+func TestSplitNodeMultiPlatformNoUnify(t *testing.T) {
+	r := makeTestResolver(map[string][]specs.Platform{
+		"aaa": {platforms.MustParse("linux/amd64")},
+		"bbb": {platforms.MustParse("linux/amd64"), platforms.MustParse("linux/riscv64")},
+	})
+
+	// the "best" choice would be the node with both platforms, but we're using
+	// a naive algorithm that doesn't try to unify the platforms
```
crazy-max, line 296: :+1:

crazy-max (approved):
Made some tests on my side and matching is way smarter than before :tada:

As said in https://github.com/docker/buildx/pull/1966#pullrequestreview-1543341240 we could check in follow-up if supported platforms reported by BuildKit are emulated or not and if so, native one should take priority.

I was thinking we would need BuildKit changes to return the native platform as a new worker opt but it doesn't seem necessary as it's always the first item of the list of supported platforms https://github.com/moby/buildkit/blob/687091bb6c8aaa0185cdc570c4db3db533f329d0/util/archutil/detect.go#L23.

build/resolver.go
```diff
@@ -0,0 +1,360 @@
+package build
```
tonistiigi, line 1: resolver as a filename in here is overused. In all codebases/pkgs it means a thing that provides a connection to image registry based on name.

build/resolver.go
```diff
@@ -0,0 +1,360 @@
...
+func (dp resolvedNode) Client() *client.Client {
+	clients, err := dp.resolver.boot(context.TODO(), []int{dp.driverIndex}, nil)
+	if err != nil {
+		panic(err)
+	}
+	return clients[0]
+}
+
+func (dp resolvedNode) Opts() gateway.BuildOpts {
+	opts, err := dp.resolver.opts(context.TODO(), []int{dp.driverIndex}, nil)
```
tonistiigi, line 41: Why are these contexts not propagated from the caller?

build/resolver.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (dp resolvedNode) Node() builder.Node {
+	return dp.resolver.nodes[dp.driverIndex]
+}
+
+func (dp resolvedNode) Client() *client.Client {
+	clients, err := dp.resolver.boot(context.TODO(), []int{dp.driverIndex}, nil)
+	if err != nil {
+		panic(err)
```
tonistiigi, line 35: No panics, even for internal logic errors. (In some edge cases they are ok where they almost behave like build-time errors but that doesn't seem to be a case like this).

build/build.go
```diff
@@ -716,10 +504,15 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 		}
 	}
 
-	m, clients, err := resolveDrivers(ctx, nodes, opt, w)
+	driverRes := newDriverResolver(nodes)
```
tonistiigi, line 507: nit: if the object only has one usable method this call should still be `resolveDrivers()`

build/resolver.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	return clients[0]
+}
+
+func (dp resolvedNode) Opts() gateway.BuildOpts {
```
tonistiigi, line 40: These are `BuildOpts/ClientBuildOpts/GatewayBuildOpts`

build/resolver.go
```diff
@@ -0,0 +1,360 @@
...
+		pk := platforms.Format(p)
+		if matchers, ok := matcherCache.Load(pk); ok {
+			return matchers.([]matcher)
+		}
+
+		// We want to match against the strictest possible platform first,
+		// descending to the least strict. If the match is not perfect, we
+		// want to boot all the nodes to get the full list of available
+		// platforms and retry all the perfect matchers before trying the
+		// inperfect ones.
```
tonistiigi, line 74: imperfect

build/resolver.go
```diff
@@ -0,0 +1,360 @@
...
+	// create a collection of ordered matchers for each platform
+	var matcherCache sync.Map
+	lookup := func(p specs.Platform) []matcher {
+		p = platforms.Normalize(p)
+		pk := platforms.Format(p)
+		if matchers, ok := matcherCache.Load(pk); ok {
+			return matchers.([]matcher)
+		}
+
+		// We want to match against the strictest possible platform first,
```
tonistiigi, line 70: How is this different from `Only()` that also sorts by priority?

build/resolver.go
```diff
@@ -0,0 +1,360 @@
...
+type nodeResolver struct {
+	nodes    []builder.Node
+	clients  flightcontrol.Group[*client.Client]
+	opt      flightcontrol.Group[gateway.BuildOpts]
+	matchers []func(specs.Platform) matcher
+}
+
+func newDriverResolver(nodes []builder.Node) *nodeResolver {
+	// create a collection of ordered matchers for each platform
+	var matcherCache sync.Map
```
tonistiigi, line 62: I'm not sure if this cache really buys us anything. The matchers are just value comparisons that are faster than map lookup (and I don't think there is a big chance of duplicates).

build/resolver.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func newArchMatcherStrict(platform specs.Platform) *archMatcher {
+	return &archMatcher{
+		platform: platform,
+		marcher:  platforms.OnlyStrict(platform),
+	}
+}
+
+func (m *archMatcher) Match(p specs.Platform) bool {
```
tonistiigi, line 355: Why would building on same arch for a different OS be faster than any other node?

Eg. if you are building for `darwin/arm64` and have nodes `linux/amd64`, `linux/arm64` then it should not go to `linux/arm64` but to the first node.

tonistiigi (commented):
One problem with this is that (imho) the `platforms.Only` implementation for arm64 is wrong in containerd. It assumes that all `arm64` mean compatibility with `arm` but that is not the case in practice, eg. no Apple chip is 32bit compatible.

We should at least make sure that if there is both `arm` node and  `arm64` node then `arm` builds will not match with `arm64`.

build/resolver.go
jedevc: The theory of this is not about faster speed, but about output artifacts.

OS of a system can be a bit of a lie sometimes, e.g. WSL, or Docker Desktop, etc. However, in those cases, while not perfect, it would be better to match to the architecture, instead of defaulting to just the first.

In your scenario, this could happen with a remote builder connected to Docker Desktop - if I build and `--load` without a `--platform` flag, then we'll just select the first builder, which might be the non-native architecture `linux/amd64`. We _should_ prefer to load the `linux/arm64` image to the container store instead.

It's not perfect by any measure, but I think it's a bit more reasonable than just picking the first. This way, we don't require that the user reconfigures the order of their builders (I'm not sure what the UX of that would even be like) to get the desired default behavior.

build/resolver.go
tonistiigi: > What we want here is to try and find a platform equal to or greater than the specified target.

Isn't this just reversing the order of the two input platforms?

build/resolver.go
tonistiigi: (🤔 I thought I already left a comment for this, either I forgot to submit or it was somewhere else)

The feature you are describing here is determining the implicit platform (if none is given by user) based on `DOCKER_CONTEXT/-o context=` value if `--load` is used. If we want this then it should be specific to the `--load` behavior and look at the context value. Without `--load` it should be the native platform of the first node.

>  which might be the non-native architecture linux/amd64. 

This is only the case when there is mismatch of configurations between builder and load target. Otherwise the first is always native. I don't understand how ignoring the host allows determining the load target.

> OS of a system can be a bit of a lie sometimes, e.g. WSL, or Docker Desktop, etc

These are server-side platforms. I don't understand where the difference could come from.

jedevc: Removed this for now, I'll open a follow-up PR to do this at some point.

crazy-max (approved):

crazy-max:
> @crazy-max @tonistiigi does this make sense to take for v0.12?

I would like this one in 0.12 as well. WDYT @tonistiigi?

tonistiigi:
😬 , I forgot to rerun CI before merging (all was green and no conflicts), and this broke everything so I had to revert. Please reopen.

Issue #1857, opened 2023-05-31
Improve buildx test coverage


jedevc:
We've now merged #1770 to add support for integration tests :tada: However, buildx has been around for several years before these tests, so we don't have anywhere near the amount of test coverage that we want. These tests are essential to track regressions, detect changes in behavior as we refactor and build new functionality, and add new modes and drivers.

In general, all _new_ functionality should be added with tests wherever feasible, but we need a plan to work through the existing backlog of code with no tests. This issue is to track what broad groups of tests we need, and how far along we are. This list isn't currently complete -- but as *a general rule of thumb*, if we document the behavior on https://docs.docker.com, we should be testing it.

Ideally, each sub-point in the list would end up as a PR, though maybe it makes sense to group some of the smaller ones together, or split some of the larger ones up.

**(note to maintainers, feel free to add anything that you think is missing :partying_face:)**

---

The first type of tests that we need are for the top-level `commands` package - at the time of writing, these currently reside in the top-level `tests` package. These are designed to test each of the buildx subcommands, such as `build`, `bake`, `inspect`, etc.

The areas that we need coverage for:

- [ ] Building with `buildx build` / `buildx bake`
  - [ ] Multi-platform builds (`--platform`)
  - [ ] Cache (`--cache-to`/`--cache-from`/`--no-cache`/`--no-cache-filter`)
  - [ ] Exporters (`--output`/`--load`/`--push`/`--tag`)
  - [ ] Build arguments / build contexts (`--build-arg`/`--build-context`)
  - [ ] Metadata recording (`--metadata-file`/`--iidfile`)
  - [ ] Cross node builds (builds that occur across multiple nodes in a builder)
  - [ ] Buildkit subrequests (`--print`)
  - [ ] Secrets (`--secret`/`--ssh`)
  - [ ] Attestations (`--attest`/`--provenance`/`--sbom`)
  - [ ] Progress output (`--progress`)
    - Not sure quite what this would look like, but some sanity checks on stdout to ensure that progress is continually printed, the ticker keeps updating, any relevant log messages correctly appear, warnings are displayed at the end of build, etc. 
  - [ ] Building using the controller interface (`BUILDX_EXPERIMENTAL` env var, `--detach` flag)
  - [ ] Tracing support
  - [ ] Source policy
- [ ] Registry manipulation with `buildx imagetools`
  - [ ] Inspecting images (`buildx imagetools inspect`)
  - [ ] Creating images (`buildx imagetools create`)
[...]

PR #1985, opened 2023-08-02
Integration test for docker buildx version

jsternberg:
An integration test for `docker buildx version` has been created. The
integration test checks that there is one line output, the output is
composed of three sections, and that these sections could feasibly be
the package path, version, and revision information.

The intention of the checks is to find obvious errors in the output like
the package path not existing or the version and revision being swapped.
It is not intended to assert that these values must be certain values
because it is assumed these values may vary depending on the build
process for buildx.

Related to #1857.


tests/version.go
```diff
@@ -0,0 +1,49 @@
...
+	// Just checking if the import path is a valid Go
+	// path should be suitable enough to make sure this is ok.
+	// Using CheckImportPath instead of CheckPath as it is less
+	// restrictive.
+	require.NoError(t, module.CheckImportPath(components[0]), "First component was not a valid import path: %+v", components[0])
+
+	// Second component should be a version.
+	// This defaults to something that's still compatible
+	// with semver.
+	require.True(t, semver.IsValid(components[1]), "Second component was not valid semver: %+v", components[1])
```
crazy-max, line 42: It seems to make sense that version is not semver when running tests. We use `git describe` to set the version: https://github.com/docker/buildx/blob/c010d3de8dc8771f89b0ec978bb58105652071d0/hack/git-meta#L7

And looking at the GHA workflow: https://github.com/docker/buildx/blob/c010d3de8dc8771f89b0ec978bb58105652071d0/.github/workflows/build.yml#L78-L80

We don't fetch tags so only commit sha will be returned. It should work if you set:

```yaml
      -
        name: Checkout
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
```

See https://github.com/actions/checkout/#usage

> ```yaml
> # Number of commits to fetch. 0 indicates all history for all branches and tags.
> # Default: 1
> fetch-depth: ''
> ```

go.mod
```diff
@@ -39,6 +39,7 @@ require (
 	github.com/zclconf/go-cty v1.10.0
 	go.opentelemetry.io/otel v1.14.0
 	go.opentelemetry.io/otel/trace v1.14.0
+	golang.org/x/mod v0.12.0
```
crazy-max, line 42: If we don't need to bump to v0.12.0, can you just keep v0.9.0? But if needed then can you do the vendoring in a dedicated commit?

thaJeztah:
Perhaps we should check that the last element is the version, and not necessarily that there's 3 components (and one line); https://www.gnu.org/prep/standards/html_node/_002d_002dversion.html#g_t_002d_002dversion

ideally the middle part (packaging info) would be between brackets, but at least convention / guidelines (see above, although admitted that is for --version) is the last element of the first line to be the version.

The output _may_ have multiple lines, which can be used to print (e.g.) licensing info / notices

jedevc:
@jsternberg just a small note - I tend to avoid trying to include PR/issue numbers in commits, just since GitHub chooses to then spam these quite heavily into those issue descriptions :cry: (see in https://github.com/docker/buildx/issues/1857)

jsternberg:
@jedevc yea I was wondering about that. I'll remove it. The merge commit should have the github pull request number which could then get traced back to the original PR and I can keep the issue description in the PR itself so it doesn't spam.

go.mod
jsternberg: Yea I can throw this back down to `v0.9.0`. I figured it was likely at `v0.9.0` because it was an implicit dependency, but it's likely better to just do any version upgrades as a different PR anyway.

jsternberg:
@thaJeztah I think changing the check to follow the standard set by GNU would be fine although it seems our current output prints the revision last when it's available so we'd probably just need to check "is there something that looks like a version in the first line?"

I do think there's some value in making the check a bit more strict though. There may be people potentially using the output for some reason and, while the output is likely not listed as stable, we shouldn't change it accidentally. I also think there's some value in the test breaking if the output changes drastically and then fixing the test to compensate.

I do also think that testing that the output is a single line might be a bit aggressive. I think I should remove that.

crazy-max (approved):

jedevc (approved):

PR #1988, opened 2023-08-04
build: error on attests on non-multiplatform driver

jedevc:
On drivers that do not support multi-platform builds (the default `docker` driver), we do not support building attestations (unless using the containerd store).

We need to check this feature before attempting to build using attestations.

(thanks @cdupuis for the find :tada:)

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -454,7 +454,7 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 			attests[k] = *v
 		}
 	}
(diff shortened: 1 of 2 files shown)


crazy-max (approved):

jedevc:
Hm, I added `needs-cherry-pick/v0.11`, but since this bug actually was there since v0.10, is that the right call?

crazy-max:
> Hm, I added `needs-cherry-pick/v0.11`, but since this bug actually was there since v0.10, is that the right call?

Yes I think it still needs to be backported

dvdksn:
I think we could add a note in attestation docs too.

PR #1998, opened 2023-08-10
build: improve error messages for docker driver

### LINKED ISSUE #1996: Docker driver multiplatform error message can confuse users
### Description

The current error when trying to build an image for multiple platforms with the Docker driver (BuildKit inside Engine _without_ the containerd snapshotter enabled) is:

```
ERROR: Multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
```

We could improve this by pointing users to our multiplatform build documentation.

dvdksn:

Improve error messages, and add documentation links, for errors related to the limited capabilities of the `docker` driver

- needs docker/docs#17902

Closes #1996

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -391,7 +391,7 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 
 	for _, e := range opt.CacheTo {
 		if e.Type != "inline" && !nodeDriver.Features(ctx)[driver.CacheExport] {
-			return nil, nil, notSupported(nodeDriver, driver.CacheExport)
+			return nil, nil, notSupported(driver.CacheExport, nodeDriver, "https://docs.docker.com/go/build-cache-backends/")
 		}
 	}
 
@@ -529,7 +529,7 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	// set up exporters
 	for i, e := range opt.Exports {
 		if e.Type == "oci" && !nodeDriver.Features(ctx)[driver.OCIExporter] {
-			return nil, nil, notSupported(nodeDriver, driver.OCIExporter)
+			return nil, nil, notSupported(driver.OCIExporter, nodeDriver, "https://docs.docker.com/go/build-exporters/")
 		}
 		if e.Type == "docker" {
 			features := docker.Features(ctx, e.Attrs["context"])
@@ -555,7 +555,7 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 					opt.Exports[i].Output = wrapWriteCloser(w)
 				}
 			} else if !nodeDriver.Features(ctx)[driver.DockerExporter] {
-				return nil, nil, notSupported(nodeDriver, driver.DockerExporter)
+				return nil, nil, notSupported(driver.DockerExporter, nodeDriver, "https:/docs.docker.com/go/build-exporters/")
 			}
 		}
 		if e.Type == "image" && nodeDriver.IsMobyDriver() {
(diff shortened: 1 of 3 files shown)


crazy-max:
If you're willing to add integration tests like you have done for build progress that would be nice

dvdksn:
@crazy-max I wanted to add that actually! But I wasn't sure how to make an integration test for a specific driver. Is it just checking integration.Sandbox.Name(), and skip if name != "docker"?

crazy-max:
> @crazy-max I wanted to add that actually! But I wasn't sure how to make an integration test for a specific driver. Is it just checking integration.Sandbox.Name(), and skip if name != "docker"?

Like https://github.com/docker/buildx/blob/86ae8ea854e639a3267399fdf3f99e9b900f6e09/tests/build.go#L185-L187

crazy-max:
@dvdksn Pushed extra commit to split tests in dedicated funcs. We need to also skip tests for `docker+containerd` case. Feel free to squash.

crazy-max (approved):

build/build.go
```diff
@@ -1560,8 +1560,22 @@ func waitContextDeps(ctx context.Context, index int, results *waitmap.Map, so *c
...
+}
+
+func exporterNotSupported(d driver.Driver, f driver.Feature) error {
+	return errors.Errorf(`%s is not supported for the %s driver.
+Switch to a different driver and try again.
+Learn more at https://docs.docker.com/go/build-exporters/`, f, d.Factory().Name())
+}
+
+func multiPlatformBuildNotSupported(d driver.Driver) error {
+	return errors.Errorf(`Multi-platform build is not supported for the %s driver without the containerd image store.
```
tonistiigi, line 1576: The other cases should also work with containerd image store if we want to mention that.

dvdksn:
Didn't realize the c8d restriction applied to all cases, so I made some refactoring

build/build.go
```diff
@@ -627,7 +627,7 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 			pp[i] = platforms.Format(p)
 		}
 		if len(pp) > 1 && !nodeDriver.Features(ctx)[driver.MultiPlatform] {
-			return nil, nil, notSupported(nodeDriver, driver.MultiPlatform)
+			return nil, nil, notSupported(driver.MultiPlatform, nodeDriver, "https://docs.docker.com/go/build-multi-platform")
```
thaJeztah, line 630: Do the go-redirect URLs need a trailing slash?

tonistiigi (approved):

build/build.go
dvdksn: Not strictly needed; the redirect handles both cases: https://github.com/docker/docs/commit/bf4a605ad14a79b9e4e935ded0e0b0773adc739f

But probably a good idea to add them. Let me fix that.

thaJeztah: Ah, yes, I was aware we were _able_ to handle both, but I think we use URLs with a trailing `/` as the "canonical" one for all our links, so thought I'd be better to be consistent and use URLs ending with a `/`. (For the non-`/go/` URLs this prevents an extra redirect).

crazy-max:
@dvdksn needs rebase :pray:

tests/build.go
```diff
@@ -376,3 +379,39 @@ func testBuildLabelNoKey(t *testing.T, sb integration.Sandbox) {
...
+	if sb.Name() != "docker" {
+		t.Skip("skipping test for non-docker workers")
+	}
+
+	dir := createTestProject(t)
+	cmd := buildxCmd(sb, withArgs("build", "--cache-to=type=registry", dir))
+	out, err := cmd.CombinedOutput()
+	require.Error(t, err, string(out))
+	require.Contains(t, string(out), "Cache export is not supported")
+}
```
thaJeztah, line 393: Wondering if we should start using a test-table for these tests; did a quick try what it'd look like; https://github.com/dvdksn/buildx/pull/2

WDYT @crazy-max ?

crazy-max: Yes sounds good but I think we should split ones about empty key from other tests

thaJeztah: Opened as https://github.com/docker/buildx/pull/2041

> but I think we should split ones about empty key from other tests

Any specific reason for that? (just wondering)

Issue #1868, opened 2023-06-07
`docker-container` boot fails with "container conflict" for multiple simultaneous builds


HarvyNBB:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

* Description:
buildkit fails with container conflict for multiple simultaneous builds

Note: I'm not sure whether this is buildx issue or buildkit issue

### Expected behaviour

buildkit should work successfully for simultaneous parallel builds by auto detecting this scenario and creating unique containers for each build

### Actual behaviour

buildkit fails with container conflict error

### Buildx version

github.com/docker/buildx v0.10.5 86bdced

### Docker info

```text
09:17:32  Client: Docker Engine - Community
09:17:32   Version:    24.0.2
09:17:32   Context:    default
09:17:32   Debug Mode: false
09:17:32   Plugins:
09:17:32    buildx: Docker Buildx (Docker Inc.)
09:17:32      Version:  v0.10.5
09:17:32      Path:     /usr/libexec/docker/cli-plugins/docker-buildx
09:17:32    compose: Docker Compose (Docker Inc.)
09:17:32      Version:  v2.18.1
09:17:32      Path:     /usr/libexec/docker/cli-plugins/docker-compose
09:17:32  
09:17:32  Server:
09:17:32   Containers: 0
09:17:32    Running: 0
09:17:32    Paused: 0
09:17:32    Stopped: 0
09:17:32   Images: 0
09:17:32   Server Version: 24.0.2
09:17:32   Storage Driver: overlay2
09:17:32    Backing Filesystem: extfs
09:17:32    Supports d_type: true
09:17:32    Using metacopy: false
09:17:32    Native Overlay Diff: false
09:17:32    userxattr: true
09:17:32   Logging Driver: json-file
09:17:32   Cgroup Driver: systemd
09:17:32   Cgroup Version: 2
09:17:32   Plugins:
09:17:32    Volume: local
09:17:32    Network: bridge host ipvlan macvlan null overlay
09:17:32    Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
09:17:32   Swarm: inactive
09:17:32   Runtimes: io.containerd.runc.v2 runc
09:17:32   Default Runtime: runc
[...]

PR #2000, opened 2023-08-10
docker-container: avoid fail if container conflict

jedevc:
Fixes the race condition where two boots are executed simultaneously across multiple processes.

Additionally, this would probably also fix https://github.com/docker/buildx/issues/1868.

We initially check to see if the container exists, but if during container creation we get a name conflict, we don't treat this error as a hard failure, and instead move immediately into waiting for the node to boot.

Not 100% sure about this, we could alternatively attempt to get a file lock when attempting to create the docker container (@crazy-max's suggestion).

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -22,6 +22,7 @@ import (
 	"github.com/docker/docker/api/types/mount"
 	"github.com/docker/docker/api/types/network"
 	dockerclient "github.com/docker/docker/client"
+	"github.com/docker/docker/errdefs"
 	dockerarchive "github.com/docker/docker/pkg/archive"
 	"github.com/docker/docker/pkg/idtools"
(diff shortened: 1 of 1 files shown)


crazy-max (approved):
LGTM. What I was thinking with locking the instance file was to avoid race cond when bootstrapping but that should not be an issue for other drivers.

jedevc:
I guess the file lock approach also doesn't work very well if the daemon is remote. e.g. the daemon is remote on a server, and multiple clients are accessing it.

crazy-max:
> I guess the file lock approach also doesn't work very well if the daemon is remote. e.g. the daemon is remote on a server, and multiple clients are accessing it.

Yes indeed. I wonder if we should have smth similar to what you have done here for k8s :thinking: (cc @AkihiroSuda)

PR #2005, opened 2023-08-16
git: wrap upstream gitutil cli

jedevc:
:warning: Requires https://github.com/moby/buildkit/pull/4106.


build/git.go
```diff
@@ -94,7 +95,7 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 	}
 
 	if setGitLabels {
-		if root, err := gitc.RootDir(); err != nil {
+		if root, err := gitc.WorkTree(ctx); err != nil {
```
crazy-max, line 98: I think we still need the `RootDir` func otherwise `root` will always be empty here.

thaJeztah:
@jedevc I see https://github.com/moby/buildkit/pull/4106 was merged; is this still something to complete? (and a rebase)?

jedevc:
Yeah, ideally this would help remove some of the duplication between buildx and buildkit's custom git logic.

Unfortunately, I don't really have any spare cycles at the moment to pick up the rebase - however, I don't *think* there's anything that's critically changed in buildx for this? So shouldn't be too bad if someone else wants to pick up this up.

Issue #1881, opened 2023-06-14
buildx build produces git not found warning even with attestations disabled


illusori:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

The `WARNING: buildx: git was not found in the system. Current commit information was not captured by the build` warning appears on systems without git installed even if attestations are disabled with `--provenance=false` and `--sbom=false`.

I'm presuming attestations are the cause, but there's no documentation for this warning and the documentation on just how attestations data is gathered is fairly threadbare. The fact that this warning gives no indication of why it's a problem and why buildx is complaining about not finding git and what to do to remedy the situation if git not being available is correct behaviour suggests that either the warning and/or documentation is inadequate.

### Expected behaviour

If attestations are switched off, buildx shouldn't be trying to gather data for them.

### Actual behaviour

Warning generated suggests that attestations data is being gathered regardless of whether it's enabled or not.

### Buildx version

github.com/docker/buildx v0.10.4 c513d34049e499c53468deac6c4267ee72948f02

### Docker info

```text
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.10.4
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx

Server:
 Containers: 3
  Running: 2
  Paused: 0
  Stopped: 1
 Images: 42
 Server Version: 24.0.2
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 3dce8eb055cbb6872793272b4f20ed16117344f8
 runc version: v1.1.7-0-g860f061
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
[...]

PR #2012, opened 2023-08-22
git: propogate failure to locate git binary

jedevc:
See https://github.com/docker/buildx/issues/1881#issuecomment-1684837909.

At the moment, a failure to find the git binary means that it's hard to actually debug this when it goes wrong.

This PR reworks the error messages from getting git attributes, so that they're a bit easier to debug, and in general, just a bit clearer.

Before:
    
    WARNING: buildx: git was not found in the system. Current commit information was not captured by the build
    
After:
    
    WARNING: current commit information was not captured by the build: git was not found in the system: <error message>


crazy-max (approved):
Thanks!

Issue #1660, opened 2023-03-03
Support Copy Multiarch Image Tag From One Repository To Another With Digest Id Intact


aggo15:
Many of my users prefer to keep the digest id of an image tag when moving/migrating image from one repository to another. This is for the purpose of tracking where the original image came from.

For single architecture image tag, it is easy to do so with command below:
```
docker image tag <registry>/<src_repo>:<source_image_tag> <registry>/<dest_repo>:<dest_image_tag>
docker push <registry>/<dest_repo>:<dest_image_tag>
```

However, for multi-arch image, it is only working within the same repository, example working command below:
```
docker buildx imagetools create -t <registry>/<src_repo>:<dest_image_tag> <registry>/<src_repo>:<source_image_tag>
```

If I were to change the destination repository, then it will return **400 bad request**.
```
docker buildx imagetools create -t <registry>/<dest_repo>:<dest_image_tag> <registry>/<src_repo>:<source_image_tag>
```

I'm able to work around this issue by manually copy over each layer by digest id to destination repository, which is a hassle to deal with. I don't see this workaround being documented somewhere as well. Work around commands below (assume the multi-arch image tag consists of armv6, armv7 and arm64 architecture):
```
# Copy ARMV6 digest
docker buildx imagetools create -t <registry>/<dest_repo>@<dest_armv6_digest <registry>/<src_repo>@<source_armv6_digest>

# Copy ARMV7 digest
docker buildx imagetools create -t <registry>/<dest_repo>@<dest_armv7_digest> <registry>/<src_repo>@<source_armv7_digest> 

# Copy ARM64 digest
docker buildx imagetools create -t <registry>/<dest_repo>@<dest_arm64_digest> <registry>/<src_repo>@<source_arm64_digest> 

# Copy source tag digest
docker buildx imagetools create -t <registry>/<dest_repo>@<dest_tag_digest> <registry>/<src_repo>@<source_tag_digest>

# Assign a name to destination tag
docker buildx imagetools create -t <registry>/<src_repo>:<dest_image_tag> <registry>/<dest_repo>@<dest_tag_digest>
```

It will be nice if the `docker buildx imagetools create` command handle copy over image tag from one repository to another, since currently it already support copy image tag within the same repository.

Here is my current docker buildx version. I'm on Windows 10 btw.
```
docker buildx version
github.com/docker/buildx v0.10.0 876462897612d36679153c3414f7689626251501

docker version
Client:
 Cloud integration: v1.0.29
[...]

PR #2013, opened 2023-08-22
tests: add imagetools tests for copying manifests and indexes

jedevc:
While in the process of investigating, https://github.com/docker/buildx/issues/1660, realized we didn't have any tests for it.

I've added some tests to ensure that manifests and indexes are copied with their digests preserved intact.

### DIFF
--- a/tests/imagetools.go
+++ b/tests/imagetools.go
@@ -5,6 +5,7 @@ import (
 	"os/exec"
 	"testing"
 
+	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/platforms"
 	"github.com/containerd/continuity/fs/fstest"
 	"github.com/moby/buildkit/util/testutil/integration"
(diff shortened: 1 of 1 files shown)


tests/imagetools.go
```diff
@@ -14,10 +15,120 @@ import (
...
+
+	cmd = buildxCmd(sb, withArgs("imagetools", "inspect", target2, "--raw"))
+	dt, err = cmd.CombinedOutput()
+	require.NoError(t, err, string(dt))
+
+	var idx2 ocispecs.Index
+	err = json.Unmarshal(dt, &idx2)
+	require.NoError(t, err)
+	require.Equal(t, images.MediaTypeDockerSchema2ManifestList, idx2.MediaType)
+	require.Equal(t, 1, len(idx2.Manifests))
```
jedevc, line 65: Out of curiosity: is this case actually correct? If the `imagetools create` input is only a single manifest, do we expect to create an index containing just it? Or would we expect to just copy it unchanged?

crazy-max: Hum I would say we just copy it unchanged imo. It looks linked to https://github.com/docker/buildx/pull/1797#discussion_r1194914910 where we default to OCI now?

jedevc: Hm, I don't think this is oci/docker media type confusion, it seems like this was always present before that PR?

I think maybe we want the just copy behaviour agreed, I think it's worth merging this PR to track the current behaviour, then fixing that as a followup, if we agree that's worth doing?

crazy-max (approved):

PR #2015, opened 2023-08-23
bake: local dockerfile support for remote definition

### LINKED ISSUE #1627: [BUG] Bake build fails when using BAKE_CMD_CONTEXT locally

crazy-max:
fixes #1627

Allow to use a local dockerfile with a remote bake definition like we do for context using `cwd://` protocol.

Can be tested with https://github.com/crazy-max/buildx-buildkit-tests/tree/main/buildx-1627:

```bash
$ cd /tmp
$ cat <<EOT >> Dockerfile.app
FROM scratch
COPY foo /foo
EOT
$ docker buildx bake https://github.com/crazy-max/buildx-buildkit-tests.git#:buildx-1627
```

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -1044,6 +1044,9 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 	if t.Dockerfile != nil {
 		dockerfilePath = *t.Dockerfile
 	}
+	if !strings.HasPrefix(dockerfilePath, "cwd://") {
+		dockerfilePath = path.Clean(dockerfilePath)
+	}
 
 	bi := build.Inputs{
 		ContextPath:    contextPath,
@@ -1054,6 +1057,20 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 		bi.DockerfileInline = *t.DockerfileInline
 	}
 	updateContext(&bi, inp)
+	if strings.HasPrefix(bi.DockerfilePath, "cwd://") {
+		// If Dockerfile is local for a remote invocation, we first check if
+		// it's not outside the working directory and then resolve it to an
+		// absolute path.
+		bi.DockerfilePath = path.Clean(strings.TrimPrefix(bi.DockerfilePath, "cwd://"))
+		if err := checkPath(bi.DockerfilePath); err != nil {
+			return nil, err
+		}
+		var err error
+		bi.DockerfilePath, err = filepath.Abs(bi.DockerfilePath)
+		if err != nil {
+			return nil, err
+		}
+	}
 	if strings.HasPrefix(bi.ContextPath, "cwd://") {
 		bi.ContextPath = path.Clean(strings.TrimPrefix(bi.ContextPath, "cwd://"))
 	}
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -1070,6 +1070,24 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 		if err != nil {
 			return nil, err
 		}
+	} else if !build.IsRemoteURL(bi.DockerfilePath) && strings.HasPrefix(bi.ContextPath, "cwd://") && (inp != nil && build.IsRemoteURL(inp.URL)) {
+		// We don't currently support reading a remote Dockerfile with a local
+		// context when doing a remote invocation because we automatically
+		// derive the dockerfile from the context atm:
+		//
+		// target "default" {
+		//  context = BAKE_CMD_CONTEXT
+		//  dockerfile = "Dockerfile.app"
+		// }
+		//
+		// > docker buildx bake https://github.com/foo/bar.git
+		// failed to solve: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount3004544897/Dockerfile.app: no such file or directory
+		//
+		// To avoid mistakenly reading a local Dockerfile, we check if the
+		// Dockerfile exists locally and if so, we error out.
+		if _, err := os.Stat(filepath.Join(path.Clean(strings.TrimPrefix(bi.ContextPath, "cwd://")), bi.DockerfilePath)); err == nil {
+			return nil, errors.Errorf("reading a dockerfile for a remote build invocation is currently not supported")
+		}
 	}
 	if strings.HasPrefix(bi.ContextPath, "cwd://") {
 		bi.ContextPath = path.Clean(strings.TrimPrefix(bi.ContextPath, "cwd://"))
--- a/tests/bake.go
+++ b/tests/bake.go
@@ -27,6 +27,7 @@ var bakeTests = []func(t *testing.T, sb integration.Sandbox){
 	testBakeRemoteCmdContextEscapeRoot,
 	testBakeRemoteCmdContextEscapeRelative,
 	testBakeRemoteDockerfileCwd,
+	testBakeRemoteLocalContextRemoteDockerfile,
 }
 
 func testBakeLocal(t *testing.T, sb integration.Sandbox) {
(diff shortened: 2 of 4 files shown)


bake/bake.go
```diff
@@ -1048,6 +1050,17 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 		bi.DockerfileInline = *t.DockerfileInline
 	}
 	updateContext(&bi, inp)
+	if strings.HasPrefix(bi.DockerfilePath, "cwd://") {
+		bi.DockerfilePath = path.Clean(strings.TrimPrefix(bi.DockerfilePath, "cwd://"))
+		if err := checkPath(bi.DockerfilePath); err != nil {
+			return nil, err
+		}
+		var err error
+		bi.DockerfilePath, err = filepath.Abs(bi.DockerfilePath)
```
crazy-max, line 1059: We need to return the absolute representation of the dockerfile path after checking its a valid one like we do for context so we can set the right dockerfile dir if context is a remote URL during build.

I was also thinking reading the file in `bi.DockerfileInline` and then set `bi.DockerfilePath = "-"` but that's hacky and would be confusing.

jedevc:
Should we add a test to make sure that https://github.com/docker/buildx/pull/1880#issuecomment-1622000397 still applies? So in the test, if `Dockerfile.app` appears without the `cwd://` prefix, we should error. 

Either - that case should error like in v0.11 (which reverted #1880), or it should pull the docker file from the remote.

crazy-max:
> Should we add a test to make sure that [#1880 (comment)](https://github.com/docker/buildx/pull/1880#issuecomment-1622000397) still applies? So in the test, if `Dockerfile.app` appears without the `cwd://` prefix, we should error.
> 
> Either - that case should error like in v0.11 (which reverted #1880), or it should pull the docker file from the remote.

Yes so for a remote bake definition like:

```hcl
target "default" {
  context = BAKE_CMD_CONTEXT
  dockerfile = "Dockerfile.app"
}
```

Invoked with:

```
docker buildx bake https://github.com/foo/bar.git
```

We should use local context but we don't support reading remote Dockerfile as expected:

```

#1 [internal] load git source https://github.com/foo/bar.git
#1 0.480 ref: refs/heads/main   HEAD
#1 0.480 dea0f6334d719a494a384a13812f604b6b79fed3       HEAD
#1 0.916 dea0f6334d719a494a384a13812f604b6b79fed3       refs/heads/main
#1 CACHED

#2 [internal] load .dockerignore
#2 transferring context:
#2 transferring context: 2B done
#2 DONE 0.0s

#3 [internal] load build definition from Dockerfile.app
#3 transferring dockerfile: 2B done
#3 DONE 0.0s
ERROR: failed to solve: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount3004544897/Dockerfile.app: no such file or directory
```

bake/bake.go
```diff
@@ -1048,6 +1050,21 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
...
+		if err != nil {
+			return nil, err
+		}
+	} else if !build.IsRemoteURL(bi.DockerfilePath) && strings.HasPrefix(bi.ContextPath, "cwd://") && (inp != nil && build.IsRemoteURL(inp.URL)) {
+		if _, err := os.Stat(filepath.Join(path.Clean(strings.TrimPrefix(bi.ContextPath, "cwd://")), bi.DockerfilePath)); err == nil {
+			return nil, errors.Errorf("reading a local dockerfile for a remote build invocation is not allowed when using a local context")
+		}
```
crazy-max, line 1066: Deny access to a local dockerfile for remote invocation with a local context. We need to check if a local Dockerfile exists early at the moment but in follow-up we should support reading a remote Dockerfile. Probably having a `DockerfileState` input like `ContextState` one.

tests/bake.go
```diff
@@ -287,3 +290,101 @@ EOT
...
+	gitutil.GitCommit(git, t, "initial commit")
+	addr := gitutil.GitServeHTTP(git, t)
+
+	out, err := bakeCmd(
+		sb,
+		withDir(dirSrc),
+		withArgs(addr, "--set", "*.output=type=cacheonly"),
+	)
+	require.Error(t, err, out)
+	require.Contains(t, out, "reading a local dockerfile for a remote build invocation is not allowed when using a local context")
```
jedevc, line 389: I think this error message should probably be something like:

```suggestion
	require.Contains(t, out, "reading a dockerfile for a remote build invocation is not currently supported")
```

Since the plan is to have it default to reading it from the remote bake definition context right?

bake/bake.go
jedevc: Just want to check - isn't this independent of what the Context is set to? Isn't it instead dependent on whether the bake file is coming from a remote?

tests/bake.go
crazy-max: Yes sounds good, will be more informative for the end user.

bake/bake.go
crazy-max: It's dependent on remote bake but also if context is not local

c-ameron:
Thanks so much for working on this! 😃

bake/bake.go
```diff
@@ -1038,6 +1037,9 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 	if t.Dockerfile != nil {
 		dockerfilePath = *t.Dockerfile
 	}
+	if !strings.HasPrefix(dockerfilePath, "cwd://") {
```
tonistiigi, line 1040: follow-up: this should work with remote URLs as well

bake/bake.go
```diff
@@ -1048,6 +1050,21 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
...
+		bi.DockerfilePath = path.Clean(strings.TrimPrefix(bi.DockerfilePath, "cwd://"))
+		if err := checkPath(bi.DockerfilePath); err != nil {
+			return nil, err
+		}
+		var err error
+		bi.DockerfilePath, err = filepath.Abs(bi.DockerfilePath)
+		if err != nil {
+			return nil, err
+		}
+	} else if !build.IsRemoteURL(bi.DockerfilePath) && strings.HasPrefix(bi.ContextPath, "cwd://") && (inp != nil && build.IsRemoteURL(inp.URL)) {
```
tonistiigi, line 1063: I'm not sure I understand this condition:

> !build.IsRemoteURL(bi.DockerfilePath) && strings.HasPrefix(bi.ContextPath, "cwd://")

If Dockerfile is not remote and context is based on local dir then why would that be invalid config?

Maybe a comment would help.

bake/bake.go
```diff
@@ -1048,6 +1050,21 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
...
+		if _, err := os.Stat(filepath.Join(path.Clean(strings.TrimPrefix(bi.ContextPath, "cwd://")), bi.DockerfilePath)); err == nil {
+			return nil, errors.Errorf("reading a dockerfile for a remote build invocation is currently not supported")
```
tonistiigi, line 1065: Is this for security?

crazy-max: Yes, see https://github.com/docker/buildx/pull/2015#discussion_r1304145146

crazy-max: Added a comment to better explain the issue.

tonistiigi (approved):

PR #2018, opened 2023-08-24
driver(docker): opt to set additional dial meta to the client

crazy-max:
if you want some changes apart, it then needs:
* closes https://github.com/docker/buildx/pull/2074
* closes https://github.com/docker/buildx/pull/2075

In some cases it can be useful to set additional dial metadata to the client.

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -399,6 +399,10 @@ func (d *Driver) Features(ctx context.Context) map[driver.Feature]bool {
 	}
 }
 
+func (d *Driver) HostGatewayIP(ctx context.Context) (net.IP, error) {
+	return nil, errors.New("host-gateway is not supported by the docker-container driver")
+}
+
 func demuxConn(c net.Conn) net.Conn {
 	pr, pw := io.Pipe()
 	// TODO: rewrite parser with Reader() to avoid goroutine switch
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -3,6 +3,7 @@ package driver
 import (
 	"context"
 	"io"
+	"net"
 
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/util/progress"
@@ -60,6 +61,7 @@ type Driver interface {
 	Rm(ctx context.Context, force, rmVolume, rmDaemon bool) error
 	Client(ctx context.Context) (*client.Client, error)
 	Features(ctx context.Context) map[Feature]bool
+	HostGatewayIP(ctx context.Context) (net.IP, error)
 	IsMobyDriver() bool
 	Config() InitConfig
 }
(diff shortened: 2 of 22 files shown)


driver/docker-container/driver.go
```diff
@@ -380,9 +387,19 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
...
 	}))
+
 	if td, ok := exp.(client.TracerDelegate); ok {
 		opts = append(opts, client.WithTracerDelegate(td))
 	}
+
+	if len(co.Meta) > 0 {
+		opts = append(opts, client.WithGRPCDialOption(grpc.WithChainUnaryInterceptor(func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+			// merge the existing context with new metadata.
+			ctx = metadata.NewOutgoingContext(ctx, co.Meta)
```
crazy-max, line 398: I think it's safe to make a new set of metadata but let me know if you prefer to append instead for extra safety.

driver/kubernetes/driver.go
```diff
@@ -218,9 +225,19 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
 	opts = append(opts, client.WithContextDialer(func(context.Context, string) (net.Conn, error) {
 		return conn, nil
 	}))
+
 	if td, ok := exp.(client.TracerDelegate); ok {
 		opts = append(opts, client.WithTracerDelegate(td))
 	}
+
+	if len(co.Meta) > 0 {
+		opts = append(opts, client.WithGRPCDialOption(grpc.WithChainUnaryInterceptor(func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
```
jedevc, line 234: I don't have all the context for this, but I think we probably want to add this to `grpc.WithChainStreamInterceptor` as well?

driver/docker-container/driver.go
crazy-max: Actually it's not safe, we need to append as we are joining metadata in buildkit like `buildkit-attachable-store-id`. Opened https://github.com/moby/buildkit/pull/4175.

driver/kubernetes/driver.go
crazy-max: As discussed with @tonistiigi, drop support for other drivers for now.

driver/driver.go
```diff
@@ -110,3 +110,15 @@ func historyAPISupported(ctx context.Context, c *client.Client) bool {
 		}
 	}
 }
+
+type ClientOption func(*ClientOptions)
+
+type ClientOptions struct {
+	Meta map[string][]string
```
tonistiigi, line 117: I think `Meta` is too generic in here. `TransportMeta` or `DialMeta`.

tonistiigi (commented):
Bring this out of draft if ready.

tonistiigi (approved):

PR #2020, opened 2023-08-24
build: add --annotation shortcut flag

jedevc:
Fixes https://github.com/docker/buildx/issues/1171.

This extracts the same logic for parsing annotations from the imagetools create command (see #1965), and allows the same flags to be attached to the build command. Additionally, we also allow these to be specified in bake's `target.annotations` field - happy to bikeshed exact syntax, I've just thrown something together (or to pull it out to a separate PR if we're not 100% sure).
    
These annotations are then merged into all provided exporters.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -589,6 +589,7 @@ type Target struct {
 	// Inherits is the only field that cannot be overridden with --set
 	Inherits []string `json:"inherits,omitempty" hcl:"inherits,optional" cty:"inherits"`
 
+	Annotations      []string           `json:"annotations,omitempty" hcl:"annotations,optional" cty:"annotations"`
 	Attest           []string           `json:"attest,omitempty" hcl:"attest,optional" cty:"attest"`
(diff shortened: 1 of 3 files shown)


util/buildflags/export.go
```diff
@@ -74,3 +78,45 @@ func ParseExports(inp []string) ([]*controllerapi.ExportEntry, error) {
 	}
 	return outs, nil
 }
+
+func ParseAnnotations(inp []string) (map[exptypes.AnnotationKey]string, error) {
+	// TODO: use buildkit's annotation parser once it supports setting custom prefix and ":" separator
+	annotationRegexp := regexp.MustCompile(`^(?:([a-z-]+)(?:\[([A-Za-z0-9_/-]+)\])?:)?(\S+)$`)
```
tonistiigi, line 84: follow up: is it really worth using this crazy regexp instead of simple splits?

tonistiigi (approved):

jedevc:
Rebased.

PR #2023, opened 2023-08-25
vendor: update git url handling with changes to buildkit

jsternberg:
The git ssh url parsing API changed a bit and broke how buildx utilized
the API. This updates the affected method with the new method of
identifying an SSH url.

PR #1934, opened 2023-07-12
Use buildkit's `client.Wait` method to bootstrap for remote driver

jedevc:

PR #2035, opened 2023-09-07
vendor: update buildkit to master

jedevc:
Updates Buildkit to the latest master commit (replaces https://github.com/docker/buildx/pull/2023)

This pulls in relevant upstream changes such as:
- https://github.com/moby/buildkit/pull/4142
  - This change requires a small update to some SSH url detection logic
- https://github.com/moby/buildkit/pull/4200
  - With this we can remove the explicit backoffs added in https://github.com/docker/buildx/pull/1934, which fixes the equivalent regression from https://github.com/moby/buildkit/issues/4164 in the remote driver
- https://github.com/moby/buildkit/pull/4113
  - Due to the API changes here, this requires some more substantial changes.
  - This pulls in support for `rawjson`.
  - However, upstream does not currently support `quiet` - since the logic for parsing display modes has been moved into the `progressui.NewDisplay`, we need an upstream patch https://github.com/moby/buildkit/pull/4213 to add this display mode back in. This is actually good - having different sets of display modes between buildx and buildkit was confusing.


jedevc:
Note: test failures will be due to https://github.com/docker/buildx/pull/2023#discussion_r1306906177, we need to update to point to the image built from `buildkit:master` once https://github.com/moby/buildkit/pull/4213 merges. https://github.com/docker/buildx/blob/da6662975f0de310887ba2f2c748cbd21753e871/Dockerfile#L9

go.mod
```diff
@@ -160,3 +160,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/moby/buildkit => ../buildkit
```
crazy-max, line 164: Looks to be an oversight :sweat_smile:

jedevc: It's intentional until #4213 merges - I'll replace with a pointer to my fork though.

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/hashicorp/go-cty-funcs v0.0.0-20200930094925-2721b1e36840
 	github.com/hashicorp/hcl/v2 v2.8.2
-	github.com/moby/buildkit v0.12.1-0.20230804094609-b49a8873179b
+	github.com/moby/buildkit v0.12.1-0.20230907220514-cbfd4023383d
```
crazy-max, line 26: Can you add a comment saying that `BUILDKIT_VERSION` in the Dockerfile should be changed as well when updating this module?

jedevc: Isn't it the same as when updating runc/containerd? The binaries should be updated along with the code, but there are legitimate cases where it may not be necessary - e.g. a change in the buildkit test harnesses, mean no `BUILDKIT_VERSION` bump is necessary.

crazy-max (approved):

Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG XX_VERSION=1.2.1
 ARG DOCKER_VERSION=24.0.2
 ARG GOTESTSUM_VERSION=v1.9.0
 ARG REGISTRY_VERSION=2.8.0
-ARG BUILDKIT_VERSION=v0.11.6
+ARG BUILDKIT_VERSION=master@sha256:58b3d496584c7a33dd129fb4f6dec7828c4f31aa7808a5be1c9835f0fbc33bfe
```
tonistiigi, line 9: Let's not update this. Vendor-only if not a release.

jedevc: See https://github.com/docker/buildx/pull/2023#discussion_r1306906177, I think this is necessary.

We could cherry-pick it to v0.12 branch and release? Otherwise, I don't see a trivial way to solve this - unless we detect the presence of the flag for the tests (@crazy-max).

crazy-max: Yes it's necessary as we are not synced between vendoring adding the `otel-socket-path` flag that is not available in a current buildkit release. We could conditionally set this flag only if available in buildkitd help output like https://github.com/moby/buildkit/compare/master...crazy-max:buildkit:integration-check-flag but I don't think it's worth it.

crazy-max: @jedevc With https://github.com/moby/buildkit/pull/4280 this is not necessary anymore. See https://github.com/docker/buildx/pull/2060.

jedevc: Pulled the changes in, done :)

crazy-max (approved):

PR #2038, opened 2023-09-11
build: support local state group

crazy-max:
follow-up https://github.com/docker/buildx/pull/1735

We currently support saving local state for a build ref but we can't associate multiple refs when building a group of targets. This PR adds support for saving the local state group so we can identify refs that belong to the same group of targets.


localstate/localstate.go
```diff
@@ -9,11 +9,25 @@ import (
 	"github.com/pkg/errors"
 )
 
-const refsDir = "refs"
+const (
+	refsDir  = "refs"
+	groupDir = "__group__"
```
crazy-max, line 14: GC of group dir not currently handled

build/build.go
```diff
@@ -805,6 +800,26 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
...
 	}
 
+	// local state
+	var lsg *localstate.StateGroup
+	if len(opt) > 1 {
+		lsg = &localstate.StateGroup{
+			Ref: identity.NewID(),
+		}
+	}
+	for k, opt := range opt {
```
tonistiigi, line 810: If this doesn't complicate it a lot then I think it can be in a goroutine.

localstate/localstate.go
```diff
@@ -62,6 +76,30 @@ func (ls *LocalState) SaveRef(builderName, nodeName, id string, st State) error
 	return ioutils.AtomicWriteFile(filepath.Join(refDir, id), dt, 0600)
 }
 
+func (ls *LocalState) ReadGroup(id string) (*StateGroup, error) {
```
tonistiigi, line 79: Where is this used? What needs to access refs per group ID?

localstate/localstate.go
```diff
@@ -9,11 +9,25 @@ import (
...
+	LocalPath string
+	// DockerfilePath is the absolute path to the Dockerfile
 	DockerfilePath string
+	// GroupID is the ID of the state group that this ref belongs to
+	GroupID string `json:",omitempty"`
+}
+
+type StateGroup struct {
+	// Ref is the unique ID of the group ref
+	Ref string `json:"-"`
```
tonistiigi, line 28: We should save the configuration of the bake command. This is variables and targets (ideally connected with the build refs). Also path to the bake file(s).

I'm not sure if a full bake definition is also needed. Maybe the `--set` options?

crazy-max: Yes variables, targets and overrides looks good to me. Bake definition can be useful. I'm looking at it.

localstate/localstate.go
crazy-max: This is an helper like `ReadRef` to be consistent.

crazy-max:
Reworked a bit build invocation in first commit as `BuildWithResultHandler` had a bunch of args already so we can now pass variadic opts when instantiating `Build` instead. Wanted this as I was thinking passing a `localstateHandleFunc` but would need https://github.com/docker/buildx/pull/1966 first as resolving the node of the build before saving its state would be necessary but seems enough to just pass `localstate.StateGroup` object instead.

Now it saves the raw definition, inputs and targets for the group. Also needs to save the target for each state ref so we can associate them with the definition.

Didn't set yet hcl variables but I guess we need a new `Variables` field to `bake.Config`. It would be ok for HCL definitions but would not have any meaning for compose files.

commands/bake.go
```diff
@@ -201,7 +197,21 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
 		return nil
 	}
 
-	resp, err := build.Build(ctx, nodes, bo, dockerutil.NewClient(dockerCli), confutil.ConfigDir(dockerCli), printer)
+	var lsg *localstate.StateGroup
+	if len(bo) > 1 {
```
crazy-max, line 201: @tonistiigi I currently set the local state group if there is more than one target but maybe it would also be useful for a single target with bake so we have the definition and inputs attached. WDYT?

build/build.go
crazy-max: Moved back to first opt iteration

crazy-max:
@tonistiigi As discussed, remove variadic opts and adds another arg to `build.Build`. Also needs to set build ref(s) early as well which requires a new Ref field to build options.

bake/bake.go
```diff
@@ -577,6 +580,11 @@ func (c Config) target(name string, visited map[string]*Target, overrides map[st
 	return tt, nil
 }
 
+type Definition struct {
```
tonistiigi, line 583: It's a bit weird now that we have `Config` and `Definition` (that are both quite meaningless names) and the difference is that one has an array, and the other has maps. Any other name ideas, ways to combine, or I guess introducing this type isn't really needed for the functionality this PR introduces.

build/build.go
```diff
@@ -433,6 +432,12 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		SourcePolicy:        opt.SourcePolicy,
 	}
 
+	if opt.Ref != "" {
+		so.Ref = opt.Ref
```
tonistiigi, line 436: This `if` can be just part of the struct definition.

bake/bake.go
```diff
@@ -1104,6 +1112,7 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 
 	bo := &build.Options{
 		Inputs:        bi,
+		Ref:           identity.NewID(),
```
tonistiigi, line 1115: setting this in `cmd/bake:204` seems more logical I think.

commands/bake.go
```diff
@@ -201,7 +197,26 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
...
+		if err != nil {
+			return err
+		}
+		lsg = &localstate.StateGroup{
+			Ref:        identity.NewID(),
+			Definition: dt,
+			Targets:    targets,
+			Inputs:     overrides,
+			Refs:       refs,
+		}
```
tonistiigi, line 216: Why not just write the file in here and not pass `lsg`?

To get the `groupID` you can just add `GroupRef` in the build options. The regular `Ref` can also be set in here if instead of other places.

commands/build.go
```diff
@@ -111,6 +112,7 @@ func (o *buildOptions) toControllerOptions() (*controllerapi.BuildOptions, error
 	}
 
 	opts := controllerapi.BuildOptions{
+		Ref:            identity.NewID(),
```
tonistiigi, line 115: This doesn't seem to be needed as already handled in `build/build` or `cmd/bake`

commands/bake.go
crazy-max: To have the local refs being saved and then the local group in the same place in `build.Build`. Otherwise we could potentially have dangling group refs.

tonistiigi: There is always a chance that something is dangling in this storage for panic. If you write ref first, then it can have a groupRef that does not exist; that is potentially a worse case because that could really cause a lookup error.

I'm just trying to reduce the area impacted by the PR. I think setting `GroupRef` value to build options is fine but would like to avoid passing big "group state" structs with their own lifecycle through different functions.

localstate/localstate.go
```diff
@@ -1,19 +1,43 @@
...
+	// GroupID is the ID of the state group that this ref belongs to
+	GroupID string `json:",omitempty"`
+}
+
+type StateGroup struct {
+	// Definition is the raw representation of the group (bake definition)
+	Definition []byte
+	// Targets are the targets invoked
+	Targets []string `json:",omitempty"`
+	// Inputs are the user inputs (bake overrides)
```
tonistiigi, line 37: So no variables? Ok by me to leave this for follow-up but comments seem to suggest that this was already added.

build/build.go
```diff
@@ -666,20 +669,14 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
...
-	if err := saveLocalState(so, opt, node, configDir); err != nil {
-		return nil, nil, err
-	}
-
 	return &so, releaseF, nil
 }
 
-func Build(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer) (resp map[string]*client.SolveResponse, err error) {
-	return BuildWithResultHandler(ctx, nodes, opt, docker, configDir, w, nil)
+func Build(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer, groupRef string) (resp map[string]*client.SolveResponse, err error) {
```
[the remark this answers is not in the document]
crazy-max: Yes looks better

localstate/localstate.go
crazy-max: Will look in follow-up. It's a bit tricky as the `matrix` feature hacks around variables so we can't just pick them.

tonistiigi (approved):

localstate/localstate.go
jedevc: I think if you want the values that the variables are evaluated to, you'd need to modify `hclparser.Parse` to return that information. I don't think there's a way to extract the vars/funcs defined in the HCL today :cry: 

Apologies, the code around there is a total mess - happy to help out :smile:

crazy-max: > you'd need to modify `hclparser.Parse` to return that information.

Yes indeed, I have started to look at it and how I could use context evaluation to extract vanilla variables. Will open a PR and let you know, thx!

PR #2041, opened 2023-09-12
use-test-tables for testing flags

thaJeztah:


tests/build.go
```diff
@@ -42,11 +42,7 @@ var buildTests = []func(t *testing.T, sb integration.Sandbox){
 	testBuildDetailsLink,
 	testBuildProgress,
 	testBuildAnnotations,
-	testBuildBuildArgNoKey,
-	testBuildLabelNoKey,
-	testBuildCacheExportNotSupported,
-	testBuildOCIExportNotSupported,
-	testBuildMultiPlatformNotSupported,
```
crazy-max, line 49: If we want test tables I think we should group `testBuildBuildArgNoKey`, `testBuildLabelNoKey` and name it `testBuildKeyPairNoKey` and for `testBuildCacheExportNotSupported`, `testBuildOCIExportNotSupported`, `testBuildMultiPlatformNotSupported` have another one named `testBuildDockerNotSupported`.

crazy-max: Ah missed your comment https://github.com/docker/buildx/pull/1998#discussion_r1322757861. So I think we should split them because they don't have the same input and purpose.

tonistiigi (commented):
@thaJeztah Needs a rebase.

There is a difference between these variants. Before tests run in isolated environments while in the new one they share same state. Unlikely that for this specific case this would actually cause a behavior difference but we should not go too far with this pattern.

PR #2048, opened 2023-09-14
feat: support memory/cpu driver options for docker-container

### LINKED ISSUE #2049: support memory driver opt for docker-container driver
### Description

as default, docker buildx builder is none-memory limit, which may cause server oom

whatwewant:
fixes #2049

limit builder container memory

use: `docker buildx create --driver docker-container --driver-opt memory=32G --name zmicro`

### DIFF
--- a/docs/reference/buildx_create.md
+++ b/docs/reference/buildx_create.md
@@ -150,9 +150,18 @@ No driver options.
 
 #### `docker-container` driver
 
-- `image=IMAGE` - Sets the container image to be used for running buildkit.
-- `network=NETMODE` - Sets the network mode for running the buildkit container.
-- `cgroup-parent=CGROUP` - Sets the cgroup parent of the buildkit container if docker is using the "cgroupfs" driver. Defaults to `/docker/buildx`.
+- `image=IMAGE` - Sets the BuildKit image to use for the container.
+- `memory=MEMORY` - Sets the amount of memory the container can use.
+- `memory-swap=MEMORY_SWAP` - Sets the memory swap limit for the container.
+- `cpu-quota=CPU_QUOTA` - Imposes a CPU CFS quota on the container.
+- `cpu-period=CPU_PERIOD` - Sets the CPU CFS scheduler period for the container.
+- `cpu-shares=CPU_SHARES` - Configures CPU shares (relative weight) of the container.
+- `cpuset-cpus=CPUSET_CPUS` - Limits the set of CPU cores the container can use.
+- `cpuset-mems=CPUSET_MEMS` - Limits the set of CPU memory nodes the container can use.
+- `network=NETMODE` - Sets the network mode for the container.
+- `cgroup-parent=CGROUP` - Sets the cgroup parent of the container if docker is using the "cgroupfs" driver. Defaults to `/docker/buildx`.
+
+Before you configure the resource limits for the container, read about [configuring runtime resource constraints for containers](https://docs.docker.com/config/containers/resource_constraints/).
 
 #### `kubernetes` driver
 
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -17,6 +17,7 @@ import (
 	"github.com/docker/buildx/util/confutil"
 	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/buildx/util/progress"
+	"github.com/docker/cli/opts"
 	dockertypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/mount"
@@ -40,6 +41,13 @@ type Driver struct {
 	factory      driver.Factory
 	netMode      string
 	image        string
+	memory       opts.MemBytes
+	memorySwap   opts.MemSwapBytes
+	cpuQuota     int64
+	cpuPeriod    int64
+	cpuShares    int64
+	cpusetCpus   string
+	cpusetMems   string
 	cgroupParent string
 	env          []string
 }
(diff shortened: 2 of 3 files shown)


driver/docker-container/driver.go
```diff
@@ -40,6 +41,13 @@ type Driver struct {
 	factory      driver.Factory
 	netMode      string
 	image        string
+	memory       string
+	memorySwap   string
+	cpuQuota     int64
+	cpuPeriod    int64
+	cpuShares    int64
+	cpusetCpus   string
+	cpusetMems   string
```
jsternberg, line 50: I think it might be better if we utilize the equivalent types from the `docker create` command: https://github.com/docker/cli/blob/0e70f1b7b831565336006298b9443b015c3c87a5/cli/command/container/opts.go#L34-L132

This plugin also has `docker/cli` as a dependency so we should be able to utilize the `github.com/docker/cli/opts` package which I think we should use instead so the options here are consistent with the equivalent options.

An alternative is to utilize the method that compose does to parse the options, but it's a bit more [adhoc](https://github.com/compose-spec/compose-go/blob/ba097a5dee2003644e4b172f63afbdb5a613b29c/loader/loader.go#L1326-L1337). The underlying implementation seems to be the same, but we may not want to reuse that section of code mostly because it is not easily reusable outside of a YAML context.

driver/docker-container/driver.go
```diff
@@ -126,6 +134,38 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
...
 		}
+		if d.memory != "" {
+			var memory uint64
+			memory, err := humanize.ParseBytes(d.memory)
+			if err != nil {
+				return err
+			}
+
+			hc.Resources.Memory = int64(memory)
+		}
```
jsternberg, line 145: Matching with my recommendation above, I think we can do this:

```go
if d.memory != "" {
    var memory opts.MemBytes
    if err := memory.Set(d.memory); err != nil {
        return err
    }
    hc.Resources.Memory = int64(memory)
}
```

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/docker/cli-docs-tool v0.6.0
 	github.com/docker/docker v24.0.5+incompatible
 	github.com/docker/go-units v0.5.0
+	github.com/dustin/go-humanize v1.0.0
```
jsternberg, line 19: I'm hesitant to add the dependencies that are added here since the functionality included in this PR is stuff that's also been done in other docker code. My preference is going to be that we utilize those methods rather than add a new way to parse the units.

The same applies to the other two dependencies here.

If there's a compelling reason why this is brand new functionality for docker and these dependencies make implementing that functionality easier, these dependencies seem solid. My preference is just going to be on not adding dependencies if there's an existing alternative.

driver/docker-container/factory.go
```diff
@@ -49,6 +50,20 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
 			}
 		case k == "image":
 			d.image = v
+		case k == "memory":
+			d.memory = v
+		case k == "memory-swap":
+			d.memorySwap = v
+		case k == "cpu-period":
+			d.cpuPeriod = int64(cast.ToInt(v))
```
jsternberg, line 58: I think `strconv.ParseInt` makes more sense here and then we don't need a dependency. The `container create` command uses the cli flags and that uses `strconv.ParseInt` internally.

jsternberg (changes requested):
I like the idea and I think these options could be helpful for those using the `docker-container` driver. I'd like to use a different method for parsing the options though to avoid pulling in new dependencies and to make it more consistent with the other docker products that have the same functionality.

driver/docker-container/driver.go
crazy-max: Agree, let's avoid adding unnecessary dep. `opts.MemBytes` can be used similar to https://github.com/docker/buildx/blob/e018f8b6fbf746e4ac4f019bc3918c92c67b927a/commands/build.go#L74

whatwewant: @jsternberg @crazy-max done, the types have been aligned to `github.com/docker/cli/opts`, and the unnecessary deps have been removed

driver/docker-container/driver.go
whatwewant: done

driver/docker-container/factory.go
whatwewant: @jsternberg done

jsternberg (approved):

jsternberg:
@tonistiigi tagging you since this touches user-facing configuration options so.

crazy-max (changes requested):
Needs docs update in https://github.com/docker/buildx/blob/master/docs/reference/buildx_create.md#docker-container-driver-1

And on docs website as well in follow-up https://docs.docker.com/build/drivers/docker-container/#synopsis (cc @dvdksn)

whatwewant:
> Needs docs update in https://github.com/docker/buildx/blob/master/docs/reference/buildx_create.md#docker-container-driver-1
> 
> And on docs website as well in follow-up https://docs.docker.com/build/drivers/docker-container/#synopsis (cc @dvdksn)

added @crazy-max @dvdksn

docs/reference/buildx_create.md
```diff
@@ -151,6 +151,13 @@ No driver options.
...
 - `image=IMAGE` - Sets the container image to be used for running buildkit.
+- `memory=MEMORY` - Sets the container memory limit to be used for running buildkit.
+- `memory-swap=MEMORY_SWAP` - Sets the container memory swap limit to be used for running buildkit.
+- `cpu-quota=CPU_QUOTA` - Sets the CPU CFS (Cmpletely Fair Scheduler) quota limit to be used for running buildkit.
+- `cpu-period=CPU_PERIOD` - Sets the CPU CFS (Cmpletely Fair Scheduler) period limit to be used for running buildkit.
+- `cpu-shares=CPU_SHARES` - Sets the CPU shares (relative weight) limit to be used for running buildkit.
+- `cpuset-cpus=CPUSET_CPUS` - Sets the CPU cores to be used for running buildkit, such as (0-3, 0,1).
+- `cpuset-mems=CPUSET_MEMS` - Sets the CPU memory nodes to be used for running buildkit, such as (0-3, 0,1).
 - `network=NETMODE` - Sets the network mode for running the buildkit container.
 - `cgroup-parent=CGROUP` - Sets the cgroup parent of the buildkit container if docker is using the "cgroupfs" driver. Defaults to `/docker/buildx`.
```
dvdksn, line 162: I'd suggest the following edits, and adding a link to the relevant page in the engine docs.

```suggestion
- `image=IMAGE` - Sets the BuildKit image to use for the container.
- `memory=MEMORY` - Sets the amount of memory the container can use.
- `memory-swap=MEMORY_SWAP` - Sets the memory swap limit for the container.
- `cpu-quota=CPU_QUOTA` - Imposes a CPU CFS quota on the container.
- `cpu-period=CPU_PERIOD` - Sets the CPU CFS scheduler period for the container.
- `cpu-shares=CPU_SHARES` - Configures CPU shares (relative weight) of the container.
- `cpuset-cpus=CPUSET_CPUS` - Limits the set of CPU cores the container can use.
- `cpuset-mems=CPUSET_MEMS` - Limits the set of CPU memory nodes the container can use.
- `network=NETMODE` - Sets the network mode for the container.
- `cgroup-parent=CGROUP` - Sets the cgroup parent of the container if docker is using the "cgroupfs" driver. Defaults to `/docker/buildx`.

Before you configure the resource limits for the container, read about [configuring runtime resource constraints for containers](https://docs.docker.com/config/containers/resource_constraints/).
```

dvdksn (commented):
(docs) LGTM

crazy-max (changes requested):
Can you squash your commits please? Otherwise LGTM

whatwewant:
@crazy-max something wrong ?

crazy-max (approved):

tonistiigi (approved):

PR #2071, opened 2023-10-06
(drivers): Introduce a client interface

silvin-lubecki:
This client abstracts from buildkit client. This will allow further enhancement like caching remote driver responses.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -164,8 +164,8 @@ func allIndexes(l int) []int {
 	return out
 }
 
-func ensureBooted(ctx context.Context, nodes []builder.Node, idxs []int, pw progress.Writer) ([]*client.Client, error) {
-	clients := make([]*client.Client, len(nodes))
+func ensureBooted(ctx context.Context, nodes []builder.Node, idxs []int, pw progress.Writer) ([]driver.Client, error) {
+	clients := make([]driver.Client, len(nodes))
 
 	baseCtx := ctx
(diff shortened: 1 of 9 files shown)


jedevc:
Could it be worth pulling in the interface into buildkit, something like this: https://github.com/moby/buildkit/pull/3820? It might help to avoid type drift over time? I know I closed the above since we don't want to support multiple different types of clients, but maybe this wrapping use case (for cached remote driver responses) could be the reason to want it?

silvin-lubecki:
> Could it be worth pulling in the interface into buildkit
I don't know about this, it doesn't fit with golang idioms about interfaces. API Consumers should own their own set of interfaces, which are most of the time a subset of the external dependency type.

tonistiigi:
I think `Client()` should always return a buildkit client. If we need any custom behavior in the implementation of specific drivers, we can add more methods to the `Driver` interface. Making these behaviors as a side effect of some buildkit client functions makes it so that buildx does not know anymore when the connection to BuildKit is actually established which can lead to unexpected pauses in the UX, differences in how drivers behave, or buildx needing to know that in some cases "buildkit client" (that is not real buildkit client anymore) behaves differently than expected.

silvin-lubecki:
Sorry didn't mean to merge this one ...

silvin-lubecki:
https://github.com/docker/buildx/pull/2072 to undo it, sorry 🙏

PR #2076, opened 2023-10-14
bake: display read definition files in build output

crazy-max:
If no bake definition are specified, it will currently consume the default ones in this specified order: https://github.com/docker/buildx/blob/05b88216251d20597e6a457f0d454dbbeb8d03eb/bake/bake.go#L47-L53

Following-up a discussion with @milas:

> If you have both `docker-bake.hcl` AND `docker-compose.yml` in a directory, Bake always parses the Compose file unless you pass `-f docker-bake.hcl` explicitly
> So if you have a syntax error in your compose file:
> ```
> ❯ docker buildx bake services
> [+] Building 0.0s (0/0)                                                                                                                  cloud:cloud-stage
> ERROR: parsing : yaml: line 3: mapping values are not allowed in this context
> ```
> But:
> ```
> ❯ docker buildx bake -f docker-bake.hcl services
> [+] Building 0.5s (10/10) FINISHED                                                                                                       cloud:cloud-stage
>  => [registry internal] connected to docker's cloud service                                                                                           0.0s
>  => => note: cloud builds are currently in beta                                                                                                       0.0s
> ```
> I actually noticed this because I was getting this in my bake:
> ```
> WARNING: The "BUILD_SERVICE_CLIENT_MTLS_CERT" variable is not set. Defaulting to a blank string.
> ```
> and was SO confused because that's not part of the `Dockerfile` or Bake file and I thought it was actually coming FROM the builder but that's a variable in the Compose file (that it wasn't even using) and that's `compose-go` writing out that message to stdout

So there is an issue with `compose-go` writing on stdout warnings: https://github.com/compose-spec/compose-go/blob/2d32c3f774bda0ad20e8d06eae493047b728fc54/template/template.go#L206. Didn't find an easy way to disable logging via https://github.com/compose-spec/compose-go/blob/2d32c3f774bda0ad20e8d06eae493047b728fc54/template/template.go#L114-L116 without setting `options.Interpolate.Substitute` and redefining templating on our side. @milas @glours Is it possible to have smth easier to consume through `loader.Options` instead like `options.Logging = false`?

It also appears in his case that consuming a `compose.yml` and `docker-bake.hcl` when invoking `docker buildx bake` is unexpected/confusing. Maybe we could put some effort and clearly state this behavior in our docs https://docs.docker.com/build/bake/reference/#file-format (cc @dvdksn).

On the other hand it's pretty common to have a compose file just for runtime and a bake hcl file to build on a repository so I tend to agree with him.

This PR will show what bake definitions are consumed in the build output to mitigate this issue and help user figures out what is the source of the definition:

```
$ docker buildx bake binaries
[+] Building 2.1s (10/10) FINISHED                                                                                                                                                                             docker:default
 => [internal] load local bake definitions                                                                                                                                                                               0.0s 
 => => reading compose.yaml 954B / 954B                                                                                                                                                                                  0.0s
 => => reading docker-bake.hcl 3.68kB / 3.68kB                                                                                                                                                                           0.0s 
 => [internal] load .dockerignore                                                                                                                                                                                        0.1s
 => => transferring context: 45B                                                                                                                                                                                         0.0s 
 => [internal] load build definition from Dockerfile                                                                                                                                                                     0.1s 
 => => transferring dockerfile: 4.66kB                                                                                                                                                                                   0.0s 
 => resolve image config for docker.io/docker/dockerfile:1
...
```

```
$ docker buildx bake binaries

#1 reading compose.yaml 954B / 954B done
#1 reading docker-bake.hcl 3.68kB / 3.68kB done
#1 DONE 0.0s

#2 transferring dockerfile: 4.66kB 0.0s done
#2 DONE 0.1s

#3 [internal] load .dockerignore
#3 transferring context: 45B 0.0s done
#3 DONE 0.1s

#4 resolve image config for docker.io/docker/dockerfile:1
#4 DONE 0.5s
...
```

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	composecli "github.com/compose-spec/compose-go/cli"
 	"github.com/docker/buildx/bake/hclparser"
@@ -18,8 +19,10 @@ import (
 	controllerapi "github.com/docker/buildx/controller/pb"
 	"github.com/docker/buildx/util/buildflags"
 	"github.com/docker/buildx/util/platformutil"
+	"github.com/docker/buildx/util/progress"
 	"github.com/docker/cli/cli/config"
 	hcl "github.com/hashicorp/hcl/v2"
+	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/pkg/errors"
@@ -54,7 +57,7 @@ func defaultFilenames() []string {
 	return names
 }
 
-func ReadLocalFiles(names []string, stdin io.Reader) ([]File, error) {
+func ReadLocalFiles(names []string, stdin io.Reader, l progress.SubLogger) ([]File, error) {
 	isDefault := false
 	if len(names) == 0 {
 		isDefault = true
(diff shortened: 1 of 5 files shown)


crazy-max:
Hum there is something to consider when we use `--print` flag. With these changes it would do:

```
$ docker buildx bake --print
#1 reading docker-bake.hcl 3.68kB / 3.68kB done
#1 DONE 0.0s
{
  "group": {
    "default": {
      "targets": [
        "binaries"
      ]
    }
  },
  "target": {
    "binaries": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
      },
      "target": "binaries",
      "platforms": [
        "local"
      ],
      "output": [
        "./bin/build"
      ]
    }
  }
}
```

Which could break some workflows using the canonical form of the definition. They could discard stderr like:

```
docker buildx bake 2>/dev/null --print
{
  "group": {
    "default": {
      "targets": [
        "binaries"
      ]
    }
  },
  "target": {
    "binaries": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
      },
      "target": "binaries",
      "platforms": [
        "local"
      ],
      "output": [
        "./bin/build"
      ]
    }
  }
}
```

But that would also discard any error output when loading the definition.

Same currently happens when loading from a remote definition:

```
docker buildx bake https://github.com/docker/buildx.git --print

#1 [internal] load remote bake definitions
#1 reading docker-bake.hcl done
#1 DONE 1.0s

#2 [internal] load git source https://github.com/docker/buildx.git
#2 0.553 ref: refs/heads/master HEAD
#2 0.553 05b88216251d20597e6a457f0d454dbbeb8d03eb       HEAD
#2 1.014 05b88216251d20597e6a457f0d454dbbeb8d03eb       refs/heads/master
#2 CACHED
{
  "group": {
    "default": {
      "targets": [
        "binaries"
      ]
    }
  },
  "target": {
    "binaries": {
      "dockerfile": "Dockerfile",
      "args": {
      },
      "target": "binaries",
      "platforms": [
        "local"
      ],
      "output": [
        "./bin/build"
      ]
    }
  }
}
```

Maybe we could discard progress output when `--print` is used?

tonistiigi (approved):

PR #2084, opened 2023-10-17
update to go 1.21.3

crazy-max:
https://tip.golang.org/doc/go1.21

Updates golangci-lint to 1.54.2 and its configuration to show all lint issues and align exclude rules with buildkit ones as well: https://github.com/golangci/golangci-lint/releases/tag/v1.54.2

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -26,12 +26,13 @@ linters:
 
 linters-settings:
   depguard:
-    list-type: blacklist
-    include-go-root: true
-    packages:
-      # The io/ioutil package has been deprecated.
-      # https://go.dev/doc/go1.16#ioutil
-      - io/ioutil
+    rules:
+      main:
+        deny:
+          # The io/ioutil package has been deprecated.
+          # https://go.dev/doc/go1.16#ioutil
+          - pkg: "io/ioutil"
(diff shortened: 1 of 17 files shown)


bake/bake.go
```diff
@@ -99,7 +99,7 @@ func ListTargets(files []File) ([]string, error) {
 	return dedupSlice(targets), nil
 }
 
-func ReadTargets(ctx context.Context, files []File, targets, overrides []string, defaults map[string]string) (map[string]*Target, map[string]*Group, error) {
+func ReadTargets(_ context.Context, files []File, targets, overrides []string, defaults map[string]string) (map[string]*Target, map[string]*Group, error) {
```
ktock, line 102: Looks like `revive` linter error? Recent revive seems too strict so maybe we can have some `exclude-rules` like this: https://github.com/moby/buildkit/blob/3dedf3c5e72c55e2a61cd98e67e0f461d545b195/.golangci.yml#L61

crazy-max: Ah gotcha thanks. I was wondering why we had so many issues. Let me check.

tonistiigi (approved):

ktock (approved):

PR #2086, opened 2023-10-18
debug: Following-up patches for error messages

ktock:
- 38137b29dd03335611fe190b4ed467e51804bd55 : fixes the following mentioned in https://github.com/docker/buildx/pull/2006#pullrequestreview-1681377351
  - > * There is an error in Dockerfile command (eg. change `RUN` to `RUN2`). I get an error. Now I run `reload`. It builds up to the Dockerfile error again but this time I don't see any error.
- b06a55cf53b0c5b34d1e1bffa058394cd673ce2d : following-up for https://github.com/docker/buildx/pull/2006/commits/0dd89f602922851a3159976972832728aafa0bd3
- 267e30a19cd5972187a6553c00e0442e8f45d150 : fixes the following mentioned in https://github.com/docker/buildx/pull/2006#pullrequestreview-1681377351
  - > * I run `debug build .` and get an error. I fix the error and issue `reload`. Now my build succeeds. I run some more interactive processes and eventually close the debugger. Now I get an error.
- 14834e6085e5227d1baface0aa057cccb34ffc29 : following-up for https://github.com/docker/buildx/pull/2006/commits/6db8569f096a89a0a2e1f637c4cd990b99c7b832

### DIFF
--- a/monitor/monitor.go
+++ b/monitor/monitor.go
@@ -302,7 +302,11 @@ func (m *monitor) startInvoke(ctx context.Context, pid string, cfg controllerapi
 	go func() {
 		// Start a new invoke
 		if err := m.invoke(ctx, pid, cfg); err != nil {
-			logrus.Debugf("invoke error: %v", err)
+			if errors.Is(err, context.Canceled) {
+				logrus.Debugf("process canceled: %v", err)
+			} else {
+				logrus.Errorf("invoke: %v", err)
+			}
 		}
 		if pid == m.attachedPid.Load() {
 			m.attachedPid.Store("")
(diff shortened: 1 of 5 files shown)


monitor/monitor.go
```diff
@@ -233,6 +239,14 @@ type monitor struct {
 	invokeIO     *ioset.Forwarder
 	invokeCancel func()
 	attachedPid  atomic.Value
+
+	lastBuildResult *LastBuildResult
```
tonistiigi, line 243: variable name `lastBuildResult` is ok but the type returned from `RunMonitor` should be different. Maybe `MonitorBuildResult`?

ktock: Fixed the type name to `MonitorBuildResult`.

tonistiigi (approved):

PR #2093, opened 2023-10-20
driver: docker-container driver uses --config correctly in rootless mode

### LINKED ISSUE #2092: Buildkitd config file is missplaced when running in dind rootless with docker-container driver
### Contributing guidelines

### Description

When using `buildx` in a `dind` rootless container with the `docker-container` driver, if you use the `--config` option of the `docker buildx create` command, the config file will be placed at `/etc/buildkit/buildkitd.toml`, but when `buildkitd` is running in a user namespace it is using `~/.config/buildkit/buildkitd.toml` as the default path.

### Expected behaviour

When using the `--config` file, the config file should be used by `buildkitd`. This could be done by placing the file at the right place, or passing the `--config` option to `buildkitd` to no rely on default values.

### Actual behaviour

The specified config file is missplaced and ignored by `buildkitd`

### Buildx version

github.com/docker/buildx v0.11.2 9872040b6626fb7d87ef7296fd5b832e8cc2ad17

### Docker info

```text
Client:
 Version:    24.0.6
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.11.2
    Path:     /usr/local/libexec/docker/cli-plugins/docker-buildx
    Version:  v2.22.0
    Path:     /usr/local/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 1
  Running: 1
  Paused: 0
  Stopped: 0
 Images: 1
 Server Version: 24.0.6
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: false
  userxattr: true
 Logging Driver: json-file
 Cgroup Driver: none
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7880925980b188f4c97b462f709d0db8e8962aff
 runc version: v1.1.9-0-gccaecfc
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  rootless
  cgroupns
 Kernel Version: 5.15.0-86-generic
 Operating System: Alpine Linux v3.18
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 30.96GiB
 Name: b2f5d95bdcd1
 ID: 12e55383-5ec7-453b-b0ba-101361240373
 Docker Root Dir: /home/rootless/.local/share/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Product License: Community Engine

WARNING: Running in rootless-mode without cgroups. Systemd is required to enable cgroups in rootless-mode.
```

### Builders list

```text
NAME/NODE DRIVER/ENDPOINT                   STATUS  BUILDKIT             PLATFORMS
test *    docker-container                                               
  test0   unix:///run/user/1000/docker.sock running v0.12.2              linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/386
default   docker                                                         
  default default                           running v0.11.6+616c3f613b54 linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/386
```

### Configuration

# How to reproduce
```
$ docker run --rm --name dind-rootless --privileged -d -e DOCKER_HOST=unix:///run/user/1000/docker.sock docker:dind-rootless
$ docker exec -it dind-rootless /bin/sh
$ echo "debug = true" > /tmp/config.toml
$ docker buildx create --name test --driver docker-container --use --config /tmp/config.toml
$ docker buildx inspect --bootstrap --builder test
$ docker logs buildx_buildkit_test0
```
We should see `debug` logs, but instead we only have `info` logs:
```
time="2023-10-20T12:09:48Z" level=info msg="auto snapshotter: using overlayfs"
time="2023-10-20T12:09:48Z" level=info msg="found worker \"rnogk4pc37pbbnz7stsvxgc2o\", labels=map[org.mobyproject.buildkit.worker.executor:oci org.mobyproject.buildkit.worker.hostname:ffeee805bf75 org.mobyproject.buildkit.worker.network:host org.mobyproject.buildkit.worker.oci.process-mode:sandbox org.mobyproject.buildkit.worker.selinux.enabled:false org.mobyproject.buildkit.worker.snapshotter:overlayfs], platforms=[linux/amd64 linux/amd64/v2 linux/amd64/v3 linux/386]"
time="2023-10-20T12:09:48Z" level=warning msg="skipping containerd worker, as \"/run/containerd/containerd.sock\" does not exist"
time="2023-10-20T12:09:48Z" level=info msg="found 1 workers, default=\"rnogk4pc37pbbnz7stsvxgc2o\""
time="2023-10-20T12:09:48Z" level=warning msg="currently, only the default worker can be used."
time="2023-10-20T12:09:48Z" level=info msg="running server on /run/buildkit/buildkitd.sock"
```

# Workaround
We can manually set the `--config` flag using `--buildkitd-flags`:
```
$ docker run --rm --name dind-rootless --privileged -d -e DOCKER_HOST=unix:///run/user/1000/docker.sock docker:dind-rootless
$ docker exec -it dind-rootless /bin/sh
$ echo "debug = true" > /tmp/config.toml
$ docker buildx create --name test --driver docker-container --use --config /tmp/config.toml --buildkitd-flags "--config /etc/buildkit/buildkitd.toml"
$ docker buildx inspect --bootstrap --builder test
$ docker logs buildx_buildkit_test0
```
We now have `debug` logs:
```
time="2023-10-20T12:10:34Z" level=info msg="auto snapshotter: using overlayfs"
time="2023-10-20T12:10:34Z" level=debug msg="running in rootless mode"
time="2023-10-20T12:10:34Z" level=info msg="found worker \"yama36v8rvc9ivwnsestxtnid\", labels=map[org.mobyproject.buildkit.worker.executor:oci org.mobyproject.buildkit.worker.hostname:7881508dd32c org.mobyproject.buildkit.worker.network:host org.mobyproject.buildkit.worker.oci.process-mode:sandbox org.mobyproject.buildkit.worker.selinux.enabled:false org.mobyproject.buildkit.worker.snapshotter:overlayfs], platforms=[linux/amd64 linux/amd64/v2 linux/amd64/v3 linux/386]"
time="2023-10-20T12:10:34Z" level=warning msg="skipping containerd worker, as \"/run/containerd/containerd.sock\" does not exist"
time="2023-10-20T12:10:34Z" level=info msg="found 1 workers, default=\"yama36v8rvc9ivwnsestxtnid\""
time="2023-10-20T12:10:34Z" level=warning msg="currently, only the default worker can be used."
time="2023-10-20T12:10:34Z" level=info msg="running server on /run/buildkit/buildkitd.sock"
time="2023-10-20T12:10:35Z" level=debug msg="remove snapshot" key=x4h6gtzy157daylfs89g1dbme snapshotter=overlayfs
time="2023-10-20T12:10:35Z" level=debug msg="schedule snapshotter cleanup" snapshotter=overlayfs
time="2023-10-20T12:10:35Z" level=debug msg="removed snapshot" key=buildkit/1/x4h6gtzy157daylfs89g1dbme snapshotter=overlayfs
time="2023-10-20T12:10:35Z" level=debug msg="snapshot garbage collected" d=7.11123ms snapshotter=overlayfs
```

### Build logs

_No response_

### Additional info

I think the root cause of this is how `buildkitd` is setting the default value for the `--config` flag:
- It defines the `--config` flag here with a default value set using the function `defaultConfigPath`: https://github.com/moby/buildkit/blob/master/cmd/buildkitd/main.go#L153-L156
- The `defaultConfigPath` is checking if `buildkitd` is running in a user namespace, and adapt the path depending on this: https://github.com/moby/buildkit/blob/3f1e44705277768edef03ff5f734d36e3c45979f/cmd/buildkitd/main.go#L414-L419
- This is causing to set `/root/.config/buildkit/buildkitd.toml` as the default value of the `--config` file
- It seems like `buildkitd` is checking later if it is running as `root` in a user namespace: https://github.com/moby/buildkit/blob/3f1e44705277768edef03ff5f734d36e3c45979f/cmd/buildkitd/main.go#L469-L473
  Should it also check this in the `defaultConfigPath` function? Should `docker buildx create` set the `--config` flag?

jsternberg:
The `docker-container` driver relies on the default config file location
for buildkit when writing the configuration file. When run in a rootless
version of docker (dind), the default location is different.

Instead of trying to figure out where the appropriate default location
is, this just writes the files to the same location and sets the
`--config` parameter explicitly. This flag is placed first so a
user-specified config option in `--buildkitd-flags` will take precedence
over the implicit config parameter.

This also fixes the `--config` option with the rootless image.
Previously, the config directory was being copied in a way that rendered
`/etc` unreadable and the configuration file wasn't readable either. It
also wasn't copied to the correct place. Now, `--config` is used to
specify the directory, `/etc` isn't included in the copied archive (so
the permissions aren't overwritten), and the directory is set as world
readable to be readable from the rootless buildkit process`.

Fixes #2092.

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -33,7 +33,8 @@ import (
 )
 
 const (
-	volumeStateSuffix = "_state"
+	volumeStateSuffix   = "_state"
+	buildkitdConfigFile = "buildkitd.toml"
 )
 
 type Driver struct {
@@ -114,9 +115,7 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 		Image: imageName,
 		Env:   d.env,
 	}
-	if d.InitConfig.BuildkitFlags != nil {
-		cfg.Cmd = d.InitConfig.BuildkitFlags
-	}
+	cfg.Cmd = getBuildkitFlags(d.InitConfig)
 
 	useInit := true // let it cleanup exited processes created by BuildKit's container API
 	if err := l.Wrap("creating container "+d.Name, func() error {
@@ -259,7 +258,9 @@ func (d *Driver) copyToContainer(ctx context.Context, files map[string][]byte) e
 		return err
 	}
 	defer srcArchive.Close()
-	return d.DockerAPI.CopyToContainer(ctx, d.Name, "/", srcArchive, dockertypes.CopyToContainerOptions{})
+
+	baseDir := path.Dir(confutil.DefaultBuildKitConfigDir)
+	return d.DockerAPI.CopyToContainer(ctx, d.Name, baseDir, srcArchive, dockertypes.CopyToContainerOptions{})
 }
 
 func (d *Driver) exec(ctx context.Context, cmd []string) (string, net.Conn, error) {
@@ -475,15 +476,34 @@ func writeConfigFiles(m map[string][]byte) (_ string, err error) {
 			os.RemoveAll(tmpDir)
 		}
 	}()
+	configDir := filepath.Base(confutil.DefaultBuildKitConfigDir)
 	for f, dt := range m {
-		f = path.Join(confutil.DefaultBuildKitConfigDir, f)
-		p := filepath.Join(tmpDir, f)
-		if err := os.MkdirAll(filepath.Dir(p), 0700); err != nil {
+		p := filepath.Join(tmpDir, configDir, f)
+		if err := os.MkdirAll(filepath.Dir(p), 0755); err != nil {
 			return "", err
 		}
-		if err := os.WriteFile(p, dt, 0600); err != nil {
+		if err := os.WriteFile(p, dt, 0644); err != nil {
 			return "", err
 		}
 	}
 	return tmpDir, nil
+}
+
+func getBuildkitFlags(initConfig driver.InitConfig) []string {
+	flags := initConfig.BuildkitFlags
+	if _, ok := initConfig.Files[buildkitdConfigFile]; ok {
+		// There's no way for us to determine the appropriate default configuration
+		// path and the default path can vary depending on if the image is normal
+		// or rootless.
+		//
+		// In order to ensure that --config works, copy to a specific path and
+		// specify the location.
+		//
+		// This should be appended before the user-specified arguments
+		// so that this option could be overwritten by the user.
+		newFlags := make([]string, 0, len(flags)+2)
+		newFlags = append(newFlags, "--config", path.Join("/etc/buildkit", buildkitdConfigFile))
+		flags = append(newFlags, flags...)
+	}
+	return flags
 }


tonistiigi:
@AkihiroSuda Looking at the issue, isn't the problem in incorrectly detecting rootless paths used by buildkit. There is no `--driver-opt` in that reproducer that would set a rootless image. Additionally, with this approach, could there be permission issues in accessing `/etc`.

jjacobelli:
> @AkihiroSuda Looking at the issue, isn't the problem in incorrectly detecting rootless paths used by buildkit. There is no `--driver-opt` in that reproducer that would set a rootless buildkit image. Additionally, with this approach, could there be permission issues in accessing `/etc`.

I would agree that maybe `buildkitd` should apply [this](https://github.com/moby/buildkit/blob/3f1e44705277768edef03ff5f734d36e3c45979f/cmd/buildkitd/main.go#L469-L473) logic when it's computing the default value for the `--config` flag instead of just checking if it's running in a user namespace: https://github.com/moby/buildkit/blob/762b2626ddafbc0874182b477f0393e8c5c3cb63/cmd/buildkitd/main.go#L414-L419

But maybe one solution could be to use a path that is writable by any user (like `/tmp/buildkitd.toml`?) and set the `--config` flag instead of relying on default values. WDYT?

crazy-max:
Yes it seems builditd configuration is saved to the default location within the container when the builder is created:

https://github.com/docker/buildx/blob/7838ade9f369b164a405628a37e8dedbe91bdd10/driver/docker-container/driver.go#L470

https://github.com/docker/buildx/blob/7838ade9f369b164a405628a37e8dedbe91bdd10/util/confutil/container.go#L20

jsternberg:
Two comments.

> I would agree that maybe buildkitd should apply [this](https://github.com/moby/buildkit/blob/3f1e44705277768edef03ff5f734d36e3c45979f/cmd/buildkitd/main.go#L469-L473) logic when it's computing the default value for the --config flag instead of just checking if it's running in a user namespace: https://github.com/moby/buildkit/blob/762b2626ddafbc0874182b477f0393e8c5c3cb63/cmd/buildkitd/main.go#L414-L419

I like this. I took a simple look and it looks like `setDefaultConfig` is used after the config file is loaded. It seems reasonable to move that check to its own function and then use it for the config file loading too.

> But maybe one solution could be to use a path that is writable by any user (like /tmp/buildkitd.toml?) and set the --config flag instead of relying on default values. WDYT?

Even with the above, I still think this is probably the best way to go. Explicitly setting `--config` is likely better and, if it's specified at the beginning, a user could overwrite it if they really wanted to. We likely don't want to rely on default behavior if we're writing a configuration file.

I am a bit confused by the `/tmp` folder requirement. My understanding is that the config file would not have to be writeable within the container and `ContainerCopy` likely worked differently for copying the contents. I can have the config file at `/tmp` easily, but I expected that the default location would continue working in rootless. I'll need to try it myself since I admit I didn't test it. So this is a good callout just in case it matters.

jjacobelli:
> I am a bit confused by the `/tmp` folder requirement. My understanding is that the config file would not have to be writeable within the container and `ContainerCopy` likely worked differently for copying the contents. I can have the config file at `/tmp` easily, but I expected that the default location would continue working in rootless. I'll need to try it myself since I admit I didn't test it. So this is a good callout just in case it matters.

I agree that having the file under `/etc` should be fine. I don't think this file needs to be writeable

jsternberg:
I tested this and it works. Here's the output before the change:

```
5b8eb54588ab:/$ docker logs buildx_buildkit_a0
time="2023-11-07T20:59:27Z" level=info msg="auto snapshotter: using overlayfs"
time="2023-11-07T20:59:27Z" level=warning msg="failed to prepare cgroup controllers: mkdir /sys/fs/cgroup/init: permission denied"                                    time="2023-11-07T20:59:27Z" level=info msg="found worker \"46hll4gwasmbzwudaryggjls1\", labels=map[org.mobyproject.buildkit.worker.executor:oci org.mobyproject.buildkit.worker.hostname:8fd426fdd760 org.mobyproject.buildkit.worker.network:host org.mobyproject.buildkit.worker.oci.process-mode:sandbox org.mobyproject.buildkit.worker.selinux.enabled:false org.mobyproject.buildkit.worker.snapshotter:overlayfs], platforms=[linux/arm64 linux/amd64 linux/amd64/v2 linux/riscv64 linux/ppc64le linux/s390
x linux/386 linux/mips64le linux/mips64 linux/arm/v7 linux/arm/v6]"
time="2023-11-07T20:59:27Z" level=warning msg="skipping containerd worker, as \"/run/containerd/containerd.sock\" does not exist"
time="2023-11-07T20:59:27Z" level=info msg="found 1 workers, default=\"46hll4gwasmbzwudaryggjls1\""
time="2023-11-07T20:59:27Z" level=warning msg="currently, only the default worker can be used."
time="2023-11-07T20:59:27Z" level=info msg="running server on /run/buildkit/buildkitd.sock"
```

Here's the logs after the change:

```
5b8eb54588ab:~$ docker logs buildx_buildkit_a0
time="2023-11-07T21:03:36Z" level=info msg="auto snapshotter: using overlayfs"
time="2023-11-07T21:03:36Z" level=debug msg="running in rootless mode"
time="2023-11-07T21:03:36Z" level=warning msg="failed to prepare cgroup controllers: mkdir /sys/fs/cgroup/init: permission denied"
time="2023-11-07T21:03:36Z" level=info msg="found worker \"g361kgmo95xdbtgnf7rjk6lr0\", labels=map[org.mobyproject.buildkit.worker.executor:oci org.mobyproject.buildkit.worker.hostname:a44636f99be4 org.mobyproject.buildkit.worker.network:host org.mobyproject.buildkit.worker.oci.process-mode:sandbox org.mobyproject.buildkit.worker.selinux.enabled:false org.mobyproject.buildkit.worker.snapshotter:overlayfs], platforms=[linux/arm64 linux/amd64 linux/amd64/v2 linux/riscv64 linux/ppc64le linux/s390x linux/386 linux/mips64le linux/mips64 linux/arm/v7 linux/arm/v6]"
time="2023-11-07T21:03:36Z" level=warning msg="skipping containerd worker, as \"/run/containerd/containerd.sock\" does not exist"
time="2023-11-07T21:03:36Z" level=info msg="found 1 workers, default=\"g361kgmo95xdbtgnf7rjk6lr0\""
time="2023-11-07T21:03:36Z" level=warning msg="currently, only the default worker can be used."
time="2023-11-07T21:03:36Z" level=info msg="running server on /run/buildkit/buildkitd.sock"
time="2023-11-07T21:03:37Z" level=debug msg="remove snapshot" key=kiq5o65de3cbqjjjmr62yti4p snapshotter=overlayfs
time="2023-11-07T21:03:37Z" level=debug msg="schedule snapshotter cleanup" snapshotter=overlayfs
time="2023-11-07T21:03:37Z" level=debug msg="removed snapshot" key=buildkit/1/kiq5o65de3cbqjjjmr62yti4p snapshotter=overlayfs
time="2023-11-07T21:03:37Z" level=debug msg="snapshot garbage collected" d=6.464584ms snapshotter=overlayfs
```

jsternberg:
I made a modification to this so now the `--config` option works properly with the rootless image.

Previously, the config directory was being copied with permissions of `0700` which meant the directory wasn't readable. It also wasn't being copied to the correct place so it wasn't being used anyway. Since the directory itself contained `/etc` and that was copied, it also overwrote the permissions from the original image and made the contents of `/etc` unreadable.

Now, it does not overwrite `/etc`, it uses `--config` to specify the directory in the command so the configuration is used, and the directory is now world readable so the rootless process can actually read the contents.

tonistiigi (approved):

tonistiigi (approved):

PR #2098, opened 2023-10-26
build: make annotations work with push flag

crazy-max:
related to https://github.com/docker/buildx/pull/2020

When implementing annotations in our build-push-action (https://github.com/docker/build-push-action/pull/992), I found out they were not working: https://github.com/docker/build-push-action/actions/runs/6642453793/job/18047273130#step:7:24

Annotations work with `--output` but it doesn't when using `--push`. This moves the parsing annotations logic to the controller so we can set annotations after `--push` has been converted to an export entry.


commands/build.go
```diff
@@ -172,16 +173,6 @@ func (o *buildOptions) toControllerOptions() (*controllerapi.BuildOptions, error
...
-	annotations, err := buildflags.ParseAnnotations(o.annotations)
-	if err != nil {
-		return nil, err
-	}
-	for _, e := range opts.Exports {
-		for k, v := range annotations {
-			e.Attrs[k.String()] = v
-		}
-	}
-
```
jsternberg, line 184: So here `--push` hasn't been translated to an export entry. Can you just link me the line where this happens for learning purposes?

controller/build/build.go
```diff
@@ -132,6 +132,17 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in controllerapi.Build
...
+
+	annotations, err := buildflags.ParseAnnotations(in.Annotations)
+	if err != nil {
+		return nil, nil, err
+	}
+	for _, o := range outputs {
+		for k, v := range annotations {
+			o.Attrs[k.String()] = v
+		}
+	}
```
jsternberg, line 144: It's delayed here (so `--push` is a true alias for `--output=type=registry`) and that's why this works.

jsternberg (approved):
Seems fine. I'd just like to see the line where `--push` get translated so I can better understand how things fit together.

favonia:
I have a very naive question: is there a reason not to add the field `Annotations` to `BuildOptions` so that `annotation`s and `output` become "independent"?

PR #2100, opened 2023-10-28
bake: improve error message for empty definition

### LINKED ISSUE #2095: Bake should check for empty definition
Running bake without any definition in working dir:

```
$ docker buildx bake
ERROR: failed to find target default
```

```
$ docker buildx bake --set *.cache-from=type=gha
ERROR: could not find any target matching '*'
```

We should instead say that no definition has been provided/found.

dvdksn:
Closes #2095

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -159,6 +159,10 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
 		return err
 	}
 
+	if len(files) == 0 {
+		return errors.New("couldn't find a bake definition")
+	}
+
 	tgts, grps, err := bake.ReadTargets(ctx, files, targets, overrides, map[string]string{
 		// don't forget to update documentation if you add a new
 		// built-in variable: docs/bake-reference.md#built-in-variables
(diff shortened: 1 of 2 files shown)


commands/bake.go
```diff
@@ -159,6 +159,10 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
 		return err
 	}
 
+	if len(files) < 1 {
```
crazy-max, line 162:
```suggestion
	if len(files) == 0 {
```

commands/bake.go
```diff
@@ -159,6 +159,10 @@ func runBake(dockerCli command.Cli, targets []string, in bakeOptions, cFlags com
...
+		return errors.Errorf("couldn't find a bake definition")
```
crazy-max, line 163:
```suggestion
		return errors.New("couldn't find a bake definition")
```

tests/bake.go
```diff
@@ -426,3 +427,9 @@ COPY foo /foo
 	require.Error(t, err, out)
 	require.Contains(t, out, "reading a dockerfile for a remote build invocation is currently not supported")
 }
+
+func testBakeEmpty(t *testing.T, sb integration.Sandbox) {
+	out, err := bakeCmd(sb)
+	require.Contains(t, out, "couldn't find a bake definition")
+	require.Error(t, err, out)
```
crazy-max, line 434: nit: I would swap these assertions

```suggestion
	require.Error(t, err, out)
	require.Contains(t, out, "couldn't find a bake definition")
```

crazy-max (approved):

Issue #2104, opened 2023-11-02
bake: Resource interpolation on matrix target


jkroepke:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

I'm trying to use [Resource interpolation](https://docs.docker.com/build/bake/configuring-build/#resource-interpolation) on my bake file together with an matrix build.

### Expected behaviour

An valid bake output:

```json
{
  "group": {
    "default": {
      "targets": [
        "windows"
      ]
    },
    "windows": {
      "targets": [
        "windows-ltsc2019",
        "windows-ltsc2022"
      ]
    }
  },
  "target": {
    "windows-ltsc2019": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "NANOSERVER_VERSION": "ltsc2019"
      },
      "tags": [
        "goldpinger:1.0.0",
        "goldpinger:latest"
      ],
      "target": "windows",
      "platforms": [
        "windows/amd64"
      ]
    },
    "windows-ltsc2022": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "NANOSERVER_VERSION": "ltsc2022"
      },
      "tags": [
        "goldpinger:1.0.0",
        "goldpinger:latest"
      ],
      "target": "windows",
      "platforms": [
        "windows/amd64"
      ]
    }
  }
}

```

### Actual behaviour

Getting an error

```
docker-bake.hcl:15
--------------------
  13 |     
  14 |     target "windows" {
  15 | >>>   tags = target.docker-metadata-action.tags
  16 |     
  17 |       name = "windows-${version}"
--------------------
ERROR: docker-bake.hcl:15,16-39: Missing map element; This map does not have an element with the key "docker-metadata-action"., and 1 other diagnostic(s)

```

### Buildx version

github.com/docker/buildx v0.11.2-desktop.5 f20ec1393426619870066baba9618cf999063886

### Docker info

```text
Client:
 Version:    24.0.6
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2-desktop.5
    Path:     /Users/jok/.docker/cli-plugins/docker-buildx
[...]

PR #2106, opened 2023-11-02
bake: fix global target access when using a matrix

jedevc:
Fixes https://github.com/docker/buildx/issues/2104

Previously, we would fail while trying to use the global "target" field when using a matrix. The contents of the matrix really don't matter for this.

What was happening was that we would copy the "target" property into the child evaluation context, so that when it was updated on the parent, it wouldn't propagate to the child. The correct behavior here is to avoid copying variables from the target evaluation context if it is the root.

Pretty simple fix (a bit of a trickier investigation), but also added a test to make sure we don't hit this again. Honestly, a bit surprised this never showed up before :scream:


tonistiigi (approved):

crazy-max (approved):
LGTM thanks!

PR #2108, opened 2023-11-07
build: wait from child targets to complete before session release

tonistiigi:
When targets in bake are chained with named build contexts we need to wait for the second target to complete the solve before the session from first build can be released. Otherwise, the first build could already release the definition from the buildkit graph before second build starts. Normally this would not be a problem because the steps from first build should get a direct cache match but there are some edge cases with local sources where another build can take ownership of the shared destination directory and then for specific graph layout BuildKit solver would find that it can possibly match cache if it would recompute content checksums for the local source, causing it to be repulled again via session.

Because `Evaluate()` is called manually in this patch it might cause a small performance regression in some edge cases because all targets that are used in build contexts need to wait for the child `Solve()` requests to return before they can export their outputs. Additionally, there is an unlikely case that the exporter like registry does not need to pull the layer at all if `Evaluate` would not be called. To fix this, buildkit API side change would be needed that separates calling exporter and releasing from build graph to separate steps.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -814,6 +814,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 	results := waitmap.New()
 
 	multiTarget := len(opt) > 1
+	childTargets := calculateChildTargets(m, opt)
 
 	for k, opt := range opt {
 		err := func(k string) error {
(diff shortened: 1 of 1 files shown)


thaJeztah:
Should the description be in the commit message as well (for future reference?)

crazy-max (approved):
:tada:

The small performance penalty to solve this issue is still worth it.

tonistiigi:
> Could we have an integration test for this case?

The case that would actually trigger a failed build is way too complicated. I haven't been able to trigger this even manually with a made-up bake file (but we have a report with a complicated bake file that did manage to do it).

PR #2112, opened 2023-11-14
Add dial-stdio command

cpuguy83:
This allows the buildx CLI to act a proxy to the configured instance. It allows external code to use buildx itself as a driver for connecting to buildkitd instances.

Instance and node selection should follow the same semantics as as `buildx build`, including taking into account the `BUILDX_BUILDER` env var and the `--builder` global flag.

### DIFF
--- a/build/dial.go
+++ b/build/dial.go
@@ -1,0 +1,62 @@
+package build
+
+import (
+	"context"
+	stderrors "errors"
+	"net"
+
+	"github.com/containerd/containerd/platforms"
+	"github.com/docker/buildx/builder"
+	"github.com/docker/buildx/util/progress"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+func Dial(ctx context.Context, nodes []builder.Node, pw progress.Writer, platform *v1.Platform) (net.Conn, error) {
+	nodes, err := filterAvailableNodes(nodes)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(nodes) == 0 {
+		return nil, errors.New("no nodes available")
+	}
+
+	var pls []v1.Platform
+	if platform != nil {
+		pls = []v1.Platform{*platform}
+	}
+
+	opts := map[string]Options{"default": {Platforms: pls}}
+	resolved, err := resolveDrivers(ctx, nodes, opts, pw)
+	if err != nil {
+		return nil, err
+	}
+
+	var dialError error
+	for _, ls := range resolved {
+		for _, rn := range ls {
+			if platform != nil {
+				p := *platform
+				var found bool
+				for _, pp := range rn.platforms {
+					if platforms.Only(p).Match(pp) {
+						found = true
+						break
+					}
+				}
+				if !found {
+					continue
+				}
+			}
+
+			conn, err := nodes[rn.driverIndex].Driver.Dial(ctx)
+			if err == nil {
+				return conn, nil
+			}
+			dialError = stderrors.Join(err)
+		}
+	}
+
+	return nil, errors.Wrap(dialError, "no nodes available")
+}
--- a/commands/dial_stdio.go
+++ b/commands/dial_stdio.go
@@ -1,0 +1,132 @@
+package commands
+
+import (
+	"io"
+	"net"
+	"os"
+
+	"github.com/containerd/containerd/platforms"
+	"github.com/docker/buildx/build"
+	"github.com/docker/buildx/builder"
+	"github.com/docker/buildx/util/progress"
+	"github.com/docker/cli/cli/command"
+	"github.com/moby/buildkit/util/appcontext"
+	"github.com/moby/buildkit/util/progress/progressui"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
+)
+
+type stdioOptions struct {
+	builder  string
+	platform string
+	progress string
+}
+
+func runDialStdio(dockerCli command.Cli, opts stdioOptions) error {
+	ctx := appcontext.Context()
+
+	contextPathHash, _ := os.Getwd()
+	b, err := builder.New(dockerCli,
+		builder.WithName(opts.builder),
+		builder.WithContextPathHash(contextPathHash),
+	)
+	if err != nil {
+		return err
+	}
+
+	if err = updateLastActivity(dockerCli, b.NodeGroup); err != nil {
+		return errors.Wrapf(err, "failed to update builder last activity time")
+	}
+	nodes, err := b.LoadNodes(ctx)
+	if err != nil {
+		return err
+	}
+
+	printer, err := progress.NewPrinter(ctx, os.Stderr, progressui.DisplayMode(opts.progress), progress.WithPhase("dial-stdio"), progress.WithDesc("builder: "+b.Name, "builder:"+b.Name))
+	if err != nil {
+		return err
+	}
+
+	var p *v1.Platform
+	if opts.platform != "" {
+		pp, err := platforms.Parse(opts.platform)
+		if err != nil {
+			return errors.Wrapf(err, "invalid platform %q", opts.platform)
+		}
+		p = &pp
+	}
+
+	defer printer.Wait()
+
+	return progress.Wrap("Proxying to builder", printer.Write, func(sub progress.SubLogger) error {
+		var conn net.Conn
+
+		err := sub.Wrap("Dialing builder", func() error {
+			conn, err = build.Dial(ctx, nodes, printer, p)
+			if err != nil {
+				return err
+			}
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+
+		defer conn.Close()
+
+		go func() {
+			<-ctx.Done()
+			closeWrite(conn)
+		}()
+
+		var eg errgroup.Group
+
+		eg.Go(func() error {
+			_, err := io.Copy(conn, os.Stdin)
+			closeWrite(conn)
+			return err
+		})
+		eg.Go(func() error {
+			_, err := io.Copy(os.Stdout, conn)
+			closeRead(conn)
+			return err
+		})
+		return eg.Wait()
+	})
+}
+
+func closeRead(conn net.Conn) error {
+	if c, ok := conn.(interface{ CloseRead() error }); ok {
+		return c.CloseRead()
+	}
+	return conn.Close()
+}
+
+func closeWrite(conn net.Conn) error {
+	if c, ok := conn.(interface{ CloseWrite() error }); ok {
+		return c.CloseWrite()
+	}
+	return conn.Close()
+}
+
+func dialStdioCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
+	opts := stdioOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "dial-stdio",
+		Short: "Proxy current stdio streams to builder instance",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.builder = rootOpts.builder
+			return runDialStdio(dockerCli, opts)
+		},
+	}
+
+	flags := cmd.Flags()
+	cmd.Flags()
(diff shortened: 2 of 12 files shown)


build/dial.go
```diff
@@ -0,0 +1,46 @@
...
+
+	if len(nodes) == 0 {
+		return nil, errors.New("no nodes available")
+	}
+
+	rand.Shuffle(len(nodes), func(i, j int) {
+		nodes[i], nodes[j] = nodes[j], nodes[i]
+	})
+
+	// TODO: filter nodes by platform
```
tonistiigi, line 28: This should be a flag for the command probably.

driver/docker/driver.go
```diff
@@ -55,6 +55,10 @@ func (d *Driver) Rm(ctx context.Context, force, rmVolume, rmDaemon bool) error {
 	return nil
 }
 
+func (d *Driver) Dial(ctx context.Context) (net.Conn, error) {
```
tonistiigi, line 58: `Client` should reuse `Dial`

commands/dial_stdio.go
```diff
@@ -0,0 +1,90 @@
...
+	return progress.Wrap("[internal] Dialing builder: "+b.Name, printer.Write, func(sub progress.SubLogger) error {
+		conn, err := build.Dial(ctx, nodes, printer)
+		if err != nil {
+			return err
+		}
+		defer conn.Close()
+
+		go func() {
+			<-ctx.Done()
+			conn.Close()
```
tonistiigi, line 58: How does one shut down connection cleanly? If we don't want to add special method then maybe context close should just close the read side of pipe?

cpuguy83: You mean attempt a half-close like below?

```go

		go func() {
			<-ctx.Done()
			if c, ok := conn.(interface{ CloseRead() error }); ok {
				c.CloseRead()
				return
			}
			conn.Close()
		}()
```

tonistiigi: No, I meant wrapping `os.Stdin` with `io.Pipe` and then calling `PipeWriter.Close()` on it.

tonistiigi: Actually, I see you call `conn.Close()` manually. Then yes, that would need to be `CloseWrite()` if conn supports it.

commands/dial_stdio.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return progress.Wrap("[internal] Dialing builder: "+b.Name, printer.Write, func(sub progress.SubLogger) error {
+		conn, err := build.Dial(ctx, nodes, printer)
+		if err != nil {
+			return err
+		}
+		defer conn.Close()
+
+		go func() {
```
jedevc, line 56: Could we extract the logic for dial-stdio in buildkit into a reusable function so we could just call that here after getting the `conn`?

Ideally, we wouldn't end up with diverging implementations in buildctl/buildx (e.g. that would make it a pain to fix bugs in the future, etc).

cpuguy83: One implementation of what exactly?
the main part of this is just wiring up copy operations.
I'm not sure its worth sharing an implementation of that?

jedevc: Ah, I think I'm just confused by the differences in handling `CloseRead`/`CloseWrite` here versus in buildctl. Given that they seem to be doing the same thing, I'm not quite sure I understand why the code is so different.

cpuguy83: @AkihiroSuda I'm curious on this case where its wrapping with a nop-closer.https://github.com/moby/buildkit/blame/5ae9b23c40a926615b6c3cc1da3f5457d2f61fd4/cmd/buildctl/dialstdio.go#L35

I don't think we necessarily need the stream to handle a half-close here (hence why its implemented here), just that it s a little cleaner in terms of the connection semantics.

Other than that I think the implementation is pretty much the same, just this is taking advantage of errgroup and buildctl is basically doing its own hand-rolled equivalent.

build/dial.go
cpuguy83: Updated to support this.
Also makes the node selection use `resolveDriverBase`

commands/dial_stdio.go
cpuguy83: Added.

driver/docker/driver.go
cpuguy83: Fixed.

commands/dial_stdio.go
AkihiroSuda: From my memory, tty stuff didn't work correctly without handling half-close there

cpuguy83: I was trying to test for using tty as stdin/stdout but this broke http2.
Not sure how to check for this.

Fundamentally though, the only real change as compared to buildctl would be to not fallback to `Close` on when `CloseRead` is not implemented.

commands/dial_stdio.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if err = updateLastActivity(dockerCli, b.NodeGroup); err != nil {
+		return errors.Wrapf(err, "failed to update builder last activity time")
+	}
+	nodes, err := b.LoadNodes(ctx)
+	if err != nil {
+		return err
+	}
+
+	printer, err := progress.NewPrinter(ctx, os.Stderr, progressui.AutoMode, progress.WithPhase("dial-stdio"), progress.WithDesc("builder: "+b.Name, "builder:"+b.Name))
```
tonistiigi, line 46: If we are going to show progressbar then mode should be controllable with `--progress` flag. 

Otherwise, I'm fine with no progress or text warning only as I think is usually the case for such commands.

build/dial.go
```diff
@@ -0,0 +1,59 @@
...
+	if len(nodes) == 0 {
+		return nil, errors.New("no nodes available")
+	}
+
+	var pls []v1.Platform
+	if platform != nil {
+		pls = []v1.Platform{*platform}
+	}
+
+	opts := map[string]Options{"": {Platforms: pls}}
```
tonistiigi, line 29: nit: call this key `"default"` (defaultTargetName value)

build/dial.go
```diff
@@ -0,0 +1,59 @@
...
+						break
+					}
+				}
+				if !found {
+					continue
+				}
+			}
+
+			conn, err := nodes[rn.driverIndex].Driver.Dial(ctx)
+			if err == nil {
```
tonistiigi, line 52: Shouldn't this error be returned instead of "no nodes available" after all nodes are processed?

There is also some discussion that these fallback rules should change https://github.com/docker/buildx/issues/2119#issuecomment-1815517521 but ok to switch that later.

driver/remote/driver.go
```diff
@@ -84,6 +85,23 @@ func (d *Driver) Client(ctx context.Context) (*client.Client, error) {
 	return client.New(ctx, d.InitConfig.EndpointAddr, opts...)
 }
 
+func (d *Driver) Dial(ctx context.Context) (net.Conn, error) {
+	if d.tlsOpts != nil {
+		// TODO: add TLS support
```
tonistiigi, line 90: Are there problems with this?

docs/reference/buildx_dial-stdio.md
```diff
@@ -0,0 +1,15 @@
...
+### Options
+
+| Name         | Type     | Default | Description                                      |
+|:-------------|:---------|:--------|:-------------------------------------------------|
+| `--builder`  | `string` |         | Override the configured builder instance         |
+| `--platform` | `string` |         | Target platform: this is used for node selection |
+
+
+<!---MARKER_GEN_END-->
+
```
tonistiigi, line 15: Let's add a bit more description/usage here. It isn't entirely obvious for new user what "dial stdio" means.

commands/dial_stdio.go
cpuguy83: I've updated this with a `--progress` flag that defaults to `quiet`.

build/dial.go
cpuguy83: Done.

driver/remote/driver.go
cpuguy83: Implemented this

docs/reference/buildx_dial-stdio.md
cpuguy83: Added extra docs.

build/dial.go
cpuguy83: Updated to include errors.

docs/reference/buildx_dial-stdio.md
tonistiigi: Not sure what changed for this

cpuguy83: 🤔 
I wrote stuff... apparently I did not commit it somehow.

cpuguy83: Pushed up again.

cpuguy83: Oh I know what happened... I did docs the wrong way.

docs/reference/buildx_dial-stdio.md
```diff
@@ -0,0 +1,49 @@
...
+| `--builder`  | `string` |         | Override the configured builder instance         |
+| `--platform` | `string` |         | Target platform: this is used for node selection |
+| `--progress` | `string` | `quiet` | Set type of progress output (auto, plain, tty).  |
+
+
+<!---MARKER_GEN_END-->
+
+## Description
+
+dial-stdio uses the stdin and stoud streams of the command to proxy to the configured builder instance.
```
tonistiigi, line 19: stdout

docs/reference/buildx_dial-stdio.md
```diff
@@ -0,0 +1,49 @@
+# docker buildx dial-stdio
+
+<!---MARKER_GEN_START-->
+Dial stdio
```
tonistiigi, line 4: "Proxy current stdio streams to builder instance"

docs/reference/buildx_dial-stdio.md
```diff
@@ -0,0 +1,49 @@
...
+| `--platform` | `string` |         | Target platform: this is used for node selection |
+| `--progress` | `string` | `quiet` | Set type of progress output (auto, plain, tty).  |
+
+
+<!---MARKER_GEN_END-->
+
+## Description
+
+dial-stdio uses the stdin and stoud streams of the command to proxy to the configured builder instance.
+It is not intended to be used by humans, but rather by other tools that want to interact with the builder instance API.
```
tonistiigi, line 20: "builder instance API." -> "builder instance via BuildKit API"

docs/reference/buildx_dial-stdio.md
```diff
@@ -0,0 +1,49 @@
...
+
+defer cmd.Wait()
+
+go func() {
+    cmd.Wait()
+    c2.Close()
+}()
+
+client.New(ctx, "", client.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
+    return c2, nil
```
tonistiigi, line 47: Why use static instance here? I think it should invoke new command when dialer gets called.

```
func dial(...) (net.Conn, error) {
  _ = exec.Command()
  ...
}

client.New(ctx, "", client.WithContextDialer(dial))
```

docs/reference/buildx_dial-stdio.md
```diff
@@ -0,0 +1,48 @@
...
+
+```go
+
+client.New(ctx, "", client.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
+    c1, c2 := net.Pipe()
+    cmd := exec.Command("docker", "buildx", "dial-stdio")
+    cmd.Stdin = c1
+    cmd.Stdout = c1
+
+    if err := cmd.Start() {
```
tonistiigi, line 35: `err := cmd.Start(); err != nil`

docs/reference/buildx_dial-stdio.md
```diff
@@ -0,0 +1,48 @@
...
+It is not intended to be used by humans, but rather by other tools that want to interact with the builder instance via BuildKit API.
+
+## Examples
+
+Example go program that uses the dial-stdio command wire up a buildkit client.
+This is for example use only and may not be suitable for production use.
+
+```go
+
+client.New(ctx, "", client.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
```
tonistiigi, line 29: nit: remove the parameters variable names if they are unused

tonistiigi (approved):

PR #2121, opened 2023-11-16
Permit '=' separator and '[ipv6]' in --add-host

robmry:
- Fixes docker/cli#4648
- See https://github.com/docker/cli/issues/4648#issuecomment-1807992740
- related: https://github.com/docker/cli/pull/4663

**- What I did**

Make it easier to specify IPv6 addresses in the '--add-host' option by permitting 'host=ip' in addition to 'host:ip', and allowing square brackets around the address.

For example:

    --add-host=hostname:127.0.0.1
    --add-host=hostname:::1
    --add-host=hostname=::1
    --add-host=hostname:[::1]

**- How I did it**

Updated the function that converts from `--add-host` format to buildx's csv format, stripping brackets from the address.

This pull request is equivalent to one in `docker/cli`, for `docker run` and legacy builds.

**- How to verify it**

New unit tests pass. And ...

```
/work # cat Dockerfile
FROM alpine:latest
RUN ping -c1 something
RUN ping -c1 somethingv6

/work # docker build --add-host something=8.8.8.8 --add-host somethingv6=[::1] . 2>&1 | cat

#1 [internal] load .dockerignore
#1 transferring context: 2B done
#1 DONE 0.0s

#2 transferring dockerfile: 104B done
#2 DONE 0.0s

#3 [internal] load metadata for docker.io/library/alpine:latest
#3 DONE 0.9s

#4 [1/3] FROM docker.io/library/alpine:latest@sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978
#4 CACHED

#5 [2/3] RUN ping -c1 something
#5 0.116 PING something (8.8.8.8): 56 data bytes
#5 0.190 64 bytes from 8.8.8.8: seq=0 ttl=62 time=73.792 ms
#5 0.190
#5 0.190 --- something ping statistics ---
#5 0.190 1 packets transmitted, 1 packets received, 0% packet loss
#5 0.190 round-trip min/avg/max = 73.792/73.792/73.792 ms
#5 DONE 0.2s

#6 [3/3] RUN ping -c1 somethingv6
#6 0.205 PING somethingv6 (::1): 56 data bytes
#6 0.205 64 bytes from ::1: seq=0 ttl=64 time=0.028 ms
#6 0.205
#6 0.205 --- somethingv6 ping statistics ---
#6 0.205 1 packets transmitted, 1 packets received, 0% packet loss
#6 0.205 round-trip min/avg/max = 0.028/0.028/0.028 ms
#6 DONE 0.2s

#7 exporting to image
#7 exporting layers done
#7 writing image sha256:ea192cc9965b812a66562225e85b437af3af1d8347daca9d65eff8a6498b679e done
#7 DONE 0.0s
```

**- Description for the changelog**

Permit '=' separator and '[ipv6]' in --add-host

### DIFF
--- a/build/utils.go
+++ b/build/utils.go
@@ -65,7 +65,10 @@ func toBuildkitExtraHosts(ctx context.Context, inp []string, nodeDriver *driver.
 	}
 	hosts := make([]string, 0, len(inp))
 	for _, h := range inp {
-		host, ip, ok := strings.Cut(h, ":")
+		host, ip, ok := strings.Cut(h, "=")
+		if !ok {
+			host, ip, ok = strings.Cut(h, ":")
+		}
 		if !ok || host == "" || ip == "" {
 			return "", errors.Errorf("invalid host %s", h)
 		}
@@ -77,8 +80,16 @@ func toBuildkitExtraHosts(ctx context.Context, inp []string, nodeDriver *driver.
 				return "", errors.Wrap(err, "unable to derive the IP value for host-gateway")
 			}
 			ip = hgip.String()
-		} else if net.ParseIP(ip) == nil {
-			return "", errors.Errorf("invalid host %s", h)
+		} else {
+			// If the address is enclosed in square brackets, extract it (for IPv6, but
+			// permit it for IPv4 as well; we don't know the address family here, but it's
+			// unambiguous).
+			if len(ip) > 2 && ip[0] == '[' && ip[len(ip)-1] == ']' {
+				ip = ip[1 : len(ip)-1]
+			}
+			if net.ParseIP(ip) == nil {
+				return "", errors.Errorf("invalid host %s", h)
+			}
 		}
 		hosts = append(hosts, host+"="+ip)
 	}
--- a/build/utils_test.go
+++ b/build/utils_test.go
@@ -1,0 +1,148 @@
+package build
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestToBuildkitExtraHosts(t *testing.T) {
+	tests := []struct {
+		doc         string
+		input       []string
+		expectedOut string // Expect output==input if not set.
+		expectedErr string // Expect success if not set.
+	}{
+		{
+			doc:         "IPv4, colon sep",
+			input:       []string{`myhost:192.168.0.1`},
+			expectedOut: `myhost=192.168.0.1`,
+		},
+		{
+			doc:   "IPv4, eq sep",
+			input: []string{`myhost=192.168.0.1`},
+		},
+		{
+			doc:         "Weird but permitted, IPv4 with brackets",
+			input:       []string{`myhost=[192.168.0.1]`},
+			expectedOut: `myhost=192.168.0.1`,
+		},
+		{
+			doc:         "Host and domain",
+			input:       []string{`host.and.domain.invalid:10.0.2.1`},
+			expectedOut: `host.and.domain.invalid=10.0.2.1`,
+		},
+		{
+			doc:         "IPv6, colon sep",
+			input:       []string{`anipv6host:2003:ab34:e::1`},
+			expectedOut: `anipv6host=2003:ab34:e::1`,
+		},
+		{
+			doc:         "IPv6, colon sep, brackets",
+			input:       []string{`anipv6host:[2003:ab34:e::1]`},
+			expectedOut: `anipv6host=2003:ab34:e::1`,
+		},
+		{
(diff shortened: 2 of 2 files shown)


build/utils.go
```diff
@@ -65,7 +65,10 @@ func toBuildkitExtraHosts(ctx context.Context, inp []string, nodeDriver *driver.
 	}
 	hosts := make([]string, 0, len(inp))
 	for _, h := range inp {
-		host, ip, ok := strings.Cut(h, ":")
+		host, ip, ok := strings.Cut(h, "=")
+		if !ok {
+			host, ip, ok = strings.Cut(h, ":")
+		}
```
crazy-max, line 71: Should we update the reference docs in https://github.com/docker/buildx/blob/cb378866587090995fba030bc75a9ca2fb4d8d26/commands/build.go#L489?

Guess it should be `... (format: "host=ip")` if this format is the preferred one now.

crazy-max (approved):
LGTM

jsternberg (approved):

robmry:
The changes for CLI and buildx are done - but docker-compose has an "extra_hosts" option that'll need the same treatment.

robmry:
Updated to use domain ".invalid" in unit tests - as per @thaJeztah's comment on the CLI review, https://github.com/docker/cli/pull/4663#discussion_r1400463372

crazy-max (approved):

thaJeztah (approved):
LGTM

thaJeztah:
I added a cherry-pick label for consideration (so that buildx and cli will align on accepted formats)

docs/reference/buildx_build.md
```diff
@@ -15,7 +15,7 @@ Start a build
 
 | Name                                                                                                                                                   | Type          | Default   | Description                                                                                         |
 |:-------------------------------------------------------------------------------------------------------------------------------------------------------|:--------------|:----------|:----------------------------------------------------------------------------------------------------|
-| [`--add-host`](https://docs.docker.com/engine/reference/commandline/build/#add-host)                                                                   | `stringSlice` |           | Add a custom host-to-IP mapping (format: `host:ip`)                                                 |
+| [`--add-host`](https://docs.docker.com/engine/reference/commandline/build/#add-host)                                                                   | `stringSlice` |           | Add a custom host-to-IP mapping (format: `host=ip`)                                                 |
```
tonistiigi, line 18: I'm not sure if we should just switch the formatting in docs as it takes time for people to upgrade to the latest versions. Doesn't look like the old formatting is deprecated as well.

robmry: Yes, I see what you mean - the online docs aren't very version-specific.

This part of the ".md" is generated from the flags.StringSliceVar() text.  We could put both in there with an "or", but that may not really help to make it obvious that the `:` is for backward compatibility - and we probably don't want an essay in there. Earlier comments have been about preferring ("soft-deprecating") the colon separator.

On the other hand, the "--add-host" in the first column here is a link to the cli/build docs, here it is in my branch ...

https://github.com/robmry/buildx/blob/align_add-host_with_cli/docs/reference/buildx_build.md

And the modified version of the page it'll link to is here ...

https://github.com/robmry/cli/blob/4648-clearer_ipv6_in_add-host/docs/reference/commandline/build.md#add-host

It says ...

> You can wrap an IPv6 address in square brackets. `=` and `:` are both valid separators. Both formats in the following example are valid:

... perhaps we should make it clear there that the `:` is for backward compatibility?

@dvdksn - any suggestions?

commands/build.go
```diff
@@ -486,7 +486,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 
 	flags := cmd.Flags()
 
-	flags.StringSliceVar(&options.extraHosts, "add-host", []string{}, `Add a custom host-to-IP mapping (format: "host:ip")`)
+	flags.StringSliceVar(&options.extraHosts, "add-host", []string{}, `Add a custom host-to-IP mapping (format: "host=ip")`)
```
dvdksn, line 489: We could keep the example as `:` here, until we officially deprecate the version. That would give users time to upgrade to later versions which would support both syntaxes.

robmry: Ok, will do ...

@crazy-max - this will undo the earlier change for your https://github.com/docker/buildx/pull/2121#discussion_r1395737201 - but I think it makes sense given https://github.com/docker/buildx/pull/2121#discussion_r1401384032 ?

crazy-max: Yes agree let's keep `:` for now

crazy-max (approved):

dvdksn (approved):

PR #2128, opened 2023-11-17
docs: update du cmd description

dvdksn:
Adds a better description for the `du` command, including asterisks, `RECLAIMABLE`, and examples

### DIFF
--- a/docs/reference/buildx_du.md
+++ b/docs/reference/buildx_du.md
@@ -20,6 +20,70 @@ Disk usage
 
 ## Examples
 
+### Show disk usage
+
+The `docker buildx du` command shows the disk usage for the currently selected
+builder.
+
+```console
+$ docker buildx du
+ID                                RECLAIMABLE    SIZE          LAST ACCESSED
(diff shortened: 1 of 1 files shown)


crazy-max (approved):
:heart: :shipit:

docs/reference/buildx_du.md
```diff
@@ -20,6 +20,70 @@ Disk usage
...
+Total:         10.36GB
+```
+
+If `RECLAIMABLE` is false, the `docker buildx du prune` command won't delete
+the record, even if you use `--all`. That's because the record is actively in
+use by some component of the builder.
+
+The asterisks (\*) in the output indicate the following:
+
+- An asterisk next to an ID (`zu7m6evdpebh5h8kfkpw9dlf2*`) indicates that the record
```
tonistiigi, line 54: We could provide some example of how this looks in the verbose record as well as it is much easier to understand that these are separate fields with their own values there, instead of knowing what `*` means.

dvdksn: Yep, I can do a follow-up

Issue #2119, opened 2023-11-15
feature request: strict platform matching


nicks:
### Description

Repro steps:

1. Create a driver with this spec:

```
{"Name":"remote","Driver":"remote","Nodes":[{"Name":"remote0","Endpoint":"tcp://localhost:1234","Platforms":[{"architecture":"arm64","os":"linux"}],"Flags":null,"DriverOpts":null,"Files":null}
```

2. Send a request to the builder for a different platform

```
docker buildx build -t out --platform linux/amd64 .
```

Expected result:
I would expect to fail because there's no available builder that matches the platform

Actual result:
buildx sends the request to the ARM builder

Additional info:
The full problem is more complicated, and involves how the Node fallback logic works when one of the builders is unavailable. Basically we're seeing problems where if platform isn't specified, and one of the builders is unavailable, the request ends up landing on a random Node and with a random Platform.

tonistiigi:
> when one of the builders is unavailable. 

But in your example you only have a single node builder. I would expect this to fail if you have a multi-node builder, platform for node is manually set to `linux/amd64` and that node is unavailable. In that case no fallback to the arm node should happen and it should fail because `amd64` node is down.

nicks:
In all my tests, it silently falls back. I'm not sure I understand all the codepaths. The problem might be this function?

https://github.com/docker/buildx/blob/cb378866587090995fba030bc75a9ca2fb4d8d26/build/build.go#L121

If some of the nodes are in an error state, but there's at least one working Node, then it filters out the bad nodes and "succeeds"

tonistiigi:
That does not look correct indeed. If the node has set a manual platform, then I think it should that node should be always used without any fallbacks.

But the problem is that if there is no `--platform` and the first node is down, then we don't know that the "manual platform" is the native (default) platform for the node. I guess we could just guess that if there is any manual platform set then the first one set is the native.

tonistiigi:
Do you agree that this issue should just be fixed by updating the node resolution logic, without any new keys like the "StrictPlatformMatching" in your first comment?

nicks:
Ya that works for me

tonistiigi:
Proposed changes for clarity:

1. Multi-node cluster where node has set a manual platform, node is in error state and user sets `--platform` that matches manual platform
- Old rule:
  - Fallback to the next active node that supports platform
- New rule
  - Fallback to the next active node that also has same manual platform. Do not fallback to detected platform. Error if no node to fallback.
[...]

PR #2131, opened 2023-11-20
build: infer platform from first node if none set

crazy-max:
closes https://github.com/docker/buildx/pull/2124

If user doesn't specify any platform, it should be inferred from the first node instead of using the local platform. In follow-up we should also look at the rules suggested in https://github.com/docker/buildx/issues/2119#issuecomment-1815517521 to take into account the preferred platform as well.

### DIFF
--- a/build/driver.go
+++ b/build/driver.go
@@ -166,35 +166,34 @@ func (r *nodeResolver) resolve(ctx context.Context, ps []specs.Platform, pw prog
 
 	perfect := true
 	nodeIdxs := make([]int, 0)
-	if len(ps) == 0 {
-		idx := r.get(platforms.DefaultSpec(), matcher, additional)
+	for _, p := range ps {
+		idx := r.get(p, matcher, additional)
 		if idx == -1 {
 			idx = 0
 			perfect = false
 		}
 		nodeIdxs = append(nodeIdxs, idx)
-	} else {
-		for _, p := range ps {
-			idx := r.get(p, matcher, additional)
-			if idx == -1 {
-				idx = 0
-				perfect = false
-			}
-			nodeIdxs = append(nodeIdxs, idx)
-		}
 	}
 
 	var nodes []*resolvedNode
-	for i, idx := range nodeIdxs {
-		node := &resolvedNode{
+	if len(nodeIdxs) == 0 {
+		nodes = append(nodes, &resolvedNode{
 			resolver:    r,
-			driverIndex: idx,
+			driverIndex: 0,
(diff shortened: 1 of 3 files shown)


build/driver_test.go
```diff
@@ -226,7 +226,7 @@ func TestSelectNodeCurrentPlatform(t *testing.T) {
 	require.NoError(t, err)
 	require.True(t, perfect)
 	require.Len(t, res, 1)
-	require.Equal(t, "bbb", res[0].Node().Builder)
+	require.Equal(t, "aaa", res[0].Node().Builder)
```
tonistiigi, line 229: The real issue is that the platform in `opt` gets reset when it should not. We need to figure out a test to make sure it doesn't happen. Looks like the old build code only expect "resolveNode" to have a platform if it is an override to the existing platform but maybe this should be made safer (more explicit) somehow.

crazy-max: Rebased on top of @jedevc changes in https://github.com/docker/buildx/compare/master...jedevc:buildx:fix-setting-wrong-platform-on-solve-opt to fix the solve opt case and remove the duplicated map.

crazy-max: I guess you talk about the `TestSelectNodeNoPlatform` and default spec case right? It seems related to https://github.com/docker/buildx/issues/2119 and proposed new rules in https://github.com/docker/buildx/issues/2119#issuecomment-1815517521 should cover this in follow-up iiuc.

jedevc (approved):

build/driver.go
```diff
@@ -19,6 +19,8 @@ import (
 )
 
 type resolvedNode struct {
+	SolveOpt *client.SolveOpt
```
tonistiigi, line 22: I think @jedevc did not like the `driverPair` name, but this is the use case for why it was added. That we pass along a struct that contains both the driver (in here it is called resolver+platform, so driver for a specific node) and the request.

I think even if we use embed structs, the separation of structs would be clearer this way.

tonistiigi (commented):
The logic side SGTM. Comment about the code organization.

build/driver.go
crazy-max: After discussing with @tonistiigi, the idea is to have a clear separation of concerns between drivers resolution and solve opts attached to the driver in build. Pushed an extra commit to move `*client.SolveOpt` to a local struct type in build.

tonistiigi (approved):

jedevc (approved):

PR #2140, opened 2023-11-25
rm: support removing multiple builders at once

crazy-max:
Similar to `docker rm`, it gives the possibility to remove one or more builder instances at once.

### DIFF
--- a/commands/rm.go
+++ b/commands/rm.go
@@ -9,7 +9,6 @@ import (
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/store/storeutil"
 	"github.com/docker/buildx/util/cobrautil/completion"
-	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/moby/buildkit/util/appcontext"
 	"github.com/pkg/errors"
@@ -18,7 +17,7 @@ import (
 )
 
 type rmOptions struct {
-	builder     string
+	builders    []string
 	keepState   bool
 	keepDaemon  bool
 	allInactive bool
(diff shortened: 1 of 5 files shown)


commands/rm.go
```diff
@@ -46,50 +45,58 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
...
-		builder.WithStore(txn),
-		builder.WithSkippedValidation(),
-	)
-	if err != nil {
-		return err
-	}
+	eg, _ := errgroup.WithContext(ctx)
+	for _, name := range in.builder {
+		func(name string) {
+			eg.Go(func() error {
```
jedevc, line 51: If one of the `rm` fails, shouldn't we continue to execute all of them, instead of exiting at the earliest opportunity? I think that's how the docker cli works as well.

crazy-max: Yes indeed good call out: https://github.com/docker/cli/blob/c1455b6751586ee994c497013b7d49b6e4ced520/cli/command/container/rm.go#L67-L80

commands/rm.go
```diff
@@ -18,7 +17,7 @@ import (
 )
 
 type rmOptions struct {
-	builder     string
+	builder     []string
```
AkihiroSuda, line 20: nit
```suggestion
	builders     []string
```

commands/rm.go
```diff
@@ -46,50 +45,68 @@ func runRm(dockerCli command.Cli, in rmOptions) error {
...
 
-	b, err := builder.New(dockerCli,
-		builder.WithName(in.builder),
-		builder.WithStore(txn),
-		builder.WithSkippedValidation(),
-	)
-	if err != nil {
-		return err
-	}
+	eg, _ := errgroup.WithContext(ctx)
```
tonistiigi, line 48: Is it intentional that you are not passing the errgroup context to goroutines?

crazy-max: Yes this is intentional so even if one builder removal fails, it will continue with the removal of the remaining builders.

tonistiigi (approved):

PR #2146, opened 2023-12-01
gitutil: find default remote by tracking branch

tonistiigi:
Using this command resolves remote based on remote-tracking branch of the currently selected branch and should be more precise in case we can't predict if the user uses origin to mark upstream or their fork.

This reverses https://github.com/docker/buildx/pull/1548 . The issue there isn't unfortunately very clear about what exact case did not show correct output before. It is possible it was just a misunderstanding about how `ls-remote --get-url` works.

https://github.com/git/git/blob/v2.43.0/remote.c#L550-L554

~The "origin" fallback probably does not do anything anymore. 95% sure it is safe to remove.~

~Looking at the git source, another difference is that if there is only one remote, then `ls-remote --get-url` returns that(even if not remote-tracking) while the current implementation would not set URL in this case. If this is not the behavior we want, it could be detected with some additional commands or we could get the remote-tracking branch directly without `ls-remote`.~

### DIFF
--- a/util/gitutil/gitutil.go
+++ b/util/gitutil/gitutil.go
@@ -78,7 +78,13 @@ func (c *Git) GitDir() (string, error) {
 }
 
 func (c *Git) RemoteURL() (string, error) {
-	// Try to get the remote URL from the origin remote first
+	// Try default remote based on remote tracking branch
+	if remote, err := c.currentRemote(); err == nil && remote != "" {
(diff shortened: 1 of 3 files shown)


tonistiigi:
The test failed for the condition of a single remote without any remote-tracking so I switched the implementation to still check for remote-tracking branch but avoid using `ls-remote`.

crazy-max (approved):
Yes it makes more sense

jedevc (approved):

PR #2150, opened 2023-12-04
builder: move builder creation logic to builder package

crazy-max:
This moves the builder creation logic to the builder package as we plan to allow builder creation in the build UI of Docker Desktop so we avoid drifting with the same logic in `command` package.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -2,7 +2,9 @@ package builder
 
 import (
 	"context"
+	"encoding/csv"
 	"encoding/json"
+	"net/url"
 	"os"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


builder/builder.go
```diff
@@ -322,3 +331,317 @@ func GetBuilders(dockerCli command.Cli, txn *store.Txn) ([]*Builder, error) {
...
+	}
+
+	if !opts.Append {
+		contexts, err := dockerCli.ContextStore().List()
+		if err != nil {
+			return nil, err
+		}
+		for _, c := range contexts {
+			if c.Name == name {
+				return nil, errors.Errorf("instance name %q already exists as context builder", name)
```
crazy-max, line 372: Same here

builder/builder.go
```diff
@@ -322,3 +330,315 @@ func GetBuilders(dockerCli command.Cli, txn *store.Txn) ([]*Builder, error) {
...
+		flags, err = shlex.Split(opts.Flags)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse buildkit flags")
+		}
+	}
+
+	var ep string
+	var setEp bool
+	switch {
+	case driverName == "kubernetes":
```
tonistiigi, line 443: Ideally, these driver specifics can be removed from this package eventually, but not new code as this is how it was in `commands` as well.

tonistiigi (approved):

PR #2154, opened 2023-12-10
docs: annotations

dvdksn:
Add/improve docs for adding annotations

- relates to docker/docs#18891


docs/reference/buildx_build.md
```diff
@@ -69,6 +69,52 @@ This page describes a subset of the new flags.
...
+matching platform. The following example adds the `foo=bar` annotation only to
+the manifest with the `linux/amd64` platform:
+
+```console
+$ docker buildx build -t TAG --annotation "manifest[linux/amd64]:foo=bar" --push .
+```
+
+Wildcards are not supported in the platform qualifier; you can't specify a type
+prefix like `manifest[linux/*]` to add annotations only to manifests which has
+`linux` as the OS platform.
```
crazy-max, line 113: Should this be rendered as a `Note`?

dvdksn: I didn't think it was significant enough to warrant a note. It was just something I noticed when testing the feature, and I thought it's probably good to call out. But I prefer notes to be things you are not allowed to miss, whereas this is more of an FYI

PR #2155, opened 2023-12-12
metrics: send metrics to the otel collector endpoint when active

jsternberg:
Introduce a meter provider to the buildx cli that will send metrics to
the otel-collector included in docker desktop if enabled.

This will send usage metrics to the desktop application but also send
metrics to a user-provided otlp receiver endpoint through the standard
environment variables.

This introduces a single metric which is the cli count for build and
bake along with the command name and a few additional attributes.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -29,8 +29,10 @@ import (
 	"github.com/docker/buildx/util/buildflags"
 	"github.com/docker/buildx/util/desktop"
 	"github.com/docker/buildx/util/ioset"
+	"github.com/docker/buildx/util/metrics"
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/buildx/util/tracing"
+	"github.com/docker/buildx/version"
 	"github.com/docker/cli-docs-tool/annotation"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
@@ -51,6 +53,9 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
 	"google.golang.org/grpc/codes"
 )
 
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -51,6 +51,12 @@ require (
 	k8s.io/client-go v0.26.7
 )
 
+require (
```
tonistiigi, line 54: This new block in here probably isn't necessary.

tonistiigi (commented):
Please add docs for how to collect/view the data and something that lists all the metrics that buildx will export.

If this `docker.cli.count` is something that currently serves more like a test value and may change in the future where more collection points are added then it should be behind `BUILDX_` opt-in env var until it is considered stable.

jsternberg:
@tonistiigi at the current moment, the metrics can't be exported or seen through a user-defined means. I'll add some docs for the metrics to get that started as I think we should allow a way to expose the metrics to end-users, but CLI metrics are a bit different from service metrics that they're generally more ephemeral.

I can add it behind an environment variable, but in its current iteration, these metrics can only be consumed by our backend without recompiling the source. I don't know if that changes the requirement to add an experimental flag for this. My fear here is that this data is already going to a backend that also has an experimental flag and we'll be adding too many experimental flags and won't get any data.

But yes. This metric is primarily meant to be a test one so I can see the metrics are working and to avoid Go telling me that the value isn't used.

go.mod
jsternberg: Let me fix this.

tonistiigi (commented):
Something is wrong with the e2e tests.

commands/build.go
```diff
@@ -926,3 +940,29 @@ func maybeJSONArray(v string) []string {
 	}
 	return []string{v}
 }
+
+func recordVersionInfo(mp metric.MeterProvider, command string) {
+	if !isExperimental() {
```
tonistiigi, line 945: nit: add extra comment in here saying that we are still in the process of testing/stabilizing these counters.

jsternberg:
End to end tests are functional now. I've split the vendor update into the PR [here](https://github.com/docker/buildx/pull/2175) so that will have to merge first.

The issue seemed to stem from how unix sockets are handled in the OTEL library. The `WithEndpoint()` option isn't intended to have a scheme and the environment variable removes the scheme before passing the option. So I mimic that behavior now. At the same time, unix sockets don't work without the scheme being passed to the endpoint so now there's custom logic for keeping the `unix://` when the scheme is `unix`.

tonistiigi (approved):

PR #2156, opened 2023-12-13
build: set local dirs as frontend attributes

crazy-max:
Set local dirs metadata if relative to ~working~ VCS directory so dockerfile path is tracked accurately.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -530,7 +530,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 	for k, opt := range opt {
 		multiDriver := len(drivers[k]) > 1
 		hasMobyDriver := false
-		gitattrs, err := getGitAttributes(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath)
+		gitattrs, addVCSLocalDir, err := getGitAttributes(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath)
 		if err != nil {
 			logrus.WithError(err).Warn("current commit information was not captured by the build")
 		}
@@ -553,6 +553,9 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 			}
 			for k, v := range gitattrs {
 				so.FrontendAttrs[k] = v
+			}
+			if addVCSLocalDir != nil {
+				addVCSLocalDir(so)
 			}
 			defers = append(defers, release)
 			reqn = append(reqn, &reqForNode{
--- a/build/git.go
+++ b/build/git.go
@@ -9,16 +9,17 @@ import (
 	"strings"
 
 	"github.com/docker/buildx/util/gitutil"
+	"github.com/moby/buildkit/client"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
 const DockerfileLabel = "com.docker.image.source.entrypoint"
 
-func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath string) (res map[string]string, _ error) {
-	res = make(map[string]string)
+func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath string) (map[string]string, func(*client.SolveOpt), error) {
+	res := make(map[string]string)
 	if contextPath == "" {
-		return
+		return nil, nil, nil
 	}
 
 	setGitLabels := false
@@ -37,7 +38,7 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 	}
 
 	if !setGitLabels && !setGitInfo {
-		return
+		return nil, nil, nil
 	}
 
 	// figure out in which directory the git command needs to run in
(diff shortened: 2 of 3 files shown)


build/git.go
```diff
@@ -111,5 +114,18 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
...
+		if root == "" {
+			root = wd
+		}
+		for k, dir := range so.LocalDirs {
+			d, err := filepath.Abs(dir)
+			if err != nil || !strings.HasPrefix(d, root) {
+				continue
+			}
+			if r, err := filepath.Rel(root, d); err == nil {
+				so.FrontendAttrs["vcs-localdir:"+k] = r
```
tonistiigi, line 127: This needs to start with `vcs:`. There is already a rule in buildkit to put all such metadata to provenance.

build/git.go
```diff
@@ -111,5 +114,18 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 		}
 	}
 
-	return
+	return res, func(so *client.SolveOpt) {
+		if root == "" {
```
tonistiigi, line 118: When will this be true?

build/git.go
```diff
@@ -111,5 +114,18 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
...
+			root = wd
+		}
+		for k, dir := range so.LocalDirs {
+			d, err := filepath.Abs(dir)
```
tonistiigi, line 122: Do these paths require `EvalSymlinks`?

build/build.go
```diff
@@ -530,7 +530,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 	for k, opt := range opt {
 		multiDriver := len(drivers[k]) > 1
 		hasMobyDriver := false
-		gitattrs, err := getGitAttributes(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath)
+		gitattrs, localDirsFunc, err := getGitAttributes(ctx, opt.Inputs.ContextPath, opt.Inputs.DockerfilePath)
```
tonistiigi, line 533: `s/localDirsFunc/addVCSLocalDir/`

tonistiigi (commented):
cc @cdupuis

build/git.go
crazy-max: If `BUILDX_GIT_LABELS=1` which is opt-in: https://github.com/docker/buildx/blob/8484fcdd5715c0683e8605e05cdffc0bfa8fde94/build/git.go#L24-L31

tonistiigi: you mean `BUILDX_GIT_INFO` ? I think if `BUILDX_GIT_INFO=0` then these paths should not be sent.

crazy-max: No I mean `BUILDX_GIT_LABELS` because root dir is computed in https://github.com/docker/buildx/blob/8484fcdd5715c0683e8605e05cdffc0bfa8fde94/build/git.go#L97

tonistiigi: This isn't related to labels, so should be opt-out with the same var that controls URL/commit.

build/git.go
```diff
@@ -93,23 +99,40 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
...
+	if setGitLabels && root != "" {
+		if dockerfilePath == "" {
+			dockerfilePath = filepath.Join(wd, "Dockerfile")
+		}
+		if !filepath.IsAbs(dockerfilePath) {
+			cwd, _ := os.Getwd()
+			dockerfilePath = filepath.Join(cwd, dockerfilePath)
+		}
+		dockerfilePath, _ = filepath.Rel(root, dockerfilePath)
+		if !strings.HasPrefix(dockerfilePath, "..") {
```
tonistiigi, line 111: `..` + `filepath.Separator` ?

build/git.go
```diff
@@ -93,23 +99,40 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
...
-				dockerfilePath = filepath.Join(cwd, dockerfilePath)
+			dir, err = filepath.Abs(dir)
+			if err != nil {
+				continue
 			}
-			dockerfilePath, _ = filepath.Rel(root, dockerfilePath)
-			if !strings.HasPrefix(dockerfilePath, "..") {
-				res["label:"+DockerfileLabel] = dockerfilePath
+			dir = filepath.ToSlash(filepath.Clean(dir))
+			if !strings.HasPrefix(dir, root) {
```
tonistiigi, line 130: Why this prefix check instead of `.Rel` + escape check?

build/git.go
crazy-max: It was added initially in https://github.com/docker/buildx/pull/1297 but instead we should just check if `.Rel` returns an error

tonistiigi: No, `Rel` can return `../` paths without error and they should not be allowed.

crazy-max: :man_facepalming:

tonistiigi (approved):

cdupuis (approved):
This looks great. Thanks for the PR. It solves a big challenge for Scout.

PR #2157, opened 2023-12-13
ci: use public bot account to push bin image

crazy-max:
@technicallyjosh Can you check if this account has access to https://hub.docker.com/r/docker/buildx-bin? Thanks


technicallyjosh (approved):
Good to go

PR #2185, opened 2024-01-11
metrics: add build duration metric

jsternberg:
This adds a build duration metric with attributes related to the build
id, build ref, driver name, and the resulting status of the build.

This also modifies some aspects of how the resource is configured by
including `service.instance.id`. This id is used to uniquely identify
the CLI invocation for use in downstream aggregators. This allows
downstream aggregators to know that the metric for an instance has not
reset and that it is a unique invocation for future aggregation. Some
work will have to be done in the aggregator to prevent the storage of
this instance id as it can cause issues with cardinality limitations,
but it's necessary for the initial reporter to include this.

The temporality selector is still the same, but has been removed from
the otlp exporter options since that one doesn't seem to do anything.  I
also recently learned the temporality didn't do what I thought it did.
I thought it would allow for multiple different invocations to be
treated as the same instance for the purposes of aggregation. It does
not work this way as the metric sdk considers each of these a gap reset.
That's the reason the instance id was added. This makes the difference
between cumulative and delta mostly cosmetic, but delta temporality has
some benefits for downstream consumers for aggregation so it's likely
good to keep.

The cli count has been removed as it doesn't contain anything new and
wasn't a useful metric. The naming of metrics has also been changed from
using `docker` as a prefix and instead relying on the instrumentation
name.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
+	"crypto/sha256"
 	_ "crypto/sha256" // ensure digests can be computed
 	"encoding/base64"
 	"encoding/hex"
@@ -55,6 +56,8 @@ import (
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/sync/errgroup"
 )
@@ -469,11 +472,11 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	return &so, releaseF, nil
 }
 
-func Build(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer) (resp map[string]*client.SolveResponse, err error) {
-	return BuildWithResultHandler(ctx, nodes, opt, docker, configDir, w, nil)
+func Build(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer, mp metric.MeterProvider) (resp map[string]*client.SolveResponse, err error) {
+	return BuildWithResultHandler(ctx, nodes, opt, docker, configDir, w, mp, nil)
 }
 
-func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer, resultHandleFunc func(driverIndex int, rCtx *ResultHandle)) (resp map[string]*client.SolveResponse, err error) {
+func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer, mp metric.MeterProvider, resultHandleFunc func(driverIndex int, rCtx *ResultHandle)) (resp map[string]*client.SolveResponse, err error) {
 	if len(nodes) == 0 {
 		return nil, errors.Errorf("driver required for build")
 	}
@@ -686,6 +689,14 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 						return err
 					}
 
+					buildAttributes := attribute.NewSet(
+						driverAttribute(dp),
+						buildIdentityAttribute(so),
+						buildRefAttribute(so),
+					)
+					pw, record := progress.Metrics(mp, pw,
+						metric.WithAttributeSet(buildAttributes))
+
 					frontendInputs := make(map[string]*pb.Definition)
 					for key, st := range so.FrontendInputs {
 						def, err := st.Marshal(ctx)
@@ -785,6 +796,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 					} else {
 						rr, err = c.Build(ctx, *so, "buildx", buildFunc, ch)
 					}
+
 					if desktop.BuildBackendEnabled() && node.Driver.HistoryAPISupported(ctx) {
 						buildRef := fmt.Sprintf("%s/%s/%s", node.Builder, node.Name, so.Ref)
 						if err != nil {
@@ -842,6 +854,9 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 							}
 						}
 					}
+
+					// Record the result of the build.
+					record(ctx)
 					return nil
 				})
 			}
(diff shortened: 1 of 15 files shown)


internal/env/env.go
```diff
@@ -0,0 +1,15 @@
+package env
+
```
tonistiigi, line 2: If you need a separate package you can use `confutil` call this `exp`

internal/metrics/metrics.go
```diff
@@ -27,40 +40,61 @@ const (
...
-		sdkmetric.WithResource(detect.Resource()),
+		sdkmetric.WithResource(Resource()),
 		sdkmetric.WithReader(reader),
 	)
-	return mp, reportFunc(reader, exps), nil
+	return withMeterProvider(ctx, mp), reportFunc(reader, exps), nil
+}
+
+func withMeterProvider(ctx context.Context, mp metric.MeterProvider) context.Context {
+	return context.WithValue(ctx, meterProviderKey, mp)
```
tonistiigi, line 81: It looks like this is just global and there is no state in the context that changes metrics (unlike tracing). In that case, I think we should just pass the meterprovider (or I'm ok with global one as well if it is initialized from `cmd`, or associated with a specific package, eg. this function can just return it if you want). Context shouldn't be used just to hide parameters that functions take.

This can also easily break with nobody noticing that some callpath did not initialize context.

internal/metrics/metrics.go
```diff
@@ -27,40 +40,61 @@ const (
...
 	}
 
-	// Use delta temporality because, since this is a CLI program, we can never
-	// know the cumulative value.
 	reader := sdkmetric.NewManualReader(
 		sdkmetric.WithTemporalitySelector(deltaTemporality),
 	)
 	mp := sdkmetric.NewMeterProvider(
-		sdkmetric.WithResource(detect.Resource()),
+		sdkmetric.WithResource(Resource()),
```
tonistiigi, line 74: Not new code but why is this duplicated from buildkit?

build/build.go
```diff
@@ -1523,3 +1540,47 @@ func ReadSourcePolicy() (*spb.Policy, error) {
 
 	return &pol, nil
 }
+
+// backendAttribute is a utility to retrieve the backend attribute from a resolvedNode.
+func backendAttribute(dp *resolvedNode) attribute.KeyValue {
```
tonistiigi, line 1545: This is `driver`

build/build.go
```diff
@@ -1523,3 +1540,47 @@ func ReadSourcePolicy() (*spb.Policy, error) {
...
+// This value should be consistent between builds.
+func buildIDAttribute(so *client.SolveOpt) attribute.KeyValue {
+	vcs := so.FrontendAttrs["vcs:source"]
+	target := so.FrontendAttrs["target"]
+	context := so.FrontendAttrs["context"]
+	filename := so.FrontendAttrs["filename"]
+
+	buildID := ""
+	if vcs != "" || target != "" || context != "" || filename != "" {
+		h := xxh3.New()
```
tonistiigi, line 1560: We use sha256 everywhere. This is not a place where performance considerations would require new dependency.

build/build.go
```diff
@@ -1523,3 +1540,47 @@ func ReadSourcePolicy() (*spb.Policy, error) {
...
+		for _, s := range []string{vcs, target, context, filename} {
+			h.WriteString(s)
+			h.Write([]byte{0})
+		}
+		buildID = hex.EncodeToString(h.Sum(nil))
```
tonistiigi, line 1565: I think we need to add something unique to node into this mix. Eg. see `tryNodeIdentifier`. Otherwise, this metric can leak data with a simple plaintext attack.

build/build.go
```diff
@@ -1523,3 +1540,47 @@ func ReadSourcePolicy() (*spb.Policy, error) {
...
+
+// backendAttribute is a utility to retrieve the backend attribute from a resolvedNode.
+func backendAttribute(dp *resolvedNode) attribute.KeyValue {
+	driverName := dp.Node().Driver.Factory().Name()
+	return attribute.String("backend", driverName)
+}
+
+// buildIDAttribute is a utility to retrieve the build id attribute from the solve options.
+// This value should be consistent between builds.
+func buildIDAttribute(so *client.SolveOpt) attribute.KeyValue {
```
tonistiigi, line 1552: I think this is more like "Identity" not "Identifier", so should not be called "ID".

build/build.go
```diff
@@ -686,6 +690,14 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 						return err
 					}
 
+					record := metrics.Measure(ctx, "build.duration", "Measures the total build duration.",
```
tonistiigi, line 693: I think it doesn't make sense track this as "buildkit solve request" duration, per build ref. For buildx side, I would think the duration is for how long buildx command ran. What about all the other error returns that make buildx fail? There is a return in between these current start-end calls as well. We can discuss more, but I think we are interested in commands, not requests.

internal/metrics/util.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+const TimeUnit string = "ms"
+
+// RecordFunc is used to record the measurement from Measure.
+type RecordFunc func(ctx context.Context, opts ...metric.RecordOption)
+
+// Measure is a utility for measuring a time duration with certain attributes.
+func Measure(ctx context.Context, name, desc string, opts ...metric.RecordOption) RecordFunc {
+	histogram, err := Meter(ctx).Int64Histogram(name,
```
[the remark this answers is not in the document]
jsternberg: This is mostly just how these things get represented in OTEL. The metric that's being stored is a histogram, but the histogram only contains one point. If this histogram was aggregated with other histograms, you would get a more traditional histogram.

build/build.go
jsternberg: I think we need to be able to correlate the same parameters between different machines. I believe the node identifier is unique between machines. Would there be a different salt parameter we could use for this purpose that would still allow us to match build identities across machines?

internal/env/env.go
jsternberg: Moved.

internal/metrics/metrics.go
```diff
@@ -27,40 +40,61 @@ const (
 // desired endpoint. It should be invoked on application shutdown.
 type ReportFunc func()
 
-// MeterProvider returns a MeterProvider suitable for CLI usage.
+// Initialize returns a context.Context with a MeterProvider suitable for CLI usage.
 // The primary difference between this metric reader and a more typical
 // usage is that metric reporting only happens once when ReportFunc
 // is invoked.
-func MeterProvider(cli command.Cli) (metric.MeterProvider, ReportFunc, error) {
+func Initialize(ctx context.Context, cli command.Cli) (context.Context, ReportFunc, error) {
```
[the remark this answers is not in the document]
jsternberg: Changed this package a bunch and moved most of the functionality to `otel/sdk/metric`. This is mostly to match with the structure of otel's libraries. The otel sdk is mostly meant to configure the more generic otel interfaces so we can wrap the otel sdk usage within there for convenience.

internal/env/env.go
```diff
@@ -0,0 +1,15 @@
+package env
```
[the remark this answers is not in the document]
jsternberg: Removed.

internal/metrics/util.go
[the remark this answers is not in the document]
jsternberg: Changed this to a counter anyway to reduce confusion. We can go back on this while we sort stuff out but it shouldn't matter.

internal/metrics/util.go
```diff
@@ -0,0 +1,33 @@
...
+	"go.opentelemetry.io/otel/metric"
+)
+
+const TimeUnit string = "ms"
+
+// RecordFunc is used to record the measurement from Measure.
+type RecordFunc func(ctx context.Context, opts ...metric.RecordOption)
+
+// Measure is a utility for measuring a time duration with certain attributes.
+func Measure(ctx context.Context, name, desc string, opts ...metric.RecordOption) RecordFunc {
```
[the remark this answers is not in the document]
jsternberg: I didn't end up finding this function useful on future changes to include more metrics so I've removed this function entirely.

internal/metrics/metrics.go
jsternberg: Removed in favor of just passing it through.

internal/metrics/metrics.go
jsternberg: I needed to add new code to include the service id in the resource but didn't want to modify buildkit for that, but I wanted to reuse the logic for the default service ID so I duplicated the code so the two were no longer linked when they had different outputs.

build/build.go
jsternberg: Changed.

build/build.go
jsternberg: Updated to use sha256.

build/build.go
jsternberg: Done.

build/build.go
jsternberg: I tried moving the location but it becomes difficult to get the parameters we need the further you move up the stack and it ends up relying on duplication between the `runControllerBuild` and `runBasicBuild` sections if I move it.

If you have a better location where this might fit, I can try to move it. I've now linked it with a progress writer because future metrics are going to read from the progress stream more directly so I think this location makes sense in that context.

jsternberg:
I've split out part of this into #2224 and also did a command-level version of this type of metric in #2225.

I'll rebase this one and keep it updated in case this is the metric we want. It's also fine if we merged all of the above.

jsternberg:
Converting this to a draft while I work on updating the PR.

jsternberg:
Just going to close this and maybe keep the branch local. I think we're opting not to go this direction right now. The code this is based on has also changed a bit for some other PRs related to this and it would probably be best to integrate this logic with those rather than try to rebase.

PR #2187, opened 2024-01-11
driver: only initialize the client once in the docker and remote drivers

jsternberg:
The docker and remote drivers both use the client internally and
therefore re-initialize the client multiple times within themselves. We
do some work to ensure the client is only initialized once through the
driver handle, but these drivers end up initializing the client multiple
times erroneously.

This updates these drivers to have their own caching code for the client
to prevent reinitialization which also has the side effect of fixing a
bug where traces were duplicated and sent multiple times.

### DIFF
--- a/driver/docker/driver.go
+++ b/driver/docker/driver.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/docker/buildx/driver"
 	"github.com/docker/buildx/util/progress"
+	"github.com/docker/buildx/util/syncutil"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/util/tracing/detect"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


tonistiigi (changes requested):
CI failures seem to be related. We would need some reference counting and possibly protection for context cancellation.

jsternberg:
I'm going to try to fix this on the client side, but I suspect that this will probably be easier to fix in two different ways:

1. Fix the underlying tracing issue where each time a client is initialized it adds it to the global list of tracers.
2. Perform the dedup on the buildkit side. Since it's typically the same client sending the traces and they're likely very close to each other, it would be pretty easy to perform the dedup in a memory-conscious way.

Although it might be as simple as just add reference counting to the close call. I guess I'll see after I try that.

jsternberg:
Going to close this because I don't think this PR works properly and it's no longer the way that I think this should be done. #2362 is a better solution that should be easier to refactor and maintain.

PR #2188, opened 2024-01-12
Mark experimental flags in `--help`

AkihiroSuda:
Prior to this commit, experimental flags were not distinguishable from regular flags in `--help`

```console
$ BUILDX_EXPERIMENTAL=1 docker buildx build --help

Start a build

Aliases:
  docker buildx build, docker buildx b

Options:
      --add-host strings              Add a custom host-to-IP mapping (format: "host:ip")
      --allow strings                 Allow extra privileged entitlement (e.g., "network.host", "security.insecure")
      --annotation stringArray        Add annotation to the image
      --attest stringArray            Attestation parameters (format: "type=sbom,generator=image")
      --build-arg stringArray         Set build-time variables
      --build-context stringArray     Additional build contexts (e.g., name=path)
      --builder string                Override the configured builder instance
      --cache-from stringArray        External cache sources (e.g., "user/app:cache", "type=local,src=path/to/dir")
      --cache-to stringArray          Cache export destinations (e.g., "user/app:cache", "type=local,dest=path/to/dir")
      --cgroup-parent string          Set the parent cgroup for the "RUN" instructions during build
      --detach                        Detach buildx server (supported only on linux) (EXPERIMENTAL)
  -f, --file string                   Name of the Dockerfile (default: "PATH/Dockerfile")
      --iidfile string                Write the image ID to the file
      --label stringArray             Set metadata for an image
      --load                          Shorthand for "--output=type=docker"
      --metadata-file string          Write build result metadata to the file
      --network string                Set the networking mode for the "RUN" instructions during build (default "default")
      --no-cache                      Do not use cache when building the image
      --no-cache-filter stringArray   Do not cache specified stages
  -o, --output stringArray            Output destination (format: "type=local,dest=path")
      --platform stringArray          Set target platform for build
      --print string                  Print result of information request (e.g., outline, targets) (EXPERIMENTAL)
      --progress string               Set type of progress output ("auto", "plain", "tty"). Use plain to show container output (default "auto")
      --provenance string             Shorthand for "--attest=type=provenance"
      --pull                          Always attempt to pull all referenced images
      --push                          Shorthand for "--output=type=registry"
  -q, --quiet                         Suppress the build output and print image ID on success
      --root string                   Specify root directory of server to connect (EXPERIMENTAL)
      --sbom string                   Shorthand for "--attest=type=sbom"
      --secret stringArray            Secret to expose to the build (format: "id=mysecret[,src=/local/secret]")
      --server-config string          Specify buildx server config file (used only when launching new server) (EXPERIMENTAL)
      --shm-size bytes                Size of "/dev/shm"
      --ssh stringArray               SSH agent socket or keys to expose to the build (format: "default|<id>[=<socket>|<key>[,<key>]]")
  -t, --tag stringArray               Name and optionally a tag (format: "name:tag")
      --target string                 Set the target build stage to build
      --ulimit ulimit                 Ulimit options (default [])
```


util/cobrautil/cobrautil.go
```diff
@@ -12,3 +16,38 @@ func HideInheritedFlags(cmd *cobra.Command, hidden ...string) {
 		_ = cmd.Flags().MarkHidden(h)
 	}
 }
+
+const AnnotationExperimentalCLI = "experimentalCLI"
```
crazy-max, line 20: I think we can keep this const private

AkihiroSuda: fixed

crazy-max (approved):
LGTM thanks! PTAL @thaJeztah

thaJeztah:
Wondering; are we also annotating these when rendering the docs at docs.docker.com (mostly curious if this would now show "experimental" twice).

crazy-max:
> Wondering; are we also annotating these when rendering the docs at docs.docker.com (mostly curious if this would now show "experimental" twice).

Hum good point, we already carry the "experimentaCLI" annotation in the yaml definition. We might just want to display "(EXPERIMENTAL)" in the cli then. @dvdksn I think in the docs website we display a blockquote for experimental commands right?

dvdksn:
yes we have a callout at the top of the page https://docs.docker.com/engine/reference/commandline/buildx_debug_build/

util/cobrautil/cobrautil.go
```diff
@@ -12,3 +16,38 @@ func HideInheritedFlags(cmd *cobra.Command, hidden ...string) {
...
+
+func MarkFlagExperimental(f *pflag.Flag) {
+	if _, ok := f.Annotations[annotationExperimentalCLI]; ok {
+		return
+	}
+	if f.Annotations == nil {
+		f.Annotations = make(map[string][]string)
+	}
+	f.Annotations[annotationExperimentalCLI] = nil
+	f.Usage += " (EXPERIMENTAL)"
```
dvdksn, line 30: I wonder if it would be better to handle this with a custom help template or function, rather than augmenting the cmd.Usage
https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#defining-your-own-help

AkihiroSuda: Maintenance cost can be higher if we switch away from the default template

AkihiroSuda:
Removed " (EXPERIMENTAL)" from yamls

tonistiigi (approved):

AkihiroSuda:
Can we merge?

dvdksn (approved):

PR #2190, opened 2024-01-12
update to go1.21.6

thaJeztah:


tonistiigi (approved):
@crazy-max Can we go back to feature release tags? The actual version can be looked up from provenance if needed.

crazy-max:
> @crazy-max Can we go back to feature release tags? The actual version can be looked up from provenance if needed.

Yes I think that's fine.

@thaJeztah Ok to switch to `1.21`?

thaJeztah:
I can change, but wondering what the big issue is; we pin all dependencies by version (so have to maintain those versions), wondering why Go has to be an exception for that.

crazy-max:
A patch releases signal changes that don't affect the public API. It guarantees backward compatibility and stability. But a security issue in the specification or implementation may come to light whose resolution requires breaking compatibility in a patch release though. So yeah I see what you mean.

PR #2194, opened 2024-01-15
Add support for DSSE envelope for attestation in imagetools

LaurentGoderre:
When running `docker buildx imagetools inspect --format '{{ JSON .SBOM }}` and the attestations use the `application/vnd.in-toto+json` mime type, the In-TOTO predicate is returned. However, then the attestation are wrapped in a DSSE envelope using the `application/vnd.in-toto.spdx+dsse` mime type, the same command returns and empty object (`{}`).

The changes add support for the DSSE mime type and base64 decode the envelope and returnb the payload object to stay consistent with images that do not use a wraper.

### DIFF
--- a/util/imagetools/loader.go
+++ b/util/imagetools/loader.go
@@ -4,7 +4,9 @@ package imagetools
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
@@ -20,6 +22,8 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
 )
+
+const inTotoGenericMime = "application/vnd.in-toto+json"
 
 var (
 	annotationReferences = []string{
@@ -274,7 +278,7 @@ type sbomStub struct {
 }
 
 func (l *loader) scanSBOM(ctx context.Context, fetcher remotes.Fetcher, r *result, refs []digest.Digest, as *asset) error {
-	ctx = remotes.WithMediaTypeKeyPrefix(ctx, "application/vnd.in-toto+json", "intoto")
+	ctx = remotes.WithMediaTypeKeyPrefix(ctx, inTotoGenericMime, "intoto")
 	as.deferredSbom = func() (*sbomStub, error) {
 		var sbom *sbomStub
 		for _, dgst := range refs {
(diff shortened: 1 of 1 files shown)


util/imagetools/loader.go
```diff
@@ -283,7 +284,9 @@ func (l *loader) scanSBOM(ctx context.Context, fetcher remotes.Fetcher, r *resul
 				return nil, errors.Errorf("referenced image %s not found", dgst)
 			}
 			for _, layer := range mfst.manifest.Layers {
-				if layer.MediaType == "application/vnd.in-toto+json" && layer.Annotations["in-toto.io/predicate-type"] == "https://spdx.dev/Document" {
+				if (layer.MediaType == "application/vnd.in-toto+json" ||
+					layer.MediaType == "application/vnd.in-toto.spdx+dsse") &&
```
crazy-max, line 288: Do you have an image we can look at with this mime type?

LaurentGoderre: [docker.io/jamescarnegie559/kms-test:latest](http://docker.io/jamescarnegie559/kms-test:latest)

util/imagetools/loader.go
```diff
@@ -292,6 +295,23 @@ func (l *loader) scanSBOM(ctx context.Context, fetcher remotes.Fetcher, r *resul
...
+
+					if layer.MediaType == "application/vnd.in-toto.spdx+dsse" {
+						var dsse struct {
+							Payload string `json:"payload"`
+						}
+						if err := json.Unmarshal(dt, &dsse); err != nil {
+							return nil, err
+						}
+
+						decoded, err := base64.StdEncoding.DecodeString(dsse.Payload)
```
tonistiigi, line 307: Should the payload be trusted if signatures (format) have not been verified?

tonistiigi (changes requested):
I think this shouldn't be SBOM specific and if we support this wrapper it should be same for provenance as well. Hopefully most of the code could be shared.

LaurentGoderre:
@tonistiigi is the provenance signed as well?

tonistiigi:
If you are talking about attestations signed by buildkit then none of them are signed atm. Afaics these are for inspecting images from custom tooling. If SBOM can be wrapped with DSSE in this case then I don't see why provenance wouldn't as well.

LaurentGoderre:
@tonistiigi the DSSE is used for signing the attestation so provenance should only need a DSSE envelope if it is signed as well.

LaurentGoderre:
@colinhemmings I am not sure if we are, I was asking.

util/imagetools/loader.go
mrjoelkamp: is there a mechanism in place to throw warnings for `imagetools inspect` commands? if so, adding a warning that the statements haven't been verified would help here

util/imagetools/loader.go
```diff
@@ -336,6 +358,23 @@ func (l *loader) scanProvenance(ctx context.Context, fetcher remotes.Fetcher, r
 					if err != nil {
 						return nil, err
 					}
+
+					if layer.MediaType == "application/vnd.in-toto.provenance+dsse" {
```
tonistiigi, line 362: Could we get this part into a shared function that is called by both provenance and SBOM code paths?

Also, define these mediatypes that are used multiple times as (private) const on top of the file.

LaurentGoderre:
@tonistiigi @whalelines I made more generic so it can be expanded to other DSSE formats.

util/imagetools/loader.go
mrjoelkamp: could potentially use `logrus` here to print a warning along the lines of `warning: sbom statements are not verified` until if/when this code checks signatures. I haven't seen many examples in the code base other than https://github.com/docker/buildx/blob/a6228ed78f3d591e56014d82c37185de4939e4c5/util/cobrautil/cobrautil.go#L37

tonistiigi (approved):

PR #2206, opened 2024-01-24
ci: test-unit job matrix for win/macos/ubuntu

crazy-max:
follow-up https://github.com/docker/buildx/pull/2205#issuecomment-1908035967

PR #2209, opened 2024-01-25
build: resolve 8.3 filename format to long one on Windows

crazy-max:
related to https://github.com/docker/buildx/pull/2206#issuecomment-1910309373

Converts Windows short pathnames to full pathnames (e.g. `C:\Users\ADMIN~1 --> C:\Users\Administrator`) so `vcs:localdir:*` and `com.docker.image.source.entrypoint` attrs are set in this case, otherwise: https://github.com/docker/buildx/actions/runs/7640538748/job/20815891223#step:5:419

Logic is taken from moby: https://github.com/moby/moby/blob/e8346c53d93d2cddf3d09910e166cdd8874070b3/integration-cli/utils_windows_test.go#L5-L27

### DIFF
--- a/build/git.go
+++ b/build/git.go
@@ -46,9 +46,9 @@ func getGitAttributes(ctx context.Context, contextPath string, dockerfilePath st
 	if filepath.IsAbs(contextPath) {
 		wd = contextPath
 	} else {
-		cwd, _ := os.Getwd()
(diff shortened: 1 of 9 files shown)


crazy-max:
closed by https://github.com/docker/buildx/pull/2206

PR #2198, opened 2024-01-17
vendor: github.com/moby/buildkit 6bd81372ad6f (master)

thaJeztah:
### vendor: github.com/containerd/containerd v1.7.12

full diff: https://github.com/containerd/containerd/compare/v1.7.11...v1.7.12

### vendor: github.com/google/go-cmp v0.6.0

full diff: https://github.com/google/go-cmp/compare/v0.5.9...v0.6.0

### vendor: github.com/google/uuid v1.5.0

full diff: https://github.com/google/uuid/compare/v1.3.1...v1.5.0

### vendor: golang.org/x/sys v0.16.0

full diff: https://github.com/golang/sys/compare/v0.15.0...v0.16.0

### vendor: golang.org/x/tools v0.14.0, golang.org/x/mod v0.13.0, golang.org/x/sync v0.4.0

full diff:

- https://github.com/golang/tools/compare/v0.10.0...v0.14.0
- https://github.com/golang/mod/compare/v0.11.0...v0.13.0
- https://github.com/golang/sync/compare/v0.3.0...v0.4.0

### vendor: github.com/moby/sys/mountinfo v0.7.1

full diff: https://github.com/moby/sys/compare/mountinfo/v0.6.2...mountinfo/v0.7.1

### vendor: github.com/docker/docker v25.0.1

- pkg/system: return even richer xattr errors

full diff: https://github.com/moby/moby/compare/v25.0.0-rc.1...v25.0.1

### vendor: github.com/docker/cli v25.0.1

- cli-plugins: move socket code into common package
- cli-plugins: don't use abstract sockets on macOS
    - fixes CLI leaving behind plugin socket mount-points
- socket: return from loop after EOF

full diff: https://github.com/docker/cli/compare/v25.0.0-rc.1...v25.0.1

### vendor: github.com/moby/buildkit 6bd81372ad6f (master)

- tests: implement NetNSDetached method

full diff: https://github.com/moby/buildkit/compare/6e200afad512...6bd81372ad6f27a8f3bf530e8cd91ea811f1f993

PR #2215, opened 2024-01-26
vendor: github.com/spf13/cobra v1.8.0

thaJeztah:
- follow-up to https://github.com/docker/buildx/pull/2198
- release notes: https://github.com/spf13/cobra/releases/tag/v1.8.0
- full diff: https://github.com/spf13/cobra/compare/v1.7.0...v1.8.0

Release notes highlights:

Features

- Support usage as plugin for tools like kubectl - this means that programs that utilize a "plugin-like" structure have much better support and usage (like for completions, command paths, etc.)
- Move documentation sources to site/content
- Add 'one required flag' group - this includes a new MarkFlagsOneRequired API for flags which can be used to mark a flag group as required and cause command failure if at least one is not used when invoked.
- Customizable error message prefix - This adds the SetErrPrefix and ErrPrefix APIs on the Command struct to allow for setting a custom prefix for errors
- feat: add getters for flag completions
- Feature: allow running persistent run hooks of all parents
- Improve API to get flag completion function

Bug fixes

- Fix typo in fish completions
- Fix grammar: 'allows to'
- powershell: escape variable with curly brackets
- Don't complete --help flag when flag parsing disabled
- Replace all non-alphanumerics in active help env var program prefix


crazy-max:
Ah they finally moved docs to a subdir!

image

thaJeztah:
Yes, they did!

We should also have a look at the "plugin" changes they added (if we can use those), the `SetErrPrefix` feature, and this one sounds potentially interesting also for our plug-ins use-case;

> Feature: allow running persistent run hooks of all parents

crazy-max:
> Yes, they did!
> 
> We should also have a look at the "plugin" changes they added (if we can use those), the `SetErrPrefix` feature, and this one sounds potentially interesting also for our plug-ins use-case;
> 
> > Feature: allow running persistent run hooks of all parents

Yeah we also need some integration tests, not just on linux but macos/windows to make sure there is no regression as follow-up of https://github.com/docker/buildx/pull/2206

crazy-max (approved):

tonistiigi (approved):

PR #2221, opened 2024-01-29
[0.12] Fix CVE-2023-47108, CVE-2023-45142, CVE-2023-48795

aepifanov:


thaJeztah:
It looks like the vendor validate is failing (might be needed to run `make vendor` for each commit to make sure vendoring is correct).

Also; we prefer using the `-X` option when cherry-picking, so that there's a reference to the original commit in master in the commit message ("cherry-picked from `<commit sha>`). This helps tracking where changes originate from, and to verify if all associated changes were included (sometimes commits may require other changes that don't immediately invalidate CI).

Separate from all of the above, I think some of these updates are preferred to be following from the BuildKit version used (I know at least grpc has been known to be sometimes subtly broken if versions don't match), but I'll leave that to buildx maintainers (cc @crazy-max @tonistiigi), and some of these CVEs were either false positives or no change in scope (considering the grpc API a privileged interface)

crazy-max:
> Separate from all of the above, I think some of these updates are preferred to be following from the BuildKit version used (I know at least grpc has been known to be sometimes subtly broken if versions don't match), but I'll leave that to buildx maintainers (cc @crazy-max @tonistiigi)

Yes indeed majority of them are preferred to be taken through BuildKit vendoring such as grpc and otel (specially hum): https://github.com/moby/buildkit/blob/master/go.mod

> and some of these CVEs were either false positives or no change in scope (considering the grpc API a privileged interface)

Yes agree with @thaJeztah, do you have details about how buildx is affected and the valid exploitation scenarios you found?

thaJeztah:
related to some of these;

- https://github.com/moby/moby/issues/47246#issuecomment-1916844668

colinhemmings:
Thank you for your contribution. After reviewing, we see that Buildx is unaffected by these issues. Introducing such a substantial change during a patch release introduces a significant test/stability risk, all for minimal value to the end user.
As such, we will not be including this change in Buildx v0.12

PR #2225, opened 2024-01-30
metrics: add build command duration metric

jsternberg:
This adds a build duration metric for the build command with attributes 
related to the buildx driver, the error type (if any), and which options 
were used to perform the build from a subset of the options.

### DIFF
--- a/build/utils.go
+++ b/build/utils.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"context"
 	"net"
-	"os"
 	"strings"
 
 	"github.com/docker/buildx/driver"
@@ -32,11 +31,6 @@ func IsRemoteURL(c string) bool {
 		return true
 	}
 	return false
-}
-
-func isLocalDir(c string) bool {
-	st, err := os.Stat(c)
-	return err == nil && st.IsDir()
 }
 
 func isArchive(header []byte) bool {
--- a/commands/build.go
+++ b/commands/build.go
@@ -3,8 +3,10 @@ package commands
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/base64"
 	"encoding/csv"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -13,6 +15,8 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/containerd/console"
 	"github.com/docker/buildx/build"
@@ -28,9 +32,11 @@ import (
 	"github.com/docker/buildx/store/storeutil"
 	"github.com/docker/buildx/util/buildflags"
 	"github.com/docker/buildx/util/cobrautil"
+	"github.com/docker/buildx/util/confutil"
 	"github.com/docker/buildx/util/desktop"
 	"github.com/docker/buildx/util/ioset"
 	"github.com/docker/buildx/util/metricutil"
+	"github.com/docker/buildx/util/osutil"
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/buildx/util/tracing"
 	"github.com/docker/cli-docs-tool/annotation"
@@ -52,6 +58,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
 	"google.golang.org/grpc/codes"
 )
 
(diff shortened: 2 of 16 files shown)


util/osutil/path_windows.go
```diff
@@ -1,10 +1,8 @@
-package build
+package osutil
 
-import "golang.org/x/sys/windows"
```
crazy-max, line 3: Looks missing

jsternberg: Good catch. I'll have to watch for that in the future. I didn't realize my IDE wasn't running `goimports` on files that didn't match my build constraints.

commands/build.go
```diff
@@ -279,6 +362,7 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
 		return err
 	}
 
+	done := timeBuildCommand(mp, options.Attributes(dockerCli))
```
tonistiigi, line 365: Make this `(build)metricsAttributes(b, dockerCli, options) attribute.Set`.

Don't need `options.newBuilder()` or other methods on `options`.

Inside the function, read the `b.Driver` directly from input without the struct/once. If you want then for the hash you can leave the lazy evaluation, but for just reading the driver property, there is no need for this overhead.

The call could also be inside `timeBuildCommand`.

commands/build.go
```diff
@@ -930,3 +1016,38 @@ func maybeJSONArray(v string) []string {
 	}
 	return []string{v}
 }
+
+// timeBuildCommand will start a timer for timing the build command. It records the time when the returned
+// function is invoked into a metric.
+func timeBuildCommand(mp metric.MeterProvider, attrs attribute.Set) func(err error) {
+	meter := metricutil.Meter(mp)
+	counter, _ := meter.Float64Counter("command.time",
+		metric.WithDescription("Measures the duration of the buildx command."),
```
tonistiigi, line 1025: "build command" seems more appropriate as looks specific to that command atm.

commands/build.go
```diff
@@ -930,3 +1016,38 @@ func maybeJSONArray(v string) []string {
...
+		counter.Add(context.Background(), dur,
+			metric.WithAttributeSet(attrs),
+			metric.WithAttributeSet(extraAttrs),
+		)
+	}
+}
+
+// errorType returns an attribute for the error type based on the error category.
+// If nil, this function returns an invalid attribute.
+func errorType(err error) string {
```
tonistiigi, line 1047: `otelErrorType/metricsErrorType` maybe to make it clearer what "type" this is.

commands/build.go
```diff
@@ -211,6 +219,88 @@ func (o *buildOptions) toDisplayMode() (progressui.DisplayMode, error) {
...
+		o.result = hex.EncodeToString(h.Sum(nil))
+	})
+	return o.result
+}
+
+// getVCSPath returns the vcs:source attribute for a context path.
+func getVCSPath(contextPath string) string {
+	var wd string
+	if filepath.IsAbs(contextPath) {
+		wd = contextPath
```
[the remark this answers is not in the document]
jsternberg: The `getGitAttributes` function in `build/git.go` doesn't seem to handle this case. Does that need to be fixed too? It seems to get passed these parameters directly.

This code will also just fail and ignore the error if the git commands don't succeed. Is that suitable or do we need to catch the URL?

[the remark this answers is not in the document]
tonistiigi: cc @crazy-max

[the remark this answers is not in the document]
jsternberg: I've updated this to include a `Stat` call before the logic. That should catch any paths that don't make sense before we start to try running git commands.

commands/build.go
jsternberg: Made the change. This isn't called inside of `timeBuildCommand` as there's another section of code that's going to use these same attributes in another PR.

commands/build.go
```diff
@@ -211,6 +219,61 @@ func (o *buildOptions) toDisplayMode() (progressui.DisplayMode, error) {
...
+type buildOptionsHash struct {
+	*buildOptions
+	result     string
+	resultOnce sync.Once
+}
+
+func (o *buildOptionsHash) String() string {
+	o.resultOnce.Do(func() {
+		target := o.target
+		contextPath := o.contextPath
```
tonistiigi, line 242: These paths need to be normalized so they always generate same hash(Eg. made absolute).

As this is leaking the system paths and Git info it still needs the salt like https://github.com/docker/buildx/pull/2185#discussion_r1465893251 . The multi-machine case doesn't apply anyway as system paths are different, and I don't think there is a way to implement it safely with some other way.

commands/build.go
```diff
@@ -211,6 +219,61 @@ func (o *buildOptions) toDisplayMode() (progressui.DisplayMode, error) {
...
+		o.result = hex.EncodeToString(h.Sum(nil))
+	})
+	return o.result
+}
+
+// getGitRemoteURL returns the git remote url attribute for a context path.
+func getGitRemoteURL(contextPath string) string {
+	// Check if this is a valid path to begin with before attempting
+	// to run git. It might be a non-filesystem URL.
+	if _, err := os.Stat(contextPath); err != nil {
```
tonistiigi, line 260: I think we should do the URL/- detection to match `LoadInputs` and unless I'm missing something, same validation would need to be called for the `vcs` codepath.

If you think this is outside of scope, then create a follow-up issue for it and assign to yourself.

commands/build.go
jsternberg: Since this code is in a new place, I don't have easy access to the node identifier anymore. I'd have to make that function public.

commands/build.go
jsternberg: I took a look at `LoadInputs` and it checks if the path is a local directory before checking for a URL. It seems to use this same code for that same functionality so I just exposed it and used that function instead. I've also included an explicit check for `-` to avoid attempting to call stat on that path and to avoid performing this logic at all in those circumstances.

We'll only do git remote url detection when it's a local directory.

commands/build.go
jsternberg: I've also now changed this so it will attempt to transform context path into a relative path. If a git directory is present, the relative path will be dependent on the root directory of the git repository.

build/utils.go
```diff
@@ -34,7 +34,7 @@ func IsRemoteURL(c string) bool {
 	return false
 }
 
-func isLocalDir(c string) bool {
+func IsLocalDir(c string) bool {
```
crazy-max, line 37: Maybe move this to `osutil`?

util/gitutil/gitutil.go
```diff
@@ -196,3 +197,10 @@ func stripCredentials(s string) string {
 	ru.User = nil
 	return ru.String()
 }
+
+func ToAbs(path string) string {
+	if !filepath.IsAbs(path) {
+		path, _ = filepath.Abs(filepath.Join(osutil.GetWd(), path))
+	}
+	return SanitizePath(path)
+}
```
crazy-max, line 206: Could be moved to `osutil` as well but needs to move `SanitizePath` there too, can be for a follow-up.

commands/build.go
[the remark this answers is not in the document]
crazy-max: > `contextPath` in here can be `-` or a URL.

Hum right this was never checked when getting Git attributes, I'm looking at it.

commands/build.go
crazy-max: @jsternberg Looks good

jsternberg:
This PR is now also semi-dependent on this PR here: https://github.com/docker/cli/pull/4875

At the current moment, we don't have a way of obtaining the alias that was used to launch buildx. That PR sets an environment variable so the plugin can find the original command name so we can include it in the metric.

jsternberg:
I've removed the section of this that's dependent on `docker/cli`. That PR seems like it's likely going to come to either a different method of passing the data or setting something like `OTEL_SERVICE_NAME` and including this information as part of the service name.

I don't want to prevent this PR from being merged while we hash that out so I've removed it.

tonistiigi (changes requested):
@jsternberg Check the windows CI

jsternberg:
Was missing an import on the Windows build because my setup doesn't auto perform imports for systems that aren't compiled on my platform (aka windows in this case). Fixed it by adding the correct import. I'll keep an eye out because some stuff got moved around and that commonly means imports get removed and readded in new places so hopefully there's no more missing areas.

tonistiigi (approved):

PR #2227, opened 2024-01-31
chore: ignore docker/docs deps with dependabot

crazy-max:
related to https://github.com/docker/buildx/pull/2222#issuecomment-1919092153

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,11 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    ignore:
+      # ignore this dependency
+      # it seems a bug with dependabot as pining to commit sha should not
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,6 +5,11 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    ignore:
+      # ignore this dependency
+      # it seems a bug with dependabot as pining to commit sha should not
+      # trigger a new version: https://github.com/docker/buildx/pull/2222#issuecomment-1919092153
+      - dependency-name: "docker/docs"
```
crazy-max, line 12: @glours You might need to do the same in compose

dvdksn (approved):

PR #2234, opened 2024-02-01
otel: create utility for storing common attributes for build options

jsternberg:
Adds a method to include common attributes for build related options
when tracking build invocations. These attributes use certain parameters
to the build command.

This also refactors some of the utility methods used by the git tool to
determine filepaths into its own separate package so they can be reused
in another place.

Also adds a test to ensure the resource is initialized correctly and
doesn't error. The otel handler logging message is suppressed on buildx
invocations so we never see the error if there's a problem with the
schema url. It's so easy to mess up the schema url when upgrading OTEL
that we need a proper test to make sure we haven't broken the
functionality.

### DIFF
--- a/build/git.go
+++ b/build/git.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/docker/buildx/util/gitutil"
+	"github.com/docker/buildx/util/osutil"
 	"github.com/moby/buildkit/client"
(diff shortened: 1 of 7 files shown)


tonistiigi:
I reviewed this as part of #2225 . This separately is hard to review as it adds functions that don't get called and it is not clear how they would be used.

PR #2235, opened 2024-02-01
metrics: measure context transfers for local source operations

jsternberg:
Measure the transfer size and duration of context transfers for various 
categories of local source transfers from the progress stream that's 
returned during the build.

Local source transfers are split into one of four categories:
* context
* dockerfile
* dockerignore
* namedcontext

Named contexts that are different names will be categorized under the same
metric.

### DIFF
--- a/util/progress/metricwriter.go
+++ b/util/progress/metricwriter.go
@@ -1,0 +1,151 @@
+package progress
+
+import (
+	"context"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/docker/buildx/util/metricutil"
+	"github.com/moby/buildkit/client"
+	"github.com/opencontainers/go-digest"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+type metricWriter struct {
+	recorders []metricRecorder
+	attrs     attribute.Set
+}
+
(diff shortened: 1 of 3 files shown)


jsternberg:
The first commit has been separated into its own PR: https://github.com/docker/buildx/pull/2234

util/progress/metricwriter.go
```diff
@@ -0,0 +1,150 @@
...
+		BaseAttributes: attrs,
+		State:          make(map[digest.Digest]*localSourceTransferState),
+	}
+	mr.TransferSize, _ = meter.Int64Counter("source.local.transfer.io",
+		metric.WithDescription("Measures the number of bytes transferred between the client and server for the context."),
+		metric.WithUnit("By"))
+
+	mr.Duration, _ = meter.Float64Counter("source.local.transfer.time",
+		metric.WithDescription("Measures the length of time spent transferring the context."),
+		metric.WithUnit("ms"))
```
jsternberg, line 74: These names are based on this section of the instrument naming document for OTEL: https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-naming

jsternberg:
@tonistiigi @crazy-max @daghack I believe this is ready for review. I've also added a check to only enable this when `BUILDX_EXPERIMENTAL` is used.

tonistiigi (approved):

PR #2238, opened 2024-02-02
vendor: github.com/containerd/console 8f6c4e4

crazy-max:
As discussed internally with @milas @laurazard @thaJeztah, the `containerd/console` terminal check was causing errors to get logged on Windows if the output was redirected (piping to another command with `more`).

```
> docker buildx build . | more
failed to get console mode for stdout: The handle is invalid.
failed to get console mode for stdout: The handle is invalid.
[+] Building 0.7s (2/3)                                          docker:default
 => => transferring dockerfile: 4.66kB                                     0.0s
 => resolve image config for docker.io/docker/dockerfile:1                 0.7s
 => [auth] docker/dockerfile:pull token for registry-1.docker.io           0.0s
```

Bump `github.com/containerd/console` to https://github.com/containerd/console/commit/8f6c4e4faef5a326d2cd907097d04c0239ee5e2f to include https://github.com/containerd/console/pull/66 that fixes this issue.


crazy-max:
We could test this with client-side integration tests on Windows related to https://github.com/docker/buildx/pull/2206#issuecomment-1912297483

crazy-max:
Ah this looks solved in https://github.com/containerd/console/pull/66. We would probably need a new release of `containerd/console`. cc @thaJeztah

thaJeztah:
@crazy-max would `github.com/moby/term` still be an alternative for this, or is that complicated?

crazy-max:
> @crazy-max would `github.com/moby/term` still be an alternative for this, or is that complicated?

No because BuildKit progress ui that buildx depends on also uses `containerd/console` so it still prints one failed to get console mode: https://github.com/moby/buildkit/blob/b09a9f1afef847b71aa01faf1b6ff9c019c75b1f/util/progress/progressui/display.go#L172

thaJeztah:
Ah, gotcha, thx.

We really need to look at some of these. I have a feeling we have multiple packages / implementations do very similar things

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/aws/aws-sdk-go-v2/config v1.18.16
 	github.com/compose-spec/compose-go/v2 v2.0.0-rc.3
-	github.com/containerd/console v1.0.3
+	github.com/containerd/console v1.0.4-0.20230706203907-8f6c4e4faef5
```
thaJeztah, line 9: Perhaps add a comment here pointing to the reason we have to use this version, e.g.;

```suggestion
	github.com/containerd/console v1.0.4-0.20230706203907-8f6c4e4faef5 // include https://github.com/containerd/console/pull/66, which is not yet in a release.
```

thaJeztah (approved):
LGTM

left a suggestion (adding a comment to provide context)

PR #2244, opened 2024-02-05
Upgrade buildkit to v0.12.5

kushmansingh:
Contains important security fixes https://github.com/moby/buildkit/releases/tag/v0.12.5


Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG XX_VERSION=1.2.1
 ARG DOCKER_VERSION=25.0.1
 ARG GOTESTSUM_VERSION=v1.9.0
 ARG REGISTRY_VERSION=2.8.0
-ARG BUILDKIT_VERSION=v0.12.4
+ARG BUILDKIT_VERSION=v0.12.5
```
crazy-max, line 9: This version of BuildKit is only used for the integration tests, I don't think this is what you expect? Also the CVEs from the release notes https://github.com/moby/buildkit/releases/tag/v0.12.5 don't concern the client (buildx).

kushmansingh: Gotcha, I see the vendored version is already ahead anyway. Thanks for the pointer

crazy-max (approved):

PR #2252, opened 2024-02-09
don't print build details when progress is rawjson

ndeloof:
when a build is ran using `--progress=rawjson` output should only contain line-delimited json stream
`PrintBuildDetails` to display an additional human-readable message at the end of the build breaks parsing this stream

Another option would be to use stdout for json and stderr for such messages

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -138,7 +138,7 @@ func runBake(ctx context.Context, dockerCli command.Cli, targets []string, in ba
 			if err == nil {
 				err = err1
 			}
-			if err == nil && progressMode != progressui.QuietMode {
+			if err == nil && progressMode != progressui.QuietMode && progressMode != progressui.RawJSONMode {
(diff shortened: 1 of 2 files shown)


crazy-max (commented):
Looks good

Can you also add it to build command: https://github.com/docker/buildx/blob/d0177c6da31fba948b7f82d2e61fe952f35de5d1/commands/build.go#L297

We might as well skip printing image id / digest in raw json mode: https://github.com/docker/buildx/blob/d0177c6da31fba948b7f82d2e61fe952f35de5d1/commands/build.go#L300

Also skip printing warnings with this mode: https://github.com/docker/buildx/blob/d0177c6da31fba948b7f82d2e61fe952f35de5d1/commands/build.go#L756

crazy-max (approved):
LGTM thanks

tonistiigi (approved):

PR #2253, opened 2024-02-09
docs: use absolute links and update link targets

dvdksn:


dvdksn:
Docs upstream workflow will be broken until main is updated

crazy-max (approved):

tonistiigi (commented):
Is this absolutely needed? It breaks the links when viewing md files in github.

dvdksn:
Alternatively we patch cli-docs-tool to rewrite the links when generating yaml. @thaJeztah had the same concern. But that could be something for a follow-up I think.

thaJeztah:
> Alternatively we patch cli-docs-tool to rewrite the links when generating yaml. @thaJeztah had the same concern. But that could be something for a follow-up I think.

- related discussion here; https://github.com/docker/cli/pull/4866#discussion_r1484428260

crazy-max (approved):

PR #2254, opened 2024-02-10
chore: switch to LocalMounts implementation

crazy-max:
* follow-up https://github.com/moby/buildkit/pull/4583

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -219,7 +219,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 			if err != nil {
 				return nil, err
 			}
-			so, release, err := toSolveOpt(ctx, np.Node(), multiDriver, opt, gatewayOpts, configDir, w, docker)
+			so, release, err := toSolveOpt(ctx, np.Node(), multiDriver, opt, gatewayOpts, configDir, addVCSLocalDir, w, docker)
 			if err != nil {
 				return nil, err
 			}
@@ -228,9 +228,6 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 			}
 			for k, v := range gitattrs {
 				so.FrontendAttrs[k] = v
-			}
-			if addVCSLocalDir != nil {
-				addVCSLocalDir(so)
 			}
 			defers = append(defers, release)
 			reqn = append(reqn, &reqForNode{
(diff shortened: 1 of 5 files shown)


build/build.go
```diff
@@ -385,11 +386,13 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	}
 	defers = append(defers, releaseLoad)
 
-	if sharedKey := so.LocalDirs["context"]; sharedKey != "" {
-		if p, err := filepath.Abs(sharedKey); err == nil {
-			sharedKey = filepath.Base(p)
+	if lm, ok := so.LocalMounts["context"]; ok {
+		if sharedKey, ok := fsutil.GetRoot(lm); ok && sharedKey != "" {
```
tonistiigi, line 390: iiuc this can just be moved inside `LoadInputs` where there is direct access to the value before it is turned into `FS`.

crazy-max: There is also `vcs:localdir` to handle. Let me know if this implementation lgty.

tonistiigi: I think pass the callback that sets the `vcs:localdir` to `toSolveOpt` (and if needed then to `LoadInputs`)

build/opt.go
```diff
@@ -32,9 +32,49 @@ import (
...
+		so.LocalMounts = map[string]fsutil.FS{}
+	}
+	so.LocalMounts[name] = lm
+	if addVCSLocalDir != nil {
+		addVCSLocalDir(name, lm.root, so)
+	}
+	return nil
+}
+
+type fs struct {
```
tonistiigi, line 57: this wrapper doesn't seem to do much. It is only called once, stores a property and then same caller reads out the property. `setLocalMount` could just set the path directly.

tonistiigi:
I removed this from the milestone as I don't think it has a user-visible impact. Lets get this in right after the release.

build/opt.go
```diff
@@ -32,9 +32,33 @@ import (
 	"github.com/moby/buildkit/util/entitlements"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"github.com/tonistiigi/fsutil"
 )
 
-func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Options, bopts gateway.BuildOpts, configDir string, pw progress.Writer, docker *dockerutil.Client) (solveOpt *client.SolveOpt, release func(), err error) {
+type solveOpt struct {
```
tonistiigi, line 38: I'm not quite sure why this wrapper type needs to exist as it doesn't do anything else but embed the other type.

For the method, it could just be a function as well.

tonistiigi (commented):
If you want you can separate out the first commit. Makes reviewing easier.

tonistiigi (approved):

PR #2259, opened 2024-02-16
Implement ability to load images by default in non-Docker build drivers.

n-g:
This eases build driver migrations, as it allows aligning the default behavior.

See also https://docs.docker.com/build/drivers/

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -56,6 +56,7 @@ type Driver struct {
 	cgroupParent  string
 	restartPolicy container.RestartPolicy
 	env           []string
+	defaultLoad   bool
 }
 
 func (d *Driver) IsMobyDriver() bool {
@@ -423,6 +424,7 @@ func (d *Driver) Features(ctx context.Context) map[driver.Feature]bool {
 		driver.DockerExporter: true,
 		driver.CacheExport:    true,
 		driver.MultiPlatform:  true,
+		driver.DefaultLoad:    d.defaultLoad,
 	}
 }
 
--- a/driver/features.go
+++ b/driver/features.go
@@ -7,3 +7,5 @@ const DockerExporter Feature = "Docker exporter"
 
 const CacheExport Feature = "Cache export"
 const MultiPlatform Feature = "Multi-platform build"
+
+const DefaultLoad Feature = "Automatically load images to the Docker Engine image store"
--- a/tests/build.go
+++ b/tests/build.go
@@ -59,6 +59,7 @@ var buildTests = []func(t *testing.T, sb integration.Sandbox){
 	testBuildMultiExporters,
 	testBuildLoadPush,
 	testBuildSecret,
+	testBuildDefaultLoad,
 }
 
 func testBuild(t *testing.T, sb integration.Sandbox) {
@@ -748,6 +749,50 @@ COPY --from=build /token /
 		require.NoError(t, err)
 		require.Equal(t, token, string(dt))
 	})
+}
+
+func testBuildDefaultLoad(t *testing.T, sb integration.Sandbox) {
+	if !isDockerWorker(sb) {
+		t.Skip("only testing with docker workers")
+	}
+
+	tag := "buildx/build:" + identity.NewID()
+
+	var builderName string
+	t.Cleanup(func() {
+		if builderName == "" {
+			return
+		}
+
+		cmd := dockerCmd(sb, withArgs("image", "rm", tag))
+		cmd.Stderr = os.Stderr
+		require.NoError(t, cmd.Run())
+
+		out, err := rmCmd(sb, withArgs(builderName))
+		require.NoError(t, err, out)
+	})
+
+	out, err := createCmd(sb, withArgs(
+		"--driver", "docker-container",
+		"--driver-opt", "default-load=true",
+	))
+	require.NoError(t, err, out)
+	builderName = strings.TrimSpace(out)
+
+	dir := createTestProject(t)
+
+	cmd := buildxCmd(sb, withArgs(
+		"build",
+		fmt.Sprintf("-t=%s", tag),
+		dir,
+	))
+	cmd.Env = append(cmd.Env, "BUILDX_BUILDER="+builderName)
+	outb, err := cmd.CombinedOutput()
+	require.NoError(t, err, string(outb))
+
+	cmd = dockerCmd(sb, withArgs("image", "inspect", tag))
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
 }
 
 func createTestProject(t *testing.T) string {
--- a/build/build.go
+++ b/build/build.go
@@ -169,6 +169,10 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 	if noMobyDriver != nil && !noDefaultLoad() && noPrintFunc(opt) {
 		var noOutputTargets []string
 		for name, opt := range opt {
+			if noMobyDriver.Features(ctx)[driver.DefaultLoad] {
+				continue
+			}
+
 			if !opt.Linked && len(opt.Exports) == 0 {
 				noOutputTargets = append(noOutputTargets, name)
 			}
--- a/build/opt.go
+++ b/build/opt.go
@@ -162,10 +162,14 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	case 1:
 		// valid
 	case 0:
-		if nodeDriver.IsMobyDriver() && !noDefaultLoad() {
-			// backwards compat for docker driver only:
-			// this ensures the build results in a docker image.
-			opt.Exports = []client.ExportEntry{{Type: "image", Attrs: map[string]string{}}}
+		if !noDefaultLoad() {
+			if nodeDriver.IsMobyDriver() {
+				// backwards compat for docker driver only:
+				// this ensures the build results in a docker image.
+				opt.Exports = []client.ExportEntry{{Type: "image", Attrs: map[string]string{}}}
+			} else if nodeDriver.Features(ctx)[driver.DefaultLoad] {
+				opt.Exports = []client.ExportEntry{{Type: "docker", Attrs: map[string]string{}}}
+			}
 		}
 	default:
 		if err := bopts.LLBCaps.Supports(pb.CapMultipleExporters); err != nil {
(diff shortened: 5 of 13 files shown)


controller/build/build.go
```diff
@@ -133,6 +133,15 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in controllerapi.Build
 		}
 	}
 
+	// ENV-guarded ability load images by default if no other output is specified.
+	// This is useful when migrating from Docker buildx drivers to others, as it aligns the default behavior.
+	if build.DefaultLoad() && len(outputs) == 0 && len(opts.Tags) > 0 {
```
tonistiigi, line 138: why the `opts.Tags` check in here?

tonistiigi (commented):
This could also be a `driver-opt` maybe?

n-g:
> This could also be a `driver-opt` maybe?

Thanks! I like this path, as it allows finer granularity when selecting the new behavior (per driver vs per env).
From what I can tell, this is the first `driver-opt` applicable to multiple driver types.
I introduced a new feature, but lmk if there is a cleaner modeling path I could employ.

controller/build/build.go
n-g: My thinking is:
Loading an image from a remote driver can be slow as networking is involved.
So, I'd like to differentiate the case where I just want to check if a build passes (and populate the cache of the remote build driver) vs I tag the image so that I can reuse it later.

driver/docker-container/driver.go
```diff
@@ -44,18 +44,19 @@ type Driver struct {
...
+	memorySwap        opts.MemSwapBytes
+	cpuQuota          int64
+	cpuPeriod         int64
+	cpuShares         int64
+	cpusetCpus        string
+	cpusetMems        string
+	cgroupParent      string
+	restartPolicy     container.RestartPolicy
+	env               []string
+	defaultLoadTagged bool
```
crazy-max, line 59: I think `defaultLoad` is enough to be consistent with `BUILDX_NO_DEFAULT_LOAD` env var: https://docs.docker.com/build/building/variables/#build-tool-configuration-variables

n-g: Thanks! Only loading tagged images was supposed to be a latency optimization as a pull from a remote driver can be slow.
But I agree that consistency is important. Updated.

driver/features.go
```diff
@@ -7,3 +7,5 @@ const DockerExporter Feature = "Docker exporter"
 
 const CacheExport Feature = "Cache export"
 const MultiPlatform Feature = "Multi-platform build"
+
+const DefaultLoad Feature = "Automatically load images"
```
crazy-max, line 11:
```suggestion
const DefaultLoad Feature = "Automatically load images to docker store"
```

@dvdksn for copy :pray:

n-g:
@tonistiigi @crazy-max
Thanks for taking a look at this!
From what I can tell, I addressed all your suggestions.
Is there anything left to do to make this PR ready to merge?
Thanks!

driver/features.go
```diff
@@ -7,3 +7,5 @@ const DockerExporter Feature = "Docker exporter"
 
 const CacheExport Feature = "Cache export"
 const MultiPlatform Feature = "Multi-platform build"
+
+const DefaultLoad Feature = "Automatically load images to docker store"
```
dvdksn, line 11:
```suggestion
const DefaultLoad Feature = "Automatically load images to the Docker Engine image store"
```

n-g: done - thanks!

driver/features.go
```diff
@@ -7,3 +7,5 @@ const DockerExporter Feature = "Docker exporter"
 
 const CacheExport Feature = "Cache export"
 const MultiPlatform Feature = "Multi-platform build"
+
+const DefaultLoad Feature = "Automatically load images to the Docker Engine image store"
```
crazy-max, line 11: @dvdksn I think "Docker store" is enough no?

```suggestion
const DefaultLoad Feature = "Automatically load images to the Docker store"
```

crazy-max (commented):
Can you squash your commits please?

Also would be nice to have integration tests in https://github.com/docker/buildx/tree/master/tests if you can or can be follow-up

Thanks!

driver/features.go
dvdksn: "Docker store" was the name of an old publisher tool of Docker Hub so that term becomes a little unfortunate for that reason. In general we always use "image store", and I think that keeping "Docker Engine" in there makes it clearer that we're referring to the local content store for the daemon.

n-g:
> Can you squash your commits please?
> 
> Also would be nice to have integration tests in https://github.com/docker/buildx/tree/master/tests if you can or can be follow-up
> 
> Thanks!

Sure! 

I initially did not realize that not all workers are tested by default with `./hack/test`.
Updated the test now, and validated it with
`TESTFLAGS="-v --parallel=1 --timeout=30m --run=//worker=docker$" TEST_DOCKERD=1 TESTFLAGS_DOCKER="-v --parallel=1 --timeout=30m" ./hack/test` locally.

PTAL

tests/build.go
```diff
@@ -675,6 +676,69 @@ func testBuildLoadPush(t *testing.T, sb integration.Sandbox) {
...
+		require.NoError(t, err, out)
+	})
+
+	// TODO: use stable buildkit image when v0.13.0 released
+	out, err := createCmd(sb, withArgs(
+		"--driver", "docker-container",
+		"--buildkitd-flags=--allow-insecure-entitlement=network.host",
+		"--driver-opt", "network=host",
+		"--driver-opt", "default-load=true",
+		"--driver-opt", "image=moby/buildkit:v0.13.0-rc3",
```
crazy-max, line 712: not needed for this test iiuc

crazy-max (changes requested):

tests/build.go
n-g: Thx, simplified to only set the driver-opt that we're testing.

tests/build.go
```diff
@@ -675,6 +676,60 @@ func testBuildLoadPush(t *testing.T, sb integration.Sandbox) {
...
+
+		cmd := dockerCmd(sb, withArgs("image", "rm", target))
+		cmd.Stderr = os.Stderr
+		require.NoError(t, cmd.Run())
+
+		out, err := rmCmd(sb, withArgs(builderName))
+		require.NoError(t, err, out)
+	})
+
+	// TODO: use stable buildkit image when v0.13.0 released
```
crazy-max, line 706: remove this as well

tests/build.go
```diff
@@ -675,6 +676,60 @@ func testBuildLoadPush(t *testing.T, sb integration.Sandbox) {
...
+	cmd.Env = append(cmd.Env, "BUILDX_BUILDER="+builderName)
+	outb, err := cmd.CombinedOutput()
+	require.NoError(t, err, string(outb))
+
+	// test docker store
+	cmd = dockerCmd(sb, withArgs("image", "inspect", target))
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
+
+	// TODO: test metadata file when supported by multi exporters https://github.com/docker/buildx/issues/2181
```
crazy-max, line 730: remove this

n-g:
> @crazy-max Could you help me re-trigger the validation workflows?

Thanks!
Let me know if there is anything else I can help with here 🙏

crazy-max (dismissed):
LGTM thanks!

tests/build.go
```diff
@@ -675,6 +676,57 @@ func testBuildLoadPush(t *testing.T, sb integration.Sandbox) {
 	// TODO: test metadata file when supported by multi exporters https://github.com/docker/buildx/issues/2181
 }
 
+func testBuildDefaultLoad(t *testing.T, sb integration.Sandbox) {
+	if sb.Name() != "docker" {
+		t.Skip("skipping test for non-docker workers")
+	}
```
crazy-max, line 682: Use helper, see https://github.com/docker/buildx/commit/bcda1dbcdfafb36c558aa4a550bf14cbd7a93ef5

tests/build.go
```diff
@@ -675,6 +676,57 @@ func testBuildLoadPush(t *testing.T, sb integration.Sandbox) {
...
+
+	registry, err := sb.NewRegistry()
+	if errors.Is(err, integration.ErrRequirements) {
+		t.Skip(err.Error())
+	}
+	require.NoError(t, err)
+
+	target := registry + "/buildx/registry:" + identity.NewID()
```
crazy-max, line 690: Setting up registry is not needed for this tests, see https://github.com/docker/buildx/commit/4d012cfa1f3f9d9037483773919c355961fe38f1

crazy-max (changes requested):
It needs a rebase with changes introduced recently sorry :sweat_smile:
See also review, I tried to push changes to your branch but looks like you don't allow edits by maintainers. Here is my branch if you want to pick these changes: https://github.com/docker/buildx/compare/master...crazy-max:buildx:default-load-opt otherwise I can carry in another PR if you prefer. Thanks

n-g:
> It needs a rebase with changes introduced recently sorry 😅 See also review, I tried to push changes to your branch but looks like you don't allow edits by maintainers. Here is my branch if you want to pick these changes: [master...crazy-max:buildx:default-load-opt](https://github.com/docker/buildx/compare/master...crazy-max:buildx:default-load-opt) otherwise I can carry in another PR if you prefer. Thanks

Thanks!
Rebased and applied your suggestions.

From what I can tell, I cannot grant edit rights retroactively.
Feel free to ping me if you need any changes.

tests/build.go
```diff
@@ -750,6 +751,46 @@ COPY --from=build /token /
...
+
+	cmd := buildxCmd(sb, withArgs(
+		"build",
+		fmt.Sprintf("-t=%s", tag),
+		dir,
+	))
+	cmd.Env = append(cmd.Env, "BUILDX_BUILDER="+builderName)
+	outb, err := cmd.CombinedOutput()
+	require.NoError(t, err, string(outb))
+}
```
crazy-max, line 792: The `image inspect` test seems to have been removed: https://github.com/docker/buildx/compare/master...crazy-max:buildx:default-load-opt#diff-12855e5a65cd0dc084324f4808d159cf2ebd514a77c6eee4dba653e1a4e94470R793-R795

n-g: Yup, sorry, bad merge. Was juggling too many things.
Updated and tested locally 🤞

crazy-max (approved):
Thanks!

PR #2260, opened 2024-02-19
Add `buildx build --opt` (EXPERIMENTAL)

AkihiroSuda:
This allows opt-in to new features of BuildKit that are not officially exposed to buildx yet.


build/build.go
```diff
@@ -226,6 +228,9 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		AllowedEntitlements: opt.Allow,
 		SourcePolicy:        opt.SourcePolicy,
 	}
+	if opt.RawOpts != nil {
+		so.FrontendAttrs = opt.RawOpts
+	}
```
crazy-max, line 233: Are frontend attributes set here intended to be overridden after?

AkihiroSuda: Updated to prioritize `RawOpts` implicit opts

crazy-max:
Seems fine if we keep this flag under experimental. Don't think we should move it out in the future?

This could also be useful for bake (can be in follow-up).

If you could add a basic integration test in https://github.com/docker/buildx/blob/master/tests/build.go that checks a frontend attribute being set through opt. Maybe similar to https://github.com/docker/buildx/blob/97e8fa7aafab7887de24ebd55d6d2acb4709ca5d/tests/build.go#L325 that would set an annotation that overrides one set by `--annotation`?

AkihiroSuda:
Added tests and bake.

crazy-max:
> Failure seems unrelated? https://github.com/docker/buildx/actions/runs/7975611949/job/21774316421?pr=2260

Yes this is linked to https://github.com/docker/buildx/pull/2253 and will be fixed when changes are made on docs repo (cc @dvdksn)

tonistiigi:
I'm not sure if we want this. Buildx uses opinionated UX and strictly defined flags/types. Allowing to set raw buildkit values directly adds confusion and makes it easier to make mistakes - pass wrong values and get to undefined configuration combinations. Definitely not as `--opt` as the meaning is more "raw buildkit frontend config option".

AkihiroSuda:
> I'm not sure if we want this. Buildx uses opinionated UX and strictly defined flags/types. Allowing to set raw buildkit values directly adds confusion and makes it easier to make mistakes - pass wrong values and get to undefined configuration combinations. Definitely not as `--opt` as the meaning is more "raw buildkit frontend config option".

This will remain experimental forever, so this will not confuse non-expert users.

If this is still not acceptable as an arg, can we implement this as an env var that takes a JSON of `[]string`?

colinhemmings:
Hi there, I agree with Tonis on this one. I'm concerned about the user experience of enabling this kind of backdoor through to Buildkit. It's confusing for users and leads to a support overhead further down the line. I know this would be experimental, but people would start using it and then get annoyed when we removed it or stopped supporting it. We already get backlash from users regarding legacy builder functionality, which has been dropped from Buildkit.
If there are specific features/use cases that are not available via Buildx, we should consider and implement those. That way, all users benefit from them with proper tested support.

PR #2266, opened 2024-02-21
driver: set network.host entitlement by default for container drivers

### LINKED ISSUE #2255: [v0.13] allow network.host daemon entitlement by default in container drivers
### Description

Based on comments in https://github.com/moby/buildkit/pull/4524#issuecomment-1937917305 , there shouldn't be a need to set `--allowed-entitlemtns network.host` as buildkitd flags when builder runs in a container as the container network is already isolated. At least when `--driver-opt network` does not equal `host`.

crazy-max:
fixes #2255

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -26,6 +26,7 @@ import (
 	"github.com/google/shlex"
 	"github.com/moby/buildkit/util/progress/progressui"
 	"github.com/pkg/errors"
+	"github.com/spf13/pflag"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
This seems to be a potentially breaking change (from security perspective), and has to be documented?

builder/builder.go
```diff
@@ -642,3 +640,37 @@ func validateBuildkitEndpoint(ep string) (string, error) {
...
+		for _, e := range allowInsecureEntitlements {
+			if e == "network.host" {
+				hasNetworkHostEntitlement = true
+				break
+			}
+		}
+
+		if !hasNetworkHostEntitlement {
+			// always set network.host entitlement as container network is
+			// isolated for docker-container and kubernetes drivers
```
AkihiroSuda, line 670: In the comment could you explain the purpose of setting the entitlement by default?

tonistiigi:
> This seems to be a potentially breaking change (from security perspective), 

It is not. The default networking for build step containers if builder was in container was already host (meaning host inside the container, not host of machine) and will remain like this in v0.13. This was without setting any `--network=host`.

builder/builder.go
```diff
@@ -642,3 +640,37 @@ func validateBuildkitEndpoint(ep string) (string, error) {
...
+
+		var hasNetworkHostEntitlement bool
+		for _, e := range allowInsecureEntitlements {
+			if e == "network.host" {
+				hasNetworkHostEntitlement = true
+				break
+			}
+		}
+
+		if !hasNetworkHostEntitlement {
```
tonistiigi, line 668: I think this condition should be "if no insecure entitlements set by user"

crazy-max: you mean we should check for any insecure entitlements? So if `security.insecure` or `network.host` or whatever is set for `--allow-insecure-entitlement` then we should skip setting `--allow-insecure-entitlement=network.host`?

tonistiigi: yes, just to leave a way for users to manually configure entitlements as well to be as they want them, without us overwriting

tonistiigi (approved):

PR #2268, opened 2024-02-21
driver: make buildkitd "config" and "flags" names consistent

crazy-max:
I found it confusing to have `config` as driver opt instead of `buildkitd-config` so we are consistent with `buildkitd-flags`. Same for variables not aligned across our code base.

This is also relevant for #2270

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -255,7 +255,7 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	supportsAttestations := bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations")) && nodeDriver.Features(ctx)[driver.MultiPlatform]
 	if len(attests) > 0 {
 		if !supportsAttestations {
-			return nil, nil, errors.Errorf("attestations are not supported by the current buildkitd")
+			return nil, nil, errors.Errorf("attestations are not supported by the current BuildKit daemon")
 		}
 		for k, v := range attests {
 			so.FrontendAttrs["attest:"+k] = v
--- a/build/build.go
+++ b/build/build.go
@@ -252,17 +252,21 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 			attests[k] = *v
 		}
 	}
-	supportsAttestations := bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations")) && nodeDriver.Features(ctx)[driver.MultiPlatform]
+
+	supportAttestations := bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations")) && nodeDriver.Features(ctx)[driver.MultiPlatform]
 	if len(attests) > 0 {
-		if !supportsAttestations {
-			return nil, nil, errors.Errorf("attestations are not supported by the current BuildKit daemon")
+		if !supportAttestations {
+			if !nodeDriver.Features(ctx)[driver.MultiPlatform] {
+				return nil, nil, notSupported("Attestation", nodeDriver, "https://docs.docker.com/go/attestations/")
+			}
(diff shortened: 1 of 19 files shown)


crazy-max:
ci failure for docs-upstream workflow is not related: #2269

dvdksn (commented):
General: should it be `--buildkitd-x` or `--buildkit-x` ?

I don't think we would expect buildx to refer to buildkit as anything other than a daemon. So maybe the `d` can be removed. I think it would make it easier to remember.

crazy-max:
> General: should it be `--buildkitd-x` or `--buildkit-x` ?
> 
> I don't think we would expect buildx to refer to buildkit as anything other than a daemon. So maybe the `d` can be removed. I think it would make it easier to remember.

Hum I think `buildkitd` is the correct assumption as we are passing them to the daemon. Would also avoid deprecating `buildkitd-flags` but I agree for casual users `buildkit` is well-known.

build/build.go
```diff
@@ -255,7 +255,7 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 	supportsAttestations := bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations")) && nodeDriver.Features(ctx)[driver.MultiPlatform]
 	if len(attests) > 0 {
 		if !supportsAttestations {
-			return nil, nil, errors.Errorf("attestations are not supported by the current buildkitd")
+			return nil, nil, errors.Errorf("attestations are not supported by the current BuildKit daemon")
```
tonistiigi, line 258: Can we make this error better and say that containerd storage needs to be enabled for `!multiPlatform`. I've seen people confused by this. Basically different error messages for the case where attestations are not supported and when multi-platform is not supported.

build/build.go
```diff
@@ -252,17 +252,22 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
...
-	supportsAttestations := bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations")) && nodeDriver.Features(ctx)[driver.MultiPlatform]
+
+	supportAttestations := bopts.LLBCaps.Contains(apicaps.CapID("exporter.image.attestations")) && nodeDriver.Features(ctx)[driver.MultiPlatform]
 	if len(attests) > 0 {
-		if !supportsAttestations {
-			return nil, nil, errors.Errorf("attestations are not supported by the current buildkitd")
+		if !supportAttestations {
+			if !nodeDriver.Features(ctx)[driver.MultiPlatform] {
+				return nil, nil, errors.Errorf(`Attestations are not supported for the %s driver.
+Switch to a different driver, or turn on the containerd image store, and try again.`, nodeDriver.Factory().Name())
```
crazy-max, line 261: @dvdksn for copy :pray:

thaJeztah:
Should we keep the old flags as hidden / deprecated for 1 .. 2 releases so that we can print a meaningful error if someone is using these? (didn't fully read through the diff, so perhaps we have something for that).

build/build.go
dvdksn: Maybe use the `notSupported` func?

https://github.com/docker/buildx/blob/f258de4c853460f6035a98d778db5cef12c32c36/build/build.go#L1429-L1433

crazy-max:
> Ah, thanks! Are we planning to officially deprecate the old one? (in that case, we could consider printing a deprecation warning when used, which can be "non-fatal" for now, but made "fatal" in a future release).

Fine to keep it imo and I don't think the cli prints such warning atm for `docker run --net-alias ...` for example?

build/build.go
dvdksn: It's referred to in the note:

image

crazy-max: I can't read :man_shrugging:

tonistiigi (approved):

crazy-max:
Needs follow-up on https://github.com/crazy-max/docker-setup-buildx-action

Issue #2256, opened 2024-02-12
[v0.13] allow driver-opt to that sets bridge network as default


tonistiigi:
### Description

BuildKit v0.13 comes with support for bridge networking. It is not default in v0.13 but is planned to become default in v0.14 . There should be a easy way to configure builder to use bridge already in v0.13 so we can do some extended testing.

TBBle:
In my [brief testing](https://github.com/moby/buildkit/issues/4637), it seems that apart from a couple of missing binaries in the BuildKit image, `--buildkitd-flags "--oci-worker-net=bridge"` was sufficient to make the Bridge mode work, so this would just be a UX feature, I guess? And given that we want bridge mode to be the default in 0.14, a rather temporary UX feature at that.

Simply adding test-suite coverage for both host and bridge modes would be interesting, I guess. I assume we only have "auto" coverage, but haven't looked. (Bridge mode helpfully covers the CNI approach, huzzah for shared implementations.)

Is there a more-general use-case for exposing `--oci-worker-net` natively from buildx UX, given that it'd be possible (but probably not common...) that the user has used other `--buildkit-flags` to switch away from the OCI worker, and might expect such a UX feature to work there too?

_Widening the scope_ to also [smooth the use of CNI](https://docs.docker.com/build/buildkit/configure/#cni-networking) by providing the desired CNI config etc. at `docker buildx create` time and hiding the need to build a new image first would seem more valuable, depending on how often people actually do that; that's basically widening into a new feature request, I reckon, and if it doesn't already exist, that's a (non-)demand signal. (It's probably also not even feasible, as I suspect that is a chicken-and-egg problem anyway.)

PR #2275, opened 2024-02-22
driver: docs to set buildkitd network mode and add tests

crazy-max:
fixes https://github.com/docker/buildx/issues/2256
follow-up https://github.com/docker/buildx/pull/2270#discussion_r1498446323 and closes #2270
needs #2268

This is an alternative to #2270 to document how to set BuildKit daemon network mode using `--buildkitd-flags` instead of introducing a new flag.

### DIFF
--- a/tests/build.go
+++ b/tests/build.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"path"
 	"path/filepath"
@@ -15,6 +16,7 @@ import (
 	"github.com/containerd/containerd/platforms"
 	"github.com/containerd/continuity/fs/fstest"
 	"github.com/creack/pty"
+	"github.com/moby/buildkit/util/appdefaults"
 	"github.com/moby/buildkit/util/contentutil"
 	"github.com/moby/buildkit/util/testutil"
 	"github.com/moby/buildkit/util/testutil/integration"
@@ -48,6 +50,7 @@ var buildTests = []func(t *testing.T, sb integration.Sandbox){
 	testBuildOCIExportNotSupported,
 	testBuildMultiPlatformNotSupported,
 	testDockerHostGateway,
+	testBuildNetworkModeBridge,
 }
 
 func testBuild(t *testing.T, sb integration.Sandbox) {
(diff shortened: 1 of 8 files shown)


tests/build.go
```diff
@@ -432,3 +434,66 @@ RUN ping -c 1 buildx.host-gateway-ip.local
...
+			ip := net.ParseIP(strings.TrimSpace(string(dt)))
+			require.NotNil(t, ip)
+
+			if tt.name == "bridge" {
+				_, subnet, err := net.ParseCIDR("10.10.0.0/16") // default subnet for bridge network in buildkit: https://github.com/moby/buildkit/blob/489f5fc6a331530daa8a31fb6ed7ba223df8ecb9/cmd/buildkitd/main.go#L474
+				require.NoError(t, err)
+				require.True(t, subnet.Contains(ip))
+			}
+		})
+	}
```
crazy-max, line 498: @tonistiigi I use this kind of test cases to check the behavior to avoid this issue encountered in https://github.com/docker/buildx/pull/2270#discussion_r1498853203

tests/build.go
```diff
@@ -432,3 +434,66 @@ RUN ping -c 1 buildx.host-gateway-ip.local
...
+			require.NoError(t, err, string(outb))
+
+			dt, err := os.ReadFile(filepath.Join(dir, "ip.txt"))
+			require.NoError(t, err)
+
+			ip := net.ParseIP(strings.TrimSpace(string(dt)))
+			require.NotNil(t, ip)
+
+			if tt.name == "bridge" {
+				_, subnet, err := net.ParseCIDR("10.10.0.0/16") // default subnet for bridge network in buildkit: https://github.com/moby/buildkit/blob/489f5fc6a331530daa8a31fb6ed7ba223df8ecb9/cmd/buildkitd/main.go#L474
```
crazy-max, line 493: Opened https://github.com/moby/buildkit/pull/4676 to avoid hardcoded value here.

tests/build.go
tonistiigi: I guess this should be fine but the case I thought is to compare the values of host and bridge, to make sure bridge was applied and we don't get the default host even when the request did not set `--network`.

So instead of two buildkitd's with different network config, only one buildkitd with bridge mode and then test that in this buildkitd with `--network host` and without it provides a different ip/interfaces.

crazy-max: Updated test

tonistiigi (approved):

PR #2280, opened 2024-02-23
build: set record provenance in response

crazy-max:
follow-up discussion with @tonistiigi
related to https://github.com/docker/buildx/pull/1681

Previously we got `containerimage.buildinfo` in the metadata file but `buildinfo` has been removed in BuildKit 0.12 (https://github.com/moby/buildkit/pull/3582): https://github.com/moby/buildkit/blob/master/docs/deprecated.md#build-information

This was useful to see all sources that were used by the build with their exact versions and also the configuration that was passed to the build.

However with the provenance attestation supported since BuildKit v0.11, we have similar (and a "bit" more) information available.

With this change we set the provenance saved along the build record in the metadata file similar to `containerimage.buildinfo` that is named `buildx.build.provenance`. Will set minimal provenance by default but can be customized using the `BUILDX_METADATA_PROVENANCE` env var.

___

```
$ cd buildx/
$ BUILDX_METADATA_PROVENANCE=max docker buildx --builder builder bake binaries --metadata-file md.json
...
#19 copying files 56.29MB 0.3s done
#19 DONE 0.3s

#20 resolve build record provenance
#20 DONE 0.0s
```

[md.json](https://github.com/docker/buildx/files/14616063/md.json)

___

```
$ cd buildx/
$ BUILDX_METADATA_PROVENANCE=max docker buildx --builder builder bake binaries-cross --metadata-file md.json
...
#28 copying files darwin/amd64 63.37MB 2.8s done
#28 copying files darwin/arm64 63.25MB 2.8s done
#28 DONE 2.8s

#29 resolve build record provenance
#29 DONE 0.0s
```

[md.json](https://github.com/docker/buildx/files/14616088/md.json)

___

```
$ cd buildkit/
$ BUILDX_METADATA_PROVENANCE=max docker buildx --builder builder bake binaries --metadata-file md.json
...
#89 [integration-tests-base 18/18] COPY --link --from=binaries / /usr/bin/
#89 merging 1.8s done
#89 DONE 1.9s

#90 [integration-tests 1/1] COPY . .
#90 DONE 1.5s

#91 resolve build record provenance
#91 DONE 0.1s
```

[md.json](https://github.com/docker/buildx/files/14776789/md.json)

___


crazy-max:
This is an early draft. Provenance can be quite huge so we might need to set the provenance when the metadata file is requested.

tests/build.go
```diff
@@ -533,13 +534,19 @@ func testBuildRef(t *testing.T, sb integration.Sandbox) {
...
 	var md mdT
 	err = json.Unmarshal(dt, &md)
 	require.NoError(t, err)
 
 	require.NotEmpty(t, md.BuildRef)
+	require.NotEmpty(t, md.BuildProvenance)
+
+	var prv slsa02.ProvenancePredicate
+	require.NoError(t, json.Unmarshal([]byte(md.BuildProvenance), &prv))
+	require.Equal(t, "https://mobyproject.org/buildkit@v1", prv.BuildType)
```
crazy-max, line 549: `buildType` is hardcoded for testing atm as importing `github.com/moby/buildkit/solver/llbsolver/provenance` brings a lot of dependencies. I think we should move provenance types to a dedicated package on BuildKit.

build/build.go
```diff
@@ -816,6 +816,11 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 						rr.ExporterResponse[k] = string(v)
 					}
 					rr.ExporterResponse["buildx.build.ref"] = buildRef
+					if confutil.IsExperimental() && node.Driver.HistoryAPISupported(ctx) {
+						if err := setRecordProvenance(ctx, c, rr, so.Ref, pw); err != nil {
+							return err
+						}
```
crazy-max, line 822: This extra request might fail for remote builders in some rare cases like DBC. We could consider a throttle logic in this case and if it doesn't succeed, warn but not fail.

crazy-max:
Size of the metadata file can be an for builds with many targets when using bake. We could either split each target to its own metadata file such as:

```hcl
# docker-bake.hcl
group "default" {
  targets = ["db", "webapp-dev"]
}
target "db" {
  dockerfile = "Dockerfile.db"
  tags = ["docker.io/username/db"]
}
target "webapp-dev" {
  tags = ["docker.io/username/webapp"]
}
```

```
docker buildx bake --metadata-file md.json
```

would produce

```
md.db.json
md.webapp-dev.json
```

or we could make our own "docker, inc" struct with a minimal provenance similar to buildinfo but I don't really like it. Want to avoid drifting and stick to provenance.

crazy-max:
@tonistiigi Updated to strip `buildConfig` and `metadata` from provenance if `BUILDX_METADATA_PROVENANCE` sets to `min`. Mode `max` sets full provenance.

build/provenance.go
```diff
@@ -0,0 +1,198 @@
...
+	"golang.org/x/sync/errgroup"
+)
+
+func setRecordProvenance(ctx context.Context, c *client.Client, sr *client.SolveResponse, ref string, pw progress.Writer) error {
+	mode := confutil.MetadataProvenance()
+	if mode == confutil.MetadataProvenanceModeNone {
+		return nil
+	}
+	pw = progress.ResetTime(pw)
+	return progress.Wrap("resolve build record provenance", pw.Write, func(l progress.SubLogger) error {
```
tonistiigi, line 32: Users would not know what this would mean. It should be something like "writing metadata file" or "resolving provenance for metadata file".

build/provenance.go
```diff
@@ -0,0 +1,198 @@
...
+		bo := backoff.NewExponentialBackOff()
+		bo.MaxElapsedTime = 5 * time.Second
+		return backoff.Retry(func() error {
+			res, err := fetchProvenance(ctx, c, ref, mode, l)
```
tonistiigi, line 36: I'm not sure if the logging of blob pulls is needed here. These are(is) always small file so I think it is more noise than useful info.

build/provenance.go
```diff
@@ -0,0 +1,198 @@
...
+					mu.Unlock()
+					return nil
+				})
+			}
+		}
+	}
+	return out, eg.Wait()
+}
+
+func provenanceDigest(res *controlapi.BuildResultInfo) *digest.Digest {
```
tonistiigi, line 121: `lookupProvenance(..)  *ocispecs.Descriptor` better I think.

build/provenance.go
```diff
@@ -0,0 +1,198 @@
...
+	if mode == confutil.MetadataProvenanceModeMax {
+		return base64.StdEncoding.EncodeToString(dt), nil
+	}
+
+	var prv provenancetypes.ProvenancePredicate
+	if err := json.Unmarshal(dt, &prv); err != nil {
+		return "", errors.Wrapf(err, "failed to unmarshal provenance")
+	}
+
+	// remove buildConfig and metadata for minimal provenance
```
tonistiigi, line 140: Should be reset the whole invocation block as well. Atm looks some partial info that I don't think would be very useful.

The builder ID shows up as empty string as well. I think we can mark it omitempty in buildkit (we could also wrap the type in here and mark it omitempty here).

build/provenance.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		l.SetStatus(st)
+	}()
+
+	now := time.Now()
+	st.Started = &now
+	l.SetStatus(st)
+
+	store := proxy.NewContentStore(c.ContentClient())
+	ra, err := store.ReaderAt(ctx, ocispecs.Descriptor{
```
tonistiigi, line 175: Assuming we can drop the logging, `content.ReadBlob` helper can be used for this.

docs/reference/buildx_bake.md
```diff
@@ -90,6 +90,76 @@ $ docker buildx bake -f docker-bake.dev.hcl db webapp-release
...
+    },
+    "containerimage.digest": "sha256:6d9ac9237a84afe1516540f40a0fafdc86859b2141954b4d643af7066d598b74"
+  }
+}
+```
+
+> **Note**
+>
+> Build record [provenance](https://docs.docker.com/build/attestations/slsa-provenance/#provenance-attestation-example)
+> (`buildx.build.provenance`) is not included by default. Set the
```
tonistiigi, line 158: My thinking was that the min mode (materials) would be included by default and user can opt-in to `max` for full provenance or disable it.

build/provenance.go
```diff
@@ -0,0 +1,198 @@
...
+					}
+					prv, err := encodeProvenance(dt, mode)
+					if err != nil {
+						return err
+					}
+					mu.Lock()
+					if out == nil {
+						out = make(map[string]string)
+					}
+					out["buildx.build.provenance/"+platform] = prv
```
tonistiigi, line 111: This doesn't seem to work. I don't get provenance keys for multi-platform builds when testing.

One thing that is bit weird in here as well is that the metadata-file doesn't currently contain the image digest for a specific platform in multi-platform builds, so digest for the artifact that the provenance is for.

build/provenance.go
crazy-max: This is mainly to display provenance size in log output and also per platform for multi-platform builds so we could use it as metrics as follow-up of @jsternberg's work. Would also be useful for Builds view in Docker Desktop (Build timing charts).

docs/reference/buildx_bake.md
crazy-max: Sounds good

build/provenance.go
crazy-max: > This doesn't seem to work. I don't get provenance keys for multi-platform builds when testing.

What cmd did you invoke? It works for me (see second example in PR desc) with:

```
docker buildx --builder builder bake binaries-cross --metadata-file md.json
```

```json
{
  "binaries-cross": {
    "buildx.build.provenance/darwin/amd64": {
      "builder": {
        "id": ""
      },
      "buildType": "https://mobyproject.org/buildkit@v1",
      "materials": [
        {
          "uri": "pkg:docker/docker/dockerfile@1",
          "digest": {
            "sha256": "dbbd5e059e8a07ff7ea6233b213b36aa516b4c53c645f1817a4dd18b83cbea56"
          }
        },
        {
          "uri": "pkg:docker/golang@1.21-alpine?platform=linux%2Famd64",
          "digest": {
            "sha256": "ed8ce6c22dd111631c062218989d17ab4b46b503cbe9a9cfce1517836e65298a"
          }
        },
        {
          "uri": "pkg:docker/tonistiigi/xx@1.4.0?platform=linux%2Famd64",
          "digest": {
            "sha256": "0cd3f05c72d6c9b038eb135f91376ee1169ef3a330d34e418e65e2a5c2e9c0d4"
          }
        }
      ],
      "invocation": {
        "configSource": {},
        "parameters": {
          "frontend": "gateway.v0",
          "args": {
            "build-arg:BUILDKIT_CONTEXT_KEEP_GIT_DIR": "1",
            "cmdline": "docker/dockerfile:1",
            "source": "docker/dockerfile:1",
            "target": "binaries"
          },
          "locals": [
            {
              "name": "context"
            },
            {
              "name": "dockerfile"
            }
          ]
        },
        "environment": {
          "platform": "linux/amd64"
        }
      }
    },
    "buildx.build.provenance/darwin/arm64": {
      "builder": {
        "id": ""
      },
      "buildType": "https://mobyproject.org/buildkit@v1",
      "materials": [
        {
          "uri": "pkg:docker/docker/dockerfile@1",
          "digest": {
            "sha256": "dbbd5e059e8a07ff7ea6233b213b36aa516b4c53c645f1817a4dd18b83cbea56"
          }
        },
        {
          "uri": "pkg:docker/golang@1.21-alpine?platform=linux%2Famd64",
          "digest": {
            "sha256": "ed8ce6c22dd111631c062218989d17ab4b46b503cbe9a9cfce1517836e65298a"
          }
        },
        {
          "uri": "pkg:docker/tonistiigi/xx@1.4.0?platform=linux%2Famd64",
          "digest": {
            "sha256": "0cd3f05c72d6c9b038eb135f91376ee1169ef3a330d34e418e65e2a5c2e9c0d4"
          }
        }
      ],
      "invocation": {
        "configSource": {},
        "parameters": {
          "frontend": "gateway.v0",
          "args": {
            "build-arg:BUILDKIT_CONTEXT_KEEP_GIT_DIR": "1",
            "cmdline": "docker/dockerfile:1",
            "source": "docker/dockerfile:1",
            "target": "binaries"
          },
          "locals": [
            {
              "name": "context"
            },
            {
              "name": "dockerfile"
            }
          ]
        },
        "environment": {
          "platform": "linux/amd64"
        }
      }
    },
    "buildx.build.provenance/linux/amd64": {
      "builder": {
        "id": ""
      },
      "buildType": "https://mobyproject.org/buildkit@v1",
      ...
```

build/provenance.go
crazy-max: Hum looks like it's not marked as omitempty in https://github.com/in-toto/in-toto-golang/blob/2ef7537b66dab4620b440536d3065e5366701f01/in_toto/slsa_provenance/common/common.go#L9. On purpose?

crazy-max: Now wrapping directly in buildx as it seems fine for the metadata file case but don't think we should omitempty in BuildKit related to https://github.com/docker/buildx/pull/2280#discussion_r1559113780

build/provenance.go
crazy-max: > One thing that is bit weird in here as well is that the metadata-file doesn't currently contain the image digest for a specific platform in multi-platform builds, so digest for the artifact that the provenance is for.

Yes it doesn't seem we set them in exporter response, this is slightly related to https://github.com/moby/buildkit/issues/4532

build/provenance.go
tonistiigi: Still these are small file pulls from the daemon, not from a remote source like registry. There isn't much difference in these values just being part of the solverresponse map or pulled from another blob. Single line about writing metadata file should be enough I think.

build/provenance.go
tonistiigi: ReadBlob/ReaderAt takes a descriptor that can be used directly instead of rebuilding (a broken one) from a digest. Additionally I think the "Digest" part hurts readability as what this function really is doing is finding the provenance for a result. It does not return a digest for already resolved provenance descriptor for example. If you need a digest you can always just use `desc.Digest`

build/build.go
```diff
@@ -82,6 +82,7 @@ type Options struct {
 	Session      []session.Attachable
 	Linked       bool // Linked marks this target as exclusively linked (not requested by the user).
 	PrintFunc    *PrintFunc
+	MetadataFile string
```
tonistiigi, line 85: As the value itself is not used (afaics) this could be `WithProvenanceResponse bool` or smth

build/provenance.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/docker/buildx/util/progress"
+	controlapi "github.com/moby/buildkit/api/services/control"
+	"github.com/moby/buildkit/client"
+	provenancetypes "github.com/moby/buildkit/solver/llbsolver/provenance/types"
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+)
+
+type ProvenancePredicate struct {
```
tonistiigi, line 25: These structs can be private I think

build/provenance.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func setRecordProvenance(ctx context.Context, c *client.Client, sr *client.SolveResponse, ref string, pw progress.Writer) error {
+	mode := confutil.MetadataProvenance()
+	if mode == confutil.MetadataProvenanceModeOff {
+		return nil
+	}
+	pw = progress.ResetTime(pw)
+	return progress.Wrap("resolving provenance for metadata file", pw.Write, func(l progress.SubLogger) error {
+		bo := backoff.NewExponentialBackOff()
```
tonistiigi, line 41: Why is the backoff needed in here? Is there some timing issue on buildkit side?

util/confutil/metadata.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// MetadataProvenance returns the provenance mode to set in the metadata file
+func MetadataProvenance() MetadataProvenanceMode {
+	switch os.Getenv("BUILDX_METADATA_PROVENANCE") {
+	case "min":
+		return MetadataProvenanceModeMin
+	case "max":
+		return MetadataProvenanceModeMax
+	case "off":
```
tonistiigi, line 26: Do we have some precedence for "off" ? We do use "disabled" in `--attest`.

Additionally I think we should run it through `ParseBool` so that `=0` and `=false` work as well.

build/build.go
crazy-max: At first I used a bool but then I was thinking we could directly handle the metadata file in build package instead of command one so I switched to this type instead. But agree that it seems more explicit to use a bool for this case.

build/provenance.go
crazy-max: I forgot to remove this my bad :man_facepalming:

crazy-max:
@dvdksn Forgot to add docs-followup label on this one 🙈 for `BUILDX_METADATA_PROVENANCE` env var that should be added in https://docs.docker.com/build/building/variables/#build-tool-configuration-variables

PR #2250, opened 2024-02-07
driver: add npipe url scheme support

iankingori:
Buildkit on windows exposes an npipe endpoint (`npipe:////./pipe/buildkitd`) which can be used to create a remote builder for WCOW.

Testing this with a simple docker file gives the result below:

image

PR #2287, opened 2024-02-27
remote: use winio DialPipeContext for named pipes

iankingori:
This PR fixes an issue where the [Dialer ](https://pkg.go.dev/net?GOOS=windows#Dialer.DialContextl) in go does not support named pipes which the buildkit daemon on Windows uses. I've added a utility DialContext function that uses  [DialPipeContext](https://github.com/microsoft/go-winio/blob/008bc6ea439f15884bef0b52f2772190c382bf46/pipe.go#L255C6-L255C21) for npipe connections on windows.

### DIFF
--- a/driver/remote/driver.go
+++ b/driver/remote/driver.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/docker/buildx/driver"
+	util "github.com/docker/buildx/driver/remote/util"
 	"github.com/docker/buildx/util/progress"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/util/tracing/detect"
@@ -99,9 +100,8 @@ func (d *Driver) Dial(ctx context.Context) (net.Conn, error) {
 		return nil, errors.Errorf("invalid endpoint address: %s", d.InitConfig.EndpointAddr)
 	}
 
-	dialer := &net.Dialer{}
+	conn, err := util.DialContext(ctx, network, addr)
 
-	conn, err := dialer.DialContext(ctx, network, addr)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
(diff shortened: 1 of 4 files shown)


tonistiigi:
@iankingori Check the linter error and remove draft if ready.

iankingori:
Thanks @jedevc, I'll drop some more context on the next one. 

I'm creating an issue to get the remote driver tests for windows up and running

iankingori:
The test failure looks like a flake, previous runs passed and only change since was a linting error.

crazy-max (approved):

driver/remote/driver_windows.go
```diff
@@ -0,0 +1,46 @@
...
+
+	var conn net.Conn
+	var err error
+
+	// dial context doesn't support named pipes
+	if network == "npipe" {
+		conn, err = winio.DialPipeContext(ctx, addr)
+	} else {
+		conn, err = dialer.DialContext(ctx, network, addr)
+	}
```
jedevc, line 32: This is the only difference?

Can we split this into a helper function, and have that be `dialer_unix.go` and `dialer_windows.go`? We shouldn't duplicate all the rest of this code.

driver/remote/driver_windows.go
```diff
@@ -0,0 +1,46 @@
+//go:build windows
+// +build windows
```
jedevc, line 2: Also, don't think we need these build constraints (though we are inconsistent with it) - the filename should be the indicator in this case.

driver/remote/driver_windows.go
iankingori: Yeah, I expected more divergence between the two but seems that's all that was needed for now. Makes sense to have it as a helper function

jedevc (approved):

tianon-sso:
Can confirm this fixes the problem I had in https://github.com/docker/buildx/pull/2250#issuecomment-1965507461!!  Thanks all! :smile: :heart:

PR #2290, opened 2024-02-28
build: allow multiple exports if supported by buildkit

### LINKED ISSUE #2181: Enable multiple exporters
Follow-up to https://github.com/moby/buildkit/pull/4134 - with this merged, we should be able to remove the explicit checks that prevent using multiple exporters.

However, there's still the question of how this should interact with `--metadata-file` - see https://github.com/moby/buildkit/issues/4532 for more information. We could:
1. ignore this issue
2. block adding this functionality until resolving that upstream
3. don't allow users to use `--metadata-file` with multiple exporters until resolving that upstream (but note that the most recent buildx release has this issue with multiple cache exporters today)

tonistiigi:
fixes #2181

### DIFF
--- a/docs/reference/buildx_build.md
+++ b/docs/reference/buildx_build.md
@@ -424,6 +424,10 @@ $ docker buildx build -o type=docker,dest=- . > myimage.tar
 $ docker buildx build -t tonistiigi/foo -o type=registry
 ```
 
+> **Note **
+>
+> Since BuildKit v0.13.0 multiple outputs can be specified by repeating the flag.
+
 Supported exported types are:
 
 #### `local`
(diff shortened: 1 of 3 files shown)


docs/reference/buildx_build.md
```diff
@@ -424,6 +424,8 @@ $ docker buildx build -o type=docker,dest=- . > myimage.tar
 $ docker buildx build -t tonistiigi/foo -o type=registry
 ```
 
+Since BuildKit v0.13.0 multiple outputs can be specified by repeating the flag.
```
crazy-max, line 427:
```suggestion
> **Note **
>
> Since BuildKit v0.13.0 multiple outputs can be specified by repeating the flag.
```

crazy-max (approved):

crazy-max:
Came across https://github.com/docker/build-push-action/issues/727, I think we forgot the `push` / `load` case. I'm looking at it.

PR #2298, opened 2024-02-29
Add tests for imagetools inspect

LaurentGoderre:


mrjoelkamp (approved):
nice! LGTM

util/imagetools/imagetools_helpers_test.go
```diff
@@ -0,0 +1,399 @@
...
+}
+
+func getImageWithAttestation(t attestationType) *result {
+	manifestList := getBaseManifests()
+
+	objManifest := ocispec.Manifest{
+		MediaType: v1.MediaTypeImageManifest,
+		Layers:    getAttestationLayers(t),
+		Annotations: map[string]string{
+			"plaform": "linux/amd64",
```
tonistiigi, line 85: "platform"

tonistiigi (commented):
We can merge this but overall I think it is less likely that a breakage will happen in one of these private helper functions that is unit tested and more likely that something breaks in handling `inspect --format` input in relation to attestations and goes unnoticed. Also, as these are private functions, they don't have a strict contract and could be modified freely (what likely requires test modification as well).

For the integration tests that check the feature, these are atm in https://github.com/docker/buildx/blob/v0.13/tests/imagetools.go . As this PR also tests images that can't be built with buildx, it makes it trickier as for that the integration test would need to manually push such an image from raw manifest objects.

LaurentGoderre:
@tonistiigi I kept the test pretty high level so they shouldn't be high maintenance. Our team is also proposing a new way to do signature of SBOM so having these tests will be helpful down the line

tonistiigi (approved):

PR #2301, opened 2024-03-01
vendor: github.com/docker/cli-docs-tool to master

crazy-max:
Just trying latest changes of cli-docs-tool for man pages generation (https://github.com/docker/cli-docs-tool/pull/14) before creating a release. See last commit for generated files.

cc @dvdksn @thaJeztah


hack/update-docs
```diff
@@ -3,7 +3,7 @@
 set -eu -o pipefail
 
 : "${BUILDX_CMD=docker buildx}"
-: "${FORMATS=md}"
+: "${FORMATS=md,yaml,man}"
```
dvdksn, line 6: This would mean we vendor all formats. Totally fine by me, it alleviates a step in updating the CLI reference downstream (we only need to re-vendor and not manually generate and copy over yaml)

crazy-max: This is just for testing, I don't want to merge this :innocent:

dvdksn: aw I was getting excited :)

docs/reference/docker-buildx-create.1
```diff
@@ -0,0 +1,325 @@
...
+	Fixed platforms for current node
+
+.PP
+\fB--use\fP[=false]
+	Set the current builder instance
+
+
+.SH EXAMPLE
+.EX
+### Append a new node to an existing builder (--append) {#append}
```
dvdksn, line 81: Looks like these first headings in the examples section don't get the .SS markup

PR #2302, opened 2024-03-01
build: handle push/load shorthands for multi exporters

crazy-max:
fixes https://github.com/docker/build-push-action/issues/727

We forgot to handle `push` and `load` shorthands in https://github.com/docker/buildx/pull/2290.

~Keeping this one in draft while I'm working on integration tests.~

~EDIT: Argh GitHub is broken https://www.githubstatus.com/incidents/wcl1sw4mzg60~

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -72,11 +72,9 @@ func runBake(ctx context.Context, dockerCli command.Cli, targets []string, in ba
 
 	overrides := in.overrides
 	if in.exportPush {
-		if in.exportLoad {
-			return errors.Errorf("push and load may not be set together at the moment")
-		}
-		overrides = append(overrides, "*.push=true")
-	} else if in.exportLoad {
+		overrides = append(overrides, "*.output=type=registry")
+	}
+	if in.exportLoad {
 		overrides = append(overrides, "*.output=type=docker")
 	}
 	if cFlags.noCache != nil {
--- a/controller/build/build.go
+++ b/controller/build/build.go
@@ -99,37 +99,37 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in controllerapi.Build
 		return nil, nil, err
 	}
 	if in.ExportPush {
-		if in.ExportLoad {
-			return nil, nil, errors.Errorf("push and load may not be set together at the moment")
+		var pushUsed bool
+		for i := range outputs {
+			if outputs[i].Type == client.ExporterImage {
+				outputs[i].Attrs["push"] = "true"
+				pushUsed = true
+			}
 		}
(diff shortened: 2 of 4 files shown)


commands/bake.go
crazy-max, line 1: Adding integration tests for bake as well. Wants to make sure override behavior is as expected.

controller/build/build.go
```diff
@@ -99,38 +99,18 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in controllerapi.Build
...
-			}}
-		} else {
-			switch outputs[0].Type {
-			case "image":
-				outputs[0].Attrs["push"] = "true"
-			default:
-				return nil, nil, errors.Errorf("push and %q output can't be used together", outputs[0].Type)
-			}
-		}
+		outputs = append(outputs, client.ExportEntry{
```
tonistiigi, line 102: I think we still need to do specific behavior based on the current value. If the existing value has image exports then these should be set to be pushed.

Additionally, if we don't do this then it is now an error for older daemons that just defined the same output multiple times.

controller/build/build.go
```diff
@@ -99,38 +99,18 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in controllerapi.Build
...
-				Attrs: map[string]string{},
-			}}
-		} else {
-			switch outputs[0].Type {
-			case "docker":
-			default:
-				return nil, nil, errors.Errorf("load and %q output can't be used together", outputs[0].Type)
-			}
-		}
+		outputs = append(outputs, client.ExportEntry{
```
tonistiigi, line 110: Similarly, if docker output already exists then this should not do anything.

crazy-max: This is a bit more than that I think. `docker` exporter can also output to a file so we need to keep this case but also be able to output to docker store if `--load`:

```
$ docker buildx build --output type=docker --output type=docker,dest=build.tar
```

controller/build/build.go
```diff
@@ -99,36 +99,52 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in controllerapi.Build
 		return nil, nil, err
 	}
 	if in.ExportPush {
-		if in.ExportLoad {
-			return nil, nil, errors.Errorf("push and load may not be set together at the moment")
-		}
 		if len(outputs) == 0 {
```
tonistiigi, line 102: This case is not needed anymore. Else case behaves same way if `len(outputs) == 0`. Same for `docker`.

crazy-max: Ah right, we are already checking if push/load is already used.

tonistiigi (approved):

PR #2313, opened 2024-03-06
ci: fix docs release vendoring

dvdksn:

### DIFF
--- a/.github/workflows/docs-release.yml
+++ b/.github/workflows/docs-release.yml
@@ -26,7 +26,6 @@ jobs:
         name: Prepare
         run: |
           rm -rf ./data/buildx/*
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-release.yml
```diff
@@ -49,13 +48,22 @@ jobs:
         name: Copy yaml
         run: |
           cp /tmp/buildx-docs/out/reference/*.yaml ./data/buildx/
+      -
+        name: Make vendoring temp dir
+        run: |
+          mkdir -p /tmp/hugo_vendor_output
```
crazy-max, line 54: Maybe we could have a `make vendor` target upstream so we don't need to explicitly do this here and we keep it consistent?

PR #2271, opened 2024-02-21
metrics: measure image transfers for image source operations

jsternberg:
This measures the transfer size and duration for image pulls along with the
time spent extracting the image contents.

PR #2316, opened 2024-03-06
metrics: measure run operations for exec operations

jsternberg:
This measures the duration of exec operations. It does not factor in
whether the operation was cached or not so this should include the
amount of time to determine whether an operation was cached.

### DIFF
--- a/util/progress/metricwriter.go
+++ b/util/progress/metricwriter.go
@@ -24,6 +24,7 @@ func newMetrics(mp metric.MeterProvider, attrs attribute.Set) *metricWriter {
 		recorders: []metricRecorder{
 			newLocalSourceTransferMetricRecorder(meter, attrs),
 			newImageSourceTransferMetricRecorder(meter, attrs),
+			newExecMetricRecorder(meter, attrs),
 		},
 		attrs: attrs,
 	}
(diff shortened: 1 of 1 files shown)


daghack (approved):

util/progress/metricwriter.go
```diff
@@ -149,3 +150,40 @@ func detectLocalSourceType(vertexName string) attribute.KeyValue {
...
+		if v.Started == nil || v.Completed == nil || !detectExecType(v.Name) {
+			continue
+		}
+
+		dur := float64(v.Completed.Sub(*v.Started)) / float64(time.Millisecond)
+		mr.Duration.Add(context.Background(), dur, metric.WithAttributeSet(mr.Attributes))
+	}
+}
+
+var reExecType = regexp.MustCompile(`^\[.*] RUN `)
```
jedevc, line 185: It feels weird to rely on regex of the progress stream for this.

This will only work with the `dockerfile` frontend, since not all exec ops will necessarily begin with `RUN`.

jedevc: Likewise in https://github.com/docker/buildx/pull/2317, https://github.com/docker/buildx/pull/2271, etc.

tonistiigi (approved):

PR #2319, opened 2024-03-07
docs: fix link to new target in dockerfile reference

dvdksn:


crazy-max (approved):

crazy-max:
@dvdksn Strange that `docs-upstream` didn't catch that issue https://github.com/docker/docs/pull/19577#issuecomment-1983034260?

dvdksn:
@crazy-max it's because that change comes from moby/buildkit, so it wasn't in main yet for testing :/ This kind of three-way error detection is hard to get right I think. We don't have a validate job in buildkit yet which would solve this. But to get a green workflow you'd have to specify a custom buildx commit to validate against...

crazy-max:
Ah right yeah makes sense

PR #2307, opened 2024-03-05
tests: handle multiple docker versions

crazy-max:
* ~needs https://github.com/moby/buildkit/pull/4736~

PR #2321, opened 2024-03-07
dockerfile: use moby-bin and cli-bin images for docker binaries

crazy-max:
picked from #2307 

we can now use `moby/moby-bin` and `dockereng/cli-bin` images for testing docker in our integration tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,15 +3,18 @@
 ARG GO_VERSION=1.21
 ARG XX_VERSION=1.4.0
 
+# for testing
 ARG DOCKER_VERSION=25.0.2
 ARG GOTESTSUM_VERSION=v1.9.0
 ARG REGISTRY_VERSION=2.8.0
 ARG BUILDKIT_VERSION=v0.12.5
 
-# xx is a helper for cross-compilation
 FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
-
 FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS golatest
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,26 +20,11 @@ ENV GOFLAGS=-mod=vendor
 ENV CGO_ENABLED=0
 WORKDIR /src
 
+FROM moby/moby-bin:$DOCKER_VERSION AS docker-engine
+FROM dockereng/cli-bin:$DOCKER_VERSION AS docker-cli
```
thaJeztah, line 24: Do you think there's a need to call out that these are automated builds, not "official releases" (for the occasional reader)?


(We should also really fix those repositories to have a readme to outline their purpose / give the right expectations 😞 - let me see if someone can work on that)

crazy-max: Added "for testing" comment on args if that sgty.

thaJeztah (approved):
LGTM, yes, that comment should do for now. I created an internal ticket to look at adding README's to those repositories.

Dockerfile
```diff
@@ -3,15 +3,18 @@
 ARG GO_VERSION=1.21
 ARG XX_VERSION=1.4.0
 
+# for testing
 ARG DOCKER_VERSION=25.0.2
 ARG GOTESTSUM_VERSION=v1.9.0
 ARG REGISTRY_VERSION=2.8.0
 ARG BUILDKIT_VERSION=v0.12.5
```
thaJeztah, line 10: Looks like we have some versions to update here (probably fine to be done separate)

crazy-max: Yes done in https://github.com/docker/buildx/pull/2322

PR #2322, opened 2024-03-07
tests: matrix with buildkit versions

crazy-max:
* ~needs or closes #2341~

Update buildkit to 0.13.1 in our Dockerfile as well.


tests/integration.go
```diff
@@ -6,11 +6,14 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/containerd/continuity/fs/fstest"
 	"github.com/moby/buildkit/util/testutil/integration"
 	"github.com/stretchr/testify/require"
 )
 
+const defaultBuildKitVersion = "v0.13.0"
```
tonistiigi, line 15: this should be `buildx-stable-1`. 

If this works then we can add extra workflow that tests specific version, or master.

crazy-max: I need semver to properly match buildkit versions and skip related tests, see last commit: https://github.com/docker/buildx/pull/2322/files#diff-6db6da28b382b90cebe937a99fedeb6c8927535c220fd1c9a40f644d4ae90ee7R82-R92

I guess I could run the image to extract buildkitd version but that sounds a bit overkill.

tests/integration.go
```diff
@@ -71,7 +78,114 @@ func dockerCmd(sb integration.Sandbox, opts ...cmdOpt) *exec.Cmd {
...
+			cmd = exec.Command(filepath.Join(destDir, "usr", "bin", "buildkitd"), "--version")
+			out, err := cmd.CombinedOutput()
+			require.NoErrorf(t, err, "failed to get BuildKit version from %q: %s", buildkitImage, string(out))
+
+			v := strings.Fields(strings.TrimSpace(string(out)))
+			if len(v) != 4 {
+				require.Fail(t, "unexpected version format: "+strings.TrimSpace(string(out)))
+			}
+			ver = v[2]
+			bkvers[sb.Name()] = ver
```
crazy-max, line 167: Will use builder inspect output to get BuildKit version (specially needed for docker driver). For other drivers that don't output BuildKit version in inspect (e.g., remote), we are invoking buildkitd binary with `--version`.

tonistiigi (approved):
Looks good but needs rebase

PR #2327, opened 2024-03-09
remote: fix connhelpers with custom dialer

### LINKED ISSUE #2324: [0.13] remote driver with kube-pod URL is stuck at inactive
### Contributing guidelines

### Description

docker-buildx 0.13.0 could not detect the buildkit 0.13.0 instance on kubernetes via `kube-pod://` URL. It works well if I kubectl port-forward the service into local and use `tcp://127.0.0.1:1234`.

### Expected behaviour

docker buildx 0.13.0 supports remote driver with kube-pod as in 0.12.1

### Actual behaviour

Remote builder is stuck at inactive when using kube-pod:// URL

### Buildx version

github.com/docker/buildx v0.13.0 Homebrew

### Docker info

```text
 Version:    25.0.3
 Context:    colima
 Debug Mode: false
 Plugins:
    Version:  v0.13.0
    Path:     /.../.docker/cli-plugins/docker-buildx
    Version:  2.24.7
    Path:     /.../.docker/cli-plugins/docker-compose

Server:
 Containers: 9
  Running: 4
  Paused: 0
  Stopped: 5
 Images: 64
 Server Version: 24.0.7
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: d8f198a4ed8892c764191ef7b3b06d8a2eeb5c7f
 runc version: v1.1.10-0-g18a0cb0
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.5.0-17-generic
 Operating System: Ubuntu 23.10
 OSType: linux
 Architecture: aarch64
 CPUs: 6
 Total Memory: 7.736GiB
 Name: colima
 ID: 0296b23e-9bd3-4597-918d-0c64e567f14e
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
```

### Builders list

```text
NAME/NODE          DRIVER/ENDPOINT                                                                                            STATUS     BUILDKIT               PLATFORMS
kube-chulki        kubernetes
 \_ kube-chulki0    \_ kubernetes:///kube-chulki?deployment=buildkit-4217ad63-0d28-4fdd-8dcf-9bf48ea1bf71-chtgj&kubeconfig=   running    v0.12.5                linux/arm64, linux/arm/v7, linux/arm/v6
 \_ kube-chulki0    \_ kubernetes:///kube-chulki?deployment=buildkit-4217ad63-0d28-4fdd-8dcf-9bf48ea1bf71-chtgj&kubeconfig=   running    v0.12.5                linux/arm64, linux/arm/v7, linux/arm/v6
 \_ kube-chulki0    \_ kubernetes:///kube-chulki?deployment=buildkit-4217ad63-0d28-4fdd-8dcf-9bf48ea1bf71-chtgj&kubeconfig=   running    v0.12.5                linux/arm64, linux/arm/v7, linux/arm/v6
 \_ kube-chulki0    \_ kubernetes:///kube-chulki?deployment=buildkit-4217ad63-0d28-4fdd-8dcf-9bf48ea1bf71-chtgj&kubeconfig=   running    v0.12.5                linux/arm64, linux/arm/v7, linux/arm/v6
remote*            remote
 \_ remote0         \_ kube-pod://buildkitd-amd64-0?namespace=buildkit&context=............                                   inactive
remote-tcp         remote
 \_ remote-tcp0     \_ tcp://127.0.0.1:1234                                                                                   running    v0.13.0                linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/amd64/v4, linux/386
colima             docker
 \_ colima          \_ colima                                                                                                 running    v0.11.7+d3e6c1360f6e   linux/arm64, linux/amd64, linux/amd64/v2
default                                                                                                                       error

Cannot load builder default: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

### Configuration

N/A

### Build logs

_No response_

### Additional info

- I deployed builkit using https://github.com/moby/buildkit/blob/v0.13.0/examples/kubernetes/statefulset.privileged.yaml
- docker-buildx 0.12.1 + buildkit 0.13.0 worked.
- docker-buildx 0.13.0 + buildkit 0.12.5 didn't work
  - kubernetes driver bootstrap a builder using `moby/buildkit:buildx-1-stabe` which is 0.12.5.. so I thought it could be compatible issue but it's actually not

Found different behavior with dial-stdio which is apparently used to connect to builder

```console
$ docker-buildx dial-stdio --builder remote-tcp --progress plain
#0 builder: remote-tcp

#1 Proxying to builder
#1 Dialing builder
#1 Dialing builder 1.8s done

$ docker-buildx dial-stdio --builder remote --progress plain
#0 builder: remote

#1 Proxying to builder
#1 Dialing builder
#1 ...

#2 [internal] waiting for connection
#2 ERROR: context deadline exceeded

#1 Proxying to builder
#1 Dialing builder 20.0s done
#1 ERROR: context deadline exceeded
------
 > Proxying to builder:
------
------
 > [internal] waiting for connection:
------
ERROR: context deadline exceeded
```

I'm not sure it's a buildkit issue or docker-buildx issue though.

tonistiigi:
fix #2324
closes #2329

With the new dial-stdio command the dialer is split from `Client` function in order to access it directly.

This breaks the custom connhelpers functionality
as support for connhelpers is a feature of the default dialer. If client defines a custom dialer then only it is used without extra modifications. This means that remote driver dialer needs to detect the
connhelpers on its own.

The relevant default dialer addition is in https://github.com/moby/buildkit/blob/a38011b9f57db4b805e6bb0322d7d923b341bc6e/client/client.go#L109-L115 .

@cpuguy83 @jsternberg 

@crazy-max For test, should we have one custom test just for connhelpers or maybe just add extra config in the matrix that runs remote driver with a `docker-container://` endpoint address?

### DIFF
--- a/builder/node.go
+++ b/builder/node.go
@@ -142,7 +142,7 @@ func (b *Builder) LoadNodes(ctx context.Context, opts ...LoadNodesOption) (_ []N
 					}
 				}
 
-				d, err := driver.GetDriver(ctx, "buildx_buildkit_"+n.Name, factory, n.Endpoint, dockerapi, imageopt.Auth, kcc, n.BuildkitdFlags, n.Files, n.DriverOpts, n.Platforms, b.opts.contextPathHash, lno.dialMeta)
+				d, err := driver.GetDriver(ctx, driver.BuilderName(n.Name), factory, n.Endpoint, dockerapi, imageopt.Auth, kcc, n.BuildkitdFlags, n.Files, n.DriverOpts, n.Platforms, b.opts.contextPathHash, lno.dialMeta)
 				if err != nil {
 					node.Err = err
 					return nil
--- a/tests/create.go
+++ b/tests/create.go
@@ -1,9 +1,13 @@
 package tests
 
 import (
+	"fmt"
+	"os"
 	"strings"
 	"testing"
 
+	"github.com/docker/buildx/driver"
+	"github.com/moby/buildkit/identity"
 	"github.com/moby/buildkit/util/testutil/integration"
 	"github.com/stretchr/testify/require"
 )
@@ -18,6 +22,7 @@ func createCmd(sb integration.Sandbox, opts ...cmdOpt) (string, error) {
 var createTests = []func(t *testing.T, sb integration.Sandbox){
 	testCreateMemoryLimit,
 	testCreateRestartAlways,
+	testCreateRemoteContainer,
 }
 
 func testCreateMemoryLimit(t *testing.T, sb integration.Sandbox) {
@@ -56,4 +61,50 @@ func testCreateRestartAlways(t *testing.T, sb integration.Sandbox) {
 	out, err := createCmd(sb, withArgs("--driver", "docker-container", "--driver-opt", "restart-policy=always"))
 	require.NoError(t, err, out)
 	builderName = strings.TrimSpace(out)
+}
+
+func testCreateRemoteContainer(t *testing.T, sb integration.Sandbox) {
+	if sb.Name() != "docker" {
+		t.Skip("skipping test for non-docker workers")
+	}
+
+	ctnBuilderName := "ctn-builder-" + identity.NewID()
+	remoteBuilderName := "remote-builder-" + identity.NewID()
+	var hasCtnBuilder, hasRemoteBuilder bool
+	t.Cleanup(func() {
+		if hasCtnBuilder {
+			out, err := rmCmd(sb, withArgs(ctnBuilderName))
+			require.NoError(t, err, out)
+		}
+		if hasRemoteBuilder {
+			out, err := rmCmd(sb, withArgs(remoteBuilderName))
+			require.NoError(t, err, out)
+		}
+	})
+
+	out, err := createCmd(sb, withArgs("--driver", "docker-container", "--name", ctnBuilderName))
+	require.NoError(t, err, out)
+	hasCtnBuilder = true
+
+	out, err = inspectCmd(sb, withArgs("--bootstrap", ctnBuilderName))
+	require.NoError(t, err, out)
+
+	cmd := dockerCmd(sb, withArgs("container", "inspect", fmt.Sprintf("%s0", driver.BuilderName(ctnBuilderName))))
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
+
+	out, err = createCmd(sb, withArgs("--driver", "remote", "--name", remoteBuilderName, fmt.Sprintf("docker-container://%s0", driver.BuilderName(ctnBuilderName))))
+	require.NoError(t, err, out)
+	hasRemoteBuilder = true
+
+	out, err = inspectCmd(sb, withArgs(remoteBuilderName))
+	require.NoError(t, err, out)
+
+	for _, line := range strings.Split(out, "\n") {
+		if v, ok := strings.CutPrefix(line, "Status:"); ok {
+			require.Equal(t, strings.TrimSpace(v), "running")
+			return
+		}
+	}
+	require.Fail(t, "remote builder is not running")
 }
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"net"
+	"strings"
 
 	"github.com/docker/buildx/store"
 	"github.com/docker/buildx/util/progress"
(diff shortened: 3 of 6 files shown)


crazy-max:
> For test, should we have one custom test just for connhelpers or maybe just add extra config in the matrix that runs remote driver with a `docker-container://` endpoint address?

I think a custom test is fine, I will open a PR with this test to make it fail.

tests/create.go
```diff
@@ -57,3 +60,28 @@ func testCreateRestartAlways(t *testing.T, sb integration.Sandbox) {
...
+	if sb.Name() != "docker" {
+		t.Skip("skipping test for non-docker workers")
+	}
+
+	var builderName string
+	t.Cleanup(func() {
+		if builderName == "" {
+			return
+		}
+		out, err := rmCmd(sb, withArgs(builderName))
```
tonistiigi, line 74: This should clean up both builders?

tests/create.go
```diff
@@ -57,3 +60,28 @@ func testCreateRestartAlways(t *testing.T, sb integration.Sandbox) {
...
+
+	out, err := createCmd(sb, withArgs("--driver", "docker-container"))
+	require.NoError(t, err, out)
+	builderName = strings.TrimSpace(out)
+
+	out, err = inspectCmd(sb, withArgs("--bootstrap", builderName))
+	require.NoError(t, err, out)
+
+	out, err = createCmd(sb, withArgs("--driver", "remote", fmt.Sprintf("docker-container://%s_%s0", driver.BuilderNamePrefix, builderName), "--bootstrap"))
+	require.NoError(t, err, out)
```
tonistiigi, line 86: For a good measure I think better to do `--name <randomname>` and then extra check with  `inspect <randomname>`. Maybe also check that the remote builder is in the running state.

crazy-max: Updated to return builder name from output and inspect after.

crazy-max:
#2329 fails as expected: https://github.com/docker/buildx/actions/runs/8215998277/job/22470095437?pr=2329#step:7:163

crazy-max (approved):

tests/create.go
```diff
@@ -57,3 +61,39 @@ func testCreateRestartAlways(t *testing.T, sb integration.Sandbox) {
...
+
+	cmd := dockerCmd(sb, withArgs("container", "inspect", fmt.Sprintf("%s%s0", driver.BuilderNamePrefix, ctnBuilderName)))
+	cmd.Stderr = os.Stderr
+	require.NoError(t, cmd.Run())
+
+	out, err = createCmd(sb, withArgs("--driver", "remote", fmt.Sprintf("docker-container://%s%s0", driver.BuilderNamePrefix, ctnBuilderName), "--bootstrap"))
+	require.NoError(t, err, out)
+	remoteBuilderName = strings.TrimSpace(out)
+
+	out, err = inspectCmd(sb, withArgs("--bootstrap", remoteBuilderName))
```
tonistiigi, line 97: This does not need bootstrap (actually the `create` in line 93 doesn't as well).

If we use a generated name for both builders then for extra safety check that the names are not the same and this line isn't just checking the first builder again.

crazy-max: > then for extra safety check that the names are not the same

Don't think they can be the same, we have a check in store logic for this when generating name: https://github.com/docker/buildx/blob/4327ee73b18949e47eb1c865daa262af095cf974/store/util.go#L40-L54

crazy-max: > This does not need bootstrap (actually the `create` in line 93 doesn't as well).

My bad forgot to remove this arg

tonistiigi: Yes, but the reason for the check is to detect the case if this code breaks.

cpuguy83 (approved):

tests/create.go
tonistiigi: Check that the second builder is in running state as well.

tonistiigi (commented):
Test LGTM

builder/node.go
```diff
@@ -142,7 +142,7 @@ func (b *Builder) LoadNodes(ctx context.Context, opts ...LoadNodesOption) (_ []N
 					}
 				}
 
-				d, err := driver.GetDriver(ctx, "buildx_buildkit_"+n.Name, factory, n.Endpoint, dockerapi, imageopt.Auth, kcc, n.BuildkitdFlags, n.Files, n.DriverOpts, n.Platforms, b.opts.contextPathHash, lno.dialMeta)
+				d, err := driver.GetDriver(ctx, driver.BuilderNamePrefix+n.Name, factory, n.Endpoint, dockerapi, imageopt.Auth, kcc, n.BuildkitdFlags, n.Files, n.DriverOpts, n.Platforms, b.opts.contextPathHash, lno.dialMeta)
```
jsternberg, line 145: Instead of using a constant and doing `driver.BuilderNamePrefix+n.Name`, maybe we should have:

```
func BuilderName(name string) string {
}
```

Functionally it would do the same thing, but it would avoid leaking that these are a simple concatenation and make the library a bit easier to use correctly.

jsternberg (approved):
LGTM. A few comments on some refactoring ideas that I think would be minor and good to implement, but nothing blocking.

jsternberg (approved):

PR #2336, opened 2024-03-13
bake: load override

crazy-max:
follow-up https://github.com/docker/buildx/pull/2330

The `--load` case is not properly handled in v0.13 and previous v0.12 as it doesn't append but overrides the outputs:

```hcl
target "foo" {
  output = [ "type=local,dest=out" ]
}

target "bar" {
}
```

```
$ docker buildx bake --load --print foo bar
```
```json
{
  "group": {
    "default": {
      "targets": [
        "foo",
        "bar"
      ]
    }
  },
  "target": {
    "bar": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        "type=docker"
      ]
    },
    "foo": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        "type=docker"
      ]
    }
  }
}
```

With this change it will append only if there is a compatible output already (image, registry) or empty:

```json
{
  "group": {
    "default": {
      "targets": [
        "foo",
        "bar"
      ]
    }
  },
  "target": {
    "bar": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        "type=docker"
      ]
    },
    "foo": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        "type=local,dest=out"
      ]
    }
  }
}
```

For example with:

```hcl
target "foo" {
  output = [ "type=image" ]
}

target "bar" {
}
```

It would result in:

```json
{
  "group": {
    "default": {
      "targets": [
        "foo",
        "bar"
      ]
    }
  },
  "target": {
    "bar": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        "type=docker"
      ]
    },
    "foo": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        "type=image"
        "type=docker"
      ]
    }
  }
}
```

Same has been done for push override so we are consistent. See new tests suite for more info.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -894,19 +894,17 @@ func (t *Target) AddOverrides(overrides map[string]Override) error {
 			}
 			t.Pull = &pull
 		case "push":
-			_, err := strconv.ParseBool(value)
+			push, err := strconv.ParseBool(value)
 			if err != nil {
 				return errors.Errorf("invalid value %s for boolean key push", value)
 			}
-			if len(t.Outputs) == 0 {
-				t.Outputs = append(t.Outputs, "type=image,push=true")
-			} else {
-				for i, output := range t.Outputs {
-					if typ := parseOutputType(output); typ == "image" || typ == "registry" {
-						t.Outputs[i] = t.Outputs[i] + ",push=" + value
-					}
-				}
+			t.Outputs = setPushOverride(t.Outputs, push)
+		case "load":
+			load, err := strconv.ParseBool(value)
+			if err != nil {
+				return errors.Errorf("invalid value %s for boolean key load", value)
 			}
+			t.Outputs = setLoadOverride(t.Outputs, load)
 		default:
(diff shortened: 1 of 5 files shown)


bake/bake.go
```diff
@@ -894,18 +894,51 @@ func (t *Target) AddOverrides(overrides map[string]Override) error {
...
+				setLoad := true
+				for _, output := range t.Outputs {
+					if typ := parseOutputType(output); typ == "docker" {
+						if v := parseOutput(output); v != nil {
+							if _, ok := v["dest"]; !ok {
+								setLoad = false
+								break
+							}
+						}
+					} else if typ != "image" && typ != "registry" {
```
tonistiigi, line 932: If we allow image and registry then I think we should allow "oci" as well.

tonistiigi (approved):

PR #2359, opened 2024-03-27
vendor: github.com/docker/docker and github.com/docker/cli v26.0.0

crazy-max:
bump docker to 26.0.0 in the dockerfile for integration tests as well.


commands/prune.go
```diff
@@ -50,7 +50,7 @@ func runPrune(ctx context.Context, dockerCli command.Cli, opts pruneOptions) err
 	}
 
 	if !opts.force {
-		if ok, err := prompt(ctx, dockerCli.In(), dockerCli.Out(), warning); err != nil {
+		if ok, err := command.PromptForConfirmation(ctx, dockerCli.In(), dockerCli.Out(), warning); err != nil {
```
crazy-max, line 53: this has been fixed in https://github.com/docker/cli/pull/4888

tonistiigi: The current version can't be used because it sets up own signal handlers internally that would mess up the buildx own signal handlers. See https://github.com/docker/cli/pull/4888#pullrequestreview-1937955718

We need to do some copy-paste of the old/forked code to work around.

tonistiigi (changes requested):

tonistiigi:
superseded by #2372

PR #2363, opened 2024-03-27
bake: git auth support for remote definitions

### LINKED ISSUE #2360: Auth support for remote bake definitions
related to https://github.com/docker/actions-toolkit/pull/288

We don't support building a remote bake definition from a private repository using a git auth token. Only SSH auth with default agent is currently supported for this case: https://github.com/docker/buildx/blob/8abef5908705e49f7ba88ef8c957e1127b597a2a/bake/remote.go#L30-L33

Loading a remote bake definition is done by a dedicated solve request before build occurs: https://github.com/docker/buildx/blob/8abef5908705e49f7ba88ef8c957e1127b597a2a/bake/remote.go#L62

So we would need to pass the secret in a proper manner through `bake` command and attach it to the session.

@tonistiigi suggests that we could have a builtin name as override that would look like this:

```
docker buildx bake \
  --set "<builtin_name>.secrets=id=GIT_AUTH_TOKEN,env=GIT_AUTH_TOKEN"
  https://github.com/docker/test-docker-action.git#remote-private
```

Where `<builtin_name>` would be unique and could not be used as target name in the bake definition. As we already restrict target name with specific chars: https://github.com/docker/buildx/blob/8abef5908705e49f7ba88ef8c957e1127b597a2a/bake/bake.go#L35

We could use one not part of it to avoid collision such as `bake@remote`:

```
docker buildx bake \
  --set "bake@remote.secrets=id=GIT_AUTH_TOKEN,env=GIT_AUTH_TOKEN"
  https://github.com/docker/test-docker-action.git#remote-private
```

Open to suggestions.

crazy-max:
fixes #2360 

Adds support for Git auth when using remote bake definitions through `BUILDX_BAKE_GIT_AUTH_TOKEN` and `BUILDX_BAKE_GIT_AUTH_HEADER` envs var that will set `GIT_AUTH_TOKEN` and `GIT_AUTH_HEADER` secrets respectively, matching https://github.com/moby/buildkit/blob/ad5a64c68f20435c52a51db73ef9246d6a3feba6/client/llb/source.go#L270-L271

Also adds `BUILDX_BAKE_GIT_SSH` to be consistent so user can provide SSH sock paths.

### DIFF
--- a/bake/remote.go
+++ b/bake/remote.go
@@ -4,6 +4,8 @@ import (
 	"archive/tar"
 	"bytes"
 	"context"
+	"os"
+	"strings"
 
 	"github.com/docker/buildx/builder"
 	controllerapi "github.com/docker/buildx/controller/pb"
@@ -23,13 +25,34 @@ type Input struct {
 }
 
 func ReadRemoteFiles(ctx context.Context, nodes []builder.Node, url string, names []string, pw progress.Writer) ([]File, *Input, error) {
-	var session []session.Attachable
+	var sessions []session.Attachable
 	var filename string
+
 	st, ok := dockerui.DetectGitContext(url, false)
 	if ok {
-		ssh, err := controllerapi.CreateSSH([]*controllerapi.SSH{{ID: "default"}})
-		if err == nil {
-			session = append(session, ssh)
+		if ssh, err := controllerapi.CreateSSH([]*controllerapi.SSH{{
+			ID:    "default",
+			Paths: strings.Split(os.Getenv("BUILDX_BAKE_GIT_SSH"), ","),
+		}}); err == nil {
+			sessions = append(sessions, ssh)
+		}
+		var gitAuthSecrets []*controllerapi.Secret
+		if _, ok := os.LookupEnv("BUILDX_BAKE_GIT_AUTH_TOKEN"); ok {
+			gitAuthSecrets = append(gitAuthSecrets, &controllerapi.Secret{
+				ID:  llb.GitAuthTokenKey,
+				Env: "BUILDX_BAKE_GIT_AUTH_TOKEN",
+			})
+		}
+		if _, ok := os.LookupEnv("BUILDX_BAKE_GIT_AUTH_HEADER"); ok {
+			gitAuthSecrets = append(gitAuthSecrets, &controllerapi.Secret{
+				ID:  llb.GitAuthHeaderKey,
+				Env: "BUILDX_BAKE_GIT_AUTH_HEADER",
+			})
+		}
+		if len(gitAuthSecrets) > 0 {
+			if secrets, err := controllerapi.CreateSecrets(gitAuthSecrets); err == nil {
+				sessions = append(sessions, secrets)
+			}
 		}
 	} else {
 		st, filename, ok = dockerui.DetectHTTPContext(url)
@@ -59,7 +82,7 @@ func ReadRemoteFiles(ctx context.Context, nodes []builder.Node, url string, name
 
 	ch, done := progress.NewChannel(pw)
 	defer func() { <-done }()
-	_, err = c.Build(ctx, client.SolveOpt{Session: session, Internal: true}, "buildx", func(ctx context.Context, c gwclient.Client) (*gwclient.Result, error) {
+	_, err = c.Build(ctx, client.SolveOpt{Session: sessions, Internal: true}, "buildx", func(ctx context.Context, c gwclient.Client) (*gwclient.Result, error) {
 		def, err := st.Marshal(ctx)
 		if err != nil {
 			return nil, err
--- a/tests/bake.go
+++ b/tests/bake.go
@@ -29,6 +29,7 @@ var bakeTests = []func(t *testing.T, sb integration.Sandbox){
 	testBakeLocal,
 	testBakeLocalMulti,
 	testBakeRemote,
+	testBakeRemoteAuth,
 	testBakeRemoteCmdContext,
 	testBakeRemoteLocalOverride,
 	testBakeLocalCwdOverride,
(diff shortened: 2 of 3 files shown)


crazy-max:
@tonistiigi While adding tests I discovered a strange behavior with git source in BuildKit when authenticated. It seems most of the requests send the authorization header but not the one when checking the default branch in https://github.com/moby/buildkit/blob/ad5a64c68f20435c52a51db73ef9246d6a3feba6/source/git/source.go#L732 even is header is set in
https://github.com/moby/buildkit/blob/ad5a64c68f20435c52a51db73ef9246d6a3feba6/source/git/source.go#L350 or https://github.com/moby/buildkit/blob/ad5a64c68f20435c52a51db73ef9246d6a3feba6/source/git/source.go#L422.

See https://github.com/docker/buildx/actions/runs/8465151975/job/23191259148#step:7:635

```
    testutilserve.go:52: Request Method: GET
    testutilserve.go:53: Request URL: /test.git/info/refs?service=git-upload-pack
    testutilserve.go:54: Request Headers:
    testutilserve.go:57:   Cache-Control: no-cache
    testutilserve.go:57:   User-Agent: git/2.43.0
    testutilserve.go:57:   Accept: */*
    testutilserve.go:57:   Accept-Encoding: deflate, gzip, br
    testutilserve.go:57:   Pragma: no-cache
    testutilserve.go:57:   Git-Protocol: version=2
    testutilserve.go:60: access token used: "cptgkeatnyrttvqj1pmrlj8sv"
    testutilserve.go:62: basic auth: user="" pass=""
```

So for now in this test I just check the very first request is authenticated but I think we need a fix on BuildKit side.

bake/remote.go
```diff
@@ -23,13 +25,42 @@ type Input struct {
...
+			sshSpecs = []*controllerapi.SSH{{ID: "default"}}
+		}
+		if ssh, err := controllerapi.CreateSSH(sshSpecs); err == nil {
+			sessions = append(sessions, ssh)
+		}
+
+		var gitAuthSecrets []*controllerapi.Secret
+		if _, ok := os.LookupEnv("BUILDX_BAKE_GIT_AUTH_TOKEN"); ok {
+			gitAuthSecrets = append(gitAuthSecrets, &controllerapi.Secret{
+				ID:  "GIT_AUTH_TOKEN",
```
crazy-max, line 50: we should expose this key in buildkit to avoid hardcoded value. Same for `GIT_AUTH_HEADER`

crazy-max: opened https://github.com/moby/buildkit/pull/4809

thaJeztah:
Will these always be for bake only, or could these be for "build" as well? (curious if the `BUILDX_BAKE_GIT_AUTH_HEADER` env vars must contain `_BAKE_`)

crazy-max:
> Will these always be for bake only, or could these be for "build" as well? (curious if the `BUILDX_BAKE_GIT_AUTH_HEADER` env vars must contain `_BAKE_`)

Only bake, this is used to build remote bake definitions. For build we already use secrets to build a remote Git context.

thaJeztah:
Ah! So.. naive question; if that doesn't require special env-vars; why does bake need them?

crazy-max:
> Re-reading; you mention remote _definitions_ (so equivalent to "remote dockerfile"); I know `docker build` / `docker buildx build` also allows building from a remote Dockerfile (Dockerfile could be remote, but build-context could be local) but not sure it currently does from a "git" URL; if that's not supported, then ... perhaps it should as well?

Yes indeed I don't think we have that for the Dockerfile atm. For bake we have "builtin" ways to use either a local or remote context, as well as merging local and remote definitions: https://docs.docker.com/build/bake/remote-definition/

@dvdksn Talking about merging local and remote bake definitions, I don't think we have this documented atm https://github.com/docker/buildx/pull/1838

bake/remote.go
```diff
@@ -23,13 +25,42 @@ type Input struct {
...
+			}
+		} else {
+			sshSpecs = []*controllerapi.SSH{{ID: "default"}}
+		}
+		if ssh, err := controllerapi.CreateSSH(sshSpecs); err == nil {
+			sessions = append(sessions, ssh)
+		}
+
+		var gitAuthSecrets []*controllerapi.Secret
+		if _, ok := os.LookupEnv("BUILDX_BAKE_GIT_AUTH_TOKEN"); ok {
```
tonistiigi, line 48: Probably out of scope, but unclear to me how the controller takes `Env` keys as input. Doesn't the controller have its own env if it is running in another process?

bake/remote.go
```diff
@@ -23,13 +25,42 @@ type Input struct {
...
+	var sessions []session.Attachable
 	var filename string
+
 	st, ok := dockerui.DetectGitContext(url, false)
 	if ok {
-		ssh, err := controllerapi.CreateSSH([]*controllerapi.SSH{{ID: "default"}})
-		if err == nil {
-			session = append(session, ssh)
+		var sshSpecs []*controllerapi.SSH
+		if v := os.Getenv("BUILDX_BAKE_GIT_SSH"); v != "" {
```
tonistiigi, line 34: ID of this needs to be set with `MountSSHSock()` if not default. Setting it to other ID would not work otherwise. So probably it should just be the path.

bake/remote.go
crazy-max: Hum good point, it seems to work when connecting to local controller. I don't think there is another process except when using `--invoke` and setting envs through invoke config: https://github.com/docker/buildx/blob/a7d59ae3328bc5da57ba7f73a65a0b5f36ffe1bb/controller/pb/controller.proto#L199

tonistiigi (approved):

PR #2373, opened 2024-04-01
metricutil: switch to using the cli meter provider

jsternberg:
The meter provider initialization that was located here has now been
moved to a common area in the docker cli. This upgrades our CLI version
and then uses this common code instead of our own version.

As a piece of additional functionality, the docker OTEL endpoint can now
be overwritten with `DOCKER_CLI_OTEL_EXPORTER_OTLP_ENDPOINT` for
testing.

This removes the OTLP exporter from the CLI that was previously locked
behind `BUILDX_EXPERIMENTAL`. I do plan for this to return, but as a
proper part of the `docker/cli` implementation rather than something
special with `buildx`.


crazy-max (commented):
This a follow-up of https://github.com/docker/cli/pull/4889 right? Can you rebase as well?

jsternberg:
Yea it's a follow up to that PR. Rebased to include the vendor update for `docker/cli`.

crazy-max (approved):

tonistiigi (approved):

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/containerd/typeurl/v2 v2.1.1
 	github.com/creack/pty v1.1.18
 	github.com/distribution/reference v0.5.0
-	github.com/docker/cli v26.0.0+incompatible
+	github.com/docker/cli v26.0.1-0.20240401150816-155dc5e4e406+incompatible
```
thaJeztah, line 17: Do we want the change backported to the 26.0  branch so that we don't have to force consumers to depend on master? (Haven't checked if we can)

cc @vvoland @laurazard

Issue #2350, opened 2024-03-21
docs: improve discoverability of build-secrets


thaJeztah:
### Description

- relates to https://github.com/moby/buildkit/discussions/4791

While writing an answer to the discussion above, I noticed that the "build secrets" section is a bit hard to discover;

<img width="821" alt="Screenshot 2024-03-21 at 10 57 02" src="https://github.com/docker/buildx/assets/1804568/5d3ddc16-135e-457b-b738-151b1fc1e250">

The ["build overview" page](https://docs.docker.com/build/) has no tile for "secrets", so users will need to discover it from the "building" images navigation:
 

Neither the [`--mount type=secret`](https://docs.docker.com/reference/dockerfile/#run---mounttypesecret), nor the [`docker build --secret`](https://docs.docker.com/reference/cli/docker/buildx/build/#secret) flag cross-references the ["build secrets" section](https://docs.docker.com/build/building/secrets/).

I think we should improve that to help discoverability;

```[tasklist]
- [ ] Add a "build secrets" or "working with secrets in your builds" tile.
- [ ] Cross-reference the "build secrets" section from both the  [`--mount type=secret`](https://docs.docker.com/reference/dockerfile/#run---mounttypesecret), and [`docker build --secret`](https://docs.docker.com/reference/cli/docker/buildx/build/#secret) sections.
- [ ] Consider adding language-specific examples in the "getting started" sections (or elsewhere?), which, for Go, could include details about working with private modules (and provide links to Go's documentation on this as well).
- [ ] Consider adding warnings to sections mentioned above about _NOT_ using `ARG` for secrets (users are still tempting to do so).
- [ ] :question: do we need some "do's / dont's" in the ["security" section](https://docs.docker.com/security/for-developers/access-tokens/)? (also see https://github.com/moby/moby/issues/13490)
```

thaJeztah:
cc @dvdksn

PR #2378, opened 2024-04-03
docs: add cross-reference about build secrets

dvdksn:
Adds a cross-reference to the build secrets page

Partly addresses #2350

### DIFF
--- a/docs/reference/buildx_build.md
+++ b/docs/reference/buildx_build.md
@@ -615,10 +615,18 @@ For more information, see [here](https://docs.docker.com/build/attestations/sbom
 --secret=[type=TYPE[,KEY=VALUE]
 ```
 
-Exposes secret to the build. The secret can be used by the build using
-[`RUN --mount=type=secret` mount](https://docs.docker.com/reference/dockerfile/#run---mounttypesecret).
+Exposes secrets (authentication credentials, tokens) to the build.
+A secret can be mounted into the build using a `RUN --mount=type=secret` mount in the
+[Dockerfile](https://docs.docker.com/reference/dockerfile/#run---mounttypesecret).
+For more information about how to use build secrets, see
+[Build secrets](https://docs.docker.com/build/building/secrets/).
 
-If `type` is unset it will be detected. Supported types are:
+Supported types are:
+
+- [`file`](#file)
+- [`env`](#env)
+
+Buildx attempts to detect the `type` automatically if unset.
 
 #### `file`
 


docs/reference/buildx_build.md
```diff
@@ -615,10 +615,18 @@ For more information, see [here](https://docs.docker.com/build/attestations/sbom
 --secret=[type=TYPE[,KEY=VALUE]
 ```
 
-Exposes secret to the build. The secret can be used by the build using
-[`RUN --mount=type=secret` mount](https://docs.docker.com/reference/dockerfile/#run---mounttypesecret).
+Exposes secrets (authentication credentials, tokens) to the build.
+A secret can mounted into the build using a `RUN --mount=type=secret` mount in the
```
crazy-max, line 619:
```suggestion
A secret can be mounted into the build using a `RUN --mount=type=secret` mount in the
```

docs/reference/buildx_build.md
```diff
@@ -615,10 +615,18 @@ For more information, see [here](https://docs.docker.com/build/attestations/sbom
...
+For more information about how to use build secrets, see
+[Build secrets](https://docs.docker.com/build/building/secrets/).
 
-If `type` is unset it will be detected. Supported types are:
+Supported types are:
+
+- [`file`](#file)
+- [`env`](#env)
+
+Buildx attempts detect the `type` automatically if unset.
```
crazy-max, line 629:
```suggestion
Buildx attempts to detect the `type` automatically if unset.
```

docs/reference/buildx_build.md
```diff
@@ -615,10 +615,18 @@ For more information, see [here](https://docs.docker.com/build/attestations/sbom
...
+For more information about how to use build secrets, see
+[Build secrets](https://docs.docker.com/build/building/secrets/).
 
-If `type` is unset it will be detected. Supported types are:
+Supported types are:
+
+- [`file`](#file)
+- [`env`](#env)
+
+Buildx attempts to detect the `type` automatically if unset.
```
thaJeztah, line 629: Not a blocker for this PR, but perhaps we could introduce a table for this. Also wondering if `type` should have an explicit `auto` option (which would be the default if unset); thinking along these lines;


Perhaps also sort these alphabetically (but looks like the paragraphs aren't either 🙈)

| type | description |
|---|---|
| `auto` (default) | automatically detect type |
| `env` | ... |
| `file` | .... |

tonistiigi (approved):

PR #2393, opened 2024-04-09
go.mod: move indirect deps to the right require block

crazy-max:
missed that in https://github.com/docker/buildx/pull/2373

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,11 +42,7 @@ require (
 	github.com/tonistiigi/fsutil v0.0.0-20240301111122-7525a1af2bb5
 	github.com/zclconf/go-cty v1.14.1
 	go.opentelemetry.io/otel v1.21.0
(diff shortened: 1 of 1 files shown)


thaJeztah:
Still curious if there's a linter for this, or if these should be reported as a bug to the go maintainers. It seems to be happening more in recent Go versions (either dependencies ending in the wrong group, or it adding a _new_ `requires` groups, and now ending up with 3 or more of them.

thaJeztah (approved):
LGTM

crazy-max:
> Still curious if there's a linter for this, or if these should be reported as a bug to the go maintainers. It seems to be happening more in recent Go versions (either dependencies ending in the wrong group, or it adding a _new_ `requires` groups, and now ending up with 3 or more of them.

Yes there should be a linter for this :disappointed:. I'm looking at it as follow-up. Not sure if go bug as users might want to organize their `go.mod` differently in some cases.

PR #2408, opened 2024-04-13
build: support statuscode response for print requests

tonistiigi:
Allows `--print` requests to respond with error statuscode that is then used for the current process.

Depends on https://github.com/moby/buildkit/pull/4840


controller/pb/controller.proto
```diff
@@ -114,6 +114,7 @@ message Secret {
 message PrintFunc {
 	string Name = 1;
 	string Format = 2;
```
crazy-max, line 116: Not related but bad indent here.

crazy-max (approved):

daghack (approved):

daghack (approved):

PR #2422, opened 2024-04-19
build: don't generate metadata file when print flag is used

crazy-max:
related to https://github.com/docker/buildx/pull/2416, we should not generate metadata file when print flag is used

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -122,27 +122,26 @@ func (o *buildOptions) toControllerOptions() (*controllerapi.BuildOptions, error
 	}
 
 	opts := controllerapi.BuildOptions{
-		Allow:                  o.allow,
-		Annotations:            o.annotations,
-		BuildArgs:              buildArgs,
-		CgroupParent:           o.cgroupParent,
-		ContextPath:            o.contextPath,
-		DockerfileName:         o.dockerfileName,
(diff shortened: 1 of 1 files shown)


tonistiigi (commented):
Shouldn't we just skip the `--metadata-file` ?

crazy-max:
> Shouldn't we just skip the `--metadata-file` ?

Yes good point we should do this as well.

build/build.go
```diff
@@ -476,7 +476,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 						rr.ExporterResponse[k] = string(v)
 					}
 					rr.ExporterResponse["buildx.build.ref"] = buildRef
-					if opt.WithProvenanceResponse && node.Driver.HistoryAPISupported(ctx) {
+					if !so.Internal && opt.WithProvenanceResponse && node.Driver.HistoryAPISupported(ctx) {
```
tonistiigi, line 479: Why do we need `!so.Internal` if `WithProvenanceResponse` is already false.

crazy-max: This was for extra safety as build record does not exist for internal builds but not needed as internal solve opt can't be set with current api.

tonistiigi (approved):

PR #2425, opened 2024-04-23
bump compose-go to v2.1.1

glours:
Use the latest version `v2.1.0` of `compose-go` with `bake`.
~This one introduces `entitlements` attribute which is managed on the Compose side~


tonistiigi:
> This one introduces entitlements attribute which is managed on the Compose side

I don't think this actually works with bake as I don't see the values passed through, or am I missing something?

Also, not sure if we should enable this (without some explicit insecure-mode opt-in) before https://github.com/docker/buildx/issues/179

crazy-max:
@glours Looks like [2.1.1](https://github.com/compose-spec/compose-go/releases/tag/v2.1.1) is out, can you bump to this version instead? Thanks.

crazy-max (approved):

glours:
Hahah @crazy-max  I just see your message, is it too late to bump the version?

crazy-max:
> Hahah @crazy-max I just see your message, is it too late to bump the version?

No worries I bumped it myself, see last commit.

PR #2426, opened 2024-04-23
docs: update link to Docker container driver

Mobiletainment:
The manual for the Docker container build driver has moved permanently:
* from https://docs.docker.com/build/building/drivers/docker-container
* to https://docs.docker.com/build/drivers/docker-container/


crazy-max:
Thanks but these docs are temporary redirect and I think we could just remove https://github.com/docker/buildx/tree/master/docs/guides and https://github.com/docker/buildx/tree/master/docs/manuals altogether (but keep https://github.com/docker/buildx/blob/master/docs/guides/debugging.md). WDYT @dvdksn?

dvdksn:
SGTM

PR #2439, opened 2024-04-30
ci(validate): split lint

crazy-max:
follow-up https://github.com/moby/buildkit/pull/4829#issuecomment-2045373882

This generates a matrix for each platform for `lint` and `lint-gopls` targets. I think I will port the logic in "Matrix" step to a subaction similar to https://github.com/docker/bake-action?tab=readme-ov-file#list-targets.

Reduces build time for the whole pipeline from ~30m:

image

To ~6m30:

image

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -17,17 +17,63 @@ on:
       - '.github/releases.json'
 
 jobs:
+  prepare:
+    runs-on: ubuntu-22.04
+    outputs:
+      includes: ${{ steps.matrix.outputs.includes }}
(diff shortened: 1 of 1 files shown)


crazy-max:
@tonistiigi 

Needs to remove "validate (lint)" from required checks in branches protection settings for the repo:

image

tonistiigi (approved):
Was also thinking that maybe it would make sense to run the lint target(s) for native arch first and then others in parallel only if first one passed. If linting already wrong for common files then not much point of even running the custom arch variants.

PR #2441, opened 2024-05-02
Update bug and feature issue templates

thompson-shaun:
## Summary 

This PR makes minor changes to the existing **bug** and **feature request** issue templates.

### Bug

- Add `kind/bug` label by default

### Feature Request

- Add checklist for requirements mimicking the **bug** template


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -3,6 +3,7 @@ name: Bug Report
 description: Report a bug
 labels:
   - status/triage
+  - kind/bug
```
jedevc, line 6: I think we originally had this, but then removed it? cc @tonistiigi 

I remember a discussion (I think in the build team internal docker slack) about wanting to ensure that all issues with the label were *confirmed* bugs, which requires a level of triage before.

thompson-shaun: 👋 @jedevc - I had a chance to read the old thread you mentioned, and the scenario described is crucial (being able to find confirmed bugs).

The goal with this was to try and help pre-sort as the issue entered into triage and be able to identify when the issue had been responded/handled in any way. In this case, a response could include requesting more information, putting the issue into a "blocked" state but no longer requiring initial sorting. This idea stemmed from status/triage being the default state for all inbound bugs and feature requests _and_ an assumption that the first response would get it out of that triage state to something else to avoid recycling issues. Would the addition of a verified (or using the existing accepted label) still address your use case? 

I am hoping for simplicity above all else on this front - any and all ideas would be amazing 😄

thompson-shaun: `kind/maybe-bug` until confirmed? 😁

jedevc: SGTM :tada: Essentially anything that helps filtering!

thompson-shaun: Added in the `kind/maybe-bug` as well as a "bug report lite" for reports that might be a little light on details out the gate.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -65,19 +66,19 @@ body:
...
+      required: true
 
   - type: textarea
     attributes:
-      label: Builders list
+      label: Docker info
       description: |
-        Output of `docker buildx ls` command.
+        Output of `docker info` command.
       render: text
```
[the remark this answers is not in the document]
thompson-shaun: When reading the attributes, I was just attempting to group `docker buildx` commands together -- no strong preference either way.

PR #2445, opened 2024-05-03
Bake: support compose ssh config

sumnerwarren:
Adds support for compose ssh config with bake. See [spec](https://github.com/compose-spec/compose-spec/blob/master/build.md#ssh). Based on #1069. This does not remove support for the `x-bake` `ssh` field.

### DIFF
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/compose-spec/compose-go/v2/dotenv"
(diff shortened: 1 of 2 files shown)


tonistiigi (approved):
fyi @ndeloof

sumnerwarren:
Looking at the [failed check](https://github.com/docker/buildx/actions/runs/8944019077/job/24651684166?pr=2445): sorting for multiple ssh keys does not appear to be stable. Do we want bake to make it stable or should we just change to an `ElementsMatch` assertion?

tonistiigi:
@sumnerwarren Is it already coming unsorted from compose? In that case, it should be fixed in there. Meanwhile, we can just add some forced sorting in here to get this merged.

sumnerwarren:
@tonistiigi Yes, it's coming from compose unsorted, at least on macos. I ran `docker compose config` repeatedly with this MRE compose file to confirm:

```
services:
  test:
    build:
      ssh:
        - default
        - key=key.pem
```

I just pushed a commit that includes sorting.

crazy-max (changes requested):
Thanks! Don't think we need a separate commit for sorting, can you squash commits please? Otherwise LGTM

crazy-max (approved):

PR #2450, opened 2024-05-08
Add stale issue and label actions workflows

thompson-shaun:
This PR adds in a couple of GitHub Actions to support common triage/support activities and help ensure stale issues are dealt with in a timely manner.

## Label actions

Use [label actions](https://github.com/marketplace/actions/label-actions) to automate responses, `@` authors, and auto-label for some common use cases. Most new labels would fall under the `auto/*` group and apply `status/*` labels as appropriate.

## Stale issues

Automatically nudge contributors if issues or pull requests have gone stale after a set number of days and close after a configured number of days post-nudge. As initially configured, the first prompt would occur after 14 days, with closure 7 days after the prompt for a total time of 21 days before closure.

### DIFF
--- a/.github/label-actions.yml
+++ b/.github/label-actions.yml
@@ -1,0 +1,91 @@
+# Configuration for Label Actions - https://github.com/dessant/label-actions
+
+# When a discussion is resolved, lock the discussion
+'status/resolved':
+  discussions:
+    close: true
+    close-reason: 'resolved'
+    lock: true
+
+# Request more information from the author
+'auto/needs-more-info':
+  comment: >
+    :wave: @{issue-author}, please provide the requested information.
+  issues:
+    label: 
+      - 'status/needs-more-info' 
+  prs:
+    label: 
+      - 'status/needs-more-info' 
+
+# Thank the author for their response, remove auto-labels
+'auto/got-info':
+  comment: >
+    :wave: @{issue-author}, thank you for your response!
+  issues:
+    unlabel: 
+      - 'auto/needs-more-info'
+      - 'status/needs-more-info'
+  prs:
+    unlabel: 
+      - 'auto/needs-more-info'
+      - 'status/needs-more-info'
+
+# Comment that the issue will be transferred to another repository
+'auto/transfer':
+  comment: >
+    :wave: @{issue-author}, this issue will be transferred to another repository.
+  lock: true
+  lock-reason: 'off-topic'
+  issues:
+    label: 
+      - 'needs/transfer'
+      - 'status/wontfix'
+
+# Thank the author for their contribution
+'auto/thank-you':
+  comment: >
+    :wave: @{issue-author}, thank you for your contribution!
+
+# Mark issue for conversion to discussion
+'auto/make-discussion':
+  comment: >
+    :wave: @{issue-author}, this issue will be converted to a discussion. If you think this was a mistake, please let us know.
+  issues:
+    label: 
+      - 'needs/convert-to-discussion'
+
+# Label as duplicate, prompt author to search
+'auto/duplicate':
+  comment: >
+    :wave: @{issue-author}, this item appears to be a duplicate of a previous one. Please check if it already exists and add any additional information as a comment.
+  issues:
+    label: 'kind/duplicate'
+  prs:
+    label: 'kind/duplicate'
+
+# Prompt author to provide a minimal reproduction
+'auto/needs-reproduction':
+  comment: >
+    :wave: @{issue-author}, please provide a minimal reproduction of the issue.
+  issues:
+    label: 
+      - 'needs/reproduction'
+      - 'status/blocked'
+  prs:
+    label: 
+      - 'needs/reproduction'
+      - 'status/blocked'
+
+# Use template
+'auto/use-template':
+  comment: >
+    :wave: @{issue-author}, please use one the provided templates.
+  issues:
+    label: 
+      - 'needs/template'
+      - 'status/blocked'
+  prs:
+    label: 
+      - 'needs/template'
+      - 'status/blocked'
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -1,0 +1,30 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 14
+          days-before-issue-close: 7
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'not-stale'
+          stale-issue-message: 'This issue is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
+          close-issue-message: 'This issue was closed because it has been stalled for 21 days with no activity.'
+          days-before-pr-stale: 14
+          days-before-pr-close: 7
+          stale-pr-message: 'This PR is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
+          close-pr-message: 'This PR was closed because it has been stalled for 21 days with no activity.'
+          stale-pr-label: 'stale'
+          exempt-pr-labels: 'not-stale'
+          exempt-all-pr-milestones: true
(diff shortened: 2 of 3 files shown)


.github/label-actions.yml
```diff
@@ -0,0 +1,91 @@
...
+# Comment that the issue will be transferred to another repository
+'auto/transfer':
+  comment: >
+    :wave: @{issue-author}, this issue will be transferred to another repository.
+  lock: true
+  lock-reason: 'off-topic'
+  issues:
+    label: 
+      - 'needs/transfer'
+      - 'status/wontfix'
```
crazy-max, line 43: I think this would only matter when transferring issues not part of the org. For issues within the org we can use the "Transfer issue" functionality: https://docs.github.com/en/issues/tracking-your-work-with-issues/transferring-an-issue-to-another-repository

image

.github/label-actions.yml
```diff
@@ -0,0 +1,91 @@
...
+  comment: >
+    :wave: @{issue-author}, thank you for your contribution!
+
+# Mark issue for conversion to discussion
+'auto/make-discussion':
+  comment: >
+    :wave: @{issue-author}, this issue will be converted to a discussion. If you think this was a mistake, please let us know.
+  issues:
+    label: 
+      - 'needs/convert-to-discussion'
```
crazy-max, line 56: I don't think this one is necessary, we already have a feature to convert an issue to discussion: https://docs.github.com/en/discussions/managing-discussions-for-your-community/moderating-discussions#converting-an-issue-to-a-discussion

image

.github/label-actions.yml
```diff
@@ -0,0 +1,91 @@
...
+      - 'status/blocked'
+  prs:
+    label: 
+      - 'needs/reproduction'
+      - 'status/blocked'
+
+# Use template
+'auto/use-template':
+  comment: >
+    :wave: @{issue-author}, please use one the provided templates.
```
crazy-max, line 83: I guess you mean a form template? Wording might be confusing for users. I think we should link to `https://github.com/docker/buildx/issues/new/choose` in `comment: ` with smth like:

```suggestion
  comment: >
    :wave: @{issue-author}, please use one the provided [forms](https://github.com/docker/buildx/issues/new/choose).
```

.github/label-actions.yml
```diff
@@ -0,0 +1,91 @@
...
+  issues:
+    label: 
+      - 'needs/template'
+      - 'status/blocked'
+  prs:
+    label: 
+      - 'needs/template'
+      - 'status/blocked'
```
crazy-max, line 91: Don't think we should have this for PRs as we don't have any template atm.

.github/workflows/label-actions.yml
```diff
@@ -0,0 +1,21 @@
...
+  contents: read
+  issues: write
+  pull-requests: write
+  discussions: write
+
+jobs:
+  action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/label-actions@v4
```
crazy-max, line 21: As this is a third-party action and we are requiring extra permissions that could have potential security implications, we should pin to commit sha similar to: https://github.com/docker/buildx/blob/8afeb56a3b00d94b86a98045b5b43f8bba94faad/.github/workflows/build.yml#L359

```suggestion
      - uses: dessant/label-actions@102faf474a544be75fbaf4df54e73d3c515a0e65  # v4.0.1
```

Would also need to audit this action to make sure there is nothing wrong with it.

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,30 @@
...
+          close-issue-message: 'This issue was closed because it has been stalled for 21 days with no activity.'
+          days-before-pr-stale: 14
+          days-before-pr-close: 7
+          stale-pr-message: 'This PR is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
+          close-pr-message: 'This PR was closed because it has been stalled for 21 days with no activity.'
+          stale-pr-label: 'stale'
+          exempt-pr-labels: 'not-stale'
+          exempt-all-pr-milestones: true
+          exempt-pr-assignees: true
+          start-date: '2024-05-01T00:00:00+0000'
```
crazy-max, line 30: Missing new line

```suggestion
          start-date: '2024-05-01T00:00:00+0000'

```

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,30 @@
...
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 14
+          days-before-issue-close: 7
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'not-stale'
+          stale-issue-message: 'This issue is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
```
crazy-max, line 20: What will happen for current issues opened? Will this add a comment for 300+ of them? I don't think we want that :see_no_evil:. Can we avoid this for old issues?

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,30 @@
...
+          close-issue-message: 'This issue was closed because it has been stalled for 21 days with no activity.'
+          days-before-pr-stale: 14
+          days-before-pr-close: 7
```
crazy-max, line 23: Not sure we want this for PRs :thinking: Some can still be relevant if we want to carry.

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,30 @@
...
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 14
```
crazy-max, line 16: 14 days look quite short and would probably comment on 300+ issues. Can we first extend to smth like 360 and decrease in follow-up?

.github/label-actions.yml
thompson-shaun: Yep, that was the hope. This set of labels was supposed to convey to the author that the issue was _going_ to be transferred. Likely a silly state and something we can express with `status/`

.github/label-actions.yml
thompson-shaun: Agreed -- removing

.github/label-actions.yml
thompson-shaun: Agreed. Prepping updates to the project processing guide which will have revised labels.

.github/workflows/close-stale-issues.yml
thompson-shaun: Great question! The [`start-date`](https://github.com/actions/stale?tab=readme-ov-file#start-date) was set in this PR to help avoid this situation. The bot will only process issues after the date. As we become more comfortable with it we can move the date backwards.

thaJeztah:
Just some quick blurbs :see_no_evil:

- Generally, I try to be more "permissive" (conservative) on _feature_ requests than _bug reports_
- :point_up: many feature requests don't go "stale", but they may be collecting feedback/use-cases (which can ultimately lead to _an_ implementation of _a_ feature - not necessarily exactly the feature that was requested). I've seen many feature requests that were not a "hard reject", but also not (yet) clear on use-cases (which are important to consider a design).
- :point_up: Admitted, some of this is from "before discussions existed", but having a public ticket for users to "add their use-case to" can be useful to get a picture of use-cases, and sometimes a "cluster" of feature-requests starts to form, which can be the start of working on a design.
- :point_right: bugs tend to be different; most bugs that are still open are either poor reports, lack a reproducer, and have not been confirmed (because of that); I do agree that 14 days is much too short, but after some months, it's ok to post if the issue still is observed by the reporter(s).
- :point_up: that said; we should take care to not automatically close if a ticket _has not been triaged at all_. I've seen some of our issue trackers where _valid tickets_ were reported, but due to limited bandwidth of engineers not been looked at, and closed automatically. That's a bad experience (effectively we're blaming the _user_ for not having their issue being looked at within a reasonable time-limit)
- :point_up: ideally we'd have a flow that excludes such cases from being automatically closed (or in reverse; them going stale being an indicator to the _engineers_ to do triage)

For the "thank you for your pull request" comment; I'd avoid adding that comment, unless the comment adds value (e.g. the PR to be adding a comment "this is what will happen next"). Even in that case, perhaps limit it to post for _first time contributors_. Such comments can add a lot of noise (which GitHub already creates a lot of).

PR #2456, opened 2024-05-14
driver/kubernetes/util: remove k8s.io/apiserver dependency

thaJeztah:
- similar to https://github.com/docker/pinata/pull/21449 (internal repo)

Use a simplified local implementation that follow the same semantics, so that we don't need k8s.io/apiserver as dependency, which sometimes gets flagged for vulnerabilities, but most of the time that's a false positive; e.g.;

Screenshot 2024-05-14 at 14 01 31


AkihiroSuda (approved):
👍

PR #2457, opened 2024-05-14
Add parameter provisioningTimeout to Kubernetes driver options.

Arsobbiak:
During running builds on the Kubernetes, we hit timeout on pod provisioning. 
Our k8s cluster is running on EKS with Karpenter. 
From time to time, node provisioning takes around one and a half minutes till it is ready to use.

Error:

```
#1 waiting for 1 pods to be ready
#1 waiting for 1 pods to be ready 110.0s done
#1 ERROR: expected 1 replicas to be ready, got 0
```
 
This PR will let to specify `provisioningTimeout` at `driver-opt` parameter.  

If that looks good, I will add a parameter to the docker documentation.


driver/kubernetes/driver.go
```diff
@@ -113,7 +114,7 @@ func (d *Driver) wait(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
-		case <-time.After(time.Duration(100+try*20) * time.Millisecond):
```
crazy-max, line 116: I think we should keep the backoff logic

Arsobbiak: It will be ok for you if I will just change `100` → `300`

driver/kubernetes/factory.go
```diff
@@ -229,6 +232,8 @@ func (f *factory) processDriverOpts(deploymentName string, namespace string, cfg
 			if err != nil {
 				return nil, "", "", false, err
 			}
+		case "provisioningTimeout":
+			deploymentOpt.ProvisioningTimeout, err = strconv.Atoi(v)
```
crazy-max, line 236: Can we use a `time.Duration` type instead? Would need to do a `time.ParseDuration` here.

driver/kubernetes/driver.go
crazy-max: > I'd be interested to know why the backoff is better when the nodes don't spin up in time.

@SimonJPegg Don't think you would want to hammer the API every `Xms` but instead backoff on each retry. Although looking at the retry logic it seems fair to not keep it.

> It will be ok for you if I will just change `100` → `300`?

I think it's fine to keep `100` as default but might be better to use a duration, see https://github.com/docker/buildx/pull/2457#discussion_r1601536109

driver/kubernetes/driver.go
```diff
@@ -41,15 +41,16 @@ type Driver struct {
...
+	minReplicas         int
+	deployment          *appsv1.Deployment
+	configMaps          []*corev1.ConfigMap
+	clientset           *kubernetes.Clientset
+	deploymentClient    clientappsv1.DeploymentInterface
+	podClient           clientcorev1.PodInterface
+	configMapClient     clientcorev1.ConfigMapInterface
+	podChooser          podchooser.PodChooser
+	defaultLoad         bool
+	provisioningTimeout int
```
crazy-max, line 53: I'm not sure on naming it `provisioningTimeout` as this is applied in a retry logic. Maybe `retryTimeout`?

tonistiigi:
PTAL #2472

Arsobbiak:
Sorry that it took me so long to come back to this PR. 
 
Thanks, @crazy-max, for pointing out that I can use `time.ParseDuration`. This way, it will be more user-friendly.  

With the chatbot's help and your comments, I believe I improved it slightly.

driver/kubernetes/driver.go
```diff
@@ -101,22 +102,33 @@ func (d *Driver) wait(ctx context.Context) error {
...
+	defer ticker.Stop()
+
+	for {
 		select {
+		case <-ticker.C:
+			if tryCounter < 100 {
+				tryCounter++
+			}
+			ticker.Stop()
+			ticker = time.NewTicker(time.Duration(100+tryCounter*20) * time.Millisecond)
```
Arsobbiak, line 117: Im not sure if that logic is correct and if there is no better way of doing that. 
Or maybe I should keep it static as `100*time.Millisecond`

PR #2467, opened 2024-05-20
build: fix resolvedNode cache and panic protection

### LINKED ISSUE #2464: Panic when re-bootstrapping driver in non Running state
### Contributing guidelines

### Description

Under very specific conditions bootstrapping a driver can panic [here](https://github.com/docker/buildx/blob/afcb609966dc748a91cb5889fa63a34034117495/driver/driver.go#L96) due to a nil `progress.Writer`.

The flow seems to be:
- In build cmd drivers are resolved explicitly [here](https://github.com/docker/buildx/blob/afcb609966dc748a91cb5889fa63a34034117495/build/build.go#L195) via `nodeResolver.Resolve`
- This calls `nodeResolver.boot` [here](https://github.com/docker/buildx/blob/afcb609966dc748a91cb5889fa63a34034117495/build/driver.go#L94) with a valid `progress.Writer`
- Later there is another boot call [here](https://github.com/docker/buildx/blob/afcb609966dc748a91cb5889fa63a34034117495/build/build.go#L357) via `resolvedNode.Client`
- Which calls `nodeResolver.boot` [here](https://github.com/docker/buildx/blob/afcb609966dc748a91cb5889fa63a34034117495/build/driver.go#L32) with a nil `progress.Writer`
- The actual problem is that later on,  [here](https://github.com/docker/buildx/blob/afcb609966dc748a91cb5889fa63a34034117495/driver/driver.go#L84), `driver.Boot` calls `driver.Info` and, if the status is not `Running`, it tries to bootstrap it again.
- This panics due to the nil `progress.Writer`

In order to reproduce this the driver must not be Running by the time the second boot call happens, which is a bit of an edge case, but possible AFAICT. 
In order to force this to reproduce we can simply patch e.g. the `remote` driver to always return Inactive.

```
func (d *Driver) Info(ctx context.Context) (*driver.Info, error) {
	return &driver.Info{
		Status: driver.Inactive,
	}, nil
}
```

### Expected behaviour

No panic. An error message related to the driver state not being Running.

### Actual behaviour

```
◊ ./bin/build/buildx create --name remote --driver remote tcp://localhost:1234
◊ ./bin/build/buildx build --builder remote --progress plain .
#0 building with "remote" instance using remote driver

#1 [internal] waiting for connection
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1017dce98]

goroutine 154 [running]:
github.com/docker/buildx/driver.Boot({0x102d4b880, 0x140003fee10}, {0x102d4b880, 0x140003fe5a0}, 0x14000581f90, {0x0?, 0x0})
        github.com/docker/buildx/driver/driver.go:96 +0x98
github.com/docker/buildx/build.(*nodeResolver).boot.func1.1({0x102d4b880?, 0x140003fee10?})
        github.com/docker/buildx/build/driver.go:244 +0x5c
github.com/moby/buildkit/util/flightcontrol.(*call[...]).run(0x102d58000)
        github.com/moby/buildkit@v0.13.0-rc3.0.20240417151852-71f99c52a669/util/flightcontrol/flightcontrol.go:121 +0x1e8
sync.(*Once).doSlow(0x0?, 0x100b95a54?)
        sync/once.go:74 +0x100
sync.(*Once).Do(0x1400061a0f0?, 0x140004d5200?)
        sync/once.go:65 +0x24
created by github.com/moby/buildkit/util/flightcontrol.(*call[...]).wait in goroutine 150
        github.com/moby/buildkit@v0.13.0-rc3.0.20240417151852-71f99c52a669/util/flightcontrol/flightcontrol.go:165 +0x504
```

### Buildx version

github.com/docker/buildx v0.14.0-36-gafcb6099.m afcb609966dc748a91cb5889fa63a34034117495.m

### Docker info

_No response_

### Builders list

```text
NAME/NODE                          DRIVER/ENDPOINT                                      STATUS     BUILDKIT   PLATFORMS
remote                             remote
 \_ remote0                         \_ tcp://localhost:1234                             running    v0.13.2    linux/arm64
default                            docker
 \_ default                         \_ default                                          running    v0.13.1    linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
desktop-linux*                     docker
 \_ desktop-linux                   \_ desktop-linux                                    running    v0.13.1    linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
```

### Configuration

# buildx checkout
◊ ./bin/build/buildx create --name remote --driver remote tcp://localhost:1234
◊ ./bin/build/buildx build --builder remote --progress plain .

### Build logs

_No response_

### Additional info

_No response_

tonistiigi:
fixes #2464

In addition to panic fix in the second commit, this fixes the inefficiency where `driver.Boot` could be called multiple times, resulting multiple docker API calls for example in the case of container driver.

I put the fixes to multiple commits in case we want to only cherry pick the second one but the first one is also pretty bad imho.

This could use further refactoring. Passing the `nil` logger should not be necessary if all the resolvedNode instances are guaranteed to already be booted and caps loaded.

### DIFF
--- a/build/driver.go
+++ b/build/driver.go
@@ -3,6 +3,7 @@ package build
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/containerd/containerd/platforms"
 	"github.com/docker/buildx/builder"
@@ -46,10 +47,22 @@ func (dp resolvedNode) BuildOpts(ctx context.Context) (gateway.BuildOpts, error)
 
 type matchMaker func(specs.Platform) platforms.MatchComparer
 
+type cachedGroup[T any] struct {
+	g       flightcontrol.Group[T]
+	cache   map[int]T
+	cacheMu sync.Mutex
+}
+
+func newCachedGroup[T any]() cachedGroup[T] {
+	return cachedGroup[T]{
+		cache: map[int]T{},
+	}
+}
+
 type nodeResolver struct {
-	nodes   []builder.Node
-	clients flightcontrol.Group[*client.Client]
-	opt     flightcontrol.Group[gateway.BuildOpts]
+	nodes     []builder.Node
+	clients   cachedGroup[*client.Client]
+	buildOpts cachedGroup[gateway.BuildOpts]
 }
 
 func resolveDrivers(ctx context.Context, nodes []builder.Node, opt map[string]Options, pw progress.Writer) (map[string][]*resolvedNode, error) {
@@ -63,7 +76,9 @@ func resolveDrivers(ctx context.Context, nodes []builder.Node, opt map[string]Op
 
 func newDriverResolver(nodes []builder.Node) *nodeResolver {
 	r := &nodeResolver{
-		nodes: nodes,
+		nodes:     nodes,
+		clients:   newCachedGroup[*client.Client](),
+		buildOpts: newCachedGroup[gateway.BuildOpts](),
 	}
 	return r
 }
@@ -179,6 +194,7 @@ func (r *nodeResolver) resolve(ctx context.Context, ps []specs.Platform, pw prog
 			resolver:    r,
 			driverIndex: 0,
 		})
+		nodeIdxs = append(nodeIdxs, 0)
 	} else {
 		for i, idx := range nodeIdxs {
 			node := &resolvedNode{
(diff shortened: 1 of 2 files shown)


build/driver.go
```diff
@@ -179,6 +194,7 @@ func (r *nodeResolver) resolve(ctx context.Context, ps []specs.Platform, pw prog
 			resolver:    r,
 			driverIndex: 0,
 		})
+		nodeIdxs = append(nodeIdxs, 0)
```
tonistiigi, line 197: This seemed to be another issue. Atm. because first boot was called with empty index array this accidentally saved us from one more extra uncached call (the real first boot happened when `opts()` was first called). Now there is one more call to `boot()` but because all calls are cached the number of calls does not matter anymore.

jedevc (approved):

PR #2475, opened 2024-05-25
remove uses of github.com/docker/docker/builder/remotecontext package

thaJeztah:
This package is part of the classic builder, and was currently only used for the IsURL utility, which is a very rudimentary check for a string having a "https://" or "http://" scheme.

This patch copies the code as non-exported functions where they're used to remove the dependency.


build/utils.go
```diff
@@ -26,8 +25,15 @@ const (
 	mobyHostGatewayName = "host-gateway"
 )
 
+// isURL returns true if the provided str is an HTTP(S) URL by checking if it
+// has a http:// or https:// scheme. No validation is performed to verify if the
+// URL is well-formed.
+func isURL(str string) bool {
```
jedevc, line 31: `isHTTPURL` or similar would be a better name here. We also have `isRemoteURL` so would be good to avoid confusion for this.

thaJeztah: Ah, yeah, I could rename; I went for keeping it as close to the original, but its a horrible function in either case, and even considered just inlining the code to even more discourage using it.

thaJeztah: Updated; PTAL 🤗

crazy-max (approved):

build/utils.go
```diff
@@ -26,8 +25,15 @@ const (
 	mobyHostGatewayName = "host-gateway"
 )
 
+// isHTTPURL returns true if the provided str is an HTTP(S) URL by checking if it
+// has a http:// or https:// scheme. No validation is performed to verify if the
+// URL is well-formed.
+func isHTTPURL(str string) bool {
+	return strings.HasPrefix(str, "https://") || strings.HasPrefix(str, "http://")
+}
```
crazy-max, line 33: I see we have `isHTTPSource` in BuildKit repo: https://github.com/moby/buildkit/blob/41c868b0c92bb176a1974f03ca45011005b02f1c/frontend/dockerfile/dockerfile2llb/convert.go#L2007-L2009

Wonder if we should not export this func in https://github.com/moby/buildkit/tree/master/util/urlutil ?

thaJeztah: Hm.. yeah, these are all a bit messy; not sure if we want to import dockerfile front-end code into the CLI's ?

I also want to avoid a bit to make it too much a "public" / exported function; I recall I added the GoDoc on this function when it was in `pkg/urlutil` (which didn't indicate "for build") after I found that there were various projects that started using it to validate if a string was a URL; the check if does is really limited, and while it's fine within the context it's used here, it's for sure not something anyone else should be using (unless they really know what they're doing).

My mean reason for this PR was to get rid of packages in buildx that are outside of "api", "client" or "pkg". 

_ideally_ we would have a single package for these kind of things that's used by both BuildKit/Buildx and the classic builder, but I don't want to re-introduce buildkit as dependency on the docker CLI; not sure where else we could put this code though 😞 


FWIW;  the other utility in this package is for checking Git URLs; that one is more complicated, but I think there may be an equivalent in BuildKit code, so again for that one .. not sure where we can put it to allow sharing, but without re-introducing BuildKit in the CLI (or elsewhere) 😞 https://github.com/docker/buildx/blob/f98ef00ec7f3fa08303aa79501749d680041d502/vendor/github.com/docker/docker/builder/remotecontext/urlutil/urlutil.go#L24-L86

crazy-max: Would https://github.com/moby/sys be a good fit?

thaJeztah: Hm... not sure; these functions are really build-specific; at least the `GitURL` ones definitely follow conventions specific to docker build.

I think the "IsURL" / "IsHTTPURL" ones are trivial enough for them to be ok to be inlined / local functions, but of course it's possible that at some point there will be more advanced rules there.


_Maybe_ `patternmatcher` repo https://github.com/moby/patternmatcher, but that's also quite a bit of a stretch I think. So alternatively something completely separate (some module maintaining "build contexts"?) but not sure.

tonistiigi (approved):
A subpackage of https://github.com/moby/buildkit/tree/master/frontend/dockerui could be a candidate for such definitions in the future.

PR #2477, opened 2024-05-25
build: loadInputs: remove redundant checks for hasTag, hasDigest

thaJeztah:
### build: loadInputs: remove redundant check for hasTag

hasTag was always true as it was set to "true" when missing, in which case
the default (`:latest`) tag was applied; https://github.com/docker/buildx/blob/37c4ff09448f30204b141a757d46ccc11982762e/build/opt.go#L500-L504

### build: loadInputs: remove redundant check for hasDigest

hasDigest would always be true when reaching this code, because the function
would return with an error when failing to resolve the digest; https://github.com/docker/buildx/blob/37c4ff09448f30204b141a757d46ccc11982762e/build/opt.go#L528-L530

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -500,7 +500,6 @@ func loadInputs(ctx context.Context, d *driver.DriverHandle, inp Inputs, addVCSL
 			localPath, tag, hasTag := strings.Cut(localPath, ":")
 			if !hasTag {
 				tag = "latest"
-				hasTag = true
 			}
 			idx := ociindex.NewStoreIndex(localPath)
 			if !hasDigest {
@@ -543,10 +542,7 @@ func loadInputs(ctx context.Context, d *driver.DriverHandle, inp Inputs, addVCSL
 			}
 			target.OCIStores[storeName] = store
 
-			layout := "oci-layout://" + storeName
-			if hasTag {
-				layout += ":" + tag
-			}
+			layout := "oci-layout://" + storeName + ":" + tag
 			if hasDigest {
 				layout += "@" + dig
 			}
--- a/build/opt.go
+++ b/build/opt.go
@@ -542,12 +542,7 @@ func loadInputs(ctx context.Context, d *driver.DriverHandle, inp Inputs, addVCSL
 			}
 			target.OCIStores[storeName] = store
 
-			layout := "oci-layout://" + storeName + ":" + tag
-			if hasDigest {
-				layout += "@" + dig
-			}
-
-			target.FrontendAttrs["context:"+k] = layout
+			target.FrontendAttrs["context:"+k] = "oci-layout://" + storeName + ":" + tag + "@" + dig
 			continue
 		}
 		st, err := os.Stat(v.Path)


thaJeztah:
Interesting; is this a bug in the linting code? It's not the cause of the failure, but I noticed the warning, and I don't see where that warning should be printed for.

https://github.com/docker/buildx/actions/runs/9236288070/job/25411984638?pr=2477

```
+ docker buildx build --platform=linux/amd64 --output=type=oci,dest=/tmp/buildx-output.g5SwiTMBk2/buildx-test-build.tar --builder=buildx-test-673362b99065c7191be9631bb810a2ec --metadata-file=/tmp/buildx-output.g5SwiTMBk2/metadata-build.json /tmp/buildx-output.g5SwiTMBk2
#0 building with "buildx-test-673362b99065c7191be9631bb810a2ec" instance using kubernetes driver

#1 transferring dockerfile: 371B done
#1 WARN: Lint Rule 'FromAsCasing': 'as' and 'FROM' keywords' casing do not match (line 1)
#1 DONE 0.0s

#2 [internal] load metadata for docker.io/library/busybox:latest
#2 ERROR: failed to do request: Head "https://registry-1.docker.io/v2/library/busybox/manifests/latest": dial tcp: lookup registry-1.docker.io on 10.43.0.10:53: read udp 10.42.0.3:37129->10.43.0.10:53: read: connection refused
------
 > [internal] load metadata for docker.io/library/busybox:latest:
------

 - Lint Rule 'FromAsCasing': 'as' and 'FROM' keywords' casing do not match (line 1)
Dockerfile:6
--------------------
   4 |     RUN echo "I am running on $BUILDPLATFORM, building for $TARGETPLATFORM" > /log
   5 |     
   6 | >>> FROM busybox AS log
   7 |     COPY --from=build /log /log
   8 |     RUN cat /log
--------------------
ERROR: failed to solve: busybox: failed to resolve source metadata for docker.io/library/busybox:latest: failed to do request: Head "https://registry-1.docker.io/v2/library/busybox/manifests/latest": dial tcp: lookup registry-1.docker.io on 10.43.0.10:53: read udp 10.42.0.3:37129->10.43.0.10:53: read: connection refused
+ docker buildx rm buildx-test-673362b99065c7191be9631bb810a2ec
buildx-test-673362b99065c7191be9631bb810a2ec removed
make: *** [Makefile:54: test-driver] Error 1
Error: Process completed with exit code 2.
```

It's also confusing that it prints both "line 1" and "line 6";

```
casing do not match (line 1)
Dockerfile:6
```

cc @daghack ☝️

crazy-max (approved):
LGTM

tonistiigi:
@jedevc Is this a bug from https://github.com/docker/buildx/commit/60c9cf74ce8baf523363137104316746daac238a#diff-7fdd319c657fbde54b83a5dfa233c5c2c417b291739ff75c05dc5d83ee7a9d24R1507 ? Doesn't look correct to assume that oci-layout has annotation with the "latest" tag. Even if it gets ignored later the client never asked for a callback via tag.

thaJeztah:
Oh! Did I mess this up, and bake in a bug? 🙈 Sorry if I did!!)

jedevc:
Mm @tonistiigi is right, the original logic is wrong - we should only default to `:latest` if no tag *or* digest is applied.

I think somehow though that this "works" - the digest is still applied and takes precendence, even though there is an incorrect tag.

tonistiigi:
@thaJeztah Re https://github.com/docker/buildx/pull/2477#issuecomment-2131304766

The warning is on line1 , not line6 where the error happened. https://github.com/docker/buildx/blob/v0.14.1/hack/test-driver#L112

PR #2478, opened 2024-05-25
build: loadInputs: extract resolving digest to a separate function

thaJeztah:
- follow-up to / stacked on https://github.com/docker/buildx/pull/2477

### build: loadInputs: extract resolving digest to a separate function

This makes the code slightly more idiomatic, but the errors produced will
change slightly to prevent having to path NamedContext as argument.


thaJeztah:
This is the follow-up to https://github.com/docker/buildx/pull/2477. Let me know if this makes sense to do (I think it's somewhat more readable / idiomatic than the existing code), but happy to make changes 😄

crazy-max (approved):
LGTM thanks!

jedevc (approved):

Issue #2481, opened 2024-05-29
`imagetools create` with single source changes copied image from `application/vnd.docker.distribution.manifest.v2+json` to `application/vnd.docker.distribution.manifest.list.v2+json`


rvoh-tismith:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

Based on documentation from https://docs.docker.com/reference/cli/docker/buildx/imagetools/create/ that says:

> The source manifests can be manifest lists or single platform distribution manifests and must already exist in the registry where the new manifest is created. If only one source is specified, create performs a carbon copy.

I would expect that running `imagetools create` against one source with a mediaType of `application/vnd.docker.distribution.manifest.v2+json` and a configType of `application/vnd.docker.container.image.v1+json` to produce a "carbon copy" that has the same mediaType and configType, however instead I end up with a copy that is now a manifest list, or in other words, it now has a mediaType of `application/vnd.docker.distribution.manifest.list.v2+json`.

### Expected behaviour

We have an image at (made-up registry and repo obviously) `foo/bar:oldimage` with the following manifest (shrunk for readability):

```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "config": {
    "mediaType": "application/vnd.docker.container.image.v1+json",
    "digest": "...",
    "size": 3060
  },
  "layers": [...]
}
```

I would expect that running the command `docker buildx imagetools create -t foo/bar:newimage foo/bar:oldimage`, which has a single source that is a single platform distribution manifest, would create a new image at `foo/bar:newimage` with a manifest that basically looks the same as the above with the same `mediaType` and `configType`.

### Actual behaviour

Instead, the image at `foo/bar:newimage` is now a manifest list and looks more like:

```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "...",
      "size": 1687,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```

### Buildx version

[...]

PR #2482, opened 2024-05-29
Add `--prefer-index` flag for`imagetools create` on a single source

rvoh-tismith:
- Fixes issue #2481 
- I updated the tests that were affected by this change to test the new logic, however there were some tests in the repo that seem to have already been failing before I made any changes, so the few tests that are failing shouldn't have anything to do with my changes
- As a response to comments below, the default behavior isn't being changed -- the new logic is now behind the flag `--prefer-index=false`
- As a response to the comments below, I removed the mediaType check that I originally just added to because I can't really see a reason for it at this point
- Updated docs

### DIFF
--- a/util/imagetools/create.go
+++ b/util/imagetools/create.go
@@ -80,12 +80,15 @@ func (r *Resolver) Combine(ctx context.Context, srcs []*Source, ann []string, pr
 	// on single source, return original bytes
 	if len(srcs) == 1 && len(ann) == 0 {
 		switch srcs[0].Desc.MediaType {
-		case images.MediaTypeDockerSchema2Manifest:
+		// if the source is already an image index or manifest list, there is no need to consider the value
+		// of preferIndex since if set to true then the source is already in the preferred format, and if false
+		// it doesn't matter since we're not going to split it into separate manifests
+		case images.MediaTypeDockerSchema2ManifestList, ocispec.MediaTypeImageIndex:
+			return dts[0], srcs[0].Desc, nil
+		default:
 			if !preferIndex {
 				return dts[0], srcs[0].Desc, nil
 			}
-		case images.MediaTypeDockerSchema2ManifestList, ocispec.MediaTypeImageIndex:
-			return dts[0], srcs[0].Desc, nil
 		}
 	}
 
--- a/util/imagetools/create.go
+++ b/util/imagetools/create.go
@@ -79,7 +79,7 @@ func (r *Resolver) Combine(ctx context.Context, srcs []*Source, ann []string) ([
 
 	// on single source, return original bytes
 	if len(srcs) == 1 && len(ann) == 0 {
-		if mt := srcs[0].Desc.MediaType; mt == images.MediaTypeDockerSchema2ManifestList || mt == ocispec.MediaTypeImageIndex {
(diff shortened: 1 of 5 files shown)


util/imagetools/create.go
```diff
@@ -79,7 +79,7 @@ func (r *Resolver) Combine(ctx context.Context, srcs []*Source, ann []string) ([
 
 	// on single source, return original bytes
 	if len(srcs) == 1 && len(ann) == 0 {
-		if mt := srcs[0].Desc.MediaType; mt == images.MediaTypeDockerSchema2ManifestList || mt == ocispec.MediaTypeImageIndex {
+		if mt := srcs[0].Desc.MediaType; mt == images.MediaTypeDockerSchema2ManifestList || mt == images.MediaTypeDockerSchema2Manifest || mt == ocispec.MediaTypeImageIndex {
```
tonistiigi, line 82: If that is the desired behavior then why is there a mediatype check at all, or at least why is `ocispec.MediaTypeImageManifest` missing?

tonistiigi (commented):
While I tend to agree that this makes more sense as a default behavior, this is removing the capability to convert image manifest into manifest list. Should we add a new flag that keeps previous behavior?

rvoh-tismith:
> While I tend to agree that this makes more sense as a default behavior, this is removing the capability to convert image manifest into manifest list. Should we add a new flag that keeps previous behavior?

@tonistiigi great point. I was too focused on my own use case, but yeah, I definitely agree that some people are probably using this explicitly hoping to convert to a manifest list, or at the very least they've probably built workflows that are expecting this behavior, so changing the default behavior is probably too risky.

So, 2 things:

1. I think a flag for this behavior is the right way to go. I'm not married to the name, but maybe something like `--preserve-single-source`, so `docker buildx imagetools create --preserve-single-source=true -t foo/bar:newimage foo/bar:oldimage`.
2. Like I mentioned above, it feels like the behavior here should somewhat align with the behavior of `docker buildx build` which only outputs a manifest list if you specify multiple platforms. What are your thoughts there? Do you think there needs to be some alignment, or maybe the same flag for `build` so that you can control this behavior in both places? (that doesn't necessarily have to be included in this pr)

Thoughts?

util/imagetools/create.go
rvoh-tismith: Ahhhhh great point! Yeah, it actually seems reasonable that there doesn't even need to be a check. Thinking through the logic, I guess I can't really think of when the logic would need to be different here. Seems like if there is one source and no annotations (and possibly if the `--preserve-single-source` flag is set to `true`, based on my response to your other comment), we should just return the original bytes. I almost feel like I'm missing something here though. It feels like this is here for a reason, but maybe it is just an oversight.

tonistiigi:
Maybe smth. like `--prefer-manifest-index` `--create-index` etc. for the behavior of always creating the index result and without the flag, it is the original manifest on a single source.

rvoh-tismith:
> Maybe smth. like `--prefer-manifest-index` `--create-index` etc. for the behavior of always creating the index result and without the flag, it is the original manifest on a single source.

Nice. I'll make some changes and push them up.

rvoh-tismith:
I split the difference and landed on `--prefer-index` lol. I also removed the mediaType checks.

tonistiigi (approved):

PR #2485, opened 2024-05-30
Add stale issue and PR labeler workflows

thompson-shaun:
This PR adds [**Close Stale Issues**](https://github.com/marketplace/actions/close-stale-issues#exempt-issue-labels) and [**PR Labeler**](https://github.com/marketplace/actions/labeler) action workflows. Durations for stale item comments and closure are initial suggestions.

A first attempt was made in a [previous PR](https://github.com/docker/buildx/pull/2450) that had great comments (thanks @crazy-max and @thaJeztah). This iteration drops the 3rd-party auto-labeler and strictly focuses on official GitHub actions. 

## Goals

- Encourage timely processing of Issues and PRs
- Automatically identify areas being modified by a PR (slight toil reduction, faster binning, may yield insights over time).

## Close Stale Issues

> Warns and then closes issues and PRs that have had no activity for a specified amount of time.

### General

The configuration as proposed will exempt Issues and PRs that have been assigned milestones (even arbitrarily long ones such `v-future`). In addition, the workflow would process only items created after the `start-date`. We can adjust this date as desired. There is also an option to exempt items that have been assigned to a user.

```yaml
exempt-milestones: true
start-date: '2024-05-01T00:00:00+0000'
```

### Issues

Notable [configuration options](https://github.com/marketplace/actions/close-stale-issues#all-options)

```yaml
# durations before stale and close actions are taken
days-before-issue-stale: <integer>
days-before-issue-close: <integer>
# labels applied as the issue transitions from active -> stale -> closed
stale-issue-label: 'stale' 
close-issue-label: 'rotten'
# labels that can be applied to avoid stale processing
exempt-issue-labels: 'frozen'
```

### Pull Requests

Notable [configuration options](https://github.com/marketplace/actions/close-stale-issues#all-options)

```yaml
# durations before stale and close actions are taken
days-before-pr-stale: <integer>
days-before-pr-close: <integer>
# labels applied as the issue transitions from active -> stale -> closed
stale-pr-label: 'stale'
close-pr-label: 'rotten'
# labels that can be applied to avoid stale processing
exempt-pr-labels: 'frozen'
```

## Pull Request Labeler

> Automatically label new pull requests based on the paths of files being changed or the branch name.

At present, this is primarily applying `area/*` labels but can be extended if it helps.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,104 @@
+
+# Add 'area/project' label to changes in basic project documentation and .github folder, excluding .github/workflows
+area/project:
+- all:
+  - changed-files:
+    - any-glob-to-any-file: 
+      - .github/**
+      - LICENSE
+      - AUTHORS
+      - MAINTAINERS
+      - PROJECT.md
+      - README.md
+      - .gitignore
+      - codecov.yml
+    - all-globs-to-all-files: '!.github/workflows/*'
+
+# Add 'area/github-actions' label to changes in the .github/workflows folder
+area/ci:
+  - changed-files:
+    - any-glob-to-any-file: '.github/workflows/**'
+
+# Add 'area/bake' label to changes in the bake
+area/bake:
+  - changed-files:
+    - any-glob-to-any-file: 'bake/**'
+
+# Add 'area/bake/compose' label to changes in the bake+compose
+area/bake/compose:
+  - changed-files:
+    - any-glob-to-any-file:
+      - bake/compose.go
+      - bake/compose_test.go
+
+# Add 'area/build' label to changes in build files
+area/build:
+  - changed-files:
+    - any-glob-to-any-file: 'build/**'
+
+# Add 'area/builder' label to changes in builder files
+area/builder:
+  - changed-files:
+    - any-glob-to-any-file: 'builder/**'
+
+# Add 'area/cli' label to changes in the CLI
+area/cli:
+  - changed-files:
+    - any-glob-to-any-file: 
+      - cmd/**
+      - commands/**
+
+# Add 'area/controller' label to changes in the controller
+area/controller:
+  - changed-files:
+    - any-glob-to-any-file: 'controller/**'
+
+# Add 'area/docs' label to markdown files in the docs folder
+area/docs:
+  - changed-files:
+    - any-glob-to-any-file: 'docs/**/*.md'
+
+# Add 'area/dependencies' label to changes in go dependency files
+area/dependencies:
+  - changed-files:
+    - any-glob-to-any-file:
+      - go.mod
+      - go.sum
+      - vendor/**
+
+# Add 'area/driver' label to changes in the driver folder
+area/driver:
+  - changed-files:
+    - any-glob-to-any-file: 'driver/**'
+
+# Add 'area/driver/docker' label to changes in the docker driver
+area/driver/docker:
+  - changed-files:
+    - any-glob-to-any-file: 'driver/docker/**'
+
+# Add 'area/driver/docker-container' label to changes in the docker-container driver
+area/driver/docker-container:
+  - changed-files:
+    - any-glob-to-any-file: 'driver/docker-container/**'
+
+# Add 'area/driver/kubernetes' label to changes in the kubernetes driver
+area/driver/kubernetes:
+  - changed-files:
+    - any-glob-to-any-file: 'driver/kubernetes/**'
+
+# Add 'area/driver/remote' label to changes in the remote driver
+area/driver/remote:
+  - changed-files:
+    - any-glob-to-any-file: 'driver/remote/**'
+
+# Add 'area/hack' label to changes in the hack folder
+area/hack:
+  - changed-files:
+    - any-glob-to-any-file: 'hack/**'
+
+# Add 'area/tests' label to changes in test files
+area/tests:
+  - changed-files:
+    - any-glob-to-any-file: 
+      - tests/**
+      - '**/*_test.go'
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,19 @@
+name: labeler
+
+concurrency:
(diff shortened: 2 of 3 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,117 @@
...
+
+# Add 'area/hack' label to changes in the hack folder
+area/hack:
+  - changed-files:
+    - any-glob-to-any-file: 'hack/**'
+
+# Add 'area/tests' label to changes in test files
+area/tests:
+  - changed-files:
+    - any-glob-to-any-file: 'tests/**'
```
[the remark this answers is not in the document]
thompson-shaun: Added

.github/labeler.yml
```diff
@@ -0,0 +1,114 @@
...
+
+## Add 'impact/docs' label to markdown files in the docs folder
+impact/docs:
+  - changed-files:
+    - any-glob-to-any-file: 'docs/**/*.md'
+
+# Add 'needs/docs-review' label to markdown files in the docs folder
+needs/docs-review:
+  - changed-files:
+    - any-glob-to-any-file: 'docs/**/*.md'
```
crazy-max, line 69: IIUC it will always set 3 labels when docs changes which is not ideal as there is no specific distinction. I think we should just use `area/docs` imo.

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
...
+- pull_request_target
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@v5
```
crazy-max, line 12: Let's keep the same syntax as other workflows:

```suggestion
name: labeler

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

on:
  pull_request_target:

jobs:
  labeler:
    permissions:
      contents: read
      pull-requests: write
    runs-on: ubuntu-latest
    steps:
      -
        name: Run
        uses: actions/labeler@v5
```

.github/workflows/close-stale-issues.yml
crazy-max, line 1: Can we use a short filename like `stale.yml`. I think this is self-explanatory.

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,34 @@
...
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
```
crazy-max, line 15: Same as previous comments and use `stale` as name to match workflow filename:

```suggestion
name: stale

on:
  schedule:
    - cron: '30 1 * * *'

permissions:
  issues: write
  pull-requests: write

jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
      -
        name: Run
        uses: actions/stale@v9
        with:
```

crazy-max (changes requested):
Some suggestions, can you also squash commits please? Thanks

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,37 @@
+name: stale
+
+on:
+  schedule:
+    - cron: '30 1 * * *'
```
crazy-max, line 5: Does not hurt to add a `workflow_dispatch` event if we want to trigger that workflow manually to make sure everything looks good:

```suggestion
  workflow_dispatch:
  schedule:
    - cron: '30 1 * * *'
```

.github/workflows/close-stale-issues.yml
```diff
@@ -0,0 +1,37 @@
...
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Run
+        uses: actions/stale@v9
```
crazy-max, line 17: Looking at this action it would be good to handle `debug-only` to check what the action would do first: https://github.com/actions/stale/blob/3f3b0175e8c66fb49b9a6d5a0cd1f8436d4c3ab6/action.yml#L131-L134

And related to previous comment we could add an input to the `workflow_dispatch` event named `dry-run`. Would be similar to: https://github.com/docker/packaging/blob/3ac603d1bfbf4f1e347c2a35e75585afae9d7276/.github/workflows/release-buildx.yml#L18-L21

image

I can push an extra commit for this if you want.

.github/labeler.yml
thompson-shaun: The `impact/docs` attempted to use the existing labels. The `needs/docs-review` is related to another PR that will be coming 🔜

.github/workflows/labeler.yml
thompson-shaun: Absolutely!

.github/workflows/close-stale-issues.yml
thompson-shaun: Go for it :flex

.github/workflows/close-stale-issues.yml
thompson-shaun: Renamed

crazy-max:
Pushed extra commits for stale workflow.

First one to prefix labels with `status/` and add them to the repo: https://github.com/docker/buildx/labels?q=status%2F

image

Move permissions to job similar to `labeler` workflow.

And last one related to https://github.com/docker/buildx/pull/2485#discussion_r1622450951 so we can manually trigger the workflow to check its behavior. Tested in:
* dry-run ON: https://github.com/crazy-max/buildx/actions/runs/9346230975
* dry-run OFF: https://github.com/crazy-max/buildx/actions/runs/9346248671

image

crazy-max (approved):
LGTM

PTAL @tonistiigi @thaJeztah as I pushed extra commits :pray:

thompson-shaun:
I'm going to break apart this submission into multiple parts to help isolate changes that might need more discussion. 

- PR Labeler
- Stale issue processing + project processing guide

PR #2487, opened 2024-05-30
Rename --print to --call and make previous name hidden

daghack:
This adds the `--call` flag, which allows for calling out to specific frontend method. This is effectively just an alias for the `--print` experimental flag that both brings the feature out of experimental and renames it, while maintaining backwards compatibility with the already existing flag.


commands/build.go
```diff
@@ -639,6 +639,9 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 	var ignoreBool bool
 	var ignoreInt int64
 
+	flags.StringVar(&options.printFunc, "call", "", "Call a specific frontend method")
+	flags.MarkHidden("call")
```
tonistiigi, line 643: I guess the previous `--print` needs to be the hidden one.

util/buildflags/printfunc.go
```diff
@@ -13,6 +13,9 @@ func ParsePrintFunc(str string) (*controllerapi.PrintFunc, error) {
 	if str == "" {
 		return nil, nil
 	}
+	if str == "check" {
```
tonistiigi, line 16: This needs a comment for why we are doing it.

tonistiigi (commented):
I guess we also want to remove experimental from `--print` (and leave it hidden)

daghack:
Ah, yes! I think I misunderstood the goal slightly! Updated!

commands/build.go
```diff
@@ -633,12 +628,18 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 		cobrautil.MarkFlagsExperimental(flags, "root", "detach", "server-config")
 	}
 
+	flags.StringVar(&options.printFunc, "call", "", "Call a specific frontend method")
```
crazy-max, line 631: As this is a new flag, we should have some documentation with basic use cases in [`Examples`](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#examples) before having a dedicated page for it on docs website (probably in "Building images" section, WDYT @dvdksn?):

image

crazy-max: Needs to also align tests like changing `testBuildPrint` to `testBuildCall` in https://github.com/docker/buildx/blob/7427adb9b0e78f7d86a73acc6d2179f2119d726e/tests/build.go#L800 and using this new flag.

dvdksn: I'm concerned about the name of this flag (and the description of it). I don't expect most users to know what a frontend method is. Buildx should have an opinionated UX, but this seems like the opposite, where it's just a very shallow layer on top of buildkit's frontends.

thompson-shaun: @colinhemmings  - can we get a better description for `call`? plz

colinhemmings: I agree that the user won't know what frontend is. This flag is for setting the build method use. So I would recommend the following help  `Set method for evaluating build ("check", "outline", "targets")`.
@dvdksn thoughts?

thompson-shaun: Updated with the suggested description @colinhemmings 🫡

dvdksn: The description is better. I still have reservations about the flag name. To me, a "call" flag doesn't really imply that this changes the build request type from the default "build an artifact from source" to something else. Neither did the `--print` flag really (maybe if you squinted hard enough), but that one was experimental. Another flag name that's more descriptive might be:

```
--mode string          Set method for evaluating build (default "build")
```

Even with a different flag name, having this switch on the `build` command is a bit misleading because the user is no longer building anything. Subrequests (as far as my understanding goes) is meant for other tasks, like analyzing source files, "introspection", or some custom output generation.

commands/build.go
```diff
@@ -633,12 +628,18 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 		cobrautil.MarkFlagsExperimental(flags, "root", "detach", "server-config")
 	}
 
+	flags.StringVar(&options.printFunc, "call", "", `Set method for evaluating build ("check", "outline", "targets")`)
```
crazy-max, line 631: nit: we could use subrequests names from BuildKit instead of hardcoding like: https://github.com/moby/buildkit/blob/03d2f9fc4074ea8bf36cea2db8d0829f5442864b/frontend/subrequests/lint/lint.go#L19

Would probably need to move these consts to another package in https://github.com/moby/buildkit/tree/03d2f9fc4074ea8bf36cea2db8d0829f5442864b/frontend/subrequests

commands/build.go
```diff
@@ -629,6 +629,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 	}
 
 	flags.StringVar(&options.printFunc, "call", "build", `Set method for evaluating build ("check", "outline", "targets")`)
+	cobrautil.BoolFunc(flags, "check", `Shorthand for "--call=check"`, callAlias(options, "check"))
```
tonistiigi, line 632: Can we do it so that the definition in here looks like:

```
flags.Var(&cobrautil.Smth{callAlias()}) , "check",  `Sho..`)
```
or
```
flags.Var(callAlias() , "check",  `Sho..`)
```

jsternberg: Yea. The main purpose of this syntax was to mimic the upstream syntax.

https://pkg.go.dev/flag#FlagSet.BoolFunc

tonistiigi: All other flags use cobra-style definition, not stdlib, so we should stick to that.

jsternberg: Done. I've changed it to directly expose `BoolFuncValue` instead so we can use `flags.Var` instead. `callAlias` then just returns `BoolFuncValue` to help it read a bit more easily.

commands/build.go
```diff
@@ -1003,6 +1004,20 @@ func maybeJSONArray(v string) []string {
 	return []string{v}
 }
 
+func callAlias(options *buildOptions, value string) cobrautil.BoolFuncValue {
+	return func(s string) error {
+		v, err := strconv.ParseBool(value)
```
tonistiigi, line 1009: `strconv.ParseBool(s)`

tonistiigi (approved):
Let's bring this in for RC. If there are still changes we want to make (or keep experimental) we can update before GA.

commands/build.go
tonistiigi: default "build" was added

jsternberg (approved):

PR #2491, opened 2024-05-31
vendor: update buildkit to v0.14.0-rc1

jsternberg:
Update buildkit dependency to v0.14.0-rc1. Update the tracing
infrastructure to use the new detect API which updates how the delegated
exporter is configured.

Signed-off-by: Jonathan A. Sternberg <jonathan.sternberg@docker.com>


tests/dialstdio.go
```diff
@@ -113,13 +113,10 @@ func testDialStdio(t *testing.T, sb integration.Sandbox) {
...
 
-	t.Run("conn=netpipe", func(t *testing.T) {
-		t.Parallel()
-		do(t, func(t *testing.T, cmd *exec.Cmd) net.Conn {
-			c1, c2 := net.Pipe()
-			cmd.Stdin = c1
-			cmd.Stdout = c1
-			return c2
-		})
+	do(t, func(t *testing.T, cmd *exec.Cmd) net.Conn {
```
tonistiigi, line 116: @cpuguy83 I don't exactly get what the original intention was as it only gets called once and nothing runs in the main test. The subtest itself is not an issue, but the parallel part is.

cpuguy83: 🤷 Probably something like "we can add more conn types" or something.
But yeah it's probably not necessary.

What's wrong with parallel here?

tonistiigi: > What's wrong with parallel here?

Sandbox (eg. buildkitd daemon started for it and `sb.Context()`) is shut down when the main test ends (in a defer). Parallel moves the subtest to a goroutine where it tries to access the daemon again after the `defer` is already called.

crazy-max (approved):

PR #2494, opened 2024-06-03
eval cmd to evaluate frontend requests

crazy-max:
alternative to and closes #2487 

Adds new root `eval` command to evaluate frontend requests and removes experimental flag `build --print`. It has a `build` subcommand that remains unchanged from the current user experience with `buildx build`.

`eval` requires `--request` flag to be set or one of its shorthands `--lint`, `--outline`, `--targets`.

Usage:

```
$ docker buildx eval --request=lint build .
$ docker buildx eval --request=lint,format=json build .
$ docker buildx eval --lint build .
$ docker buildx eval --targets build --build-arg=BAR=678 --target=foo .
```


docs/reference/buildx_eval_build.md
```diff
@@ -0,0 +1,56 @@
...
+| [`--target`](https://docs.docker.com/reference/cli/docker/image/build/#target)                                                                     | `string`      |           | Set the target build stage to build                                                                 |
+| `--ulimit`                                                                                                                                         | `ulimit`      |           | Ulimit options                                                                                      |
+
+
+<!---MARKER_GEN_END-->
+
+## Description
+
+Same as [`build`](buildx_build.md), but evaluate the build configuration without
+building the image.
```
crazy-max, line 56: @dvdksn Could add some docs below this description with some quick examples maybe? Similar to ones use in our integration tests: https://github.com/docker/buildx/pull/2494/files#diff-9a2da9eceb368adfae46838969a2274770a558a410fa374707616b0c5737d9b6R21

commands/build.go
```diff
@@ -597,11 +619,6 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 
 	flags.StringArrayVar(&options.platforms, "platform", platformsDefault, "Set target platform for build")
 
-	if confutil.IsExperimental() {
-		flags.StringVar(&options.printFunc, "print", "", "Print result of information request (e.g., outline, targets)")
-		cobrautil.MarkFlagsExperimental(flags, "print")
-	}
```
crazy-max, line 603: Compared to https://github.com/docker/buildx/pull/2494/files#diff-9a2da9eceb368adfae46838969a2274770a558a410fa374707616b0c5737d9b6R21, I just remove this flag as it was under experimental mode anyway.

docs/reference/buildx_eval_build.md
dvdksn: Yes, below this we would need to have at least three examples, one each for `lint`, `targets`, and `outline`. We should also link out to the lint rules reference (from the lint example).

Additionally, if we are going with this approach, we need to add a new page under the manuals section, maybe under buildkit or dockerfile, that describes build request vs subrequests, and how subrequests can be implemented in frontends to support alternate strategies for a build invocation.

colinhemmings:
> On the other hand, this means that subrequest handling becomes decoupled from the main build command. The semantics of the command changes too: "build" becomes the noun in `docker eval build`, and leaves the door open to other evaluations in the future (e.g. `eval worker`). In theory, this could also make it easier to add bake support and nicer abstractions/shorthands for subrequests, which leverage the canonical `eval` command under the hood.

@dvdksn I like the idea of the decoupling with build becoming the noun. This came up during an interview; however,, in that case,, the verb would be the action the user was trying to achieve, for example, `docker check build,` `docker check bake`, or `docker outline build.` 

Is there a different term than eval that would be more relevant to the actions the user is trying to achieve?

PR #2501, opened 2024-06-06
remote: ensure that client connection is not established twice

tonistiigi:
Because remote driver implements Info() by calling Client() internally, two instances on Client are created backed by separate TCP connection. This hack avoids it and improves performance.

### DIFF
--- a/driver/remote/driver.go
+++ b/driver/remote/driver.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/docker/buildx/driver"
@@ -14,6 +15,7 @@ import (
 	"github.com/docker/buildx/util/progress"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/client/connhelper"
+	"github.com/moby/buildkit/util/tracing/delegated"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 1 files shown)


driver/remote/driver.go
```diff
@@ -78,12 +85,18 @@ func (d *Driver) Rm(ctx context.Context, force, rmVolume, rmDaemon bool) error {
...
-			return d.Dial(ctx)
-		}),
-	}, opts...)
-	return client.New(ctx, "", opts...)
+	d.clientOnce.Do(func() {
+		opts = append([]client.ClientOpt{
+			client.WithContextDialer(func(ctx context.Context, _ string) (net.Conn, error) {
+				return d.Dial(ctx)
+			}),
+			client.WithTracerDelegate(delegated.DefaultExporter),
```
tonistiigi, line 93: This is pretty 😬 , normally called by `DriverHandle`. Mostly weird as other drivers don't need to call it.

crazy-max: But do not `client.WithTracerDelegate` already set as opt in https://github.com/docker/buildx/blob/cfef22ddf0ccff52a8aced77c6115249d78b914b/driver/remote/driver.go#L80 via https://github.com/docker/buildx/blob/cfef22ddf0ccff52a8aced77c6115249d78b914b/driver/manager.go#L167

tonistiigi: This is how it works in regular mode and for other drivers. But for remote driver the call path can be that `Info()` is called that internally calls `Client()` (and that client is now cached). In that case no delegate is set up.

Maybe a possible solution would be to move `ClientOpt` as part of `InitConfig`

crazy-max: > Maybe a possible solution would be to move `ClientOpt` as part of `InitConfig`

Ah right yeah that would be neat

tonistiigi: This still would mean that every driver would need to handle the `ClientOpt` and if they don't then at least the tracing would not work.

crazy-max (approved):

PR #2502, opened 2024-06-06
vendor: github.com/compose-spec/compose-go/v2 v2.1.2

thaJeztah:
- relates to https://github.com/compose-spec/compose-go/pull/635
- relates to https://github.com/docker/cli/pull/4744
- relates to https://github.com/mitchellh/mapstructure/issues/349

Replaces uses of the github.com/mitchellh/mapstructure module, which was deprecated by the owner and moved to new maintainership at github.com/go-viper/mapstructure.

The old module is still referenced as indirect dependency (through docker/cli and theupdateframework/notary), but not used in code, and should eventually go away.

full diff: https://github.com/compose-spec/compose-go/compare/v2.1.1...v2.1.2


crazy-max (approved):
Thanks!

PR #2507, opened 2024-06-10
Update the lint metrics to match agains the rule URL

daghack:
Update the lint metrics to match agains the rule URL rather than a prefix on the lint rule.
Required to support  https://github.com/moby/buildkit/pull/5012


util/progress/metricwriter.go
```diff
@@ -446,14 +448,22 @@ func newLintMetricRecorder(meter metric.Meter, attrs attribute.Set) *lintMetricR
...
 func (mr *lintMetricRecorder) Record(ss *client.SolveStatus) {
 	for _, warning := range ss.Warnings {
-		m := reLintMessage.FindSubmatch(warning.Short)
+		m := reLintMessage.FindSubmatch([]byte(warning.URL))
 		if m == nil {
 			continue
 		}
 
-		ruleName := string(m[1])
+		ruleName := kebabToCamel(string(m[1]))
```
tonistiigi, line 466: This should need a length check (already in old code)

tonistiigi (changes requested):

util/progress/metricwriter.go
daghack: Do you mean a length check on the result from `FindSubmatch`?

util/progress/metricwriter.go
```diff
@@ -446,14 +448,22 @@ func newLintMetricRecorder(meter metric.Meter, attrs attribute.Set) *lintMetricR
...
+	}
+	return strings.Join(words, "")
+}
+
 func (mr *lintMetricRecorder) Record(ss *client.SolveStatus) {
 	for _, warning := range ss.Warnings {
-		m := reLintMessage.FindSubmatch(warning.Short)
-		if m == nil {
+		m := reLintMessage.FindSubmatch([]byte(warning.URL))
+		if m == nil || len(m) < 2 {
```
tonistiigi, line 462: nit: technically `nil` check is not needed anymore

util/progress/metricwriter.go
tonistiigi: Yes, to avoid panic if `[]` is returned.

util/progress/metricwriter.go
daghack: Removed redundant nil!

tonistiigi (approved):

PR #2508, opened 2024-06-11
test: setup integration tests coverage

crazy-max:
follow-up https://github.com/moby/buildkit/pull/4938

https://app.codecov.io/gh/docker/buildx/pull/2508/indirect-changes

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,10 +27,36 @@ WORKDIR /src
 
 FROM gobase AS gotestsum
 ARG GOTESTSUM_VERSION
-ENV GOFLAGS=
-RUN --mount=target=/root/.cache,type=cache \
-  GOBIN=/out/ go install "gotest.tools/gotestsum@${GOTESTSUM_VERSION}" && \
-  /out/gotestsum --version
+ENV GOFLAGS=""
+RUN --mount=target=/root/.cache,type=cache <<EOT
+  set -ex
+  go install "gotest.tools/gotestsum@${GOTESTSUM_VERSION}"
+  go install "github.com/wadey/gocovmerge@latest"
+  mkdir /out
+  /go/bin/gotestsum --version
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -27,10 +27,41 @@ WORKDIR /src
...
 FROM gobase AS gotestsum
 ARG GOTESTSUM_VERSION
-ENV GOFLAGS=
-RUN --mount=target=/root/.cache,type=cache \
-  GOBIN=/out/ go install "gotest.tools/gotestsum@${GOTESTSUM_VERSION}" && \
-  /out/gotestsum --version
+ENV GOFLAGS=""
+RUN --mount=target=/root/.cache,type=cache <<EOT
+  set -ex
+  xx-go install "gotest.tools/gotestsum@${GOTESTSUM_VERSION}"
```
tonistiigi, line 33: Why is this using `xx-go` to install. Iiuc these tools should always be for native arch. And I think they are atm as I don't see `ARG TARGETPLATFORM` defined anywhere.

codecov.yml
```diff
@@ -1 +1,15 @@
 comment: false
+
+coverage:
+  status:
+    project:  # settings affecting project coverage
+      default:
+        target: auto  # auto % coverage target
+        threshold: 1%  # allow for 1% reduction of coverage without failing
```
tonistiigi, line 8: What's the behavior of this?

I'm ok with bot leaving comments if coverage goes down but I don't think we should auto-block PRs based on this.

crazy-max: I was just aligning our codecov conf with BuildKit one. I think for BuildKit we added this years ago because reports were inaccurate. We could try removing it.

tonistiigi: I'm just not sure what it even does. Not just the `1%` but this block in general.

tonistiigi (approved):

PR #2509, opened 2024-06-11
build: retain local and tar exporter attributes

crazy-max:
When investigating https://github.com/moby/buildkit/pull/5017 I found that `local` and `tar` didn't have the `dest` attribute set. I think we should keep it even if not used as we want to keep track of it in the build record.


tonistiigi (commented):
I'm not sure we should be sending this to buildkit daemon. It is a client local path. If anything, it should be part of localstate.

crazy-max:
> I'm not sure we should be sending this to buildkit daemon. It is a client local path. If anything, it should be part of localstate.

Yes indeed that's not safe, I'm looking at moving this in localstate

PR #2516, opened 2024-06-11
remove use of deprecated containerd/containerd/errdefs

thaJeztah:
This package has moved to a separate module. Also added linting rules to prevent accidental reintroduction.


crazy-max (approved):
Thanks!

thaJeztah:
I added the cherry-pick label, in case there will be a v0.15.1; this should be "no risk", but helps getting the deprecated package out of the tree; let me know if you want me to prepare a cherry-pick

PR #2522, opened 2024-06-13
Make multi-type annotation settings match docs

treuherz:
The Docker docs in multiple places describe passing an annotation at the command line like "index,manifest:com.example.name=my-cool-image", and say that this will result in the annotation being applied to both the index and the manifest. It doesn't seem like this was actually implemented, and instead it just results in an annotation key with "index,manifest:" at the beginning being applied to the manifest.

This change splits the part of the key before the colon by comma, and creates an annotation for each type/platform given, so the implementation should now match the docs.

### DIFF
--- a/util/buildflags/export.go
+++ b/util/buildflags/export.go
@@ -81,8 +81,10 @@ func ParseExports(inp []string) ([]*controllerapi.ExportEntry, error) {
 
 func ParseAnnotations(inp []string) (map[exptypes.AnnotationKey]string, error) {
 	// TODO: use buildkit's annotation parser once it supports setting custom prefix and ":" separator
-	annotationRegexp := regexp.MustCompile(`^((?:[a-z-]+(?:\[[A-Za-z0-9_/-]+\])?)(?:,[a-z-]+(?:\[[A-Za-z0-9_/-]+\])?)*:)?(\S+)$`)
+
+	// type followed by optional platform specifier in square brackets
 	annotationTypeRegexp := regexp.MustCompile(`^([a-z-]+)(?:\[([A-Za-z0-9_/-]+)\])?$`)
+
 	annotations := make(map[exptypes.AnnotationKey]string)
 	for _, inp := range inp {
 		k, v, ok := strings.Cut(inp, "=")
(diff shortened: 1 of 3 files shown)


treuherz:
I get the same test failures locally with and without this change - I think these must be my setup rather than an actual problem. Hopefully the validation will be green but I'll mark as draft again if not.

treuherz:
A dep changed from `// indirect` to direct  - fixed, re-run vendor etc. Sorry about that

util/buildflags/export_test.go
```diff
@@ -0,0 +1,118 @@
...
+					},
+					Key: "com.example",
+				}: "a",
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			got, gotErr := ParseAnnotations(test.in)
```
tonistiigi, line 89: `s/gotErr/err/`

util/buildflags/export_test.go
```diff
@@ -0,0 +1,118 @@
...
+			if test.wantErr != "" {
+				require.ErrorContains(t, gotErr, test.wantErr)
+			} else {
+				assert.NoError(t, gotErr)
```
tonistiigi, line 93: `NoError` should use `require`

util/buildflags/export.go
```diff
@@ -81,7 +81,8 @@ func ParseExports(inp []string) ([]*controllerapi.ExportEntry, error) {
 
 func ParseAnnotations(inp []string) (map[exptypes.AnnotationKey]string, error) {
 	// TODO: use buildkit's annotation parser once it supports setting custom prefix and ":" separator
-	annotationRegexp := regexp.MustCompile(`^(?:([a-z-]+)(?:\[([A-Za-z0-9_/-]+)\])?:)?(\S+)$`)
+	annotationRegexp := regexp.MustCompile(`^((?:[a-z-]+(?:\[[A-Za-z0-9_/-]+\])?)(?:,[a-z-]+(?:\[[A-Za-z0-9_/-]+\])?)*:)?(\S+)$`)
```
tonistiigi, line 84: This needs comments about why there are multiple and what's the difference

util/buildflags/export_test.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+			for k, v := range got {
+				gotKVs = append(gotKVs, kv{k, v})
+			}
+
+			sortFunc := func(a, b kv) int { return cmp.Compare(a.Key.String(), b.Key.String()) }
+			slices.SortFunc(wantKVs, sortFunc)
+			slices.SortFunc(gotKVs, sortFunc)
+
+			if diff := gocmp.Diff(wantKVs, gotKVs); diff != "" {
```
tonistiigi, line 113: I think you can just use `assert.Equal` in here. No need for `gocmp`.

util/buildflags/export_test.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			got, gotErr := ParseAnnotations(test.in)
+			if test.wantErr != "" {
+				require.ErrorContains(t, gotErr, test.wantErr)
+			} else {
+				assert.NoError(t, gotErr)
+			}
+
+			// Can't compare maps with pointer in their keys, need to extract and sort the map entries
```
tonistiigi, line 96: might make sense to make this covert code into a function and then convert both maps. Atm there is duplication where exactly the same commands need to run on both maps.

util/buildflags/export.go
treuherz: The regexp is pretty inscrutable, I might try breaking it up into cuts and splits and just using a smaller regexp for pulling out the platform specifier, if it’s all the same to you

util/buildflags/export_test.go
treuherz: done

util/buildflags/export_test.go
treuherz: done

util/buildflags/export_test.go
treuherz: done

tonistiigi (approved):

PR #2534, opened 2024-06-17
bake: print warnings on progress

tonistiigi:
Add progress warnings to the output for `bake` commands same way as they show up in `build`.

```
 - UndefinedVar: Usage of undefined variable '$PAHT' (did you mean $PATH?) (line 2)
```

```
 1 warning found:
 - UndefinedVar: Usage of undefined variable '$PAHT' (did you mean $PATH?) (line 2)
Variables should be defined before their use
More info: https://docs.docker.com/go/dockerfile/rule/undefined-var/
Dockerfile:2
--------------------
   1 |     FROM alpine
   2 | >>> ENV PATH=$PAHT:/foo
   3 |     RUN env && stop
   4 |
--------------------
```

--

There is a bit of an issue with this atm that when two targets invoked by `bake` use the same Dockerfile they are likely to generate the same warnings. Some deduplication would be needed.

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -124,8 +124,12 @@ func runBake(ctx context.Context, dockerCli command.Cli, targets []string, in ba
 	}
 
 	progressMode := progressui.DisplayMode(cFlags.progress)
-	printer, err := progress.NewPrinter(ctx2, os.Stderr, progressMode,
+	var printer *progress.Printer
+	printer, err = progress.NewPrinter(ctx2, os.Stderr, progressMode,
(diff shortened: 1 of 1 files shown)


tonistiigi:
https://github.com/docker/buildx/pull/2551/files#diff-005a09d9d83039339e043f46660ad7cebb258f5c9e2163bf5863a7567dae226bR190 started to add dedupe functionality that can be reused in here (@crazy-max )

tonistiigi:
Issue with duplication was fixed in #2551

crazy-max (approved):

Issue #438, opened 2020-11-12
bake: more structured fields instead csv


errordeveloper:
Currently some bake config files (and Go structs) inherited mini-language syntax from flags like `--secret`, `--output` and `--ssh`. It would be much more user-friendly (mostly in terms of readability) if these fields had more concrete structure native to JSON/HCL and Go.

PR #2556, opened 2024-06-26
bake: add call methods support and printing

tonistiigi:
### Define call method in the bake target

```
target "validate" {
	call = "check"
}
```

```
docker buildx bake validate
```

### Set call/check on all active targets

```
docker buildx bake --call=outline
docker buildx bake --check
docker buildx bake --call check,format=json
```

The text output of `call=outline` should be updated in follow-up to include bake definition and variables. Getting variables seems tricky and I think needs HCL parser updates.

```
» docker buildx bake binaries test --call=outline 
[+] Building 1.0s (11/11) FINISHED                                                                                                                                           0.0s
binaries

TARGET: binaries

BUILD ARG                  VALUE   DESCRIPTION
GO_VERSION                 1.21
XX_VERSION                 1.4.0
GO_EXTRA_FLAGS
BUILDKIT_SBOM_SCAN_STAGE   true


test

TARGET: test-coverage

BUILD ARG    VALUE   DESCRIPTION
GO_VERSION   1.21
XX_VERSION   1.4.0
```

JSON output contains both Bake definition and the method result.

### List all bake targets

```
docker buildx bake --list-targets
```

```
TARGET				DESCRIPTION
binaries
binaries-cross
default				binaries
image
image-cross
image-local
integration-test
integration-test-base
lint
lint-gopls			Run gopls analyzers
meta-helper
```

### List all bake variables

```
docker buildx bake --list-variables
```

```
VARIABLE			VALUE	DESCRIPTION
DESTDIR				./bin	Destination directory for build artifacts
DOCS_FORMATS			md
GOLANGCI_LINT_MULTIPLATFORM
GO_VERSION			<null>
HTTPS_PROXY
HTTP_PROXY
NO_PROXY
TEST_BUILDKIT_TAG		<null>
TEST_COVERAGE			<null>
```

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -15,6 +15,7 @@ import (
 	"github.com/containerd/console"
 	"github.com/containerd/platforms"
 	"github.com/docker/buildx/bake"
+	"github.com/docker/buildx/bake/hclparser"
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/builder"
 	"github.com/docker/buildx/controller/pb"
@@ -39,6 +40,7 @@ type bakeOptions struct {
 	overrides   []string
 	printOnly   bool
 	listTargets bool
+	listVars    bool
 	sbom        string
 	provenance  string
 
@@ -161,19 +163,19 @@ func runBake(ctx context.Context, dockerCli command.Cli, targets []string, in ba
 			if err != nil {
 				return
 			}
-		}
-		if progressMode != progressui.QuietMode && progressMode != progressui.RawJSONMode {
-			desktop.PrintBuildDetails(os.Stderr, printer.BuildRefs(), term)
-		}
-		if resp != nil && len(in.metadataFile) > 0 {
-			dt := make(map[string]interface{})
-			for t, r := range resp {
-				dt[t] = decodeExporterResponse(r.ExporterResponse)
+			if progressMode != progressui.QuietMode && progressMode != progressui.RawJSONMode {
+				desktop.PrintBuildDetails(os.Stderr, printer.BuildRefs(), term)
 			}
-			if warnings := printer.Warnings(); len(warnings) > 0 && confutil.MetadataWarningsEnabled() {
-				dt["buildx.build.warnings"] = warnings
+			if resp != nil && len(in.metadataFile) > 0 {
+				dt := make(map[string]interface{})
+				for t, r := range resp {
+					dt[t] = decodeExporterResponse(r.ExporterResponse)
+				}
+				if warnings := printer.Warnings(); len(warnings) > 0 && confutil.MetadataWarningsEnabled() {
+					dt["buildx.build.warnings"] = warnings
+				}
+				err = writeMetadataFile(in.metadataFile, dt)
 			}
-			err = writeMetadataFile(in.metadataFile, dt)
 		}
 	}()
 
@@ -193,8 +195,8 @@ func runBake(ctx context.Context, dockerCli command.Cli, targets []string, in ba
 		"BAKE_LOCAL_PLATFORM": platforms.Format(platforms.DefaultSpec()),
 	}
 
-	if in.listTargets {
-		cfg, err := bake.ParseFiles(files, defaults)
+	if in.listTargets || in.listVars {
+		cfg, pm, err := bake.ParseFiles(files, defaults)
 		if err != nil {
 			return err
 		}
@@ -204,7 +206,11 @@ func runBake(ctx context.Context, dockerCli command.Cli, targets []string, in ba
 		if err != nil {
 			return err
 		}
-		return printTargetList(dockerCli.Out(), cfg)
+		if in.listTargets {
+			return printTargetList(dockerCli.Out(), cfg)
+		} else if in.listVars {
+			return printVars(dockerCli.Out(), pm.AllVariables)
+		}
 	}
 
 	tgts, grps, err := bake.ReadTargets(ctx, files, targets, overrides, defaults)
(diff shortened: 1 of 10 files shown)


dvdksn:
Should `call` be a block type in the bake file when defining subfields? And should it be possible to set control directives here?

```hcl
# Short form
target "default" {
  call = "check"
}

# Long form
target "default" {
  call = {
    method = "check"
    format = "json"
    error = true # ignorestatus / BUILDKIT_DOCKERFILE_CHECK=error=true
    skip = [ "NoEmptyContinuation", "StageNameCasing" ] # BUILDKIT_DOCKERFILE_CHECK=skip=StageNameCasing,NoEmptyContinuations
  }
}
```

crazy-max:
> Should `call` be a block type in the bake file when defining subfields?

Yes in the future we would like objects instead of csv values in HCL. This is tracked in https://github.com/docker/buildx/issues/438

commands/bake.go
```diff
@@ -282,10 +439,15 @@ func bakeCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags.StringArrayVarP(&options.files, "file", "f", []string{}, "Build definition file")
 	flags.BoolVar(&options.exportLoad, "load", false, `Shorthand for "--set=*.output=type=docker"`)
 	flags.BoolVar(&options.printOnly, "print", false, "Print the options without building")
+	flags.BoolVar(&options.listTargets, "list-targets", false, "List available targets")
+	flags.BoolVar(&options.listVars, "list-variables", false, "List defined variables")
```
crazy-max, line 443: Should we open another PR for these flags? I don't think this is the same use case as `call` related to https://github.com/docker/buildx/issues/1072

Also I think it should be a single flag `--list` defaulting to `targets` as value but can also be `variables`? I'm looking at https://github.com/casey/just#command-line-options

tonistiigi:
> Should call be a block type in the bake file when defining subfields? And should it be possible to set control directives here?

`format=json` is for the whole output atm, so it can be set with `--call` but in bake definition it is just a method name. Should there be multiple targets with different format, there is no way to print it as if user requests JSON we shouldn't print random text in the middle of the output.

`ignorestatus` and `#check=error=true` are not quite the same field as well. I'm not sure if `ignorestatus` is even worth documenting atm.

commands/bake.go
crazy-max: Maybe we could consider using existing `--print` flag so we could pass `--print=targets` or `--print=variables` and `--print` would default to `--print=definition`?

crazy-max:
:eyes: https://github.com/docker/buildx/actions/runs/9781238278/job/27004845569?pr=2556#step:7:468

```
=== FAIL: tests TestIntegration/TestBakeCallCheckFlag/worker=docker-container (4.20s)
    bake.go:1086: 
        	Error Trace:	/src/tests/bake.go:1086
        	            				/src/vendor/github.com/moby/buildkit/util/testutil/integration/run.go:96
        	            				/src/vendor/github.com/moby/buildkit/util/testutil/integration/run.go:211
        	            	invalid character '/' after top-level value
        	Test:       	TestIntegration/TestBakeCallCheckFlag/worker=docker-container
    --- FAIL: TestIntegration/TestBakeCallCheckFlag/worker=docker-container (4.20s)
```

commands/bake.go
colinhemmings: Unless we have a clear use case, I would be keen not to overcomplicate things at this point. We can always make additional updates in the future

colinhemmings (approved):

dvdksn (approved):

crazy-max:
Will open a follow-up when #2562 is merged to add cross linking from bake to build

commands/bake.go
crazy-max: `list-targets` and `list-variables` have been marked as experimental in the meantime.

crazy-max (approved):
@tonistiigi PTAL with my last commit

crazy-max (changes requested):
Hum now happens in https://github.com/docker/buildx/actions/runs/9781672942/job/27006277267?pr=2556#step:7:428

```
=== Failed
=== FAIL: tests TestIntegration/TestBakeCallCheckFlag/worker=docker-container (2.04s)
    bake.go:1086: 
        	Error Trace:	/src/tests/bake.go:1086
        	            				/src/vendor/github.com/moby/buildkit/util/testutil/integration/run.go:96
        	            				/src/vendor/github.com/moby/buildkit/util/testutil/integration/run.go:211
        	            	invalid character '/' after top-level value
        	Test:       	TestIntegration/TestBakeCallCheckFlag/worker=docker-container
```

tonistiigi:
@crazy-max Something is printing to stdout on `format=json` I think

tests/bake.go
```diff
@@ -951,3 +956,163 @@ target "default" {
...
+	out, err = bakeCmd(
+		sb,
+		withDir(dir),
+		withArgs("build", "another", "--call", "check,format=json"),
+	)
+	require.Error(t, err, out)
+
+	var res map[string]any
+	err = json.Unmarshal([]byte(out), &res)
+	require.NoError(t, err)
```
crazy-max, line 1086: @tonistiigi could you add: to see what's going on with unmarshalling issue?

```suggestion
	require.NoError(t, err, out)
```

crazy-max (approved):
LGTM

opened https://github.com/docker/buildx/issues/2575 to make sure we don't miss it before GA

crazy-max:
Docs is missing for variable description field: https://github.com/docker/buildx/commit/233b869c63844ddabc4100b2fc2e2f7e4bdd6aa4#diff-dd01f2c9767e65fba215bfdbb1317fea48358318fa27866339a4dc3ef7757742R30

We should have an example in https://github.com/docker/buildx/blob/master/docs/bake-reference.md#variable similar to https://github.com/docker/buildx/blob/master/docs/bake-reference.md#targetdescription

PR #2577, opened 2024-07-03
vendor: update buildkit to v0.15.0-rc1

tonistiigi:


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/docker/buildx
 
-go 1.21
+go 1.21.0
+
+toolchain go1.22.5
```
tonistiigi, line 5: 😠

daghack (approved):

go.mod
crazy-max: you can safely remove `toolchain go1.22.5`

crazy-max: ah nevermind you already removed it

PR #2578, opened 2024-07-04
driver: allow arbitrary client opts

crazy-max:
cc @fiam


tonistiigi (commented):
I'm not sure this actually works because the client is cached. The options would only matter for the very first call.

tonistiigi (approved):

Issue #2597, opened 2024-07-13
optimize k8s config initialization


tonistiigi:
In the `LoadNodes` function https://github.com/docker/buildx/blob/v0.16.0/builder/node.go#L122-L150 before it initializes the Driver it goes into a bunch of k8s config loading routines that is then passed to the driver initialization.

It loads all contexts to retrieve potential k8s config from it and then passes it to various k8s libraries. Problem is that this code runs every time, even if my current builder is not using k8s driver. It also looks like even if I don't have any k8s config in my contexts, these operations are still not no-ops. In the profiler I can see it calling into some yaml libraries to do some parsing. I assume this happens with some default config as I don't have any custom config on my machine (that I know about).

K8s-specific logic should ideally happen only when loading k8s driver, or at least be skippable in the `LoadNodes`.

PR #2602, opened 2024-07-16
optimize k8s config initialization

karthikkurella:
Fix: #2597 

#Problem:
* The LoadNodes function performs unnecessary Kubernetes (k8s) configuration loading routines even when the k8s driver is not in use, leading to performance overhead due to unnecessary YAML parsing.

# Solution:
* Modify the LoadNodes function to conditionally execute k8s-specific logic only when the k8s driver is being used.

# Changes:

Add Conditional Check:

* Check if the node's endpoint contains "kubernetes" before loading the k8s configuration.
* Move k8s Config Loading Logic:

Encapsulate the k8s configuration loading within the conditional check to prevent unnecessary operations.
This ensures that k8s-specific logic is only executed when required, reducing unnecessary overhead.

### DIFF
--- a/builder/node.go
+++ b/builder/node.go
@@ -121,31 +121,31 @@ func (b *Builder) LoadNodes(ctx context.Context, opts ...LoadNodesOption) (_ []N
 
 				contextStore := b.opts.dockerCli.ContextStore()
 
+				// Check if the driver is a Kubernetes driver
 				var kcc driver.KubeClientConfig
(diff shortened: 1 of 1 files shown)


karthikkurella:
@tonistiigi Please review the changes.

Thank you.

crazy-max:
Sorry I missed this PR, I think we want to tend for the implementation from #2606 that moves k8s config handling to the driver package. But thanks!

Issue #2598, opened 2024-07-15
bake: panic in progress printer


crazy-max:
https://github.com/docker/buildx/actions/runs/9915811338/job/27397269631?pr=2595#step:6:259

```
  panic: runtime error: invalid memory address or nil pointer dereference
  [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x13d7ba7]
  
  goroutine 10 [running]:
  github.com/docker/buildx/util/progress.(*Printer).Warnings(...)
  	github.com/docker/buildx/util/progress/printer.go:65
  github.com/docker/buildx/commands.runBake.func2()
  	github.com/docker/buildx/commands/bake.go:149 +0x27
  github.com/docker/buildx/util/progress.NewPrinter.func1()
  	github.com/docker/buildx/util/progress/printer.go:129 +0x20a
  created by github.com/docker/buildx/util/progress.NewPrinter in goroutine 1
  	github.com/docker/buildx/util/progress/printer.go:112 +0x2b1
```

https://github.com/docker/buildx/blob/10c9ff901c1c255b58969dd2ade4789acbbab3bb/util/progress/printer.go#L65

tonistiigi:
Looks like https://github.com/docker/buildx/blob/10c9ff901c1c255b58969dd2ade4789acbbab3bb/commands/bake.go#L255 runs before https://github.com/docker/buildx/blob/10c9ff901c1c255b58969dd2ade4789acbbab3bb/commands/bake.go#L149

Note that there is another similar section now in https://github.com/docker/buildx/blob/10c9ff901c1c255b58969dd2ade4789acbbab3bb/commands/bake.go#L206


This is a new code in v0.16.0 https://github.com/docker/buildx/pull/2534 . Previously warnings were not printed for Bake.

PR #2603, opened 2024-07-16
bake: check printer before printing warnings

crazy-max:
fixes https://github.com/docker/buildx/issues/2598

A simple fix for a patch release but we should do a proper one as follow-up to avoid this situation with `printer` in the first place.


tonistiigi (commented):
I don't think this is safe. Iiuc the `printer = nil` and `WithOnClose()` happen in different goroutines. So even if nil check passes it might be nil once it gets to `printWarnings`

crazy-max:
> I don't think this is safe. Iiuc the `printer = nil` and `WithOnClose()` happen in different goroutines. So even if nil check passes it might be nil once it gets to `printWarnings`

Ah true, I'm taking another look

crazy-max:
superseded by https://github.com/docker/buildx/pull/2608

tonistiigi (approved):

PR #2606, opened 2024-07-16
builder: move kube config handling to k8s driver package

crazy-max:
fixes https://github.com/docker/buildx/issues/2597

### DIFF
--- a/driver/kubernetes/factory.go
+++ b/driver/kubernetes/factory.go
@@ -2,19 +2,22 @@ package kubernetes
 
 import (
 	"context"
+	"os"
 	"strconv"
 	"strings"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
-
 	"github.com/docker/buildx/driver"
 	"github.com/docker/buildx/driver/bkimage"
+	ctxkube "github.com/docker/buildx/driver/kubernetes/context"
 	"github.com/docker/buildx/driver/kubernetes/manifest"
 	"github.com/docker/buildx/driver/kubernetes/podchooser"
 	dockerclient "github.com/docker/docker/client"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


driver/kubernetes/factory.go
```diff
@@ -107,6 +133,45 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
...
+			tryToUseKubeConfigInCluster = true
+		}
+	}
+	if tryToUseKubeConfigInCluster {
+		ccInCluster := ClientConfigInCluster{}
+		if _, err := ccInCluster.ClientConfig(); err == nil {
+			logrus.Debug("using kube config in cluster")
+			cc = ccInCluster
+		}
+	}
```
crazy-max, line 170: didn't touch this logic from builder package yet, should look at it more closely as follow-up.

driver/manager.go
```diff
@@ -28,38 +26,19 @@ type BuildkitConfig struct {
...
-	DockerAPI        dockerclient.APIClient
-	KubeClientConfig KubeClientConfig
-	BuildkitdFlags   []string
-	Files            map[string][]byte
-	DriverOpts       map[string]string
-	Auth             Auth
-	Platforms        []specs.Platform
-	ContextPathHash  string // can be used for determining pods in the driver instance
-	DialMeta         map[string][]string
+	// TODO: This object needs updates to be generic for different drivers
```
crazy-max, line 30: I guess we can remove this comment now that kube config is removed from there?  There is still `ContextPathHash` but looks generic enough imo.

crazy-max:
@AkihiroSuda When you have time could you take another look please?

driver/manager.go
AkihiroSuda: 👍

driver/kubernetes/factory.go
```diff
@@ -46,18 +69,50 @@ func (*factory) Priority(ctx context.Context, endpoint string, api dockerclient.
...
+	var err error
+	var cc ClientConfig
+	if f.cc != nil {
+		cc = f.cc
+	} else {
+		cc, err = ctxkube.ConfigFromEndpoint(cfg.EndpointAddr, cfg.ContextStore)
+		if err != nil {
+			// err is returned if n.Endpoint is non-context name like "unix:///var/run/docker.sock".
+			// try again with name="default".
+			// FIXME(@AkihiroSuda): n should retain real context name.
```
AkihiroSuda, line 81: What was "n"? 

Maybe the variable was renamed since then? (hard to check from phone now)

AkihiroSuda (approved):
Thanks

driver/kubernetes/factory.go
crazy-max: Seems it was the node https://github.com/docker/buildx/blob/63073b65c01c415db8b34053ac3f0b7f2c14d27a/commands/util.go#L90

As we now have `ContextStore` in driver `InitConfig` maybe it would be fine to have `ContextName` as well and use it here?

PR #2608, opened 2024-07-17
bake: fix printer handling

### LINKED ISSUE #2598: bake: panic in progress printer

crazy-max:
relates to #2598

Removes extra go routine that is racy when used along the other one from `progress.WithOnClose`.

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -29,7 +29,6 @@ import (
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/buildx/util/tracing"
 	"github.com/docker/cli/cli/command"
-	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/identity"
 	"github.com/moby/buildkit/util/progress/progressui"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


commands/bake.go
```diff
@@ -238,26 +207,23 @@ func runBake(ctx context.Context, dockerCli command.Cli, targets []string, in ba
 		return err
 	}
 
-	def := struct {
+	dtdef, err := json.MarshalIndent(struct {
```
tonistiigi, line 210: Passing this encoded JSON around seems weird. I think we can just call encode in SaveLocalState.

tonistiigi (commented):
I think this mostly is good but if we want to cherry-pick then I think for v0.16 we should go with something like

```
if p := printer; p != nil {
```

tonistiigi:
@crazy-max yes

commands/bake.go
```diff
@@ -468,12 +424,34 @@ func bakeCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
...
 }
 
-func saveLocalStateGroup(dockerCli command.Cli, ref string, lsg localstate.StateGroup) error {
+func saveLocalStateGroup(dockerCli command.Cli, in bakeOptions, targets []string, bo map[string]build.Options, overrides []string, def any) error {
+	prm := confutil.MetadataProvenance()
+	if len(in.metadataFile) == 0 {
+		prm = confutil.MetadataProvenanceModeDisabled
+	}
+	groupRef := identity.NewID()
+	var refs []string
```
tonistiigi, line 433: nit: `refs := make([]string, 0, len(bo))`

tonistiigi (approved):

commands/bake.go
crazy-max: fixed

PR #2610, opened 2024-07-17
metrics: add metrics for bake command

jsternberg:
This adds metrics for the bake command using a different method of calculating the build identifier but with the same attributes otherwise.

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -4,12 +4,16 @@ import (
 	"bytes"
 	"cmp"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"slices"
+	"sort"
 	"strings"
+	"sync"
 	"text/tabwriter"
 
 	"github.com/containerd/console"
@@ -26,6 +30,7 @@ import (
 	"github.com/docker/buildx/util/confutil"
 	"github.com/docker/buildx/util/desktop"
 	"github.com/docker/buildx/util/dockerutil"
+	"github.com/docker/buildx/util/osutil"
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/buildx/util/tracing"
 	"github.com/docker/cli/cli/command"
@@ -33,6 +38,7 @@ import (
 	"github.com/moby/buildkit/util/progress/progressui"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"go.opentelemetry.io/otel/attribute"
 )
 
 type bakeOptions struct {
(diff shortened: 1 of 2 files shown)


commands/bake.go
```diff
@@ -588,3 +604,62 @@ func printTargetList(w io.Writer, cfg *bake.Config) error {
...
+		var files []string
+		if len(o.files) == 0 {
+			files = []string{osutil.GetWd()}
+		} else {
+			files = make([]string, len(o.files))
+			for i, file := range o.files {
+				files[i] = osutil.ToAbs(file)
+			}
+			sort.Strings(files)
+		}
```
crazy-max, line 641: Not sure that would work for remote bake definitions, maybe using `cmdContext` would work.

jsternberg: Can you expand on what you mean by this? I'm not very familiar with remote bake definitions. I'm guessing this is referring to `osutil.ToAbs`?

crazy-max: > I'm guessing this is referring to `osutil.ToAbs`?

Yes I mean `osutil.*` calls that would not be right for remote bake definition set here: https://github.com/docker/buildx/blob/e2f6808457ab9d8de4a85f0f26b0028411ef20fc/commands/bake.go#L68 and retrieved here https://github.com/docker/buildx/blob/e2f6808457ab9d8de4a85f0f26b0028411ef20fc/commands/bake.go#L474

Stat of these definitions are not on the host but memory: https://github.com/docker/buildx/blob/e2f6808457ab9d8de4a85f0f26b0028411ef20fc/bake/remote.go#L186-L194

Hence why I think we should use `cmdContext` and just the relative path of `files`.

jsternberg: Ok I think I understand now. Do you think that it would be better to just include the `url` and `cmdContext` attributes in the hash instead of trying to resolve the path of the files? The intention is just to differentiate different bake commands in different directories on the same machine. I don't think we need to resolve the absolute path to these files.

I'll mock something up and see how it looks.

crazy-max (changes requested):
Some lint issues otherwise looks pretty good! https://github.com/docker/buildx/actions/runs/10286248497/job/28466467965?pr=2610#step:5:368

```
#18 [lint 1/1] RUN --mount=target=/go/src/github.com/docker/buildx     --mount=target=/root/.cache,type=cache,id=lint-cache-darwin/amd64     xx-go --wrap &&     golangci-lint run
#18 85.65 commands/bake.go:649:3: ineffectual assignment to cpy (ineffassign)
#18 85.65 		cpy = s
#18 85.65 		^
#18 ERROR: process "/bin/sh -c xx-go --wrap &&     golangci-lint run" did not complete successfully: exit code: 1
```

crazy-max:
Not for this PR but I wonder we could have integration tests pushing metrics to a local endpoint in the sandbox similar to minio integration for s3 remote cache https://github.com/moby/buildkit/pull/3398?

jsternberg:
We don't have any integration tests like that but I think we really need that at this point.

jsternberg:
https://github.com/docker/buildx/pull/2610#pullrequestreview-2228117696

Ha I must have been really tired for this one. The `cpy = s` is supposed to be `s = cpy`. Just going to change it to `return cpy` just to avoid the entire thing.

crazy-max (approved):

PR #2620, opened 2024-07-20
tests: build multiplatform

idnandre:
add test build and bake multiplatform (--platform), to improve test coverage based on #1857

### DIFF
--- a/tests/bake.go
+++ b/tests/bake.go
@@ -56,6 +56,7 @@ var bakeTests = []func(t *testing.T, sb integration.Sandbox){
 	testListVariables,
 	testBakeCallCheck,
 	testBakeCallCheckFlag,
+	testBakeMultiPlatform,
 }
 
 func testBakePrint(t *testing.T, sb integration.Sandbox) {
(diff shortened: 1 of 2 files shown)


tests/bake.go
```diff
@@ -887,6 +888,56 @@ target "def" {
...
+		t,
+		fstest.CreateFile("docker-bake.hcl", bakefile, 0600),
+		fstest.CreateFile("Dockerfile", dockerfile, 0600),
+		fstest.CreateFile("foo", []byte("foo"), 0600),
+	)
+
+	cmd := buildxCmd(sb, withDir(dir), withArgs("bake"), withArgs("--set", fmt.Sprintf("*.output=type=image,name=%s,push=true", target)))
+	out, err := cmd.CombinedOutput()
+
+	if !isMobyWorker(sb) {
```
thaJeztah, line 922: @crazy-max how does this check detect whether the docker daemon supports multi-arch? Having a quick look; `isMobyWorker` and `isMobyContainerdSnapWorker` look identical, but if docker is running with the containerd image-store enabled, it supports multi-arch? https://github.com/docker/buildx/blob/56cb197c0a186e4382eba26537af238e19fe880c/tests/integration.go#L97-L105

Should we have a `supportsMultiArch` check instead?

thaJeztah: OH! Nevermind; `hasFeature` vs `!hasFeature` 🙈 I missed the `!` 🙈 😂 


Still, maybe a `supportsMultiArch` or `hasMultiarchSupport` could make sense (more descriptive?)

crazy-max: This is not just multiarch but also cache and oci exporters that are supported with containerd snap: https://github.com/docker/buildx/blob/aa35c954f3f1a346ef71a1e07f942d2a46740b1f/driver/docker/driver.go#L92-L95

crazy-max (approved):
LGTM thanks!

tonistiigi (approved):

PR #2621, opened 2024-07-22
Dockerfile: update to docker 27.1.1

thaJeztah:
Also adding a DOCKER_CLI_VERSION build-arg, so that we can set versions
independently for (untagged) pre-releases.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ ARG GO_VERSION=1.22
 ARG XX_VERSION=1.4.0
 
 # for testing
-ARG DOCKER_VERSION=27.0.3
+ARG DOCKER_VERSION=27.1.1
+ARG DOCKER_CLI_VERSION=${DOCKER_VERSION}
 ARG GOTESTSUM_VERSION=v1.9.0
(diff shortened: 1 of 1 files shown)


crazy-max (approved):

Dockerfile
```diff
@@ -4,7 +4,8 @@ ARG GO_VERSION=1.22
 ARG XX_VERSION=1.4.0
 
 # for testing
-ARG DOCKER_VERSION=27.0.3
+ARG DOCKER_VERSION=27.1.0
+ARG DOCKER_CLI_VERSION=${DOCKER_CLI_VERSION-${DOCKER_VERSION}}
```
tonistiigi, line 8: nit: this is just `ARG DOCKER_CLI_VERSION=$DOCKER_VERSION`

thaJeztah: Oh! You're right; not sure what I was thinking there. Let me update; and also update to v27.1.1

thaJeztah:
Updated to v27.1.1, and fixed the nit from https://github.com/docker/buildx/pull/2621#discussion_r1688725640

crazy-max (approved):

PR #2636, opened 2024-07-30
docs: fix metadata section for build command

crazy-max:
`buildx.build.warnings` note is not in the right section :see_no_evil:


dvdksn (approved):
Oops probably a bad rebase on my part

thompson-shaun (approved):

PR #2551, opened 2024-06-24
build: opt to set progress warnings in response

crazy-max:
alternative to https://github.com/docker/buildx/pull/2511 and similar to https://github.com/docker/buildx/pull/2280, we can set build checks warnings in the metadata file. Does not set them by default but can be enabled by using the `BUILDX_METADATA_WARNINGS=1` environment variable.

This will be useful with our GitHub Actions so we can generate GitHub Annotations without having to do a subrequest build like in https://github.com/crazy-max/docker-build-push-action/pull/39/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR132

@tonistiigi Also added extra commit to dedup warnings for bake related to https://github.com/docker/buildx/pull/2534#issue-2358036103

```dockerfile
# syntax=docker/dockerfile-upstream:master
frOM busybox as base
cOpy Dockerfile .

from scratch
COPy --from=base \
  /Dockerfile \
  /
```

```
$ BUILDX_METADATA_WARNINGS=1 docker buildx build . --metadata-file md.json
...
#12 exporting to image
#12 exporting layers done
#12 writing image sha256:d0bb2193d26744307a7003c5397e277d1fa73ee979851b0c2ed2afad89d2e44b done
#12 DONE 0.0s

#13 resolving provenance for metadata file
#13 DONE 0.0s

 3 warnings found (use --debug to expand):
 - ConsistentInstructionCasing: Command 'frOM' should be consistently cased (line 2)
 - ConsistentInstructionCasing: Command 'cOpy' should be consistently cased (line 3)
 - ConsistentInstructionCasing: Command 'COPy' should be consistently cased (line 6)

View build details: docker-desktop://dashboard/build/default/default/wuiiyvfzfof6q93b69sp97n2l
```

<details>

<summary><code>md.json</code></summary>

```json
{
  "buildx.build.provenance": {
    "buildType": "https://mobyproject.org/buildkit@v1",
    "materials": [
      {
        "uri": "pkg:docker/docker/dockerfile-upstream@master",
        "digest": {
          "sha256": "c2da757515621f3767427a157f351149ff48a69e0550afa08fa768feaaf466c7"
        }
      },
      {
        "uri": "pkg:docker/busybox@latest?platform=linux%2Famd64",
        "digest": {
          "sha256": "9ae97d36d26566ff84e8893c64a6dc4fe8ca6d1144bf5b87b2b85a32def253c7"
        }
      }
    ],
[...]

PR #2640, opened 2024-08-05
support metadata file with call flag for build and bake commands

crazy-max:
Atm when using `--call` or `--check`, the metadata file is not generated:

```
$ docker buildx build --call check --file ./test/bad.Dockerfile --metadata-file md.json ./test
...
  12 |     CMD [ "echo", "Hello, Sweden!" ]
  13 |     ENTRYPOINT my-program start
--------------------

WARNING: JSONArgsRecommended - https://docs.docker.com/go/dockerfile/rule/json-args-recommended/
JSON arguments recommended for ENTRYPOINT to prevent unintended behavior related to OS signals
bad.Dockerfile:13
--------------------
  11 |     CMD [ "echo", "Hello, Norway!" ]
  12 |     CMD [ "echo", "Hello, Sweden!" ]
  13 | >>> ENTRYPOINT my-program start
  14 |
--------------------

$ cat md.json
cat: md.json: No such file or directory
```

I think we should allow it for build warnings related to https://github.com/docker/buildx/pull/2551

So we can ingest call result if we want to do extra work with similar to https://github.com/docker/build-push-action/pull/1197.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -367,19 +367,22 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions) 
 			return errors.Wrap(err, "writing image ID file")
 		}
 	}
+	if options.metadataFile != "" {
+		dt := decodeExporterResponse(resp.ExporterResponse)
+		if opts.PrintFunc == nil {
+			if warnings := printer.Warnings(); len(warnings) > 0 && confutil.MetadataWarningsEnabled() {
+				dt["buildx.build.warnings"] = warnings
+			}
+		}
+		if err := writeMetadataFile(options.metadataFile, dt); err != nil {
+			return err
+		}
+	}
 	if opts.PrintFunc != nil {
 		if exitcode, err := printResult(dockerCli.Out(), opts.PrintFunc, resp.ExporterResponse); err != nil {
 			return err
 		} else if exitcode != 0 {
 			os.Exit(exitcode)
-		}
-	} else if options.metadataFile != "" {
-		dt := decodeExporterResponse(resp.ExporterResponse)
-		if warnings := printer.Warnings(); len(warnings) > 0 && confutil.MetadataWarningsEnabled() {
-			dt["buildx.build.warnings"] = warnings
-		}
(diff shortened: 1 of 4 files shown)


daghack (approved):

commands/build.go
```diff
@@ -733,10 +736,16 @@ func writeMetadataFile(filename string, dt interface{}) error {
...
 	for k, v := range exporterResponse {
-		dt, err := base64.StdEncoding.DecodeString(v)
-		if err != nil {
-			out[k] = v
-			continue
+		var dt []byte
+		var err error
+		if k == "result.json" {
+			dt = []byte(v)
+		} else {
```
jsternberg, line 743: This area seems new. Can you expand on what this does? I'd also suggest maybe moving this section of code to a small helper function so it's more self-documenting and then you can have the function directly return `([]byte, error)` instead of using `var` and the branching logic here.

jsternberg (commented):
Most of it seems straightforward but there was one new section I had a question about.

commands/build.go
crazy-max: > Can you expand on what this does?

When doing subrequests, the response looks like this: https://github.com/moby/buildkit/blob/f6eb72f2f5db07ddab89ac5e2bd3939a6444f4be/frontend/dockerui/requests.go#L100-L102

And with this PR we are setting this in the metadata file but as this is a string type https://github.com/moby/buildkit/blob/f6eb72f2f5db07ddab89ac5e2bd3939a6444f4be/frontend/dockerui/requests.go#L88, we need byte for this specific key.

I can put a comment if you wish

jsternberg: Yea moving this to a comment and a small helper function I think would be good. Maybe something like `decodeExporterResponse` that contains this if/else logic or whatever naming makes more sense.

jsternberg (approved):

tonistiigi:
I'm pretty sure I did this intentionally, eg. because the default exporter is `cacheonly` with `--call`, and provenance is skipped. Not 100% anymore which is better.

PR #2641, opened 2024-08-05
metricwriter: compile regular expressions only on first use

### LINKED ISSUE #2596: metrics regexp usage for progress can be optimized
Currently https://github.com/docker/buildx/blob/master/util/progress/metricwriter.go uses heavy regexp functions a lot.

This show both in inittrace, and with Find/Match functions when retrieving progress. It is present independently if tracing is enabled or not.

```
init github.com/docker/buildx/util/progress @30 ms, 0.077 ms clock, 41992 bytes, 224 allocs
```

For init we should avoid `regexp.Compile` in the init functions as it allocates lots of memory and blocks rest of the loading. While individually this isn't that significant, these bad practices by packages add up. This can be solved by adding `sync.Once` so the regexp instances are created one time on first access.

For the matchers, it looks like all the regexp always try to match against all of the vertex records coming in progress. This should only need to happen once per vertex. If vertex is type=exec it can not be type=source at the same time and does not need to be repeatedly rechecked. If a type for a specific digest has already been found, it does not need to be rechecked again when new logs are arriving.

I think the `sync.Once` can cover all the regexp together that then resolves the type info (and possible type specific values parsed from name) that is then memorized by digest.

While this would also be faster by doing a simpler string comparison in most cases, I don't think that is a good idea as the code becomes more error-prone. Instead, we should finish up returning type as a separate field in progress in buildkit side as was discussed in the original proposal for this. The buildx side of the code can already prepare for this (as it would need to keep regexp code for backwards compatibility anyway), so that there is a single function that determines the type based on string parsing that can then be skipped if the type is already set by buildkit.

@jsternberg

jsternberg:
Compile the regular expressions only on first use rather than implicitly as part of the `init()` function of the package. This prevents taking a speed hit on the initialization of the package regardless if this type is used and moves it to the time when a regular expression is first used.

Fixes #2596.

### DIFF
--- a/util/progress/metricwriter.go
+++ b/util/progress/metricwriter.go
@@ -17,6 +17,31 @@ import (
 	"golang.org/x/text/language"
 )
 
+type rePatterns struct {
+	LocalSourceType *regexp.Regexp
+	ImageSourceType *regexp.Regexp
+	ExecType        *regexp.Regexp
+	ExportImageType *regexp.Regexp
+	LintMessage     *regexp.Regexp
+}
+
+var re = sync.OnceValue(func() *rePatterns {
+	return &rePatterns{
+		LocalSourceType: regexp.MustCompile(
+			strings.Join([]string{
+				`(?P<context>\[internal] load build context)`,
+				`(?P<dockerfile>load build definition)`,
+				`(?P<dockerignore>load \.dockerignore)`,
+				`(?P<namedcontext>\[context .+] load from client)`,
+			}, "|"),
+		),
+		ImageSourceType: regexp.MustCompile(`^\[.*] FROM `),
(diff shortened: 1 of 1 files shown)


crazy-max:
I assume this fixes #2596?

crazy-max (approved):

tonistiigi:
The `init` side of this was handled but this part was not

```
For the matchers, it looks like all the regexp always try to match against all of the vertex records coming in progress. This should only need to happen once per vertex. If vertex is type=exec it can not be type=source at the same time and does not need to be repeatedly rechecked. If a type for a specific digest has already been found, it does not need to be rechecked again when new logs are arriving.
```

It's not a high priority to update this but I guess we should leave the issue open.

PR #2647, opened 2024-08-12
build: print out the number of warnings after completing a rule check

daghack:
Atm when building with `--check`, when there are no warnings we generate the output
`Check completed, no warnings found.`

```                                                                                                                   0.0s
 => => transferring context: 88B                                                                                                                         0.0s
Check completed, no warnings found.
```

When there are warnings, we leap right into displaying them.
```                                                                                                               0.0s
 => => transferring context: 88B 

WARNING: FromAsCasing - https://docs.docker.com/go/dockerfile/rule/from-as-casing/ ....
```

This updates the output in the later case to contain a similar "Check completed" message which lists the warning count.
```
 => => transferring context: 88B 
Check completed, 1 warning found!

WARNING: FromAsCasing - https://docs.docker.com/go/dockerfile/rule/from-as-casing/ ....
```

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -891,17 +891,29 @@ func printResult(w io.Writer, f *controllerapi.CallFunc, res map[string]string) 
 	case "subrequests.describe":
 		return 0, printValue(w, subrequests.PrintDescribe, subrequests.SubrequestsDescribeDefinition.Version, f.Format, res)
 	case "lint":
-		err := printValue(w, lint.PrintLintViolations, lint.SubrequestLintDefinition.Version, f.Format, res)
-		if err != nil {
-			return 0, err
-		}
-
 		lintResults := lint.LintResults{}
 		if result, ok := res["result.json"]; ok {
 			if err := json.Unmarshal([]byte(result), &lintResults); err != nil {
 				return 0, err
 			}
 		}
+
+		warningCount := len(lintResults.Warnings)
+		if f.Format != "json" && warningCount > 0 {
+			var warningCountMsg string
+			if warningCount == 1 {
+				warningCountMsg = "1 warning has been found!"
+			} else if warningCount > 1 {
+				warningCountMsg = fmt.Sprintf("%d warnings have been found!", warningCount)
+			}
+			fmt.Fprintf(w, "Check complete, %s\n", warningCountMsg)
+		}
+
+		err := printValue(w, lint.PrintLintViolations, lint.SubrequestLintDefinition.Version, f.Format, res)
+		if err != nil {
+			return 0, err
+		}
+
 		if lintResults.Error != nil {
 			// Print the error message and the source
 			// Normally, we would use `errdefs.WithSource` to attach the source to the
--- a/tests/bake.go
+++ b/tests/bake.go
@@ -59,6 +59,7 @@ var bakeTests = []func(t *testing.T, sb integration.Sandbox){
 	testBakeCallCheckFlag,
 	testBakeCallMetadata,
 	testBakeMultiPlatform,
+	testBakeCheckCallOutput,
 }
 
 func testBakePrint(t *testing.T, sb integration.Sandbox) {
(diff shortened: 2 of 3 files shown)


commands/build.go
```diff
@@ -891,17 +891,29 @@ func printResult(w io.Writer, f *controllerapi.PrintFunc, res map[string]string)
...
 				return 0, err
 			}
 		}
+
+		if f.Format != "json" {
+			var warningCountMsg string
+			warningCount := len(lintResults.Warnings)
+			if warningCount == 1 {
+				warningCountMsg = "1 warning has been found!"
+			} else if len(lintResults.Warnings) > 1 {
```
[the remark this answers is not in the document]
daghack: 👍

commands/build.go
```diff
@@ -891,17 +891,29 @@ func printResult(w io.Writer, f *controllerapi.PrintFunc, res map[string]string)
...
+
+		if f.Format != "json" {
+			var warningCountMsg string
+			warningCount := len(lintResults.Warnings)
+			if warningCount == 1 {
+				warningCountMsg = "1 warning has been found!"
+			} else if warningCount > 1 {
+				warningCountMsg = fmt.Sprintf("%d warnings have been found!", warningCount)
+			}
+			fmt.Fprintf(w, "Check completed, %s\n", warningCountMsg)
```
crazy-max, line 909: Can we have an integration test for both running check with and without warning to validate this?

iiuc this will print `Check complete, ` if no warnings are found and this as well https://github.com/docker/buildx/blob/07a85a544b4b1209377c2a74a9f5a91cb9d31d04/commands/?build.go#L924?

crazy-max, line 909: iiuc this will print `Check complete, ` if no warnings are found and this as well https://github.com/docker/buildx/blob/07a85a544b4b1209377c2a74a9f5a91cb9d31d04/commands/?build.go#L924?

crazy-max (commented):
Can we have an integration test for both running check with and without warning to validate this?

commands/build.go
daghack: Added!

daghack: Good catch, fixed!

daghack:
> Can we have an integration test for both running check with and without warning to validate this?

Added!

daghack:
@crazy-max @tonistiigi ping

tonistiigi (commented):
@daghack Did you check the output of this for `buildx bake` as well?

daghack:
@tonistiigi Good catch. Will update!

crazy-max:
> @daghack Did you check the output of this for `buildx bake` as well?

Yes another integration for bake would be ideal to make sure

daghack:
@crazy-max @tonistiigi Added matching integration checks for bake! :)

tests/bake.go
```diff
@@ -1274,3 +1275,87 @@ target "default" {}
...
+		dir := tmpdir(
+			t,
+			fstest.CreateFile("docker-bake.hcl", bakefile, 0600),
+			fstest.CreateFile("Dockerfile", dockerfile, 0600),
+		)
+
+		cmd := buildxCmd(
+			sb,
+			withDir(dir),
+			withArgs("bake", "--call", "check", "--metadata-file", filepath.Join(dir, "md.json")),
```
tonistiigi, line 1297: Why is the `--metadata-file` in these checks? Seems to be unused?

tests/bake.go
```diff
@@ -1274,3 +1275,87 @@ target "default" {}
...
```
tonistiigi, line 1352: How is the output for the case when multiple targets are built?

tests/bake.go
daghack: Removed!

tests/bake.go
daghack: Added test for this!

crazy-max (approved):

PR #2649, opened 2024-08-13
bake: add basename, dirname and sanitize functions

tonistiigi:
These functions help with dealing with path inputs and using parts of them to configure targets.

For example, this improves cases like https://github.com/moby/buildkit/pull/5248/files#diff-870f6fe23fc034f008f5203ee1e628d7bfa65a49095d5a4688db498328449ed2R238 where current best approach is to take hash and lose readability of the variable.

### DIFF
--- a/bake/hclparser/stdlib.go
+++ b/bake/hclparser/stdlib.go
@@ -2,6 +2,8 @@ package hclparser
 
 import (
 	"errors"
+	"path"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/go-cty-funcs/cidr"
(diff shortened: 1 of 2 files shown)


dvdksn (commented):
do we want to create a reference doc for the stdlib functions we ship? Right now we only refer to go-cty which is incorrect https://docs.docker.com/build/bake/funcs/#standard-library

tonistiigi:
> (follow-up) do we want to create a reference doc for the stdlib functions we ship?

This would be nice indeed. Ideally, we could maybe generate that list as we do with the flags.

crazy-max (approved):

PR #2660, opened 2024-08-16
commands: add debug as persistent flag

tonistiigi:
Allows using `--debug` to enable debug logging under any subcommand. Currently it needed to be set as
`docker --debug buildx` meaning only way to enable debug in standalone mode was to set env variable instead and updating existing commands to add `--debug` was cumbersome.

### DIFF
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -18,6 +18,7 @@ Build from a file
 | [`--builder`](#builder)             | `string`      |         | Override the configured builder instance                                                            |
 | [`--call`](#call)                   | `string`      | `build` | Set method for evaluating build (`check`, `outline`, `targets`)                                     |
 | [`--check`](#check)                 | `bool`        |         | Shorthand for `--call=check`                                                                        |
+| `-D`, `--debug`                     | `bool`        |         | Enable debug logging                                                                                |
 | [`-f`](#file), [`--file`](#file)    | `stringArray` |         | Build definition file                                                                               |
 | `--load`                            | `bool`        |         | Shorthand for `--set=*.output=type=docker`                                                          |
 | [`--metadata-file`](#metadata-file) | `string`      |         | Write build result metadata to a file                                                               |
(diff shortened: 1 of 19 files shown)


docs/reference/buildx_bake.md
```diff
@@ -18,6 +18,7 @@ Build from a file
 | [`--builder`](#builder)             | `string`      |         | Override the configured builder instance                                                            |
 | [`--call`](#call)                   | `string`      | `build` | Set method for evaluating build (`check`, `outline`, `targets`)                                     |
 | [`--check`](#check)                 | `bool`        |         | Shorthand for `--call=check`                                                                        |
+| `-D`, `--debug`                     | `bool`        |         | Enable debug logging                                                                                |
```
thaJeztah, line 21: Do we want both the `-D` and the `--debug` to be inherited, or should we limit to only the long form (`--debug`) for now? (can we?)

cc @jalonsogo

tonistiigi: I think it should be invisible to user if this flag is defined in the cli or buildx binary. `docker --debug buildx` still uses the defined in CLI.

crazy-max (commented):
Could we have an integration test for checks with debug?

docs/reference/buildx_bake.md
thaJeztah: Mostly looking here that the `-D` is documented for the top-level, but for the subcommands it's a new flag, and in most (if not all) cases we refer to it as `--debug`.

Where possibly I try to avoid shorthand flags if they're not _critical_ to have (it's easy to add them later if there's a strong need), more so if they're case-sensitive (if there's both a `-d` and ` -D` flag, which can be confusing).

tonistiigi: >  but for the subcommands it's a new flag, and in most (if not all) cases we refer to it as --debug.

`--debug` is also a new flag for subcommand.

I don't think any user would understand why one flag formatting works in one place but not another.

crazy-max (approved):

PR #2663, opened 2024-08-23
build: avoid possible panic when reading git info

tonistiigi:
Not all the error cases from getGitAttributes returned appendNoneFunc. When nil was returned it caused a panic.


crazy-max (approved):
:fearful:

PR #2666, opened 2024-08-29
bake: enable support for entitlements

tonistiigi:
Ref #179

Add support for `security.insecure` and `network.host` entitlements via bake. User needs to confirm elevated privileges through a prompt or CLI flags.

### DIFF
--- a/bake/entitlements.go
+++ b/bake/entitlements.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"slices"
 	"strings"
 
@@ -143,8 +142,8 @@ func (c EntitlementConf) Prompt(ctx context.Context, out io.Writer) error {
 	}
 
 	args := append([]string(nil), os.Args...)
-	if filepath.Base(args[0]) == "docker-buildx" {
-		args[0] = "docker"
+	if v, ok := os.LookupEnv("DOCKER_CLI_PLUGIN_ORIGINAL_CLI_COMMAND"); ok && v != "" {
+		args[0] = v
 	}
 	idx := slices.Index(args, "bake")
 
--- a/bake/entitlements.go
+++ b/bake/entitlements.go
@@ -1,0 +1,176 @@
+package bake
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/containerd/console"
+	"github.com/docker/buildx/build"
+	"github.com/moby/buildkit/util/entitlements"
+	"github.com/pkg/errors"
+)
+
+type EntitlementKey string
+
+const (
+	EntitlementKeyNetworkHost      EntitlementKey = "network.host"
+	EntitlementKeySecurityInsecure EntitlementKey = "security.insecure"
+	EntitlementKeyFSRead           EntitlementKey = "fs.read"
+	EntitlementKeyFSWrite          EntitlementKey = "fs.write"
+	EntitlementKeyFS               EntitlementKey = "fs"
+	EntitlementKeyImagePush        EntitlementKey = "image.push"
+	EntitlementKeyImageLoad        EntitlementKey = "image.load"
+	EntitlementKeyImage            EntitlementKey = "image"
+	EntitlementKeySSH              EntitlementKey = "ssh"
+)
+
+type EntitlementConf struct {
+	NetworkHost      bool
+	SecurityInsecure bool
+	FSRead           []string
+	FSWrite          []string
+	ImagePush        []string
+	ImageLoad        []string
+	SSH              bool
+}
+
+func ParseEntitlements(in []string) (EntitlementConf, error) {
+	var conf EntitlementConf
+	for _, e := range in {
+		switch e {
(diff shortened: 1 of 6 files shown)


bake/entitlements.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/moby/buildkit/util/entitlements"
+	"github.com/pkg/errors"
+)
+
+type EntitlementKey string
+
+const (
+	EntitlementKeyNetworkHost      EntitlementKey = "network.host"
+	EntitlementKeySecurityInsecure EntitlementKey = "security.insecure"
+	EntitlementKeyFSRead           EntitlementKey = "fs.read"
```
tonistiigi, line 23: Even though these additional entitlements are defined here, the current PR does not perform any validation for them yet.

tonistiigi:
While the `entitlements` key in HCL is new and could not be used before, `network: host` was enabled via compose and may potentially break if entitlements is not passed by the user.

```
 # cat docker-compose.yml
services:
  app:
    build:
      network: "host"
 #
 # docker buildx bake app --print
 => [internal] load local bake definitions                                                 0.0s
 => => reading docker-compose.yml 50B / 50B                                                0.0s
{
  "group": {
    "default": {
      "targets": [
        "app"
      ]
    }
  },
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "entitlements": [
        "network.host"
      ]
    }
  }
}
 # docker buildx bake app
 => [internal] load local bake definitions                                                 0.0s
 => => reading docker-compose.yml 50B / 50B                                                0.0s

 - Running build containers that can access host network

In order to not see this message in the future pass "--allow=network.host" to grant requested privileges.

docker buildx bake --allow=network.host app

Do you want to grant requested privileges and continue? [y/N]
```

crazy-max:
> While the `entitlements` key in HCL is new and could not be used before, `network: host` was enabled via compose and may potentially break if entitlements is not passed by the user.

I think this is fine looking at the recommendation printed during build.

> Additionally, I guess we now want to expose `network = "host"` via HCL/JSON as well? Note that the recommended way is for Dockerfile to define when it needs host network by setting `RUN --network=host` instead of overwriting the network for the whole build here.

Yes we need that

bake/entitlements.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	if term {
+		fmt.Fprintf(out, "\nIn order to not see this message in the future pass %q to grant requested privileges.\n", strings.Join(flags, " "))
+	} else {
+		fmt.Fprintf(out, "\nPass %q to grant requested privileges.\n", strings.Join(flags, " "))
+	}
+
+	args := append([]string(nil), os.Args...)
+	if filepath.Base(args[0]) == "docker-buildx" {
```
crazy-max, line 146: If the idea is to check if buildx runs as docker plugin, then it would be better to use `plugin.RunningStandalone()`

crazy-max:
@dvdksn Seems a docs link is broken: https://github.com/docker/buildx/actions/runs/10636095098/job/29487219028?pr=2666#step:6:467

```
 #29 [validate-upstream 5/5] RUN htmltest
#29 0.066 htmltest started at 04:36:56 on public
#29 0.066 ========================================================================
#29 1.050 reference/cli/docker/buildx_bake/index.html
#29 1.050   hash does not exist --- reference/cli/docker/buildx_bake/index.html --> /build/attestations/slsa-provenance/#provenance-attestation-example
#29 1.050 reference/cli/docker/buildx_build/index.html
#29 1.205   hash does not exist --- reference/cli/docker/buildx_build/index.html --> /build/attestations/slsa-provenance/#provenance-attestation-example
#29 1.205 reference/cli/docker/buildx/build/index.html
#29 1.923   hash does not exist --- reference/cli/docker/buildx/build/index.html --> /build/attestations/slsa-provenance/#provenance-attestation-example
#29 1.923 reference/cli/docker/buildx/bake/index.html
#29 1.954   hash does not exist --- reference/cli/docker/buildx/bake/index.html --> /build/attestations/slsa-provenance/#provenance-attestation-example
#29 1.954 ========================================================================
#29 12.36 ✘✘✘ failed in 12.29200319s
#29 12.36 4 errors in 2598 documents
```

dvdksn:
@crazy-max let me fix those in a follow-up

tonistiigi:
I'm not sure what this error is about. `buildx bake validate-docs` passes for me locally.

bake/entitlements.go
tonistiigi: This isn't quite it. It is because when `docker` invokes the command then the arg is always `docker-buildx` so we shouldn't just replace it based on any command name.

But I now see that there is `DOCKER_CLI_PLUGIN_ORIGINAL_CLI_COMMAND` . I'll see if I can reuse it for this.

tonistiigi:
Looks like the docs thing went away. Green now.

crazy-max:
> Looks like the docs thing went away. Green now.

Yes this has been sorted with https://github.com/docker/buildx/pull/2652

crazy-max:
Needs follow-up on actions-toolkit repo https://github.com/docker/actions-toolkit/blob/38d1dce1ff8cfbf9c74491fd4bc3df0ca9b66b77/src/types/buildx/bake.ts#L26

PR #2671, opened 2024-09-04
bake: add network mode support to HCL and docs for network and entitlements

tonistiigi:
follow-up #2666

### DIFF
--- a/docs/bake-reference.md
+++ b/docs/bake-reference.md
@@ -505,6 +505,25 @@ $ docker buildx bake --print -f - <<< 'target "default" {}'
 }
 ```
 
+### `target.entitlements`
+
+Entitlements are permissions that the build process requires to run.
+
+Currently supported entitlements are:
+
+- `network.host`: Allows the build to use commands that access the host network. In Dockerfile, use [`RUN --network=host`](https://docs.docker.com/reference/dockerfile/#run---networkhost) to run a command with host network enabled.
+
+- `security.insecure`: Allows the build to run commands in privileged containers that are not limited by the default security sandbox. Such container may potentially access and modify system resources. In Dockerfile, use [`RUN --security=insecure`](https://docs.docker.com/reference/dockerfile/#run---security) to run a command in a privileged container.
+
+```hcl
+target "integration-tests" {
+  # this target requires privileged containers to run nested containers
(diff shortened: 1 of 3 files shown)


docs/bake-reference.md
```diff
@@ -749,6 +768,27 @@ target "app" {
 }
 ```
 
+### `target.network`
+
+Specify the network mode for the whole build request. This will override the the default network mode
```
crazy-max, line 773:
```suggestion
Specify the network mode for the whole build request. This will override the default network mode
```

crazy-max (changes requested):
Small typo otherwise LGTM :rocket:

@dvdksn PTAL for copy

docs/bake-reference.md
```diff
@@ -505,6 +505,25 @@ $ docker buildx bake --print -f - <<< 'target "default" {}'
 }
 ```
 
+### `target.entitlements`
+
+Entitlements are permissions that the build process requires to run.
+
+Currently supported entitlements are:
+
+- `network.host`: Allows the build to use commands that access the host network. In Dockerfile, use `RUN --network=host` to run a command with host network enabled.
```
dvdksn, line 514:
```suggestion
- `network.host`: Allows the build to use commands that access the host network. In Dockerfile, use [`RUN --network=host`](https://docs.docker.com/reference/dockerfile/#run---networkhost) to run a command with host network enabled.
```

docs/bake-reference.md
```diff
@@ -505,6 +505,25 @@ $ docker buildx bake --print -f - <<< 'target "default" {}'
...
+
+- `security.insecure`: Allows the build to run commands in privileged containers that are not limited by the default security sandbox. Such container may potentially access and modify system resources. In Dockerfile, use `RUN --security=insecure` to run a command in a privileged container.
```
dvdksn, line 516:
```suggestion
- `security.insecure`: Allows the build to run commands in privileged containers that are not limited by the default security sandbox. Such container may potentially access and modify system resources. In Dockerfile, use [`RUN --security=insecure`](https://docs.docker.com/reference/dockerfile/#run---security) to run a command in a privileged container.
```

dvdksn (approved):

crazy-max (approved):

Issue #2542, opened 2024-06-20
source snippet paths for warnings/errors should be relative to context/filename


tonistiigi:
Buildkit does not see the full path of Dockerfile is context is a subdir or a local path is passed with `-f`. 

Command like

```
docker buildx build --check ./src/ld 
```

Gives me warnings
```
WARNING: WorkdirRelativePath - https://docs.docker.com/go/dockerfile/rule/workdir-relative-path/
Relative workdir "./apple-libtapi" can have unexpected results if the base image changes
Dockerfile:19
--------------------
  17 |     ARG LIBTAPI_VERSION
  18 |     RUN git clone https://github.com/tpoechtrager/apple-libtapi --depth 1 -b ${LIBTAPI_VERSION}
  19 | >>> WORKDIR ./apple-libtapi
  20 |     RUN --mount=target=/tmp/libtapi-cmake-args.patch,source=libtapi-cmake-args.patch \
  21 |         git apply /tmp/libtapi-cmake-args.patch
--------------------
```

Note the `Dockerfile:19` while the actual path to the Dockerfile is `./src/ld/Dockerfile:19`. In this case it is especially confusing as there is another Dockerfile in root directory as well. This also makes debugging harder as for example VSCode can detect these paths and can directly open a correct line, but atm it would open a line from wrong file for this case.

PR #2672, opened 2024-09-04
build: display Dockerfile path on check warnings

daghack:
fixes https://github.com/docker/buildx/issues/2542

Currently, when printing warnings from the rule checks, it will only report the base Dockerfile name.
 
`docker build --check -f foo/bar/Dockerfile .`

```WARNING: ConsistentInstructionCasing - https://docs.docker.com/go/dockerfile/rule/consistent-instruction-casing/
Command 'copy' should match the case of the command majority (uppercase)
Dockerfile:2
--------------------
   1 |     FROM alpine
   2 | >>> copy ./tmp /tmp
   3 |
   4 |
--------------------
```

This updates printing for the warnings to include a more specific path if one is available.

```WARNING: ConsistentInstructionCasing - https://docs.docker.com/go/dockerfile/rule/consistent-instruction-casing/
Command 'copy' should match the case of the command majority (uppercase)
foo/bar/Dockerfile:2
--------------------
   1 |     FROM alpine
   2 | >>> copy ./tmp /tmp
   3 |
   4 |
--------------------
```

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -147,18 +147,19 @@ func toRepoOnly(in string) (string, error) {
 	return strings.Join(out, ","), nil
 }
 
-func Build(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer) (resp map[string]*client.SolveResponse, err error) {
+func Build(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer) (resp map[string]*client.SolveResponse, dockerfileMappings map[string]string, err error) {
 	return BuildWithResultHandler(ctx, nodes, opt, docker, configDir, w, nil)
 }
 
-func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer, resultHandleFunc func(driverIndex int, rCtx *ResultHandle)) (resp map[string]*client.SolveResponse, err error) {
+func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[string]Options, docker *dockerutil.Client, configDir string, w progress.Writer, resultHandleFunc func(driverIndex int, rCtx *ResultHandle)) (resp map[string]*client.SolveResponse, dockerfileMappings map[string]string, err error) {
+	dockerfileMappings = map[string]string{}
 	if len(nodes) == 0 {
-		return nil, errors.Errorf("driver required for build")
+		return nil, nil, errors.Errorf("driver required for build")
 	}
 
 	nodes, err = filterAvailableNodes(nodes)
 	if err != nil {
-		return nil, errors.Wrapf(err, "no valid drivers found")
+		return nil, nil, errors.Wrapf(err, "no valid drivers found")
 	}
 
 	var noMobyDriver *driver.DriverHandle
@@ -194,7 +195,7 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 
 	drivers, err := resolveDrivers(ctx, nodes, opt, w)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	defers := make([]func(), 0, 2)
@@ -227,14 +228,14 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opt map[s
 			opt.Platforms = np.platforms
 			gatewayOpts, err := np.BuildOpts(ctx)
 			if err != nil {
-				return nil, err
+				return nil, nil, err
 			}
-			so, release, err := toSolveOpt(ctx, np.Node(), multiDriver, opt, gatewayOpts, configDir, w, docker)
+			so, release, dockerfileMapping, err := toSolveOpt(ctx, np.Node(), multiDriver, opt, gatewayOpts, configDir, w, docker)
 			if err != nil {
-				return nil, err
+				return nil, nil, err
 			}
 			if err := saveLocalState(so, k, opt, np.Node(), configDir); err != nil {
-				return nil, err
+				return nil, nil, err
 			}
 			addGitAttrs(so)
 			defers = append(defers, release)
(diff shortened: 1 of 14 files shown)


tests/build.go
```diff
@@ -1291,6 +1291,29 @@ cOpy Dockerfile .
 		require.Error(t, cmd.Run(), stdout.String(), stderr.String())
 		require.Contains(t, stdout.String(), "Check complete, 2 warnings have been found!")
 	})
+
+	t.Run("check for Dockerfile name when target is set", func(t *testing.T) {
```
crazy-max, line 1295: Don't think that's the right description

daghack: Ah, good call! Definitely not the correct usage of the term `target` in this context! Updated.

tonistiigi (commented):
I think a problem with this one is that we are trying to parse the inputs from the user twice and assume that they get the same result. One is happening internally when builder converts input to `LocalMounts` inside `loadInputs` (that can create temp directories etc.), and other now here on converting the path for printing.

A better solution could be that when `loadInputs` sets up local mount by calling `setLocalMount`, then that mapping is then saved in that function so that it can be reversed later when printing. I think sourceInfo also has the `Definition` block so we can actually confirm the name of the local and match it with correct client directory.

This is better because:
- Recude duplication and possible fragmentation where one side gets changed or some formats are not supported in one implementation.
- Cases like inline Dockerfiles and stdin can show "(inline)" as a filename instead of bogus "Dockerfile".
- It also works with named contexts, instead of correct hardcoded context+dockerfile variants.


If this is considered too complex for initial PR we could leave that solution to future follow-up, but in that case I think we should only have exception for the one path, where we already define the name for the Dockerfile we expect to receive in the `commands` pkg and only pass this name instead of both raw `contextPath` and `dockerfileName` and then trying to figure out what is remote and how to combine them while printing. We should at least handle `-` as well.

daghack:
> If this is considered too complex for initial PR

Not too complex, I think! Ty you for the review!

daghack:
Massively refactored in order to take the approach suggested by @tonistiigi. :)

build/opt.go
```diff
@@ -353,23 +353,34 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt Op
 		so.Internal = true
 	}
 
-	return &so, releaseF, nil
+	return &so, releaseF, dockerfileMapping, nil
 }
 
-func loadInputs(ctx context.Context, d *driver.DriverHandle, inp Inputs, pw progress.Writer, target *client.SolveOpt) (func(), error) {
+type DockerfileMapping struct {
```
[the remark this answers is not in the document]
daghack: Updated to do it this way - I still ended up returning a `build.Inputs` because the top-level calls we need to return this information from use `controllerapi.BuildOptions` as an arg, and I didn't think updating that was the right call (because adding the info to GRPC-land seems like the wrong approach when we don't need to communicate this information across client/server boundary).

build/opt.go
```diff
@@ -440,6 +441,11 @@ func loadInputs(ctx context.Context, d *driver.DriverHandle, inp Inputs, pw prog
 
 	if inp.DockerfileInline != "" {
 		dockerfileReader = io.NopCloser(strings.NewReader(inp.DockerfileInline))
+		dockerfileSrcName = "inline"
+	} else if inp.DockerfilePath == "-" {
+		dockerfileSrcName = "stdin"
```
crazy-max, line 446: What will be the output for `inline` and `stdin` case? I guess it should be omitted like:

```
Command 'copy' should match the case of the command majority (uppercase)
--------------------
   1 |     FROM alpine
   2 | >>> copy ./tmp /tmp
   3 |
   4 |
--------------------
```

Something like this would be confusing:

```
Command 'copy' should match the case of the command majority (uppercase)
inline:2
--------------------
   1 |     FROM alpine
   2 | >>> copy ./tmp /tmp
   3 |
   4 |
--------------------
```

If someone has a Dockerfile named `inline`

daghack: Removing the line entirely requires either a `buildkit` PR or refactoring how `buildx` is printing warnings to effectively duplicate that functionality. I'm not opposed, but if that's the goal then it might be best to merge this and then make a corresponding issue in `buildkit` for the follow-up work.

I was trying to model what @tonistiigi had suggested, here. For MY part, I think that ambiguity around two very specifically named Dockerfiles (in this case, `stdin` and `inline`) is less confusing than simply not showing anything in these cases. That being said, it could be updated to something a little more clear and even less likely to be a filename?

```
Command 'copy' should match the case of the command majority (uppercase)
Inline Dockerfile:2
--------------------
...
```

```
Command 'copy' should match the case of the command majority (uppercase)
Dockerfile (inline):2
--------------------
...
```

crazy-max: > That being said, it could be updated to something a little more clear and even less likely to be a filename?

Yes that sounds good, let's check that for a follow-up then

daghack:
@crazy-max noted that this does not work for the `--debug` flag. After going over and through the code, it seems like that will be a surprisingly tricky lift to get this functionality applying there as well; the current plan is to make a follow up issue / separate PR for that body of work.

daghack:
@tonistiigi Looking for a review on this bad boy still. :)

build/build.go
```diff
@@ -95,12 +95,14 @@ type CallFunc struct {
...
-	ContextState     *llb.State
-	DockerfileInline string
-	NamedContexts    map[string]NamedContext
+	ContextPath          string
+	DockerfilePath       string
+	InStream             *SyncMultiReader
+	ContextState         *llb.State
+	DockerfileInline     string
+	NamedContexts        map[string]NamedContext
+	DockerfileMappingSrc string
```
tonistiigi, line 104: Add a comment that these are filled in by the builder

tonistiigi (approved):

build/build.go
daghack: Added :)

tonistiigi:
@daghack PTAL linter error

daghack:
> @daghack PTAL linter error

Ah, RIP. I now have an opinion about `gofmt` and how it interacts with comments in structs. 😂

tonistiigi (approved):

PR #2678, opened 2024-09-06
Update PROJECT mermaid daigrams

thompson-shaun:
There appears to be some issues with GitHub rendering mermaid diagrams where a node references itself.

```
apple --> apple
```

Results in

```mermaid
apple --> apple
```


thompson-shaun:
Render appears to be happy now -> https://github.com/thompson-shaun/buildx/blob/mermaid-fix/PROJECT.md#issue-lifecyle

Issue #2683, opened 2024-09-11
buildkitd.default.toml is not taken into account


dvdksn:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [ ] ... the documentation does not mention anything about my problem
- [ ] ... there are no open or closed issues that are related to my problem

### Description

By default, buildx should use a default buildkitd config if it exists in any of the following locations:

```
$BUILDX_CONFIG/buildkitd.default.toml
$DOCKER_CONFIG/buildx/buildkitd.default.toml
~/.docker/buildx/buildkitd.default.toml
```

But it does not seem like buildx is interpreting this config at all.

### Expected behaviour

Settings in e.g. `~/.docker/buildx/buildkitd.default.toml` should be automatically picked up with `buildx create`

### Actual behaviour

`buildkitd.default.toml` is ignored

### Buildx version

github.com/docker/buildx v0.17.0 78c8c28cf5685dd37d2a514147c28030763b8a03

### Builders list

```text
NAME/NODE              DRIVER/ENDPOINT     STATUS     BUILDKIT   PLATFORMS
container*             docker-container                          
 \_ container0          \_ desktop-linux   running    3a70550    linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
focused_cerf           docker-container                          
 \_ focused_cerf0       \_ desktop-linux   inactive              
default                docker                                    
 \_ default             \_ default         running    v0.15.2    linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64
desktop-linux          docker                                    
 \_ desktop-linux       \_ desktop-linux   running    v0.15.2    linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64
cloud-docker-default   cloud               error
```

### Configuration

```hcl
# docker-bake.hcl
target default {
  entitlements = ["security.insecure", "network.host"]
}
```

```toml
# ~/.docker/buildx/buildkitd.default.toml
insecure-entitlements = [ "network.host", "security.insecure" ]
```

```console
[...]

PR #2684, opened 2024-09-11
inspect: display buildkit daemon configuration file

crazy-max:
follow-up https://github.com/docker/buildx/issues/2683#issuecomment-2343001508

Displays buildkit daemon configuration file content used when creating a builder in inspect output:

```toml
# ~/.docker/buildx/buildkitd.default.toml

# debug enables additional debug logging
debug = true
# trace enables additional trace logging (very verbose, with potential performance impacts)
trace = true
# insecure-entitlements allows insecure entitlements, disabled by default.
insecure-entitlements = [ "network.host", "security.insecure" ]

[log]
  # log formatter: json or text
  format = "text"
```

```
$ docker buildx create --name builder --driver-opt "image=moby/buildkit:v0.16.0"
builder
```

```
$ docker buildx inspect builder --bootstrap
Name:          builder
Last Activity: 2024-09-11 08:31:27 +0000 UTC

Nodes:
Name:                  builder0
Endpoint:              unix:///var/run/docker.sock
Driver Options:        image="moby/buildkit:v0.16.0"
Status:                running
BuildKit daemon flags: --allow-insecure-entitlement=network.host
BuildKit version:      v0.16.0
Platforms:             linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
Features:
 Automatically load images to the Docker Engine image store: false
 Cache export:                                               true
 Docker exporter:                                            true
 Multi-platform build:                                       true
 OCI exporter:                                               true
Labels:
 org.mobyproject.buildkit.worker.executor:         oci
 org.mobyproject.buildkit.worker.hostname:         e5772cf17e94
 org.mobyproject.buildkit.worker.network:          host
 org.mobyproject.buildkit.worker.oci.process-mode: sandbox
 org.mobyproject.buildkit.worker.selinux.enabled:  false
 org.mobyproject.buildkit.worker.snapshotter:      overlayfs
GC Policy rule#0:
 All:           false
 Filters:       type==source.local,type==exec.cachemount,type==source.git.checkout
 Keep Duration: 48h0m0s
 Keep Bytes:    488.3MiB
GC Policy rule#1:
 All:           false
 Keep Duration: 1440h0m0s
 Keep Bytes:    94.06GiB
GC Policy rule#2:
 All:        false
 Keep Bytes: 94.06GiB
GC Policy rule#3:
 All:        true
 Keep Bytes: 94.06GiB
File#buildkitd.toml:
 > debug = true
 > insecure-entitlements = ["network.host", "security.insecure"]
 > trace = true
 >
 > [log]
 >   format = "text"
 >
```

### DIFF
--- a/commands/inspect.go
+++ b/commands/inspect.go
@@ -126,6 +126,12 @@ func runInspect(ctx context.Context, dockerCli command.Cli, in inspectOptions) e
 						fmt.Fprintf(w, "\tKeep Bytes:\t%s\n", units.BytesSize(float64(rule.KeepBytes)))
 					}
 				}
+				for f, dt := range nodes[i].Files {
+					fmt.Fprintf(w, "File#%s:\n", f)
+					for _, line := range strings.Split(string(dt), "\n") {
+						fmt.Fprintf(w, "\t> %s\n", line)
+					}
+				}
 			}
 		}
 	}
--- a/tests/inspect.go
+++ b/tests/inspect.go
@@ -1,6 +1,9 @@
 package tests
 
 import (
+	"os"
+	"path"
+	"regexp"
 	"strings"
 	"testing"
 
@@ -19,6 +22,7 @@ var inspectTests = []func(t *testing.T, sb integration.Sandbox){
 	testInspect,
 	testInspectBuildkitdFlags,
 	testInspectNetworkHostEntitlement,
+	testInspectBuildkitdConf,
 }
 
 func testInspect(t *testing.T, sb integration.Sandbox) {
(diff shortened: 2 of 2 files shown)


commands/inspect.go
```diff
@@ -126,6 +126,14 @@ func runInspect(ctx context.Context, dockerCli command.Cli, in inspectOptions) e
 						fmt.Fprintf(w, "\tKeep Bytes:\t%s\n", units.BytesSize(float64(rule.KeepBytes)))
 					}
 				}
+				if debug.IsEnabled() {
```
crazy-max, line 129: Not quite sure if we should do this only when debug is enabled :thinking:

tonistiigi: I think if file is set, then just show it.

hack/test-driver
```diff
@@ -105,7 +105,7 @@ if [ "$DRIVER" != "docker" ]; then
 fi
 
 # inspect and bootstrap
-buildxCmd inspect --bootstrap --builder="${builderName}"
+buildxCmd --debug inspect --bootstrap --builder="${builderName}"
```
tonistiigi, line 108: Why is this needed?

tonistiigi (approved):

crazy-max:
needs follow-up on actions-toolkit repo: https://github.com/docker/actions-toolkit/blob/dd3654acca27188c34a78ee9270a5670056123dd/src/buildx/builder.ts#L85

PR #2685, opened 2024-09-11
builder: do not set network.host entitlement flag if already set in buildkitd conf

crazy-max:
follow-up https://github.com/docker/buildx/issues/2683#issuecomment-2343100380

On container builder creation, it's possible to define buildkitd configuration path that will injected and used inside the container.

If insecure-entitlements opts are set in this configuration they will be skipped because we are always setting the `--allow-insecure-entitlement network.host` https://github.com/docker/buildx/pull/2266 if user does not set it explicitly and buildkitd overrides configuration values with flags: https://github.com/moby/buildkit/blob/3a7055008a5e58a2abbe0e0c21c919d9e014e062/cmd/buildkitd/main.go#L583-L584.

With this change we are now detecting if `network.host` insecure entitlement is set within the buildkitd configuration and skip setting `--allow-insecure-entitlement network.host` flag.

cc @dvdksn

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -435,7 +435,16 @@ func Create(ctx context.Context, txn *store.Txn, dockerCli command.Cli, opts Cre
 		return nil, err
 	}
 
-	buildkitdFlags, err := parseBuildkitdFlags(opts.BuildkitdFlags, driverName, driverOpts)
+	buildkitdConfigFile := opts.BuildkitdConfigFile
+	if buildkitdConfigFile == "" {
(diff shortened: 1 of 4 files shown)


crazy-max:
@tonistiigi As follow-up I think it would make sense to merge insecure entitlements from flags and configuration in BuildKit https://github.com/moby/buildkit/blob/3a7055008a5e58a2abbe0e0c21c919d9e014e062/cmd/buildkitd/main.go#L583-L584. WDYT?

tonistiigi:
v1.17.1 for this?

PR #2686, opened 2024-09-11
dockerfile, ci: update buildkit to latest stable

crazy-max:


.github/workflows/build.yml
```diff
@@ -45,9 +45,9 @@ jobs:
           - master
           - latest
           - buildx-stable-1
```
crazy-max, line 47: `buildx-stable-1` still `v0.15.2` but will be updated soon

tonistiigi (approved):

PR #2713, opened 2024-09-26
protobuf: remove gogoproto

jsternberg:
Removes gogo/protobuf from buildx and updates to a version of
moby/buildkit where gogo is removed.

This also changes how the proto files are generated. This is because
newer versions of protobuf are more strict about name conflicts. If two
files have the same name (even if they are relative paths) and are used
in different protoc commands, they'll conflict in the registry.

Since protobuf file generation doesn't work very well with
`paths=source_relative`, this removes the `go:generate` expression and
just relies on the dockerfile to perform the generation.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/docker/buildx
 
-go 1.21.0
+go 1.22.0
+
+toolchain go1.22.2
```
crazy-max, line 5: remove toolchain, see https://github.com/docker/buildx/pull/2577#pullrequestreview-2157133499

build/url.go
```diff
@@ -43,7 +44,7 @@ func createTempDockerfileFromURL(ctx context.Context, d *driver.DriverHandle, ur
 		if err != nil {
 			return nil, err
 		}
-		if stat.Size() > 512*1024 {
+		if proto.Size(stat) > 512*1024 {
```
tonistiigi, line 47: This is meant to be file size. We also should probably increase the limit (in a different PR).

commands/prune.go
```diff
@@ -148,10 +149,14 @@ func pruneCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&options.all, "all", "a", false, "Include internal/frontend images")
 	flags.Var(&options.filter, "filter", `Provide filter values (e.g., "until=24h")`)
-	flags.Var(&options.keepStorage, "keep-storage", "Amount of disk space to keep for cache")
+	flags.Var(&options.minStorage, "min-storage", "Minimum amount of disk space to keep for cache")
```
tonistiigi, line 152: This part needs updates based on https://github.com/moby/buildkit/pull/5359 

Not a blocker but need to make sure we don't ship this version.

controller/pb/progress.go
```diff
@@ -124,3 +127,38 @@ func FromControlStatus(resp *StatusResponse) *client.SolveStatus {
...
+	if ts == nil {
+		return nil
+	}
+	return timestamppb.New(*ts)
+}
+
+func digestSliceFromPB(elems []string) []digest.Digest {
+	clone := make([]digest.Digest, len(elems))
+	for i, e := range elems {
+		clone[i] = digest.Digest(e)
```
tonistiigi, line 153: Ideally all digest casts(when data comes from API) go through `digest.Parse`. There was a similar follow-up to buildkit PR.

tonistiigi (approved):
Some follow-ups but I think we can bring this in.

build/url.go
jsternberg: https://github.com/docker/buildx/pull/2716

PR #2716, opened 2024-10-03
build: raise maximum size limit for dockerfile and fix size check

jsternberg:
Raise the maximum size limit for the dockerfile and correct the size check. The size check was intended to use the size attribute from the file stat, but the original gogo version confused the `Size()` method (which returned the size of the proto message) with the `Size` attribute (which was named `Size_`).

During the conversion, we noticed the mistake but kept the incorrect behavior for the sake of keeping the conversion simple.

This also raises the maximum limit because 512 kB is likely a bit too conservative. The limit has been raised to 2 MB and the limit has been included in the error message.

### DIFF
--- a/build/url.go
+++ b/build/url.go
@@ -7,12 +7,14 @@ import (
 
 	"github.com/docker/buildx/driver"
 	"github.com/docker/buildx/util/progress"
+	"github.com/docker/go-units"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/client/llb"
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/pkg/errors"
-	"google.golang.org/protobuf/proto"
 )
+
+const maxDockerfileSize = 2 * 1024 * 1024 // 2 MB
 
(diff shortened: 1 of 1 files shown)


build/url.go
```diff
@@ -44,8 +48,8 @@ func createTempDockerfileFromURL(ctx context.Context, d *driver.DriverHandle, ur
 		if err != nil {
 			return nil, err
 		}
-		if proto.Size(stat) > 512*1024 {
-			return nil, errors.Errorf("Dockerfile %s bigger than allowed max size", url)
+		if stat.Size > maxDockerfileSizeBytes {
+			return nil, errors.Errorf("Dockerfile %s bigger than allowed max size (%s)", url, maxDockerfileSizeHumanReadable)
```
tonistiigi, line 52: For string value use `maxDockerfileSizeBytes * units.B` https://pkg.go.dev/github.com/tonistiigi/units or `units.HumanSize` https://pkg.go.dev/github.com/docker/go-units#HumanSize . Both are already vendored.

invisiblepancake (approved):

tonistiigi (approved):

PR #2724, opened 2024-10-08
hack: generate vtproto files for buildx

jsternberg:
Integrates vtproto into buildx. The generated files dockerfile has been modified to copy the buildkit equivalent file to ensure files are laid out in the appropriate way for imports.

An import has also been included to change the grpc codec to the version in buildkit that supports vtproto. This will allow buildx to utilize the speed and memory improvements from that.


commands/prune.go
```diff
@@ -149,12 +150,13 @@ func pruneCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
...
-	flags.Var(&options.minStorage, "min-storage", "Minimum amount of disk space to keep for cache")
-	flags.Var(&options.maxStorage, "max-storage", "Maximum amount of disk space to keep for cache")
+	flags.Var(&options.reservedSpace, "reserved-space", "Amount of disk space always allowed to keep for cache")
+	flags.Var(&options.minFreeSpace, "min-free-space", "Target amount of free disk space after pruning")
+	flags.Var(&options.maxUsedSpace, "max-used-space", "Maximum amount of disk space allowed to keep for cache")
 	flags.BoolVar(&options.verbose, "verbose", false, "Provide a more verbose output")
 	flags.BoolVarP(&options.force, "force", "f", false, "Do not prompt for confirmation")
 
-	flags.Var(&options.maxStorage, "keep-storage", "Amount of disk space to keep for cache")
+	flags.Var(&options.reservedSpace, "keep-storage", "Amount of disk space to keep for cache")
```
tonistiigi, line 159: We can mark this hidden I think.

There is no problem with both using `&options.reservedSpace` as target? Maybe we should check this separately and also detect if both flags are set.

jsternberg: `MarkDeprecated` makes it hidden too along with a message. I don't believe there's any problem with both being set. The second will just overwrite the first and then they'll get a deprecated message. I don't think we need to detect if both are set because they should have the same behavior.

commands/prune.go
```diff
@@ -149,12 +150,13 @@ func pruneCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags := cmd.Flags()
 	flags.BoolVarP(&options.all, "all", "a", false, "Include internal/frontend images")
 	flags.Var(&options.filter, "filter", `Provide filter values (e.g., "until=24h")`)
-	flags.Var(&options.minStorage, "min-storage", "Minimum amount of disk space to keep for cache")
```
thaJeztah, line 152: Do these need a deprecation as well?

tonistiigi: No, these were never in a release.

tonistiigi (approved):

Issue #2386, opened 2024-04-08
secret handling documentation for env missing attribute keys


betelgeuse:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

https://docs.docker.com/reference/cli/docker/buildx/build/#secret

Under env type secret there's a text: "env - Secret environment variable. id used if unset, otherwise will look for src, source if id unset."

This implies that there's undocumented src and source attributes.

### Expected behaviour

Document the src and source attributes

### Actual behaviour

They are not documented under env

### Buildx version

4b3c3c8401f53e7f2263ad0555df79a4516cd1c7

### Builders list

```text
N/A
```

### Configuration

N/A

### Additional info

https://github.com/docker/buildx/blob/4b3c3c8401f53e7f2263ad0555df79a4516cd1c7/docs/reference/buildx_build.md#L647

PR #2740, opened 2024-10-16
docs: clarify options for secret types (file, env)

dvdksn:
Improve descriptions of the attributes for the `--secret` flag for `buildx build`.

- Clarify how buildx determines whether to use `type=file` or `type=env`.
- Clarify that relative filepaths for `type=file` are calculated based on CWD where the buildx command is executed.
- Add `src` and `source` as possible attributes for `type=env`, but clarify that they require `type=env` to be set explicitly if you want to source from an env var (otherwise `type=file` is assumed).

## Related issues

- Closes docker/buildx#2386

### DIFF
--- a/docs/reference/buildx_build.md
+++ b/docs/reference/buildx_build.md
@@ -626,7 +626,7 @@ For example, the following Dockerfile contains four stages:
 ```dockerfile
 # syntax=docker/dockerfile:1
 
-FROM oven/bun:1 as base
+FROM oven/bun:1 AS base
 WORKDIR /app
 
 FROM base AS install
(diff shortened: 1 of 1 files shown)


dvdksn:
check failures are unrelated, but would be nice to make sure `validate-docs` can run;

```
ERROR: failed to authorize: failed to fetch oauth token: unexpected status from POST request to https://auth.docker.io/token: 502 Bad Gateway
```

I lack permissions to restart the failed jobs.

crazy-max (approved):
Looks good thanks!

tonistiigi (approved):

PR #2742, opened 2024-10-17
build: add OTEL span around build function

tonistiigi:
Small improvement for OTEL trace to group all the grpc requests made by build function under the same parent.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -500,7 +500,9 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
 						resultHandle, rr, err = NewResultHandle(ctx, cc, *so, "buildx", buildFunc, ch)
 						resultHandleFunc(dp.driverIndex, resultHandle)
 					} else {
+						span, ctx := tracing.StartSpan(ctx, "build")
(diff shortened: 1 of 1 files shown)


build/build.go
```diff
@@ -500,7 +500,9 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
 						resultHandle, rr, err = NewResultHandle(ctx, cc, *so, "buildx", buildFunc, ch)
 						resultHandleFunc(dp.driverIndex, resultHandle)
 					} else {
+						span, ctx := tracing.StartSpan(ctx, "Build")
```
crazy-max, line 503: Does casing matters?

```suggestion
						span, ctx := tracing.StartSpan(ctx, "build")
```

tonistiigi: 🤷 Is there a convention? Or should we define one?

crazy-max: Don't think there is any convention but it's lowercasing from what I see on buildkit repo: https://github.com/search?q=repo%3Amoby%2Fbuildkit%20tracing.StartSpan&type=code

Issue #1889, opened 2020-01-15
[proposal] implement docker build --pull=missing|never|always


thaJeztah:

anordal:
Any news on this? From the user's perspective, BuildKit broke offline, the OG backend is now deprecated, and as of Docker 29.4, there is still no --pull=missing:

> docker buildx build --pull=missing .
> invalid argument "missing" for "--pull" flag: strconv.ParseBool: parsing "missing": invalid syntax

There is thankfully the workaround of pinning the base images by sha256sum instead of by tag, but tags were satisfactory and used to work, while sha256sums is a bit of maintenance burden that feels unnecessary: Surely, it must be possible to figure out that a base image with the right tag exists locally without reaching out to the internet each time.

beanscg:
Bean Labs took a public-source pass on this issue.

The narrowest useful next step looks like: Offer to identify the exact docs page and propose a minimal wording/example clarification before any patch.

Public evidence checked:
- Open public issue in `docker/buildx`; repo license `Apache-2.0`; no assignee; 10 comments.
- Labels: kind/enhancement, area/cli, status/needs-maintainer-decision.
- Locator from title/labels; future commenter should ask for preferred docs/test file before drafting.

A good acceptance check would be a narrow public change that directly exercises the issue's reported boundary.

No PR opened from this pass.

PR #2744, opened 2024-10-17
docs: bake pull attr is a boolean

dvdksn:


crazy-max (approved):
Good catch!

This might change in the future https://github.com/docker/buildx/issues/1889 :eyes:

PR #2745, opened 2024-10-17
config: fix file/folder ownership

crazy-max:
alternative and closes https://github.com/docker/buildx/pull/2739

* fixes https://github.com/docker/for-mac/issues/7454
* fixes https://github.com/docker/for-mac/issues/7356
* fixes https://github.com/docker/for-mac/issues/7277
* fixes https://github.com/docker/for-mac/issues/7362
* fixes https://github.com/docker/for-mac/issues/7195
* fixes https://github.com/docker/for-mac/issues/7144
* fixes https://github.com/docker/for-mac/issues/7186
* fixes https://github.com/docker/for-mac/issues/7309

When buildx runs through `sudo`, file and folder created are owned by root user. This is fine if `HOME` is not kept in `/etc/sudoers` config which is the default behavior on unix systems and would therefore write to `/root` but this is not the case on macOS:

```
Defaults        env_keep += "HOME MAIL"
```

```
$ sudo env|sort
HOME=/Users/crazymax
LOGNAME=root
MAIL=/var/mail/root
PATH=...
SHELL=/bin/sh
SUDO_COMMAND=/usr/bin/env
SUDO_GID=20
SUDO_UID=502
SUDO_USER=crazymax
TERM=xterm-256color
USER=root
```

Checked as well when using macos GitHub Runners:

image

If we try to create a builder with `sudo`, it would write with `root` owner in user's home:

```
$ sudo docker buildx create --name foo
foo
$ tree -pugfi ~/.docker/buildx
[drwx------ crazymax staff   ]  /Users/crazymax/.docker/buildx
[drwx------ crazymax staff   ]  /Users/crazymax/.docker/buildx/activity
[-rw------- crazymax staff   ]  /Users/crazymax/.docker/buildx/activity/desktop-linux
[-rw------- root     staff   ]  /Users/crazymax/.docker/buildx/activity/foo
[-rw------- crazymax staff   ]  /Users/crazymax/.docker/buildx/current
[drwx------ crazymax staff   ]  /Users/crazymax/.docker/buildx/defaults
[drwx------ crazymax staff   ]  /Users/crazymax/.docker/buildx/instances
[-rw------- root     staff   ]  /Users/crazymax/.docker/buildx/instances/foo
```

If we then try to remove this builder without sudo we got:

```
$ docker buildx rm foo
failed to remove foo: open /Users/crazymax/.docker/buildx/instances/foo: permission denied
ERROR: failed to remove one or more builders
```

Same if try to list builders:

```
$ docker buildx ls
ERROR: open /Users/crazymax/.docker/buildx/instances/foo: permission denied
```

To fix this issue and avoid changing file permissions like https://github.com/docker/buildx/pull/2739, we need to check if buildx is running through `sudo` and set right owner when writing file or folder in configuration directory.


util/confutil/config.go
```diff
@@ -1,39 +1,140 @@
...
 
-// ConfigDir will look for correct configuration store path;
-// if `$BUILDX_CONFIG` is set - use it, otherwise use parent directory
-// of Docker config file (i.e. `${DOCKER_CONFIG}/buildx`)
-func ConfigDir(dockerCli command.Cli) string {
-	if buildxConfig := os.Getenv("BUILDX_CONFIG"); buildxConfig != "" {
-		logrus.Debugf("using config store %q based in \"$BUILDX_CONFIG\" environment variable", buildxConfig)
-		return buildxConfig
+var (
+	sudoerUID = -1
```
tonistiigi, line 18: Put this in a struct. 
```
type user struct{uid, gid}
var sudoer *user
```

util/confutil/config_unix.go
```diff
@@ -0,0 +1,46 @@
...
+
+package confutil
+
+import (
+	"os"
+	"os/user"
+	"strconv"
+)
+
+func init() {
```
tonistiigi, line 12: Put this in `NewConfig` instead of `init`. We can also check that configdir is under the home dir. If ran with `DOCKER_CONFIG=/foobar` then we don't need to convert things I think.

util/confutil/config.go
```diff
@@ -1,39 +1,140 @@
...
+			if err := os.Chown(filepath.Join(c.dir, filepath.Join(parts[:i+1]...)), sudoerUID, sudoerGID); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// WriteFile writes data to a file within the config dir
+func (c *Config) WriteFile(filename string, data []byte, perm os.FileMode) error {
```
tonistiigi, line 90: Why do we need non-atomic write?

util/confutil/config.go
```diff
@@ -1,39 +1,140 @@
...
+func (c *Config) MkdirAll(dir string, perm os.FileMode) error {
+	d := filepath.Join(c.dir, dir)
+	if err := os.MkdirAll(d, perm); err != nil {
+		return err
+	}
+	if sudoerUID != -1 && sudoerGID != -1 {
+		// apply chown to each directory level
+		parts := strings.Split(dir, string(filepath.Separator))
+		for i := range parts {
+			if err := os.Chown(filepath.Join(c.dir, filepath.Join(parts[:i+1]...)), sudoerUID, sudoerGID); err != nil {
```
tonistiigi, line 81: Shouldn't this only run if `MkdirAll` actually created the directory component?

util/confutil/config.go
```diff
@@ -1,39 +1,140 @@
...
+// Dir will look for correct configuration store path;
+// if `$BUILDX_CONFIG` is set - use it, otherwise use parent directory
+// of Docker config file (i.e. `${DOCKER_CONFIG}/buildx`)
+func (c *Config) Dir() string {
+	return c.dir
+}
+
+// BuildKitConfigFile returns the default BuildKit configuration file path
+func (c *Config) BuildKitConfigFile() (string, bool) {
+	f := filepath.Join(c.dir, "buildkitd.default.toml")
```
tonistiigi, line 64: Not new code but nicer if this string is in const.

util/confutil/config_unix.go
crazy-max: > If ran with `DOCKER_CONFIG=/foobar` then we don't need to convert things I think.

Hum right I missed this case. Maybe also check if `DOCKER_CONFIG` is not part of home dir as well.

util/confutil/config.go
crazy-max: This is the current behavior for https://github.com/docker/buildx/blob/9f0ebd2643482722e54eae975df4193df5b456f1/util/confutil/node.go#L23

But we can switch to atomic.

util/confutil/config.go
crazy-max: Hum maybe but that would fix currently broken permissions on folders for some users. We could check if last directory level has right perms and skip this iteration?

tonistiigi: I don't think last component makes a difference as `MkdirAll` should have same behavior for all components.

Maybe run `stat` in here and if the owner is root but the conditions are correct for sudo then chown.

util/confutil/config.go
tonistiigi: Yes, this can be atomic.

util/confutil/config.go
crazy-max: As this logic already exists in `fsutil` in `copy` package, I switched to `fs.MkdirAll`: https://github.com/tonistiigi/fsutil/blob/3f140a1299b0928e425ca320d5b8eb396eee0d66/copy/mkdir.go#L10

util/confutil/config.go
```diff
@@ -1,39 +1,144 @@
...
+		return nil
+	}
+	st, err := os.Stat(d)
+	if err != nil {
+		return err
+	}
+	downer := osutil.FileOwner(st)
+	if downer != nil && (downer.UID != c.chowner.UID || downer.GID != c.chowner.GID) {
+		return os.Chown(d, c.chowner.UID, c.chowner.GID)
+	}
```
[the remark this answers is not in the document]
crazy-max: @tonistiigi I see you have a pending change that seems exactly what I need https://github.com/tonistiigi/fsutil/pull/214 but for ownership instead of timestamp :eyes:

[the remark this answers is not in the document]
tonistiigi: For this change if you want to use `fsutil` implementation, can't you just pass a chowner? Why do you need to get a list of created dirs?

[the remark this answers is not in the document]
tonistiigi: Ah, you already did that part. This question was about fixing old broken permissions? You can't do that with the `fsutil` implementation as well, as it only returns the directories it created. Not old directories.

util/confutil/config.go
```diff
@@ -1,39 +1,144 @@
...
+	configDir := co.dir
+	if configDir == "" {
+		configDir = os.Getenv("BUILDX_CONFIG")
+		if configDir == "" {
+			configDir = filepath.Join(filepath.Dir(dockerCli.ConfigFile().Filename), "buildx")
+		}
+	}
+
+	var usr *osutil.User
+	home, _ := os.UserHomeDir()
```
tonistiigi, line 52: This homedir logic seems duplicated. I would just make sudo detection private helper in here and directly pass the `configDir` to the helper in here to check if the sudoer is needed or not.

util/confutil/config.go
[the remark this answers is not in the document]
crazy-max: True it makes sense in fsutil only for newly created dirs. I keep this logic then.

util/confutil/config.go
```diff
@@ -1,39 +1,142 @@
...
+		}
+	}
+	d := filepath.Join(c.dir, dir)
+	if err := fs.MkdirAll(d, perm, chown, nil); err != nil {
+		return err
+	}
+	if c.chowner == nil {
+		return nil
+	}
+	st, err := os.Stat(d)
```
tonistiigi, line 92: Add a comment in here that this part is to fix the owners own already existing directories. This can also be before the `MkdirAll` function that makes it clearer.

util/confutil/config.go
```diff
@@ -1,39 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+	downer := fileOwner(st)
```
tonistiigi, line 96: just `owner` :)

tonistiigi (approved):

PR #2746, opened 2024-10-17
pprof: take cpu and memory profiles by setting environment variables

jsternberg:
When run in standalone mode, the environment variables `DOCKER_BUILDX_CPU_PROFILE` and `DOCKER_BUILDX_MEM_PROFILE` will cause profiles to be written by the CLI.

### DIFF
--- a/cmd/buildx/debug.go
+++ b/cmd/buildx/debug.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"context"
+	"os"
+	"runtime"
+	"runtime/pprof"
+
+	"github.com/moby/buildkit/util/bklog"
+	"github.com/sirupsen/logrus"
+)
+
+func setupDebugProfiles(ctx context.Context) (stop func()) {
+	var stopFuncs []func()
+	if fn := setupCPUProfile(ctx); fn != nil {
+		stopFuncs = append(stopFuncs, fn)
+	}
+	if fn := setupHeapProfile(ctx); fn != nil {
+		stopFuncs = append(stopFuncs, fn)
+	}
+	return func() {
+		for _, fn := range stopFuncs {
+			fn()
+		}
+	}
+}
+
+func setupCPUProfile(ctx context.Context) (stop func()) {
+	if cpuProfile := os.Getenv("BUILDX_CPU_PROFILE"); cpuProfile != "" {
+		f, err := os.Create(cpuProfile)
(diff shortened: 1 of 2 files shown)


cmd/buildx/debug.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return func() {
+		for _, fn := range stopFuncs {
+			fn()
+		}
+	}
+}
+
+func setupCPUProfile(ctx context.Context) (stop func()) {
+	if cpuProfile := os.Getenv("DOCKER_BUILDX_CPU_PROFILE"); cpuProfile != "" {
```
crazy-max, line 29: I think this could just be `BUILDX_CPU_PROFILE` to be consistent with other envs: https://docs.docker.com/build/building/variables/#build-tool-configuration-variables

jsternberg: Done. I also added a snippet to capture the log when this is run as a plugin too. It was previously just for standalone runs.

crazy-max (approved):

dvdksn:
is this for 0.18.0?

cmd/buildx/main.go
```diff
@@ -40,6 +40,9 @@ func init() {
 }
 
 func runStandalone(cmd *command.DockerCli) error {
+	stopProfiles := setupDebugProfiles(context.TODO())
```
tonistiigi, line 43: Why not call this from `main()` directly to avoid duplicate calls.

tonistiigi: This can also be a single defer line

tonistiigi (commented):
There are couple of places in `commands` package that call `os.Exit()` directly and would skip these defers. Can be follow-up but needs tracking issue then.

thompson-shaun:
> is this for 0.18.0?

As long as it's ready + meged 💪

cmd/buildx/main.go
jsternberg: This is just a style thing. I find this easier to read.

jsternberg:
Refactored the run into a single function that calls `runStandalone` or `runPlugin`. The main reason for duplicating that code was because of the calls to `os.Exit` in the main function. We can tackle that with a future clean up since there are also some other places that I think call `os.Exit` when they shouldn't.

tonistiigi (approved):

PR #2748, opened 2024-10-18
Improve consistency of linter rules messages

felipecrs:
I spend some time reviewing the messages of these rules, and in my opinion this PR improves them.

It's mainly about keeping consistency between the messages.


felipecrs:
Oh God, just realized it should have been done in the buildkit repo. Never mind.

PR #2758, opened 2024-10-28
bake: initial set of composable bake attributes

jsternberg:
This allows using either the csv syntax or object syntax to specify
certain attributes.

This applies to the following fields:
- output
- cache-from
- cache-to
- secret
- ssh

There are still some remaining fields to translate. Specifically
ulimits, annotations, and attest.

Related to #438.

### DIFF
--- a/bake/hcl_test.go
+++ b/bake/hcl_test.go
@@ -17,6 +17,7 @@ func TestHCLBasic(t *testing.T) {
 		target "db" {
 			context = "./db"
 			tags = ["docker.io/tonistiigi/db"]
+			output = ["type=image"]
 		}
 
 		target "webapp" {
@@ -25,6 +26,9 @@ func TestHCLBasic(t *testing.T) {
 			args = {
 				buildno = "123"
 			}
+			output = [
+				{ type = "image" }
+			]
 		}
 
 		target "cross" {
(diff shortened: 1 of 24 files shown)


jsternberg:
Very incomplete first draft. This requires some changes to the `gohcl` package in the `hcl` package for it to work. The biggest thing is overriding the `ImpliedType` method and copying the full implementation from `go-cty` so that we can customize the behavior for certain types. In this case, I create a `FromCtyValue` interface and that works for our purposes.

The only changes to `gohcl` is creating a `DecodeOptions` which allows us to inject our own version of `ImpliedType` into the decoder. I also made it so we can change convert, but modifying the convert implementation was a lot more involved than modifying `ImpliedType` and using the existing capsule type.

bake/types.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (e *ExportEntry) FromCtyValue(in cty.Value) error {
+	// Attempt to read as a cty.Map(cty.String) first as that's our primary form.
+	if conv, err := convert.Convert(in, cty.Map(cty.String)); err == nil {
+		// Build the csv string for parsing. We do this mostly because
+		// ParseExports expects that format and it's easier to do that then reimplement it
+		// in multiple locations.
+		var sb strings.Builder
+		for name, val := range conv.AsValueMap() {
```
tonistiigi, line 24: We shouldn't encode objects to `csv` string. That is backwards as the (typed) object is what we need.

If you want to reduce duplication then there can be generic `csv -> map[k]v` convert and then map to struct, although not sure if worth it.

jsternberg: Updated the prototype so the `map[k]v` version is the canonical form.

To reduce duplication, I've created a separate `ExportEntry` type with the same attributes. This type will implement `encoding.TextUnmarshaler` and `json.Unmarshaler` to control the marshaling and unmarshaling process in a consistent way. The type also has a `ToPB()` method to convert the type to the controller API version.

The csv parsing has been extracted into `UnmarshalText` and the original `ParseExports` method has been changed to be a loop that does this:

```
var out ExportEntry
if err := out.UnmarshalText([]byte(s)); err != nil {
    return nil, err
}
outs = append(outs, out.ToPB())
```

The a `FromCtyValue` method has been created so we can still detect it. But, we now only need one implementation that uses `UnmarshalText` and JSON unmarshaling and we should be able to reuse this implementation for each of these types.

jsternberg:
I've modified this to include a small fork of the `gohcl` package temporarily while we resolve any legal stuff getting the CLA signed for upstream.

crazy-max:
> I've modified this to include a small fork of the `gohcl` package temporarily while we resolve any legal stuff getting the CLA signed for upstream.

https://github.com/docker/buildx/blob/master/bake/hclparser/merged.go is also a fork from upstream. Maybe we could move it to an `hcl` package under `hclparser` one and then move `LICENSE` and `gohcl` in `hcl`?

```
bake/
└── hclparser
    └── hcl
        ├── gohcl
        │   ├── ...
        │   ├── ...
        ├── LICENSE
        └── merged.go
```

jsternberg:
This PR is becoming fairly large. There are a few more attributes that have to be migrated to work, but it might be good to review and merge what we have to land this feature. I'm changing this to ready to review while I work on the remaining attributes.

crazy-max:
Overall LGTM but still testing other cases like the `matrix` one so bare with me but would be nice to have more test cases.

jsternberg:
Putting this back to draft for the time being since the variable references don't work.

jsternberg:
Ready for review again. Variable resolution works. The key part is HCL doesn't support the capsule types when resolving variables so there's now code that converts the capsule types to native types when resolving variables.

bake/hcl_test.go
```diff
@@ -597,6 +601,82 @@ func TestHCLAttrsCustomType(t *testing.T) {
...
+
+	require.Equal(t, 1, len(c.Targets))
+	require.Equal(t, []string{"type=oci,dest=../out.tar"}, stringify(c.Targets[0].Outputs))
+	require.Equal(t, []string{"type=local,src=path/to/cache", "user/app:cache"}, stringify(c.Targets[0].CacheFrom))
+	require.Equal(t, []string{"type=local,dest=path/to/cache"}, stringify(c.Targets[0].CacheTo))
+	require.Equal(t, []string{"id=mysecret,src=/local/secret", "id=mysecret2,env=TOKEN"}, stringify(c.Targets[0].Secrets))
+	require.Equal(t, []string{"default", "key=path/to/key"}, stringify(c.Targets[0].SSH))
+}
+
+func TestHCLAttrsCapsuleTypeVars(t *testing.T) {
```
tonistiigi, line 643: Add also cases for
- accessing another target, not just properties of the same target
- using global variables in the object values
- mixing csv and objects (whatever the behavior is just to lock it down)

tonistiigi: - array assignment `cache-to =  target.app.cache-from`

jsternberg: Done.

tonistiigi (approved):

PR #2728, opened 2024-10-08
gha: set default permissions to "contents: read"

thaJeztah:
make the OpenSSF scorecard slightly happier;
https://securityscorecards.dev/viewer/?uri=github.com/docker/buildx

    Warn: no topLevel permission defined: .github/workflows/build.yml:1
    Warn: topLevel 'security-events' permission set to 'write': .github/workflows/codeql.yml:13
    Warn: no topLevel permission defined: .github/workflows/docs-release.yml:1
    Warn: no topLevel permission defined: .github/workflows/docs-upstream.yml:1
    Warn: no topLevel permission defined: .github/workflows/e2e.yml:1
    Warn: no topLevel permission defined: .github/workflows/labeler.yml:1
    Warn: no topLevel permission defined: .github/workflows/validate.yml:1

PR #2761, opened 2024-10-29
ci: fix workflow permissions

crazy-max:
relates to https://github.com/docker/buildx/pull/2728#issuecomment-2443587745

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -229,8 +229,6 @@ jobs:
     permissions:
       # required to write sarif report
       security-events: write
-      # required to check out the repository
-      contents: read
     steps:
       -
         name: Checkout
@@ -404,6 +402,9 @@ jobs:
 
   release:
     runs-on: ubuntu-24.04
+    permissions:
+      # required to create GitHub release
+      contents: write
     needs:
       - test-integration
       - test-unit
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -21,12 +21,10 @@ env:
 
 jobs:
   codeql:
+    runs-on: ubuntu-24.04
     permissions:
       actions: read
-      contents: read
       security-events: write
-
-    runs-on: ubuntu-24.04
     steps:
       -
         name: Checkout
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -229,8 +229,6 @@ jobs:
     permissions:
       # required to write sarif report
       security-events: write
-      # required to check out the repository
-      contents: read
```
thaJeztah, line 233: Will this remove the `contents: read` ? The way I understood it from their docs (but need to re-read) is that setting _any_ permissions resets all the other ones already set; https://github.com/docker/buildx/blob/704b2cc52dea1f2d2d7540d119144296ef520a17/.github/workflows/build.yml#L3-L10

In this case (my interpretation) it would mean that it only has `security-events: write`, but no longer `contents: read`.

.github/workflows/build.yml
```diff
@@ -404,6 +402,9 @@ jobs:
 
   release:
     runs-on: ubuntu-24.04
+    permissions:
+      # required to create GitHub release
+      contents: write
```
thaJeztah, line 407: Same question here

.github/workflows/build.yml
crazy-max: Jobs inherit from global permissions set for the workflow, hence why we got this error when trying to create a GitHub Release: https://github.com/docker/buildx/pull/2728#issuecomment-2443587745

.github/workflows/build.yml
crazy-max: This overrides the global perm `contents: read` from the workflow within this job.

tonistiigi (approved):

.github/workflows/build.yml
thaJeztah: Right, but I wondered if it _extended_ the top-level permissions or _overrides_ them. I know we added the `contents: read` because it was needed to run these in private forks (where `contents: read` is not available by default)

crazy-max: I'm pretty sure it extends, I will check on runner app repo.

crazy-max: You right it doesn't :man_facepalming:: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions

> If you specify the access for any of these permissions, all of those that are not specified are set to `none`.

Issue #2755, opened 2024-10-28
Allow to run bake {--list-targets, --list-vars} without having a docker daemon / builder instance available (similar to --print)


neumantm:
### Description

buildx bake already supports running with `--print` even if there is no docker daemon / builder instance available.

This exception is handled here: https://github.com/docker/buildx/blob/2bdf451b680126ea962c6bc4ee7bef19ce3f1a18/commands/bake.go#L119

I think it would be great if the same would apply to `--list-targets` and `--list-vars`.

I assume this is easy to add by just changing the above-mentioned condition as **I assume** the logic for these flags doesn't actually needs a builder and the function returns after processing either of these flags:

https://github.com/docker/buildx/blob/2bdf451b680126ea962c6bc4ee7bef19ce3f1a18/commands/bake.go#L188-L192

I would be open to create a PR, if there is no argument against this.

crazy-max:
Yes sounds good to me, feel free to open a PR, thanks!

PR #2763, opened 2024-10-29
Skip Builder Init For Bake List Flags

neumantm:
Add the flags `--list-targets` and `--list-variables` to the cases where initializing the builder can be skipped.

This allows the listing of targets and variables when no builder is available.

Resolves: docker/buildx#2755


crazy-max (approved):
LGTM thanks!

tonistiigi (approved):

PR #2766, opened 2024-10-29
prune: detect if buildkit supports newer storage filters

tonistiigi:


crazy-max (approved):

jsternberg (commented):
👍 late but I thought I'd give a thumbs up too.

PR #2777, opened 2024-11-04
Add ability to output json lists in metadata build file

LaurentGoderre:
Supports metadata descriptors list in BuildKit https://github.com/moby/buildkit/pull/5454

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -763,8 +763,11 @@ func decodeExporterResponse(exporterResponse map[string]string) map[string]inter
 		}
 		var raw map[string]interface{}
 		if err = json.Unmarshal(dt, &raw); err != nil || len(raw) == 0 {
-			out[k] = v
-			continue
+			var rawList []map[string]interface{}
+			if err = json.Unmarshal(dt, &rawList); err != nil || len(rawList) == 0 {
+				out[k] = v
+				continue
+			}
 		}
 		out[k] = json.RawMessage(dt)
 	}


commands/build.go
```diff
@@ -737,8 +737,30 @@ func checkWarnedFlags(f *pflag.Flag) {
...
 
-func writeMetadataFile(filename string, dt interface{}) error {
-	b, err := json.MarshalIndent(dt, "", "  ")
+func writeMetadataFile(filename string, response map[string]interface{}) error {
+	out := make(map[string]interface{})
+	for k, v := range response {
+		var dt []byte
+		j, ok := v.(string)
+		if ok {
+			d, err := base64.StdEncoding.DecodeString(string(j))
```
tonistiigi, line 746: There is already a decoder in `decodeExporterResponse` in the same code path, so it shouldn't be duplicated.

commands/build.go
```diff
@@ -737,8 +737,30 @@ func checkWarnedFlags(f *pflag.Flag) {
 	}
 }
 
-func writeMetadataFile(filename string, dt interface{}) error {
-	b, err := json.MarshalIndent(dt, "", "  ")
+func writeMetadataFile(filename string, response map[string]interface{}) error {
+	out := make(map[string]interface{})
+	for k, v := range response {
+		var dt []byte
+		j, ok := v.(string)
```
tonistiigi, line 744: If this means that any base64 string is decoded then I don't think that can be done as false positives can happen when regular string happens to be valid base64. Valid JSON array could work though, assuming that previous versions of buildx does not produce an error (previous versions failing to decode is unavoidable I guess).

LaurentGoderre:
@tonistiigi thanks! I missed that function!

commands/build.go
tonistiigi: I'm not sure if this comment was addressed. Looks like any value can be base64 encoded now. To avoid false positives, should be limited to JSON arrays and objects. Eg. "MTU0" should not be decoded to "154"

tonistiigi: But it does change the valid values from JSON objects to any value afaics

tonistiigi (approved):

Issue #2315, opened 2024-03-06
bake: improve error message when using incorrect format for setting labels 


nicks:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

setting labels with bake fails with "labels require name"

### Expected behaviour

```
docker buildx bake --set=*.labels=x.y=name
```

should give me an error message or link to a doc on how to specify labels correctly

### Actual behaviour

This works:

```
docker buildx build --label=x.y=name .
```

This fails:

```
docker buildx bake --set=*.labels=x.y=name
ERROR: labels require name
```

i wasn't able to find any examples in the docs about how to set a label from the cli

i eventually figured out that the right syntax is

```
docker buildx bake --set=*.labels.x.y=name
```

### Buildx version

github.com/docker/buildx v0.11.2-desktop.5 f20ec1393426619870066baba9618cf999063886

### Docker info

```text
Client: Docker Engine - Community
 Version:    25.0.3
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.11.2-desktop.5
    Path:     /home/nick/.pinata/downloads/docker-plugins/v0.11.2-desktop.5/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.3-desktop.1
    Path:     /usr/lib/docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container. (Docker Inc.)
    Version:  0.0.22
    Path:     /usr/lib/docker/cli-plugins/docker-debug
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.0
    Path:     /usr/lib/docker/cli-plugins/docker-dev
  extension: Manages Docker extensions (Docker Inc.)
    Version:  v0.2.22
    Path:     /usr/lib/docker/cli-plugins/docker-extension
  feedback: Provide feedback, right in your terminal! (Docker Inc.)
    Version:  v1.0.4
    Path:     /usr/lib/docker/cli-plugins/docker-feedback
  harmonia: Extend your localhost with cloud workspaces (Docker Inc.)
    Version:  0.8.30-2-g68be490-m
[...]

PR #2778, opened 2024-11-04
bake: improve error when using incorrect format for setting labels

jsternberg:
Improves the error message when using an incorrect format for setting labels. This includes the intended format directly in the error message instead of assuming the user knows what the format is.

Fixes https://github.com/docker/buildx/issues/2315.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -725,10 +725,12 @@ type Target struct {
 	linked bool
 }
 
-var _ hclparser.WithEvalContexts = &Target{}
-var _ hclparser.WithGetName = &Target{}
-var _ hclparser.WithEvalContexts = &Group{}
-var _ hclparser.WithGetName = &Group{}
+var (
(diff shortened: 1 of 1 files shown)


tonistiigi (commented):
Alternatively this could be in the style of "invalid format for _ , expecting _"

jsternberg:
Done. Updated the message. I think that likely communicates more explicitly.

bake/bake.go
```diff
@@ -865,23 +867,23 @@ func (t *Target) AddOverrides(overrides map[string]Override) error {
 			t.Dockerfile = &value
 		case "args":
 			if len(keys) != 2 {
-				return errors.Errorf("args require name")
+				return errors.Errorf("invalid format for args, expecting args.${name}")
```
tonistiigi, line 870: Maybe `args.<name>=<value>`

tonistiigi (approved):

PR #2790, opened 2024-11-14
bake: check for empty build network with compose

crazy-max:
relates to https://github.com/docker/docs/pull/21410#discussion_r1842300599


tonistiigi (commented):
What case is it fixing? Is the issue that compose parser is reusing some of the pointers it returns?

crazy-max:
> What case is it fixing? Is the issue that compose parser is reusing some of the pointers it returns?

This is similar to what we got with hcl in https://github.com/docker/buildx/pull/2689 but this time for compose

tonistiigi (approved):

PR #2794, opened 2024-11-18
bake: basic variable validation

crazy-max:
fixes https://github.com/docker/buildx/issues/1346

This adds basic validation for variables similar to what terraform does: https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules

As follow-up and with work being made in #2758 we could handle additional types to variables and handle some kind of pre-validation if type is not compatible but I think HCL already handles this kind of diag for evaluation already.

___

```hcl
variable "FOO" {
  validation {
    condition = FOO != ""
    error_message = "FOO is required."
  }
}

target "default" {
  args = {
    FOO = FOO
  }
}
```

```
$ docker buildx bake --print
#1 reading docker-bake.hcl 148B / 148B done
#1 DONE 0.0s
docker-bake.hcl:3
--------------------
   1 |     variable "FOO" {
   2 |       validation {
   3 | >>>     condition = FOO != ""
   4 |         error_message = "FOO is required."
   5 |       }
--------------------
ERROR: docker-bake.hcl:3,17-26: Validation failed; FOO is required.
```

```
$ FOO=baz docker buildx bake --print
#1 reading docker-bake.hcl 148B / 148B done
#1 DONE 0.0s
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "FOO": "baz"
      }
    }
  }
}
```

___

Multiple validations case:

```hcl
variable "FOO" {
  validation {
    condition = FOO != ""
    error_message = "FOO is required."
  }
  validation {
    condition = strlen(FOO) > 4
    error_message = "FOO must be longer than 4 characters."
  }
}

target "default" {
  args = {
    FOO = FOO
  }
}
```

```
$ docker buildx bake --print
#1 reading docker-bake.hcl 260B / 260B done
#1 DONE 0.0s
docker-bake.hcl:3
--------------------
   1 |     variable "FOO" {
   2 |       validation {
   3 | >>>     condition = FOO != ""
   4 |         error_message = "FOO is required."
   5 |       }
--------------------
ERROR: docker-bake.hcl:3,17-26: Validation failed; FOO is required., and 1 other diagnostic(s)
```

```
$ FOO=bar docker buildx bake --print
#1 reading docker-bake.hcl 260B / 260B done
#1 DONE 0.0s
docker-bake.hcl:7
--------------------
   5 |       }
   6 |       validation {
   7 | >>>     condition = strlen(FOO) > 4
   8 |         error_message = "FOO must be longer than 4 characters."
   9 |       }
--------------------
ERROR: docker-bake.hcl:7,17-32: Validation failed; FOO must be longer than 4 characters.
```

```
FOO=barbar docker buildx bake -f docker-bake.hcl --print
#1 reading docker-bake.hcl 264B / 264B done
#1 DONE 0.0s
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "FOO": "barbar"
      }
    }
  }
}
```

___

Dependency on other variables:

```hcl
variable "FOO" {}
variable "BAR" {
  validation {
    condition = FOO != ""
    error_message = "BAR requires FOO to be set."
  }
}

target "default" {
  args = {
    BAR = BAR
  }
}
```

```
$ docker buildx bake --print
#1 reading docker-bake.hcl 183B / 183B done
#1 DONE 0.0s
docker-bake.hcl:4
--------------------
   2 |     variable "BAR" {
   3 |       validation {
   4 | >>>     condition = FOO != ""
   5 |         error_message = "BAR requires FOO to be set."
   6 |       }
--------------------
ERROR: docker-bake.hcl:4,17-26: Validation failed; BAR requires FOO to be set.
```

```
$ FOO=baz docker buildx bake --print
#1 reading docker-bake.hcl 183B / 183B done
#1 DONE 0.0s
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "args": {
        "BAR": ""
      }
    }
  }
}
```

### DIFF
--- a/bake/bake_test.go
+++ b/bake/bake_test.go
@@ -1856,3 +1856,155 @@ func TestNetNone(t *testing.T) {
 	require.Len(t, bo["app"].Allow, 0)
 	require.Equal(t, "none", bo["app"].NetworkMode)
 }
+
+func TestVariableValidation(t *testing.T) {
+	fp := File{
+		Name: "docker-bake.hcl",
+		Data: []byte(`
+variable "FOO" {
+  validation {
+    condition = FOO != ""
+    error_message = "FOO is required."
+  }
+}
+target "app" {
+  args = {
+    FOO = FOO
+  }
+}
+`),
+	}
+
+	ctx := context.TODO()
+
+	t.Run("Valid", func(t *testing.T) {
+		t.Setenv("FOO", "bar")
+		_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, nil, nil)
+		require.NoError(t, err)
+	})
+
+	t.Run("Invalid", func(t *testing.T) {
+		_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, nil, nil)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "FOO is required.")
+	})
+}
+
+func TestVariableValidationMulti(t *testing.T) {
+	fp := File{
+		Name: "docker-bake.hcl",
+		Data: []byte(`
+variable "FOO" {
+  validation {
+    condition = FOO != ""
+    error_message = "FOO is required."
+  }
+  validation {
+    condition = strlen(FOO) > 4
+    error_message = "FOO must be longer than 4 characters."
+  }
+}
+target "app" {
+  args = {
+    FOO = FOO
+  }
+}
+`),
+	}
+
+	ctx := context.TODO()
+
+	t.Run("Valid", func(t *testing.T) {
+		t.Setenv("FOO", "barbar")
+		_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, nil, nil)
+		require.NoError(t, err)
+	})
+
+	t.Run("InvalidLength", func(t *testing.T) {
+		t.Setenv("FOO", "bar")
+		_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, nil, nil)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "FOO must be longer than 4 characters.")
+	})
+
+	t.Run("InvalidEmpty", func(t *testing.T) {
+		_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, nil, nil)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "FOO is required.")
+	})
+}
+
+func TestVariableValidationWithDeps(t *testing.T) {
+	fp := File{
+		Name: "docker-bake.hcl",
+		Data: []byte(`
+variable "FOO" {}
+variable "BAR" {
+  validation {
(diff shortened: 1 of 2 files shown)


bake/hclparser/hclparser.go
```diff
@@ -541,6 +547,31 @@ func (p *parser) resolveBlockNames(block *hcl.Block) ([]string, error) {
...
+		message, diags := v.Validation.Message.Value(ectx)
+		if diags.HasErrors() {
+			return diags
+		}
+		return hcl.Diagnostics{
+			&hcl.Diagnostic{
+				Severity: hcl.DiagError,
+				Summary:  "Validation failed",
+				Detail:   message.AsString(),
+				Subject:  v.Validation.Condition.Range().Ptr(),
```
crazy-max, line 568: Was wondering if we should make the subject the whole var block with `v.Body.MissingItemRange().Ptr(),` or just the condition like here?

bake/bake_test.go
```diff
@@ -1856,3 +1856,37 @@ func TestNetNone(t *testing.T) {
...
 	require.Equal(t, "none", bo["app"].NetworkMode)
 }
+
+func TestVariableValidation(t *testing.T) {
+	fp := File{
+		Name: "docker-bake.hcl",
+		Data: []byte(`
+variable "FOO" {
+  validation {
+    condition = FOO != ""
```
tonistiigi, line 1866: Does the condition only allow current variable or links to others(like the value)? If it does allow then could we add tests.

bake/hclparser/hclparser.go
```diff
@@ -25,11 +25,17 @@ type Opt struct {
...
 type variable struct {
-	Name        string         `json:"-" hcl:"name,label"`
-	Default     *hcl.Attribute `json:"default,omitempty" hcl:"default,optional"`
-	Description string         `json:"description,omitempty" hcl:"description,optional"`
-	Body        hcl.Body       `json:"-" hcl:",body"`
-	Remain      hcl.Body       `json:"-" hcl:",remain"`
+	Name        string              `json:"-" hcl:"name,label"`
+	Default     *hcl.Attribute      `json:"default,omitempty" hcl:"default,optional"`
+	Description string              `json:"description,omitempty" hcl:"description,optional"`
+	Validation  *variableValidation `json:"validation,omitempty" hcl:"validation,block"`
```
tonistiigi, line 31: Shouldn't this be array type?

bake/bake_test.go
crazy-max: Yes it allows access to other variables, I will add additional tests

bake/hclparser/hclparser.go
crazy-max: Yes we should, implemented and added tests.

tonistiigi (approved):

PR #2796, opened 2024-11-19
bake: add filesystem entitlements support

### LINKED ISSUE #2697: Possible to escape working directory with `cwd://` for remote plaintext Bake definitions
### Contributing guidelines

### Description

Bake should error out if you try to escape the current working directory in a remote Bake definition. This validation only occurs for remote Bake files in Git. For remote plaintext files, no validation occurs.

### Expected behaviour

ERROR: path /private/tmp is outside of the working directory, please set BAKE_ALLOW_REMOTE_FS_ACCESS=1

### Actual behaviour

It copies the files fine.

### Buildx version

github.com/docker/buildx 056cf8a 056cf8a7ca083d91eccf9200e7e2c99ff170bbaf

### Docker info

_No response_

### Builders list

```text
Name:          container
Last Activity: 2024-09-13 13:46:41 +0000 UTC

Nodes:
Name:                  container0
Endpoint:              desktop-linux
Driver Options:        default-load="true" image="moby/buildkit:master"
Status:                running
BuildKit daemon flags: --allow-insecure-entitlement=network.host
BuildKit version:      3a70550
```

### Configuration

```hcl
target "default" {
  context = "cwd:///tmp"
  dockerfile-inline = <<EOT
FROM alpine
WORKDIR /src
COPY . .
RUN ls -l && stop
EOT
}

target "named" {
  contexts = {
    tmp = "cwd:///tmp"
  }
  dockerfile-inline = <<EOT
FROM alpine
WORKDIR /src
COPY --from=tmp . .
RUN ls -l && stop
EOT
}
```

tonistiigi:
Adds support for entitlements when bake definition accesses local files. Accessing files within the current working directory is allowed by default.

part of #179
closes #2697

In this version, privileged path accesses for local bake definitions only warns, unless `BUILDX_BAKE_ENTITLEMENTS_FS` is set. In next version, the intention is to make this case fail as with other entitlements.


tonistiigi:
@profnandaa FYI, if you would like to double-check any of the Windows path stuff. We can also ping you on follow-up PR>

tonistiigi:
Discussing with @crazy-max the difference seems to be that in windows `EvalSymlinks` has extra `toNorm` call https://github.com/golang/go/blob/c37d9a0f9313ff2936b1b0b82476416e6cb03c62/src/path/filepath/symlink_windows.go#L113 that is missing in our custom `evaluateToExistingPath` implementation.

PR #2797, opened 2024-11-20
ci: update runner to macos-14

crazy-max:
relates to https://github.com/actions/runner-images/issues/10721

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -158,7 +158,7 @@ jobs:
       matrix:
         os:
           - ubuntu-24.04
-          - macos-12
+          - macos-14
           - windows-2022
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -158,7 +158,7 @@ jobs:
       matrix:
         os:
           - ubuntu-24.04
-          - macos-12
+          - macos-14
```
AkihiroSuda, line 161: Why not 15?

crazy-max: It's still in beta https://github.com/actions/runner-images?tab=readme-ov-file#available-images, preferred to be on stable for now.

AkihiroSuda (approved):

PR #2814, opened 2024-11-22
bake: various fixes for composable attributes

### LINKED ISSUE #2822: [v0.19.0] bake: Unsuitable value: EOF
```hcl
variable "DESTDIR" {
  default = ""
}

target "default" {
  output = [DESTDIR]
}
```

```
$ docker buildx bake --print
#1 reading docker-bake.hcl 81B / 81B done
#1 DONE 0.0s
docker-bake.hcl:6
--------------------
   4 |
   5 |     target "default" {
   6 | >>>   output = [DESTDIR]
   7 |     }
   8 |
--------------------
ERROR: docker-bake.hcl:6,12-13: Unsuitable value type; Unsuitable value: EOF
```

Done some debugging and seems related to changes in `gohcl` fork. This error is triggered at https://github.com/docker/buildx/blob/71c78897192730a2a22f74310f15d0fbf93a157f/bake/hclparser/gohcl/decode.go#L295-L305

Image

As we are using our own decode options for implied types in https://github.com/docker/buildx/blob/71c78897192730a2a22f74310f15d0fbf93a157f/bake/hclparser/hclparser.go#L992, the bug might be around: https://github.com/docker/buildx/blob/master/bake/hclparser/type_implied.go

With previous release:

```
$ docker buildx bake --print
#1 reading docker-bake.hcl 81B / 81B done
#1 DONE 0.0s
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile"
    }
  }
}
```

### LINKED ISSUE #2823: bake: cache-to/cache-from attributes output credentials with --print since Buildx 0.19.0
also related to https://github.com/docker/buildx/pull/2758

Give the following definition:

```hcl
target "default" {
  cache-from = [
    "type=s3,region=eu-west-1,bucket=mybucket"
  ]
  cache-to = [
    "type=s3,region=eu-west-1,bucket=mybucket",
    "type=inline"
  ]
}
```

And using `--print` flag to output canonical representation:

```
$ docker buildx bake --print
#1 reading docker-bake.hcl 266B / 266B done
#1 DONE 0.0s
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "cache-from": [
        {
          "access_key_id": "[REDACTED]",
          "bucket": "mybucket",
          "region": "eu-west-1",
          "secret_access_key": "[REDACTED]",
          "type": "s3"
        },
        {
          "ref": "user/repo:cache",
          "type": "registry"
        }
      ],
      "cache-to": [
        {
          "access_key_id": "[REDACTED]",
          "bucket": "mybucket",
          "region": "eu-west-1",
          "secret_access_key": "[REDACTED]",
          "type": "s3"
        },
        {
          "type": "inline"
        }
      ],
      "output": [
        {
          "type": "cacheonly"
        }
      ]
    }
  }
}
```

It prints the credentials where it should not (marked as `[REDACTED]`.

With previous release:

```
$ docker buildx bake --print
#1 reading ./__tests__/.fixtures/bake-03.hcl 266B / 266B done
#1 DONE 0.0s
{
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "cache-from": [
        "type=s3,region=eu-west-1,bucket=mybucket",
        "user/repo:cache"
      ],
      "cache-to": [
        "type=s3,region=eu-west-1,bucket=mybucket",
        "type=inline"
      ],
      "output": [
        "type=cacheonly"
      ]
    }
  }
}
```

Didn't check if `secret` attributes behaves in a similar way.

jsternberg:
This changes how the composable attributes are implemented and provides
various fixes to the first iteration.

Cache-from and cache-to now no longer print sensitive values that are
automatically added. These automatically added attributes are added when
the protobuf is created rather than at the time of parsing so they will
no longer be printed. If they are part of the original configuration
file, they will still be printed.

Empty strings will now be skipped. This was the original behavior and
composable attributes removed this functionality accidentally. This
functionality is now restored.

This also expands the available syntax that works with each of the
composable attributes. It is now possible to interleave the csv syntax
with the object syntax without any problems. The canonical form is still
the object syntax and variables are resolved according to that syntax.

Fixes #2823.
Fixes #2822.
Fixes https://github.com/docker/buildx/issues/2858

Replaces #2832.
Replaces #2833.


crazy-max (approved):
Tested on my side on various projects and looks great!

tests/bake.go
```diff
@@ -149,6 +152,124 @@ RUN echo "Hello ${HELLO}"
...
+			"HCL",
+			"docker-bake.hcl",
+			[]byte(`
+target "build" {
+  args = {
+    HELLO = "foo"
+  }
+
+  cache-from = [
+    "type=gha",
```
crazy-max, line 171: small nit, In https://github.com/docker/buildx/pull/2827/files#diff-095bae3e0563ad1f84bf1ca114f4687cc22db9bbbdd54299055015838038f235R84 I specifically sets `token` attribute in the definition to make sure it's being printed as intended if you want to add this new case.

util/buildflags/utils.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/gocty"
+)
+
+type comparable[E any] interface {
+	Equal(other E) bool
+}
+
+func removeDupes[E comparable[E]](s []E) []E {
```
tonistiigi, line 12: This seems to be duplicated. Additionally (not for this PR) there is also `removeDupesStr` but it uses map. I think this could also avoid N^2 complexity but not sure if worth it as the inputs are likely small.

bake/bake.go
```diff
@@ -739,12 +739,12 @@ func (t *Target) normalize() {
 	t.Annotations = removeDupesStr(t.Annotations)
 	t.Attest = removeAttestDupes(t.Attest)
 	t.Tags = removeDupesStr(t.Tags)
-	t.Secrets = removeDupes(t.Secrets)
+	t.Secrets = t.Secrets.Normalize()
 	t.SSH = removeDupes(t.SSH)
```
tonistiigi, line 743: `t.SSH.Normalize()` ?

bake/bake.go
```diff
@@ -1333,13 +1333,8 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
...
 
-	secrets := make([]*controllerapi.Secret, len(t.Secrets))
-	for i, s := range t.Secrets {
-		secrets[i] = s.ToPB()
-	}
-	bo.SecretSpecs = secrets
-
-	secretAttachment, err := controllerapi.CreateSecrets(secrets)
+	bo.SecretSpecs = t.Secrets.ToPB()
+	secretAttachment, err := controllerapi.CreateSecrets(bo.SecretSpecs)
```
tonistiigi, line 1337: Not sure why similar `SSH` code is different from this.

jsternberg: There were just a lot of changes and I missed this. Fixed.

util/buildflags/utils.go
jsternberg: The purpose of this was it could utilize the `Equal` method that was added so it wasn't so dependent on strings. Due to the small number of inputs, the N^2 complexity isn't likely to be an issue. This might actually be faster than `removeDupesStr` just because maps have some issues when used for deduplicating small inputs due to overhead from memory allocation and hash key lookups.

This was duplicated and I've removed the duplicate from `bake/bake.go`. When I originally did this, it was in `bake/bake.go` and then I started moving stuff. It was easier to just copy it at the time. Since I missed the change to call `Normalize()` for `t.SSH` as pointed out in the other comment, the linter didn't catch that the original call was no longer needed since all calls had been moved to this file. I've changed that to use `Normalize()` and removed the duplicate.

tests/bake.go
jsternberg: Good idea. Done.

tonistiigi (approved):

PR #2812, opened 2024-11-22
bake: add wildcard to fs entitlements to allow any paths

crazy-max:
follow-up https://github.com/docker/buildx/pull/2796#issuecomment-2491867068

PR #2815, opened 2024-11-25
bake: fix entitlement test when running from symlink temp

tonistiigi:
As the paths returned by validator have the symlinks resolved, the test needs to resolve the symlinks also in the expected values. Previously this would fail if t.TempDir() or os.GetWd() returned a path that contained a symlink.

The issue was purely in the test and not in the entitlements validation logic.

### DIFF
--- a/bake/entitlements_test.go
+++ b/bake/entitlements_test.go
@@ -180,10 +180,18 @@ func TestValidateEntitlements(t *testing.T) {
 	dir2, err := osutil.GetLongPathName(t.TempDir())
 	require.NoError(t, err)
 
+	// the paths returned by entitlements validation will have symlinks resolved
+	expDir1, err := filepath.EvalSymlinks(dir1)
+	require.NoError(t, err)
+	expDir2, err := filepath.EvalSymlinks(dir2)
+	require.NoError(t, err)
+
 	escapeLink := filepath.Join(dir1, "escape_link")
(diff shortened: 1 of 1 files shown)


bake/entitlements_test.go
```diff
@@ -180,10 +180,18 @@ func TestValidateEntitlements(t *testing.T) {
 	dir2, err := osutil.GetLongPathName(t.TempDir())
 	require.NoError(t, err)
 
+	// the paths returned by entitlements validation will have symlinks resolved
+	expDir1, err := filepath.EvalSymlinks(dir1)
```
crazy-max, line 184: In this case I think we can remove `osutil.GetLongPathName` for `dir1` related to https://github.com/docker/buildx/pull/2812#issuecomment-2493632641

bake/entitlements_test.go
```diff
@@ -180,10 +180,18 @@ func TestValidateEntitlements(t *testing.T) {
...
+	require.NoError(t, err)
+	expDir2, err := filepath.EvalSymlinks(dir2)
+	require.NoError(t, err)
+
 	escapeLink := filepath.Join(dir1, "escape_link")
 	require.NoError(t, os.Symlink("../../aa", escapeLink))
 
 	wd := osutil.GetWd()
```
crazy-max, line 192: Same here, we can get back to `os.Getwd()`

crazy-max (approved):

PR #2832, opened 2024-12-02
bake: do not attempt to unmarshal empty attributes

crazy-max:


bake/bake_test.go
```diff
@@ -2019,6 +2019,24 @@ target "app" {
...
+  output = [""]
+}
+`),
+	}
+
+	ctx := context.TODO()
+
+	_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, nil, nil)
+	require.NoError(t, err)
+}
```
crazy-max, line 2038: This test works in current state but when looking at the `app` target it contains an empty `ExportEntry`, while previously there were none. So I guess we need to make some changes in the HCL parser as well.

crazy-max: It fails as expected: https://github.com/docker/buildx/actions/runs/12120057362/job/33787924177?pr=2832#step:6:1326

```
=== FAIL: bake TestEmptyAttribute (0.00s)
    bake_test.go:2038: 
        	Error Trace:	/home/runner/work/buildx/buildx/bake/bake_test.go:2038
        	Error:      	"[]" should have 0 item(s), but has 1
        	Test:       	TestEmptyAttribute
```

bake/bake_test.go
```diff
@@ -2019,6 +2019,26 @@ target "app" {
...
+  output = [""]
+}
+`),
+	}
+
+	ctx := context.TODO()
+
+	m, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, nil, nil)
+	require.Equal(t, 1, len(m))
+	require.Len(t, m["app"].Outputs, 0)
```
crazy-max, line 2038: It fails as expected: https://github.com/docker/buildx/actions/runs/12120057362/job/33787924177?pr=2832#step:6:1326

```
=== FAIL: bake TestEmptyAttribute (0.00s)
    bake_test.go:2038: 
        	Error Trace:	/home/runner/work/buildx/buildx/bake/bake_test.go:2038
        	Error:      	"[]" should have 0 item(s), but has 1
        	Test:       	TestEmptyAttribute
```

PR #2833, opened 2024-12-02
bake: cache-from/cache-to options no longer print sensitive values

jsternberg:
This refactors how the cache-from/cache-to composable attributes work so they no longer print sensitive values that are automatically added.

This also expands the available syntax that works with the cache options. It is now possible to interleave the csv syntax with the object syntax without any problems. The canonical form is still the object syntax and variables are resolved according to that syntax.

`cache-from` and `cache-to` now correctly ignore empty string inputs so these can be used with variables.

Fixes https://github.com/docker/buildx/issues/2823.

Partial fix for https://github.com/docker/buildx/issues/2822.

### DIFF
--- a/bake/hcl_test.go
+++ b/bake/hcl_test.go
@@ -606,7 +606,7 @@ func TestHCLAttrsCapsuleType(t *testing.T) {
 	target "app" {
 		cache-from = [
 			{ type = "registry", ref = "user/app:cache" },
-			{ type = "local", src = "path/to/cache" },
+			"type=local,src=path/to/cache",
 		]
 
 		cache-to = [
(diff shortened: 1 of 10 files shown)


bake/hcl_test.go
```diff
@@ -674,7 +674,7 @@ func TestHCLAttrsCapsuleTypeVars(t *testing.T) {
 		output = [ "type=oci,dest=../${foo}.tar" ]
 
 		secret = [
-			{ id = target.app.output[0].type, src = "/local/secret" },
+			{ id = target.app.output[0].type, src = "/${target.app.cache-from[1].type}/secret" },
```
crazy-max, line 677: @dvdksn I'm not sure we have examples showing how to reference attributes. I think the example with `cache-to = [ target.app.cache-from[0] ]` is a good one.

crazy-max (approved):

bake/hcl_test.go
dvdksn: We have a simpler example: https://docs.docker.com/build/bake/inheritance/#reusing-single-attributes-from-targets

We don't show interpolation though, maybe we should

jsternberg:
I'm going to put this into draft. I have a more complete fix for this but it'll also involve other changes.

jsternberg:
Batching in this change with some other fixes too in #2814. I'll close this one.

PR #2834, opened 2024-12-02
bake: fix entitlements path checks for local outputs

tonistiigi:
Previous check based on dest attributes was not correct as the attributes already get converted before validation happens.

Because the local path is not preserved for single-file outputs and gets replaced by io.Writer, a temporary array variable was needed. This value should instead be added to ExportEntry struct in BuildKit in future revision.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -194,7 +194,7 @@ func ListTargets(files []File) ([]string, error) {
 	return dedupSlice(targets), nil
 }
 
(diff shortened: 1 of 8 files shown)


tonistiigi:
Added another commit to automatically allow paths that are set via `--set` flags(for outputs, secrets and remote cache). This is to avoid the need to use both `--set` and `--allow` for the same paths. This also helps with the testsuite where this is a common pattern.

crazy-max (approved):
Good call on allowing automatically for overrides

PR #2835, opened 2024-12-03
docs: bake v0.19 entitlements

dvdksn:
- **docs(bake): improve docs on "call" and "description" in bake file**
- **docs: add docs for bake --allow**

### DIFF
--- a/docs/bake-reference.md
+++ b/docs/bake-reference.md
@@ -221,8 +221,10 @@ The following table shows the complete list of attributes that you can assign to
 | [`attest`](#targetattest)                       | List    | Build attestations                                                   |
 | [`cache-from`](#targetcache-from)               | List    | External cache sources                                               |
 | [`cache-to`](#targetcache-to)                   | List    | External cache destinations                                          |
+| [`call`](#targetcall)                           | String  | Specify the frontend method to call for the target.                  |
 | [`context`](#targetcontext)                     | String  | Set of files located in the specified path or URL                    |
 | [`contexts`](#targetcontexts)                   | Map     | Additional build contexts                                            |
+| [`description`](#targetdescription)             | String  | Description of a target                                              |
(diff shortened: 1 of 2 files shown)


docs/reference/buildx_bake.md
```diff
@@ -50,6 +50,66 @@ guide for introduction to writing bake files.
 
 ## Examples
 
+### <a name="allow"></a> Allow extra privileged entitelement (--allow)
```
tonistiigi, line 53: entitlement

docs/reference/buildx_bake.md
```diff
@@ -50,6 +50,66 @@ guide for introduction to writing bake files.
...
+
```
tonistiigi, line 54: Would be good to explain here a bit why allowing entitlements is needed and in what cases it protects the user.

crazy-max (approved):
LGTM thanks!

docs/bake-reference.md
```diff
@@ -481,6 +490,25 @@ FROM baseapp
...
+target "lint" {
+    description = "Runs golangci-lint to detect style errors"
+    args = {
+        GOLANGCI_LINT_VERSION = null
+    }
+    dockerfile = "lint.Dockerfile"
+}
+```
+
+This attribute is useful when combined with the `targets` frontend method,
```
tonistiigi, line 508: This isn't actually `targets` frontend method, but `bake --list-targets` flag as these descriptions are for bake targets, not Dockerfile stages.

tonistiigi (commented):
New entitlements content looks good

docs/bake-reference.md
dvdksn: Updated

dvdksn:
Updated from `--list-targets` to `--list=targets`

crazy-max:
rebased to fix conflict

tonistiigi (approved):

PR #2848, opened 2024-12-09
bake: implement composable attributes for attestations

jsternberg:


jsternberg:
Depends on https://github.com/docker/buildx/pull/2814.

util/buildflags/attests.go
```diff
@@ -10,6 +12,167 @@ import (
 	"github.com/tonistiigi/go-csvvalue"
 )
 
+type Attests []*Attest
```
tonistiigi, line 15: These types with `ToPB/String/(Un)Marshal*/FromToCty` are good candidates for unit tests. Some part of the helpers can probably be shared. I'm worried about cases where somebody makes a change in one of these functions but forgets to make a matching change in another.

tonistiigi:
We can leave the unit tests for follow-up as it contains fields not in this PR.

tonistiigi (approved):

PR #2830, opened 2024-11-29
Dockerfile: update to docker v27.4.0

thaJeztah:

PR #2851, opened 2024-12-10
dockerfiles: pin alpine version

crazy-max:
Similar to what we have on buildkit repo so we avoid side effects like https://github.com/docker/buildx/pull/2830#issuecomment-2528706259. Also bump alpine and xx version on remaining dockerfiles.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # syntax=docker/dockerfile:1
 
 ARG GO_VERSION=1.23
+ARG ALPINE_VERSION=3.21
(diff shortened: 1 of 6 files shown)


tonistiigi:
Why do we need this now that xx is on 1.6.1?

crazy-max:
> Why do we need this now that xx is on 1.6.1?

In case something else happens again in next Alpine release, so pinning would mitigate that kind of issue in the future.

tonistiigi (approved):

PR #2857, opened 2024-12-13
bump compose-go v2.4.6

ndeloof:


crazy-max (changes requested):
Missing `make vendor`

crazy-max (approved):
LGTM thanks!

PR #2861, opened 2024-12-16
[v0.19] bake: remove empty values set by --set

### LINKED ISSUE #2858: Regression in handling of empty lists on command line
### Contributing guidelines

### Description

Between v0.18.0 and v0.19.2, handling of null-values for keys from the command line changed, breaking existing configurations.

### Expected behaviour

We have a `docker-bake.yaml` which sets the `cache-to` key:

```hcl
target "default" {
  cache_to = [
    "type=local,dest=.docker-cache"
  ]
}
```

In some circumstances (some Github workflow builds, for example) we want to _not_ cache the results.   We would do this with:

```console
$ docker buildx bake --set "*.cache-to="
```

(or the equivalent through the `docker/bake-action` workflow action).

The expected behavior is:

```console
# Default bakefile
$ docker buildx bake --print

target "default" {
  cache_to = [
    "type=local,dest=.docker-cache"
  ]
}


# Bakefile without cache
$ docker buildx bake --print --set "*.cache="

target "default" {
  cache_to = []
}
```

### Actual behaviour

Actual behavior:

```console
# Bakefile without cache
> docker buildx bake --print --set "*.cache-to="
 => [internal] load local bake definitions                                                                                    0.0s
 => => reading docker-bake.hcl 77B / 77B                                                                                      0.0s
ERROR: EOF
```

### Buildx version

```
github.com/docker/buildx v0.19.2 1fc5647
```

### Docker info

```text
 Version:    27.4.0
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.19.2
    Version:  v2.31.0
```

### Builders list

```text
NAME/NODE                DRIVER/ENDPOINT                   STATUS    BUILDKIT   PLATFORMS
container-builder*       docker-container
 \_ container-builder0    \_ unix:///var/run/docker.sock   running   v0.17.1    linux/amd64 (+3), linux/arm64, linux/arm (+2), linux/ppc64le, (4 more)
default                  docker
 \_ default               \_ default                       running   v0.17.3    linux/amd64 (+3), linux/arm64, linux/arm (+2), linux/ppc64le, (4 more)
```

### Configuration

```dockerfile
FROM alpine
```

```hcl
target "default" {
  cache_to = [
    "type=local,dest=.docker-cache"
  ]
}
```

```console
$ docker buildx bake
$ docker buildx bake --set "*.cache-to="
```

### Build logs

```text

```

### Additional info

_No response_

tonistiigi:
regression from https://github.com/docker/buildx/pull/2846/commits/b2c0c26c26791019ffa0fc1153626d515553ca3d
fixes #2858 for v0.19

These parser functions are called for `--set` to
resolve entitlement paths that would be automatically added and will fail for empty value.

The empty values would already be ignored but in v0.19 is done after merging the `--set` values and then calling parse again.

### DIFF
--- a/bake/bake_test.go
+++ b/bake/bake_test.go
@@ -2028,3 +2028,20 @@ target "app" {
 	_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, nil, nil, &EntitlementConf{})
 	require.NoError(t, err)
 }
+
+// https://github.com/docker/buildx/issues/2858
+func TestOverrideEmpty(t *testing.T) {
+	fp := File{
+		Name: "docker-bake.hcl",
+		Data: []byte(`
+target "app" {
+  output = ["./bin"]
+}
+`),
+	}
+
+	ctx := context.TODO()
+
+	_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, []string{"app.output="}, nil, &EntitlementConf{})
+	require.NoError(t, err)
+}
--- a/util/buildflags/cache.go
+++ b/util/buildflags/cache.go
@@ -14,6 +14,9 @@ import (
 func ParseCacheEntry(in []string) ([]*controllerapi.CacheOptionsEntry, error) {
 	outs := make([]*controllerapi.CacheOptionsEntry, 0, len(in))
 	for _, in := range in {
+		if in == "" {
+			continue
+		}
 		fields, err := csvvalue.Fields(in, nil)
 		if err != nil {
 			return nil, err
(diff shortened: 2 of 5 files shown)


crazy-max:
https://github.com/docker/buildx/actions/runs/12347799964/job/34460332018?pr=2861#step:6:183

Needs to pick the xx update

crazy-max (approved):

bake/bake_test.go
```diff
@@ -2028,3 +2028,20 @@ target "app" {
...
+  output = ["./bin"]
+}
+`),
+	}
+
+	ctx := context.TODO()
+
+	_, _, err := ReadTargets(ctx, []File{fp}, []string{"app"}, []string{"app.output="}, nil, &EntitlementConf{})
+	require.NoError(t, err)
+}
```
crazy-max, line 2047: Fyi @jsternberg, we should pick this to https://github.com/docker/buildx/pull/2814

Issue #2853, opened 2024-12-11
Broken buildkitd config doesn't prevent creating a new builder


dvdksn:
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

### Description

Buildx does not prevent you from creating a new builder with a broken buildkitd config, but then it does not allow you to remove the builder if you did create it.

### Expected behaviour

I should not be able to create a broken builder.

### Actual behaviour

```console
$ cat cfg.toml 
[worker.oci]
	gc = "maybe"
$ docker buildx create --name=bad_builder --buildkitd-config cfg.toml 
ERROR: failed to parse config: (5, 5): Can't convert maybe(string) to bool
$ docker buildx ls
NAME/NODE           DRIVER/ENDPOINT     STATUS    BUILDKIT   PLATFORMS
default             docker                                   
 \_ default          \_ default         running   v0.17.3    linux/amd64 (+2), linux/arm64, linux/arm (+2), linux/ppc64le, (3 more)
desktop-linux*      docker                                   
 \_ desktop-linux    \_ desktop-linux   running   v0.17.3    linux/amd64 (+2), linux/arm64, linux/arm (+2), linux/ppc64le, (3 more)
bad_builder         docker-container    error                

Cannot load builder bad_builder: failed to parse config: (5, 5): Can't convert maybe(string) to bool
$ docker buildx rm bad_builder
failed to remove bad_builder: failed to parse config: (5, 5): Can't convert maybe(string) to bool
ERROR: failed to remove one or more builders
```

### Buildx version

github.com/docker/buildx 3e3242c 3e3242cfdda23ab9077dde54479a74d26fcc2242

### Docker info

```text

```

### Builders list

```text
NAME/NODE           DRIVER/ENDPOINT     STATUS    BUILDKIT   PLATFORMS
default             docker                                   
 \_ default          \_ default         running   v0.17.3    linux/amd64 (+2), linux/arm64, linux/arm (+2), linux/ppc64le, (3 more)
desktop-linux*      docker                                   
 \_ desktop-linux    \_ desktop-linux   running   v0.17.3    linux/amd64 (+2), linux/arm64, linux/arm (+2), linux/ppc64le, (3 more)
bad_builder         docker-container    error
```

[...]

PR #2864, opened 2024-12-16
builder: validate buildkit configuration

crazy-max:
fixes https://github.com/docker/buildx/issues/2853

We should validate buildkit configuration before creating the builder.

### DIFF
--- a/util/confutil/config.go
+++ b/util/confutil/config.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/docker/pkg/ioutils"
+	"github.com/moby/buildkit/cmd/buildkitd/config"
 	"github.com/pelletier/go-toml"
 	"github.com/pkg/errors"
 	fs "github.com/tonistiigi/fsutil/copy"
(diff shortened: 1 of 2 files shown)


util/confutil/config.go
```diff
@@ -151,7 +152,11 @@ func LoadConfigTree(fp string) (*toml.Tree, error) {
 	defer f.Close()
 	t, err := toml.LoadReader(f)
 	if err != nil {
-		return t, errors.Wrap(err, "failed to parse config")
+		return t, errors.Wrap(err, "failed to parse buildkit config")
+	}
```
thaJeztah, line 156: Wondering; will this be problematic when crating a builder using an older version of buildkit? (i.e. would validation only work correctly if the version matches the one that's vendored?)

crazy-max: I don't think this will be an issue as we are backward compat. Also we are already using this logic to get the registry config in https://github.com/docker/buildx/blob/567361d4948b06c559b41797c58e683f9a587583/store/storeutil/storeutil.go#L125-L128

thaJeztah: Ah yeah, I wasn't sure, so thought  I'd mention it. I recall that I ran into changes in the garbage-collect configuration that changed, and for which I needed to make changes in moby/moby, because it used those; https://github.com/moby/moby/pull/48634#issuecomment-2407072949

tonistiigi: Can we add a test for this, making sure a new unknown key doesn't cause a validation error.

crazy-max: @tonistiigi Done

PR #2866, opened 2024-12-17
ci: e2e bake

crazy-max:
Build some projects using bake with advanced features to catch regressions before release.


tonistiigi:
😬 this takes 17 minutes for a single job? If so, we can't run this for every PR.

crazy-max:
> 😬 this takes 17 minutes for a single job? If so, we can't run this for every PR.

Ah yes the xx build is quite long, I will find another target

crazy-max:
Added GHA cache export for this target to reduce build time:

image

Should always being cached as we are pinning the git context to a fixed revision.

tonistiigi (approved):

PR #2871, opened 2024-12-18
bake: test empty override

jsternberg:
Cherry pick of afc9cebb4861d02086e8f4d043256cfcebf1d3a9.

### DIFF
--- a/bake/bake_test.go
+++ b/bake/bake_test.go
@@ -2019,24 +2019,43 @@ target "app" {
 	})
 }
 
-// https://github.com/docker/buildx/pull/428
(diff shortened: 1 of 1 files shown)


tonistiigi (commented):
These are not testing the return value of `ReadTargets` like the old one did (same can be added to override case).

jsternberg:
@tonistiigi done and I also added a check to make sure that `app` was actually in the list of targets. This is because the length of `m["foo"].Outputs` would also be zero so we want to verify "app" is actually there.

tonistiigi (approved):

PR #2893, opened 2025-01-07
bump compose-go to v2.4.7

glours:


crazy-max (approved):
LGTM thanks!

About test change I will take a look if we can define the working dir when parsing the compose file as follow-up but I'm pretty sure this will break remote bake definition with local context handling.

PR #2891, opened 2025-01-07
Add buildx history command

tonistiigi:
implements https://github.com/docker/buildx/issues/2711

These commands allow working with build records
of completed and running builds.

Currently supports:

```
history ls
history logs
history rm
history open
history inspect
history inspect attachment
```

PR #2904, opened 2025-01-10
Add history trace command

tonistiigi:
Includes wip trace command on top of https://github.com/docker/buildx/pull/2891

The goal was for this command to show a trace of the build directly, for example, in Jaeger UI, but it looks like there isn't a simple API that would allow uploading a trace to an existing Jaeger image (their upload feature is an ephemeral client-side-only solution).

Options would be to upload the trace spans again via controller APIs or build our own UI-only image that we could pair with a simple server that can return the trace JSON. Separating from #2891 in order to not block that PR.

The Jaeger conversion is mostly reuse of some old code. Looks like there is also a `github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/jaeger` package now that probably has similar functionality.


tonistiigi:
Added support for viewing traces using Jaeger UI.

```
// open trace for a ref in UI
docker buildx history trace <ref>

// open trace for last build in UI
docker buildx history trace

// save trace to file (eg. manual upload)
docker buildx history trace > trace.json

// open comparison trace with another build
docker buildx history trace --compare=otherref ref

// compare last build with the one before it
docker buildx history trace --compare '^1'
```

I think the UX is pretty ok and better than manual uploads and running `jaeger-all-in-one`. Maintainence-wise, there may be some questions though.

The old otelutil conversion code may need some updates as well. Looks like some info, like process data, could be deduplicated. I quickly looked at `github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/jaeger` but I don't think we have the original format for that converter, so still some work needed to use that.

crazy-max:
https://github.com/docker/buildx/actions/runs/13173913374/job/36769042423?pr=2904#step:6:1301

```
 === RUN   TestJaegerData
    jaeger_test.go:26: 
        	Error Trace:	D:/a/buildx/buildx/util/otelutil/jaeger_test.go:26
        	Error:      	Not equal:
```

I can't repro this one on a windows machine :thinking:
I'm taking a closer look.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/docker/buildx
 
-go 1.23.0
+go 1.23.1
```
crazy-max, line 3: It should be 1.23.0. Seems related to https://github.com/tonistiigi/jaeger-ui-rest/blob/bb2801509720cbedaac29fbefeeef9ba24e54c45/go.mod#L3

Looking at code on this repo I don't think there is any requirement for this specific patch release.

.github/workflows/build.yml
```diff
@@ -174,6 +174,11 @@ jobs:
     env:
       SKIP_INTEGRATION_TESTS: 1
     steps:
+      -
+        name: Setup Git config
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
```
crazy-max, line 181: Ok seems on Windows runner eol is set to CRLF by default so needs this step before checking out.

crazy-max:
https://github.com/docker/buildx/actions/runs/13242497270/job/36960897099?pr=2904#step:6:31

```
Bringing machine 'default' up with 'libvirt' provider...
==> default: Box 'generic/freebsd14' could not be found. Attempting to find and install...
    default: Box Provider: libvirt
    default: Box Version: >= 0
==> default: Box file was not detected as metadata. Adding it directly...
==> default: Adding box 'generic/freebsd14' (v0) for provider: libvirt
    default: Downloading: https://vagrantcloud.com/generic/freebsd14

The box failed to unpackage properly. Please verify that the box
file you're trying to add is not corrupted and that enough disk space
is available and then try again.
The output from attempting to unpackage (if any):

bsdtar: Error opening archive: Unrecognized archive format
Error: Process completed with exit code 1.
```

This one seems related to https://github.com/hashicorp/vagrant/issues/13571, I'm taking a look

go.mod
crazy-max: Oh would need 1.22.0 actually looking at https://github.com/docker/buildx/pull/2978/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R3

crazy-max: Opened https://github.com/tonistiigi/jaeger-ui-rest/pull/1

crazy-max (approved):

crazy-max:
needs docs follow-up https://github.com/docker/buildx/blob/master/docs/reference/buildx_history_trace.md

PR #2905, opened 2025-01-10
bake: infer git auth token from remote files to build request

crazy-max:
When `BUILDX_BAKE_GIT_AUTH_*` env is set to fetch a remote bake definition, we should infer this git auth token to build request if a Git context is detected.


tonistiigi (commented):
This should be only if the context is git based I think. A remote bake definition may not always use the same remote URL as a context.

crazy-max:
> This should be only if the context is git based I think. A remote bake definition may not always use the same remote URL as a context.

I see but that could be unreliable if I use a remote bake def that has a target with subdir context like:

```hcl
target "default" {
  context = "./subdir"
}
```

But I guess I can just check context does not start with `cwd://`

tonistiigi:
`bo` would probably already have the context set correctly. If it is a subdir, then it would be a subdir of the git context in this case.

crazy-max:
> `bo` would probably already have the context set correctly. If it is a subdir, then it would be a subdir of the git context in this case.

Yes indeed, let me check this

crazy-max:
@tonistiigi Should be good now

bake/bake.go
```diff
@@ -1338,7 +1338,23 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 	}
 	bo.Platforms = platforms
 
-	bo.SecretSpecs = t.Secrets.ToPB()
+	secrets := t.Secrets
+	if build.IsRemoteURL(bi.ContextPath) {
```
tonistiigi, line 1342: I don't think this is correct. I think if you have a subpath in bake and command invoked directly then the contextpath in here is a subdir but the contextstate has been replaced with something like `llb.Git().Copy("subdir")`

crazy-max: Ah you right it would not work for this case

bake/bake.go
```diff
@@ -1179,6 +1179,16 @@ func updateContext(t *build.Inputs, inp *Input) {
 	t.ContextState = &st
 }
 
+func isRemoteContext(t build.Inputs, inp *Input) bool {
+	if build.IsRemoteURL(t.ContextPath) {
+		return true
+	}
+	if inp != nil && build.IsRemoteURL(inp.URL) && !strings.HasPrefix(t.ContextPath, "cwd://") {
```
tonistiigi, line 1186: isn't `inp.URL` always URL if it is set?

crazy-max: Yes if a remote bake def is set but if context is set to git context it is not (first cond in `isRemoteContext`):

```
docker buildx bake --set *.context=https://github.com/docker/buildx.git --print
{
  "group": {
    "default": {
      "targets": [
        "binaries"
      ]
    }
  },
  "target": {
    "binaries": {
      "dockerfile": "Dockerfile",
      "args": {
      },
      "target": "binaries",
      "platforms": [
        "local"
      ],
      "output": [
        {
          "dest": "./bin/build",
          "type": "local"
        }
      ]
    }
  }
}
```

tonistiigi: Isn't that the `build.IsRemoteURL(t.ContextPath) {`, how does that affect `inp.URL`?

crazy-max: > Isn't that the `build.IsRemoteURL(t.ContextPath) {`,

Yes it is but `inp.URL` is only set when remote bake def arg is set: https://github.com/docker/buildx/blob/026ac2313c75cb374a25b22474e1ba364a3fa323/commands/bake.go#L557

It doesn't take into account git urls in context within bake def or override.

tonistiigi (approved):

PR #2906, opened 2025-01-10
bake: support setting the context to null in bake

### LINKED ISSUE #2611: [bake] Cannot specify empty context for target
### Contributing guidelines

### Description

When using the legacy builder, you could read the Dockerfile from stdin to avoid sending any build context:
https://github.com/docker/cli/blob/24.0/docs/reference/commandline/build.md?plain=1#L267-L270
> This will read a Dockerfile from `STDIN` without context. Due to the lack of a
context, no contents of any local directory will be sent to the Docker daemon.
Since there is no context, a Dockerfile `ADD` only works if it refers to a
remote URL.

There does not appear to be any similar way to request a null build context for a bake target. I am aware that other enhancements to the build system (ie only sending files from context when referenced by a COPY command) have made this largely unnecessary, but it would be nice to be able to explicitly request an empty/null context for the target.

### Expected behaviour

Able to set a null context for the target.

### Actual behaviour

cannot set an empty target. Null or empty string for the `context` are translated to the default `.`, and use of `-` is rejected with `context from stdin not allowed in bake`

### Buildx version

github.com/docker/buildx v0.16.0 10c9ff901c1c255b58969dd2ade4789acbbab3bb

### Docker info

```text
 Version:    25.0.3
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.16.0
    Path:     /home/brandond/.docker/cli-plugins/docker-buildx
    Version:  v2.24.5

Server:
 Containers: 7
  Running: 7
  Paused: 0
  Stopped: 0
 Images: 20
 Server Version: 25.0.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: ae07eda36dd25f8a1b98dfbf587313b99c0190bb
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
  apparmor
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.6.0-1001-aws
 Operating System: Ubuntu 24.04 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 30.57GiB
 Name: dev01.khaus.io
 ID: BMMG:V7I2:TKCO:ITI4:RQQE:W53K:3K25:KHWL:3TYH:RJ45:W6ZC:OPUS
 Debug Mode: false
 Username: brandond
 Experimental: false
 Insecure Registries:
  172.17.0.2:5000
  127.0.0.0/8
 Registry Mirrors:
  http://172.17.0.2:5000/
 Live Restore Enabled: false
```

### Builders list

```text
NAME/NODE      DRIVER/ENDPOINT                   STATUS    BUILDKIT   PLATFORMS
builder*       docker-container
 \_ builder0    \_ unix:///var/run/docker.sock   running   v0.15.0    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/amd64/v4, linux/386
default        docker
 \_ default     \_ default                       running   v0.12.5    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/amd64/v4, linux/386
```

### Configuration

```hcl
group "default" {
  targets = ["stage1", "stage2"]
}

target "stage1" {
  context = null
  dockerfile = "Dockerfile"
  output = ["type=cacheonly"]
}

target "stage2" {
  tags = ["example:latest"]
  output = ["type=docker"]
  dockerfile-inline = "FROM stage1\nCOPY . /source/"
  contexts = {
    "stage1" = "target:stage1"
  }
}
```

### Build logs

_No response_

### Additional info

_No response_

jsternberg:
Setting the context to null in bake will cause the dockerfile to be sent over stdin to the build and will prevent using the default context of "." similar to how this would be used on the command line.

When the JSON is printed from bake, it will reproduce the null so subsequent uses can keep the value.

Fixes #2611.

### DIFF
--- a/bake/bake_test.go
+++ b/bake/bake_test.go
@@ -46,7 +46,7 @@ target "webapp" {
 		require.Equal(t, 1, len(m))
 
 		require.Equal(t, "Dockerfile.webapp", *m["webapp"].Dockerfile)
-		require.Equal(t, ".", *m["webapp"].Context)
+		require.Equal(t, ".", *m["webapp"].ContextPath())
 		require.Equal(t, ptrstr("webDEP"), m["webapp"].Args["VAR_INHERITED"])
 		require.Equal(t, true, *m["webapp"].NoCache)
 		require.Equal(t, "128m", *m["webapp"].ShmSize)
@@ -79,7 +79,7 @@ target "webapp" {
 			require.NoError(t, err)
 
 			require.Equal(t, "Dockerfile.webapp", *m["webapp"].Dockerfile)
-			require.Equal(t, ".", *m["webapp"].Context)
+			require.Equal(t, ".", *m["webapp"].ContextPath())
 
 			_, isSet := m["webapp"].Args["VAR_UNSET"]
 			require.False(t, isSet, m["webapp"].Args["VAR_UNSET"])
@@ -121,7 +121,7 @@ target "webapp" {
 
 		m, g, err := ReadTargets(ctx, []File{fp}, []string{"webapp"}, []string{"webapp.context=foo"}, nil, &EntitlementConf{})
 		require.NoError(t, err)
-		require.Equal(t, "foo", *m["webapp"].Context)
+		require.Equal(t, "foo", *m["webapp"].ContextPath())
 		require.Equal(t, 1, len(g))
 		require.Equal(t, []string{"webapp"}, g["default"].Targets)
 	})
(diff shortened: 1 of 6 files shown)


bake/bake_test.go
```diff
@@ -610,7 +610,7 @@ func TestHCLContextCwdPrefix(t *testing.T) {
 	require.Equal(t, 1, len(m))
 	require.Contains(t, m, "app")
 	assert.Equal(t, "test", *m["app"].Dockerfile)
-	assert.Equal(t, "foo", *m["app"].Context)
+	assert.Equal(t, "cwd://foo", *m["app"].ContextPath())
```
crazy-max, line 613: > `cwd://foo`

Was this a bug or is it a behavior change? I think a `path.Clean(bi.ContextPath)` is missing in this else cond: https://github.com/docker/buildx/pull/2906/files#diff-fbe2fa23203f2ff05f06b6c0731ea90529f90906807483db76a79f6018781465R1422

PR #2920, opened 2025-01-15
dockerfile: update buildkit to 0.19.0-rc2

crazy-max:


.github/workflows/build.yml
```diff
@@ -54,9 +54,9 @@ jobs:
           - master
           - latest
           - buildx-stable-1
+          - v0.19.0-rc2
           - v0.18.2
           - v0.17.2
-          - v0.16.0
```
thaJeztah, line 59: Wondering if at some point we  should make this something  like;

- latest
- latest - 1
- oldest


(so remove versions in between)

thaJeztah (approved):
LGTM

.github/workflows/build.yml
crazy-max: Yes we could look again at this matrix

PR #2922, opened 2025-01-15
Allow default for --output flag.

sudiptob2:
`zstd` compression is quite popular compared to the default `gzip`. But to use `zstd`, we need to pass several parameters. For example:

```bash
docker buildx build --output type=image,oci-mediatypes=true,compression=zstd,push=true,force-compression=true --provenance=false .
```

We can use Bash aliases or scripts to make this easier, but it would be simpler if `buildx` allowed setting defaults for `--output` flag.

A possible implementation has been added to this PR.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -585,6 +585,10 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 	if v := os.Getenv("DOCKER_DEFAULT_PLATFORM"); v != "" {
 		platformsDefault = []string{v}
 	}
+	var outputDefault []string
+	if v := os.Getenv("DOCKER_DEFAULT_OUTPUT"); v != "" {
+		outputDefault = []string{v}
+	}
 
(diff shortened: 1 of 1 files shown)


polarathene:
> ```shell
> --provenance=false
> ```

This is due to wanting to avoid the docker-specific additions to the index manifest right? It should be avoidable with BuildKit's `oci-artifact=true` AFAIK.

That said upstream has also discussed intent to switch to both OCI settings as default in future, which would simplify this a bit further for you.

---

Related [`--compression` option](https://github.com/docker/buildx/issues/3072).

jsternberg:
I don't think this is a good solution for adding default options. I think bake provides a much more compelling way to wrap more complex build outputs with options and is preferred over an environment variable. I'm going to close this for the moment.

It might make sense to have an environment variable or other way to configure default compression options as I can imagine that being something you might want to change locally to potentially speed up build times by lowering the compression. I'm not really open to the idea of having a default output be controlled by an environment variable though.

PR #2931, opened 2025-01-20
history(inspect): print build name

crazy-max:
Prints build name for `history inspect` command so it can relate with the `history ls` one.

### DIFF
--- a/commands/history/inspect.go
+++ b/commands/history/inspect.go
@@ -109,6 +109,8 @@ func runInspect(ctx context.Context, dockerCli command.Cli, opts inspectOptions)
 		delete(attrs, k)
 	}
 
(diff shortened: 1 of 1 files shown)


tonistiigi:
Name is not an actual field of the build. It's a reduction of a collection of fields in the same short string. Inspect should show the original fields with full values.

crazy-max:
> Name is not an actual field of the build. It's a reduction of a collection of fields in the same short string. Inspect should show the original fields with full values.

This is needed for the build summary in GHA: https://github.com/docker/actions-toolkit/blob/e6e18dee2531192e00536d6f21c792e6d01c0b3a/src/github.ts#L272-L299

image

And if we are following this logic then we should do the same for `Build Steps` no?

```
Build Steps:    59/59 (69% cached)
```

would be smth like:

```
Completed Steps:    59
Total Steps:        59
Cached Steps:       35
```

PR #2937, opened 2025-01-21
buildflags: marshal attestations into json with extra attributes correctly

### LINKED ISSUE #2936: Composable attributes for provenance attestation not correctly handled
relates to: https://github.com/docker/actions-toolkit/actions/runs/12888049135/job/35932056137?pr=572#step:2:3060

```hcl
target "default" {
  attest = [
    "type=provenance,mode=max",
    "type=sbom,disabled=true"
  ]
}
```

```
$ docker buildx bake --print
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "attest": [
        {
          "type": "provenance"
        },
        {
          "disabled": true,
          "type": "sbom"
        }
      ],
      "context": ".",
      "dockerfile": "Dockerfile"
    }
  }
}
```

`"mode": "max"` is missing in the output.

Also relates to https://github.com/docker/buildx/pull/2848#issuecomment-2552275394 to add tests.

jsternberg:
`MarshalJSON` would not include the extra attributes because it iterated over the target map rather than the source map.

Also fixes JSON unmarshaling for SSH and secrets. The intention was to
unmarshal into the struct, but `UnmarshalText` takes priority over the
default struct unmarshaling so it didn't work as intended.

Tests have been added for all marshaling and unmarshaling methods.

Fixes #2936.

### DIFF
--- a/util/buildflags/attests.go
+++ b/util/buildflags/attests.go
@@ -91,7 +91,7 @@ func (a *Attest) ToPB() *controllerapi.Attest {
 
 func (a *Attest) MarshalJSON() ([]byte, error) {
 	m := make(map[string]interface{}, len(a.Attrs)+2)
-	for k, v := range m {
+	for k, v := range a.Attrs {
 		m[k] = v
 	}
 	m["type"] = a.Type
--- a/util/buildflags/attests_test.go
+++ b/util/buildflags/attests_test.go
@@ -1,0 +1,79 @@
+package buildflags
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/zclconf/go-cty/cty"
+)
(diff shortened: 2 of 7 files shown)


tonistiigi (approved):
Tests?

jsternberg:
Tests have now been added. I found an issue with the JSON unmarshaling for SSH and Secrets so those have also been fixed.

Not included in this PR is to support the string format for the JSON format. Currently, the JSON format requires the object syntax. It would be possible to allow a mixing like we do for the HCL syntax. If we want that, we should create an issue to track that work and it can be included in another PR.

crazy-max (approved):

tonistiigi (approved):

PR #2957, opened 2025-01-31
don't warn user about missing --allows when running with progress=rawjson

ndeloof:
When ran with `rawjson` progress mode, client only expect output to contain json messages. warning about missing `--allow` should then be skipped.


crazy-max (approved):

tonistiigi (approved):
This should also reconfigure `logrus` logger probably.

PR #2964, opened 2025-02-03
history: inspect json and go template format

crazy-max:


tonistiigi:
https://github.com/docker/buildx/compare/master...tonistiigi:buildx:history-inspect-format?expand=1 is my wip for adding format. Main difference is that there is a new struct for history inspection instead of raw HistoryRecord with other (random) values. That struct can be formatted and provides similar output to the default text (that also now comes from printing the struct).

I think this is preferred. If lgty you may finish that instead.

commands/history/inspect.go
```diff
@@ -92,7 +94,135 @@ func runInspect(ctx context.Context, dockerCli command.Cli, opts inspectOptions)
...
+	}
+
+	workers, err := c.ListWorkers(ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed to list workers")
+	}
+workers0:
+	for _, w := range workers {
+		for _, p := range w.Platforms {
+			out.DefaultPlatform = platforms.FormatAll(platforms.Normalize(p))
```
tonistiigi, line 155: This is not part of the history record. If needed then this can be set into platform key if one is unset.

crazy-max:
> https://github.com/docker/buildx/compare/master...tonistiigi:buildx:history-inspect-format?expand=1 is my wip for adding format. Main difference is that there is a new struct for history inspection instead of raw HistoryRecord with other (random) values. That struct can be formatted and provides similar output to the default text (that also now comes from printing the struct).
> 
> I think this is preferred. If lgty you may finish that instead.

Yes looks good to me, I carry over your changes and keep you posted.

crazy-max:
@tonistiigi Updated on top of your branch.

Steps counts were missing in `out` and also added build name and error logs that are needed for GHA summary. I didn't add error sources but I think we could. Should probably be base64 encoded?

Also missing materials and attachments. I guess we would want this in `out.config`?

commands/history/inspect.go
```diff
@@ -318,44 +572,21 @@ func runInspect(ctx context.Context, dockerCli command.Cli, opts inspectOptions)
...
+			for _, l := range out.Error.Logs {
+				fmt.Fprintln(dockerCli.Out(), "> "+l)
 			}
+			fmt.Fprintln(dockerCli.Out())
 		}
-
 		if debug.IsEnabled() {
-			fmt.Fprintf(dockerCli.Out(), "\n%+v\n", stack.Formatter(retErr))
-		} else if len(stack.Traces(retErr)) > 0 {
+			fmt.Fprintf(dockerCli.Out(), "\n%+v\n", stack.Formatter(out.Error.statusErr))
```
crazy-max, line 588: Could we always have the full stack as well in json output?

tonistiigi: Yes, don't set the raw error in `out.Error.statusErr`. Instead already split it into separate properties like `Stack` .

commands/history/inspect.go
```diff
@@ -146,131 +245,286 @@ func runInspect(ctx context.Context, dockerCli command.Cli, opts inspectOptions)
...
+	out.Config.RestRaw = unusedAttrs
 
-	for _, k := range unusedAttrs {
-		fmt.Fprintf(tw, "%s:\t%s\n", k, attrs[k])
+	if opts.format == formatter.JSONFormatKey {
+		enc := json.NewEncoder(dockerCli.Out())
+		enc.SetIndent("", "  ")
+		return enc.Encode(out)
+	} else if opts.format != formatter.RawFormatKey {
+		return errors.Errorf("unsupported format %q", opts.format)
```
tonistiigi, line 416: Why unsupported? I'd expect any other `format` to be template that is executed with `out` as the value.

crazy-max: Oh I didn't thought about using Go template, I will take a look if we can use the CLI formatter.

commands/history/inspect.go
```diff
@@ -282,81 +437,209 @@ func runInspect(ctx context.Context, dockerCli command.Cli, opts inspectOptions)
...
-		fmt.Fprintln(dockerCli.Out())
 	}
 
-	if rec.ExternalError != nil {
-		dt, err := content.ReadBlob(ctx, store, ociDesc(rec.ExternalError))
+	if opts.format == formatter.JSONFormatKey {
+		enc := json.NewEncoder(dockerCli.Out())
+		enc.SetIndent("", "  ")
+		return enc.Encode(out)
+	} else if opts.format != formatter.RawFormatKey {
```
tonistiigi, line 474: The default should be `PrettyFormatKey` I guess.

tonistiigi:
JSON should be usable as function as well like `{{json .}}`

tonistiigi (approved):

PR #2979, opened 2025-02-11
vendor: update buildkit to 0e3037c0182e

tonistiigi:


vendor/github.com/moby/buildkit/util/testutil/integration/run.go
```diff
@@ -257,7 +258,16 @@ func copyImagesLocal(t *testing.T, host string, images map[string]string) error
 				defer closer()
 			}
 		} else {
-			desc, provider, err = contentutil.ProviderFromRef(from)
+			dockerConfig := config.LoadDefaultConfigFile(os.Stderr)
```
tonistiigi, line 261: @crazy-max This brings in auth support for tests but still need to update CI side to pass the config into container as well.

crazy-max (approved):

PR #2986, opened 2025-02-12
remove import of x/exp

tonistiigi:
This is some kind of `goimports` bug that went unnoticed.

ref https://github.com/compose-spec/compose-go/issues/735


crazy-max (approved):

crazy-max:
Maybe we could have a forbiddigo in golangci lint conf?

PR #2987, opened 2025-02-12
update test buildkit ref to v0.20.0-rc1

tonistiigi:


.github/workflows/build.yml
```diff
@@ -54,6 +54,7 @@ jobs:
           - master
           - latest
           - buildx-stable-1
+          - v0.20.0-rc1
           - v0.19.0
           - v0.18.2
           - v0.17.2
```
crazy-max, line 60: we can remove this one

tonistiigi:
I think we can leave this in draft anyway, just check that the CI has passed and update to the next RC when it is ready. Merge after there is GA.

crazy-max:
Pushed commit to test BuildKit 0.20.0-rc3

crazy-max:
https://github.com/docker/buildx/pull/3013

PR #2988, opened 2025-02-12
docker-container: check error from response body when pulling image

crazy-max:
fixes https://github.com/moby/moby/issues/49453

On `ImageCreate` we should check if the pull response body has an error.


driver/docker-container/driver.go
```diff
@@ -94,14 +96,14 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
...
+		resp, err := d.DockerAPI.ImageCreate(ctx, imageName, image.CreateOptions{
 			RegistryAuth: ra,
 		})
 		if err != nil {
 			return err
 		}
-		_, err = io.Copy(io.Discard, rc)
-		return err
+		defer resp.Close()
+		return jsonmessage.DisplayJSONMessagesToStream(resp, streams.NewOut(io.Discard), nil)
```
vvoland, line 106: We can avoid the cli dependency:

```suggestion
		return jsonmessage.DisplayJSONMessagesStream(resp, io.Discard, 0, false, nil)
```

tonistiigi (approved):

PR #2994, opened 2025-02-14
support for device entitlement in build and bake

tonistiigi:
Buildx side of https://github.com/moby/buildkit/pull/5742

Allow access to CDI Devices in Buildkit v0.20.0+ for devices that are not automatically allowed to be used by everyone in BuildKit configuration.

```
--allow device grants access to any device.
--allow device=kind|name grants access to specific device.
--allow device=kind|name,alias=kind|name allows mapping kind to a specific device or one device to another. Alias is the name requested by the build and device is the actual device that is being enabled.
```

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -593,7 +593,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 
 	flags.StringSliceVar(&options.extraHosts, "add-host", []string{}, `Add a custom host-to-IP mapping (format: "host:ip")`)
 
-	flags.StringSliceVar(&options.allow, "allow", []string{}, `Allow extra privileged entitlement (e.g., "network.host", "security.insecure")`)
+	flags.StringArrayVar(&options.allow, "allow", []string{}, `Allow extra privileged entitlement (e.g., "network.host", "security.insecure")`)
 
 	flags.StringArrayVarP(&options.annotations, "annotation", "", []string{}, "Add annotation to the image")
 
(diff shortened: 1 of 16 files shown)


commands/build.go
```diff
@@ -593,7 +593,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debug.D
 
 	flags.StringSliceVar(&options.extraHosts, "add-host", []string{}, `Add a custom host-to-IP mapping (format: "host:ip")`)
 
-	flags.StringSliceVar(&options.allow, "allow", []string{}, `Allow extra privileged entitlement (e.g., "network.host", "security.insecure")`)
+	flags.StringArrayVar(&options.allow, "allow", []string{}, `Allow extra privileged entitlement (e.g., "network.host", "security.insecure")`)
```
tonistiigi, line 596: This is technically a breaking change but I think `StringSlice` was accidental. Alias syntax can not be supported with the slice.

crazy-max: Yes I think it should have been `StringArray` type in the first place for build. With bake it looks good: https://github.com/docker/buildx/blob/ef73c64d2cb103d5a4ba00a08e0c66a7cb4f33dc/commands/bake.go#L470

Would need to adapt build push action to ignore comma:
* https://github.com/docker/build-push-action/blob/ca877d9245402d1537745e0e356eab47c3520991/README.md?plain=1#L215
* https://github.com/docker/build-push-action/blob/ca877d9245402d1537745e0e356eab47c3520991/src/context.ts#L49

Looking at https://grep.app/search?regexp=true&q=build.*--allow.*%28security%5C.insecure%7Cnetwork%5C.host%29 it seems people don't use csv values.

There are some in GHA workflows https://grep.app/search?f.path=.github%2Fworkflows%2F&regexp=true&q=allow%3A+.*%28security%5C.insecure%7Cnetwork%5C.host%29 but we can manage this in our action.

crazy-max (approved):

Issue #2989, opened 2025-02-12
bake: overriding annotations


jonapich:
### Description

The [official documentation](https://docs.docker.com/build/bake/overrides/) states that `annotations` cannot be overridden.

However, [the code](https://github.com/docker/buildx/blob/1dc5f0751bfdf504eb6c393a031b33f968d0d61c/bake/bake.go#L701) claims that only "inherits" cannot be overridden.

In practice, using `--set` with annotations doesn't work:

```
$ docker buildx bake --print
{
  "target": {
    "default": {
      "annotations": [
        "index-descriptor,manifest-descriptor,index,manifest:com.myorg.ci.tags.stable=stable"
      ],
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "tag1"
      ]
    }
  }
}

$ docker buildx bake --print \
  --set *.tags='foo' \
  --set *.annotations='index-descriptor,manifest-descriptor,index,manifest:com.myorg.ci.tags.stable=bar'
{
  "target": {
    "default": {
      "annotations": [
        "index-descriptor,manifest-descriptor,index,manifest:com.myorg.ci.tags.stable=stable"
      ],
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "foo"
      ]
    }
  }
}
```

But it actually works if it's included in a `docker-bake.override.hcl` file:
```
$ cat docker-bake.override.hcl
target default {
  annotations = [
    annotate("tags.stable", "bar"),
  ]
  tags = ["foo"]
}

function annotate {
  params = [key, value]
  result = "index-descriptor,manifest-descriptor,index,manifest:com.myorg.ci.${key}=${value}"
}

$ docker buildx bake --print
{
  "target": {
    "default": {
      "annotations": [
        "index-descriptor,manifest-descriptor,index,manifest:com.myorg.ci.tags.stable=stable",
        "index-descriptor,manifest-descriptor,index,manifest:com.myorg.ci.tags.stable=bar"
      ],
      "context": ".",
      "dockerfile": "Dockerfile",
      "tags": [
        "foo"
      ]
    }
  }
}
```

_In this case, the annotations are merged together._

**Is this intended, or just an oversight?**

---
## Use case

I am writing an automated release pipeline for general CI/CD use, and I'd like to store metadata on the image as a way to carry information around:

1. The user merges to main and triggers a GitHub workflow in their repository.
2. The workflow calls the automated release pipeline workflow (what I'm working on)
[...]

PR #2997, opened 2025-02-14
bake: allow annotations to be set on the command line

jsternberg:
Annotations were not merged correctly. The overrides in `ArrValue` would be merged, but the section of code setting them from the command line did not include `annotations` in the list of available attributes so the command line option was completely discarded.

Fixes https://github.com/docker/buildx/issues/2989.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -556,9 +556,9 @@ func (c Config) newOverrides(v []string) (map[string]map[string]Override, error)
 			o := t[kk[1]]
 
 			// IMPORTANT: if you add more fields here, do not forget to update
-			// docs/bake-reference.md and https://docs.docker.com/build/bake/overrides/
+			// docs/reference/buildx_bake.md (--set) and https://docs.docker.com/build/bake/overrides/
 			switch keys[1] {
-			case "output", "cache-to", "cache-from", "tags", "platform", "secrets", "ssh", "attest", "entitlements", "network":
+			case "output", "cache-to", "cache-from", "tags", "platform", "secrets", "ssh", "attest", "entitlements", "network", "annotations":
 				if len(parts) == 2 {
 					o.ArrValue = append(o.ArrValue, parts[1])
 				}
--- a/bake/bake_test.go
+++ b/bake/bake_test.go
@@ -34,6 +34,9 @@ target "webapp" {
 	args = {
 		VAR_BOTH = "webapp"
 	}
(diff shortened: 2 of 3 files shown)


bake/bake.go
```diff
@@ -558,7 +558,7 @@ func (c Config) newOverrides(v []string) (map[string]map[string]Override, error)
 			// IMPORTANT: if you add more fields here, do not forget to update
 			// docs/bake-reference.md and https://docs.docker.com/build/bake/overrides/
 			switch keys[1] {
-			case "output", "cache-to", "cache-from", "tags", "platform", "secrets", "ssh", "attest", "entitlements", "network":
+			case "output", "cache-to", "cache-from", "tags", "platform", "secrets", "ssh", "attest", "entitlements", "network", "annotations":
```
crazy-max, line 561: Should we do the same for `labels`?

bake/bake.go
```diff
@@ -558,7 +558,7 @@ func (c Config) newOverrides(v []string) (map[string]map[string]Override, error)
 			// IMPORTANT: if you add more fields here, do not forget to update
 			// docs/bake-reference.md and https://docs.docker.com/build/bake/overrides/
```
crazy-max, line 559: not related but comment is not quite correct:

```suggestion
			// docs/reference/buildx_bake.md (--set) and https://docs.docker.com/build/bake/overrides/
```

jsternberg: Ah I was looking for where to change this.

crazy-max (approved):

bake/bake.go
jsternberg: For what it's worth, I don't think labels has the same issue. It uses `Value` from the overrides: https://github.com/docker/buildx/blob/52f503e8068d4a09e7c1b17e4b88ae75b1f64651/bake/bake.go#L891-L898

Compare to `annotations`: https://github.com/docker/buildx/blob/52f503e8068d4a09e7c1b17e4b88ae75b1f64651/bake/bake.go#L973-L974

So this might already work for `labels`? It looks like it's documented in there to work.

jsternberg: Just to clarify, this `switch` is a list of "use `ArrValue` to fill in the overrides" and the default is to use `Value` if you're not in the list.

crazy-max: Ah yes indeed, thanks

PR #2983, opened 2025-02-11
vendor: update buildkit to c5d687177

tonistiigi:

PR #3001, opened 2025-02-17
cache: enable gha cache backend if cache service v2 detected

crazy-max:
After testing latest RC, gha cache backend is always disabled when cache service v2 is detected. We forgot to check if `url_v2` is set in https://github.com/docker/buildx/pull/2983.

Also fixes gha cache url handling to always set both `url` and `url_v2` if possible to match fallback of github actions tookit.

### DIFF
--- a/util/buildflags/cache.go
+++ b/util/buildflags/cache.go
@@ -207,6 +207,7 @@ func addGithubToken(ci *controllerapi.CacheOptionsEntry) {
 	}
 	version, ok := ci.Attrs["version"]
 	if !ok {
+		// https://github.com/actions/toolkit/blob/2b08dc18f261b9fdd978b70279b85cbef81af8bc/packages/cache/src/internal/config.ts#L19
 		if v, ok := os.LookupEnv("ACTIONS_CACHE_SERVICE_V2"); ok {
 			if b, err := strconv.ParseBool(v); err == nil && b {
 				version = "2"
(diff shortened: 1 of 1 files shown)


crazy-max:
Hum it fails to solve in our e2e test with: https://github.com/docker/buildx/actions/runs/13375533199/job/37353936009?pr=3001#step:9:388

```
#6 [xx internal] load metadata for docker.io/library/alpine:latest
#6 DONE 0.7s
ERROR: target ld64-static-tgz: failed to solve: failed to configure gha cache exporter: url not set for github actions cache
```

I think that's because we only set `url_v2` but the gha cache exporter requires url to be set: https://github.com/moby/buildkit/blob/de39e457a2c91179913c2eb303543f240ca793b3/cache/remotecache/gha/gha.go#L66-L69

So we need a fix on BuildKit as well.

crazy-max:
> Hum it fails to solve in our e2e test with: https://github.com/docker/buildx/actions/runs/13375533199/job/37353936009?pr=3001#step:9:388

In this case it fails because we are pushing an empty `url` attribute against previous BuildKit version (0.19.0) that doesn't support `url_v2`. In such case we should always set the `url` attribute. Looking at github toolkit source code https://github.com/actions/toolkit/blob/2b08dc18f261b9fdd978b70279b85cbef81af8bc/packages/cache/src/internal/config.ts#L28-L33 they set either `ACTIONS_CACHE_URL` or `ACTIONS_RESULTS_URL` if found. Although I doubt  `ACTIONS_RESULTS_URL` would work with v1 :thinking:

crazy-max:
> I think we should do the same in buildctl https://github.com/moby/buildkit/blob/de39e457a2c91179913c2eb303543f240ca793b3/cmd/buildctl/build/util.go#L28-L41

Done in https://github.com/moby/buildkit/pull/5754

tonistiigi (approved):

PR #3014, opened 2025-02-20
Dockerfile: update to docker v28.0.0

crazy-max:
closes #3009

Also adds alt for previous Docker 27 release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,6 +76,16 @@ jobs:
           - worker: docker+containerd  # same as docker, but with containerd snapshotter
             pkg: ./tests
             mode: experimental
+          - worker: "docker@27.5"
+            pkg: ./tests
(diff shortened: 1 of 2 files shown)


thaJeztah:
> Also adds alt for previous Docker 27 release.

Ah, thanks! I was looking at that a few times, and I gathered we wanted to keep v26.1 in there at least because that's still the one used in GitHub actions, but adding an extra one probably makes sense.

Would it make sense to handle this through a table in GitHub actions (instead of continuing to add more to the Dockefile?)

thaJeztah (approved):
LGTM

crazy-max:
> I gathered we wanted to keep v26.1 in there at least because that's still the one used in GitHub actions, but adding an extra one probably makes sense.

Yes I kept 26 specially for GHA :+1:

PR #3025, opened 2025-02-25
bake: deprecate x-bake compose extension

crazy-max:
https://github.com/docker/buildx/pull/721 was introduced to allow the use of additional fields that were not supported by the official specification at the time. Now that the [Compose specification for build](https://docs.docker.com/reference/compose-file/build/) is ahead, we can deprecate this extension and eventually remove it in a future release.

Note that `no-cache-filter` and `output` are not yet supported by the official specification.

### DIFF
--- a/bake/compose.go
+++ b/bake/compose.go
@@ -16,6 +16,7 @@ import (
 	dockeropts "github.com/docker/cli/opts"
 	"github.com/docker/go-units"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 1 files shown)


glours (approved):
👍

jsternberg (approved):

tonistiigi:
I don't get this. Is this expecting that `bake` will never add new fields in the definition. Plus, there are many fields in bake definition not supported in compose build block.

crazy-max:
> Is this expecting that `bake` will never add new fields in the definition? Plus, there are many fields in bake definition not supported in compose build block.

No it just means that the official compose specification needs to add new fields to build block. `x-bake` was meaningful when there was no parity between bake fields and the official spec at the time.

Only missing fields from the official spec that are supported today in `x-bake` extension are `no-cache-filter` and `output`. We can postpone deprecation until these fields made into the official spec.

crazy-max:
Put this one in draft until missing fields are supported by the compose spec.

thompson-shaun:
We'll need to replace this PR with another that uses the warning path

PR #3027, opened 2025-02-27
Fix attest extra arguments

LaurentGoderre:
The new parsing of attest argument broke support for extra args for the attest command in buildkit as shown with these tests:

https://github.com/moby/buildkit/blob/master/frontend/attestations/parse_test.go#L31-L58

Note that the case should preserved for these args as well as they are set as env var in the scanner.


crazy-max:
> attest argument broke support for extra args for the attest command in buildkit as shown with these tests:
> 
> https://github.com/moby/buildkit/blob/master/frontend/attestations/parse_test.go#L31-L58

Hum indeed, I forgot the extra arguments support which has been introduced in https://github.com/moby/buildkit/pull/5372. I think we also need docs changes:
* https://github.com/moby/buildkit/blob/master/docs/attestations/sbom.md
* https://docs.docker.com/build/metadata/attestations/sbom/#sbom-generator
* https://docs.docker.com/scout/how-tos/view-create-sboms/#attest

util/buildflags/export.go
```diff
@@ -259,9 +260,16 @@ func (w *csvBuilder) Write(key, value string) {
 	if w.sb.Len() > 0 {
 		w.sb.WriteByte(',')
 	}
-	w.sb.WriteString(key)
-	w.sb.WriteByte('=')
-	w.sb.WriteString(value)
+
+	var attr strings.Builder
```
tonistiigi, line 264: Why the extra `strings.Builder` in here? What is the effect of `TrimSpace` in 272?

util/buildflags/export.go
```diff
@@ -259,9 +260,16 @@ func (w *csvBuilder) Write(key, value string) {
...
+	writer := csv.NewWriter(&attr)
```
tonistiigi, line 265: Btw, every call to `csv.NewWriter` does a 4KB memory allocation. So maybe only call it if there is a comma in value?

util/buildflags/attests.go
```diff
@@ -159,6 +159,8 @@ func (a *Attest) UnmarshalText(text []byte) error {
 				return errors.Wrapf(err, "invalid value %s", field)
 			}
 			a.Disabled = disabled
+		case "generator":
```
tonistiigi, line 162: Why is there special handling for "generator" key in here?

util/buildflags/export.go
LaurentGoderre: The csv writer creates the proper encoding that handles the commas and the quotes but adds an extra new line at the end so it needs to be trimmed

util/buildflags/export.go
LaurentGoderre: A comma or a quote maybe?

util/buildflags/attests.go
LaurentGoderre: I believe that key is sensitive in BuildKit

https://github.com/moby/buildkit/blob/824bc35b9771357e548ca913f8790854b70cf1f5/control/control.go#L479

LaurentGoderre: Yeah, if I modify these tests here:

https://github.com/moby/buildkit/blob/824bc35b9771357e548ca913f8790854b70cf1f5/frontend/attestations/parse_test.go#L19

and add 

```go
{
	name: "test",
	values: map[string]string{
		"attest:sbom": "GENERATOR=docker.io/foo/bar",
	},
	expected: map[string]map[string]string{
		"sbom": {
			"GENERATOR": "docker.io/foo/bar",
		},
	},
},
```

I get

```
 Error:          Not equal:
                                expected: map[string]map[string]string{"sbom":map[string]string{"GENERATOR":"docker.io/foo/bar"}}
                                actual  : map[string]map[string]string{"sbom":map[string]string{"GENERATOR":"docker.io/foo/bar", "generator":"docker/buildkit-syft-scanner:stable-1"}}
```

LaurentGoderre: Yeah, you're right. It's not a big impact

util/buildflags/export.go
LaurentGoderre: Done!

util/buildflags/export.go
```diff
@@ -259,9 +260,17 @@ func (w *csvBuilder) Write(key, value string) {
...
-	w.sb.WriteByte('=')
-	w.sb.WriteString(value)
+
+	pair := key + "=" + value
+	if strings.ContainsRune(pair, ',') || strings.ContainsRune(pair, '"') {
+		var attr strings.Builder
+		writer := csv.NewWriter(&attr)
+		writer.Write([]string{pair})
+		writer.Flush()
+		pair = strings.TrimSpace(attr.String())
```
tonistiigi, line 270: You may add a comment in here that csv writer adds a newline. Not entirely obvious.

LaurentGoderre: Done!

tonistiigi (approved):

crazy-max (approved):

PR #3028, opened 2025-02-27
buildflags: do not incorrectly lowercase the attestation attributes

jsternberg:


cdupuis:
Is this the same as #3027?

tonistiigi:
@jsternberg We should get the test updates from #3027 (looks like they need some fixes though)

LaurentGoderre:
I think #3027 is more complete because it's not just the case but the handling of comma in fields.

jsternberg:
Took the tests from #3027 and also fixed the reproduction of the CSV fields with handling commas. Thanks for the callout @LaurentGoderre.

The main difference between the two was the behavior of lowering the attribute and adding `generator` as one of the keys for the switch statement. There was a mistake in translating the previous code and we never should have been lowercasing unknown keys.

util/buildflags/export.go
```diff
@@ -259,9 +259,17 @@ func (w *csvBuilder) Write(key, value string) {
 	if w.sb.Len() > 0 {
 		w.sb.WriteByte(',')
 	}
+
+	needsQuotes := strings.ContainsRune(value, ',') || strings.ContainsRune(key, ',')
```
LaurentGoderre, line 263: What happens if there is comma and quotes in the key or value?

LaurentGoderre:
See https://github.com/docker/buildx/pull/3027/files/9b2e0c78dcd94c482d80e5cbac1adf0353f2f732#diff-3fbb3f4c8ec197482c149eb07b7b19e2cc7f4facca460a6a02e781caf5fac089

The handling of the generator key might also be needed to be safe

util/buildflags/export.go
LaurentGoderre: nvm

jsternberg:
We'll just go with #3027 when it's ready to merge.

Issue #901, opened 2022-01-04
bake --set does not allow multiple tags, cache-to or cache-from


MaximeWeyl:
builx build can take several tags, cache-to or cache-from arguments.
docker-bake.hcl allows this too

but it does not seam possible to override them properly with buildx bake --set which is intended for single arguments, from my understanding. It would be nice to be able to override only one given arg, delete an arg, or append an arg.

tonistiigi:
You can set multiple values by multiple set I believe. But yes, there is no append vs replace semantics. There is a proposal to add `+=` separator.

Issue #631, opened 2021-06-13
How can I append tags with --set ?


back-2-95:
I have target like:

```
target "fpm-73" {
  ...
  tags = ["repo/php:7.3-fpm", "repo/php:7.3-fpm-latest"]
}
```

I'm getting Minor version outside of `docker-bake.hcl`, e.g. `7.3.27` so I'd like to append a tag to the target tags:

```
docker buildx bake --print --pull --push --set fpm-73.tags=repo/php:7.3.27
```

Now with that ^^ the tags list from hcl file gets overriden totally.

back-2-95:
Got around this with:

```
PHP_MINOR_VERSION=7.3.27 docker buildx bake --print --pull --push
```

And in `docker-bake.hcl`:

```
variable PHP_MINOR_VERSION {}

target "fpm-73" {
  ...
  tags = ["repo/php:7.3-fpm", "repo/php:${PHP_MINOR_VERSION}-fpm-latest", "repo/php:7.3-fpm-latest"]
}
```

tonistiigi:
I don't think this is currently possible. We've discussed things like `--set fpm-73.tags+=repo/php:7.3.27`

errordeveloper:
> We've discussed things like `--set fpm-73.tags+=repo/php:7.3.27`

Personally I find mini-languages in flags a bit over the top, `--set` flag is already complex enough. Using HCL variables seems like a viable solution.

back-2-95:
Yeah, I have been quite happy with HCL variable thing there ^^
It also reminds passing args to Dockerfile.

thompson-shaun:
Related to https://github.com/docker/buildx/issues/901?

PR #3031, opened 2025-02-28
bake: support += operator to append with overrides

crazy-max:
* fixes https://github.com/docker/buildx/issues/631
* fixes https://github.com/docker/buildx/issues/901

Adds support for `+=` operator to append through `--set` flag for supported array value typed fields.

```hcl
target "default" {
  platforms = ["linux/amd64"]
}
```

```
$ docker buildx bake --set *.platform+=linux/arm64 --print
#1 reading docker-bake.hcl 51B / 51B done
#1 DONE 0.0s
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "platforms": [
        "linux/amd64",
        "linux/arm64"
      ]
    }
  }
}
```

Note that it already appends for `entitlements`, `annotations` and `attest` fields. This has been documented to avoid confusion but wonder if we should make them consistent as follow-up. Would be a breaking change though so maybe first we should warn about this case?


bake/bake_test.go
```diff
@@ -148,6 +173,49 @@ target "webapp" {
...
+		m, _, err := ReadTargets(ctx, []File{fp}, []string{"webapp"}, []string{"webapp.platform+=linux/arm64", "webapp.platform+=linux/riscv64"}, nil, &EntitlementConf{})
+		require.NoError(t, err)
+		require.Equal(t, []string{"linux/amd64", "linux/arm64", "linux/riscv64"}, m["webapp"].Platforms)
+	})
+
+	t.Run("PlatformAppendMultiLastOverride", func(t *testing.T) {
+		m, _, err := ReadTargets(ctx, []File{fp}, []string{"webapp"}, []string{"webapp.platform+=linux/arm64", "webapp.platform=linux/riscv64"}, nil, &EntitlementConf{})
+		require.NoError(t, err)
+		require.Equal(t, []string{"linux/arm64", "linux/riscv64"}, m["webapp"].Platforms)
+	})
```
crazy-max, line 198: This case is edgy but don't think current implementation is right. I think if last override does not append then it should only take this value and not append with previous one:

```hcl
target "default" {
  platforms = ["linux/amd64"]
}
```

```
$ docker buildx bake --set *.platform+=linux/arm64 --set *.platform=linux/riscv64 --print
#1 reading docker-bake.hcl 51B / 51B done
#1 DONE 0.0s
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "platforms": [
        "linux/arm64",
        "linux/riscv64"
      ]
    }
  }
}
```

Should be instead:

```
$ docker buildx bake --set *.platform+=linux/arm64 --set *.platform=linux/riscv64 --print
#1 reading docker-bake.hcl 51B / 51B done
#1 DONE 0.0s
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "platforms": [
        "linux/riscv64"
      ]
    }
  }
}
```

Maybe that's fine for this PR to start with but I think we should refactor the override logic for such case.

bake/bake_test.go
```diff
@@ -148,6 +173,49 @@ target "webapp" {
...
+
+	t.Run("SecretsAppend", func(t *testing.T) {
+		t.Setenv("FOO", "foo")
+		t.Setenv("BAR", "bar")
+		m, _, err := ReadTargets(ctx, []File{fp}, []string{"webapp"}, []string{"webapp.secrets+=id=BAR,env=BAR"}, nil, &EntitlementConf{})
+		require.NoError(t, err)
+		require.Len(t, m["webapp"].Secrets, 2)
+		require.Equal(t, "FOO", m["webapp"].Secrets[0].ID)
+		require.Equal(t, "BAR", m["webapp"].Secrets[1].ID)
+	})
```
crazy-max, line 217: Not for this PR but if we append a secret with the same ID it doesn't seem to override the existing one from the bake definition but append instead:

```hcl
target "default" {
  secret = [
    "id=FOO,env=FOO"
  ]
}
```

```
$ FOO=foo BAR=bar docker buildx bake -f secret.hcl --set *.secrets+=id=FOO,env=BAR --print
#1 reading secret.hcl 59B / 59B done
#1 DONE 0.0s
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "secret": [
        {
          "id": "FOO",
          "env": "FOO"
        },
        {
          "id": "FOO",
          "env": "BAR"
        }
      ]
    }
  }
}
```

Building with this Dockerfile:

```dockerfile
FROM busybox
RUN --mount=type=secret,id=FOO cat /run/secrets/FOO
```

We have:

```

#1 reading secret.hcl 59B / 59B done
#1 DONE 0.0s

#2 transferring dockerfile: 102B done
#2 DONE 0.1s

#3 [internal] load metadata for docker.io/library/busybox:latest
#3 DONE 0.0s

#4 [internal] load .dockerignore
#4 transferring context:
#4 transferring context: 2B done
#4 DONE 0.1s

#5 [stage-0 1/2] FROM docker.io/library/busybox:latest
#5 DONE 0.0s

#6 [stage-0 2/2] RUN --mount=type=secret,id=FOO cat /run/secrets/FOO
#6 0.335 bar
#6 DONE 0.4s

#7 exporting to image
#7 exporting layers 0.0s done
#7 writing image sha256:0d8fffa9b8b5bee5151adc9fd83c9477653621c751f143a5fe6e44600bbbed52 done
#7 DONE 0.1s
```

Which is ok as it takes the last one defined but we should dedup early.

crazy-max: Opened #3032 and maybe we should do the same for `ssh`?

bake/bake_test.go
tonistiigi: is there ever a case where `=` and `+=` for same override key make sense? It would not be append anyway. Even if you follow the order of arguments, then one way it is a replace, and the other way some overrides get ignored.

bake/bake.go
```diff
@@ -534,7 +535,8 @@ func (c Config) newOverrides(v []string) (map[string]map[string]Override, error)
 		}
 
 		pattern := keys[0]
-		if len(parts) != 2 && keys[1] != "args" {
+		attrName := strings.TrimSuffix(keys[1], "+")
+		if len(parts) != 2 && attrName != "args" {
```
tonistiigi, line 539: out of scope: why isn't the `labels` behaving the same

bake/bake.go
```diff
@@ -534,7 +535,8 @@ func (c Config) newOverrides(v []string) (map[string]map[string]Override, error)
 		}
 
 		pattern := keys[0]
-		if len(parts) != 2 && keys[1] != "args" {
+		attrName := strings.TrimSuffix(keys[1], "+")
```
tonistiigi, line 538: this doesn't look very logical to do multiple `TrimSuffix`. Instead there should be one `TrimSuffix` for `parts[0]`

bake/bake.go
crazy-max: I guess `annotations` and `labels` don't behave like `args` because key can contain invalid chars when used with override like `.` Reminds me of this conversation with @jsternberg https://github.com/docker/buildx/pull/2997#pullrequestreview-2618430667

tonistiigi: Why would invalid characters matter, the "parser" uses `splitn(,3)`. Annotations would be different as that is an array, not a map.

tonistiigi (approved):

Issue #2711, opened 2024-09-25
Proposal: buildx history command


tonistiigi:
### Goals
- Provide access to data stored by BuildKit via history record (provenance, logs, traces etc.) to `docker buildx` users. History APIs were added to BuildKit in version v0.11 
- Allow buildx users to better understand the internals of their builds
- Allow replaying the logs without running the whole build again with `--progress=plain` (in default TTY mode logs disappear once a build step completes successfully). The same can be used to print logs without the multiplexing switches that happen when build steps run in parallel.

### Build summary (could be separate PR)

Currently, when running buildx build or buildx bake in a machine where Docker Desktop is installed, “View build details: docker-desktop://dashboard/build/…” will be printed after build is complete allowing access to [Docker Desktop Build Views](https://docs.docker.com/desktop/use-desktop/builds/). This should be replaced with more generic “summary” feature.

Summary should contain:
- Duration of the build (especially important for --progress=plain that does not log this).
- Link to open build in Docker Desktop (if it is installed)
- Exported outputs. Eg. image names that were exported to docker or pushed to registry. Paths for local exports. Optionally in debug mode could even show example commands for running the image or inspecting with with buildx imagetools if it was pushed. If possible image can also be opened in Docker Desktop.
- Example command to replay the logs. docker buildx history logs <ref>
- Probably some summary of an error (eg. build step name that failed), but no need to duplicate whole error information again that is already printed.

```
Build completed: 21.0s (53 steps, 34 cached)
View build details: docker-desktop://dashboard/build/..
Outputs:
	Image    foobar                   (Exported to Docker)
	Image    docker.io/foo/bar:v2     (Pushed)
Replay logs: docker buildx history logs abc123def567
```

### docker buildx history ls

Lists all the builds BuildKit history API knows about. Most recent builds are shown on top.

```
BUILD ID          NAME             STATUS          CREATED      DURATION           
abcd              binaries         Completed       5 min ago    1m03s        Open
abcef             binaries         Error           2 hours ago  3s           Open
```

Open link can be used to open build in Docker Desktop if supported.

Note that this command shows only the build for the active builder in Buildx, not like the [build UI in Docker Desktop](https://docs.docker.com/desktop/use-desktop/builds/), that shows builds across all your builders.

### docker buildx history logs [\<id>]

Replay the logs for a specified build.

The logs would appear with the same formatting as `--progress=plain` , but there is no need to ever do multiplexing of build steps. Every step will only appear once in the output and print its whole log.

[...]

PR #3039, opened 2025-03-04
history: add history import command

tonistiigi:
part of #2711


commands/history/import.go
```diff
@@ -0,0 +1,109 @@
...
+	return err
+}
+
+func importCmd(dockerCli command.Cli, _ RootOptions) *cobra.Command {
+	var options importOptions
+
+	cmd := &cobra.Command{
+		Use:   "import [OPTIONS] < bundle.dockerbuild",
+		Short: "Import a build into Docker Desktop",
+		Args:  cobra.NoArgs,
```
crazy-max, line 98: I think we should use args as file input instead of a flag and require at least one arg. Multiple files should be supported like Docker Desktop does atm imo.

commands/history/import.go
```diff
@@ -0,0 +1,109 @@
...
+	if len(refs) == 0 {
+		return errors.New("no build records found in the bundle")
+	}
+
+	for i, ref := range refs {
+		url := desktop.BuildURL(fmt.Sprintf(".imported/_/%s", ref))
+		fmt.Fprintln(dockerCli.Err(), url)
+		if i == 0 {
+			err = browser.OpenURL(url)
+		}
```
crazy-max, line 86: If multiple files are provided I think we should skip this.

commands/history/import.go
tonistiigi: That's not usually the convention I think when the file is optional and defaults to stdin. If file is arg then for stdin, the user should write `import -`

commands/history/import.go
tonistiigi: Why?

crazy-max: Because who knows what user wants to see? Is it the last one or the first one? But I guess that's fine to open the very first imported record in Docker Desktop.

crazy-max: Although I wonder if there are more than one record imported we could instead open the Builds view and filter by imported builds :thinking:

util/desktop/paths_linux.go
```diff
@@ -0,0 +1,25 @@
...
+const (
+	socketName    = "docker-desktop-build.sock"
+	socketPath    = ".docker/desktop"
+	wslSocketPath = "/mnt/wsl/docker-desktop/shared-sockets/host-services"
+)
+
+func BuildServerAddr() (string, error) {
+	if os.Getenv("WSL_DISTRO_NAME") != "" {
+		return "unix://" + filepath.Join(wslSocketPath, socketName), nil
+	}
```
crazy-max, line 19: Pushed this change for testing WSL support but doesn't work. I'm taking a look if this can be solved but in the meantime we could disable support for `import` command on WSL and explain that it should be invoked on Windows host instead while this is being investigated.

crazy-max (approved):

Issue #2730, opened 2024-10-10
`docker inspect --bootstrap` ignores an error and exits successfully when one of two nodes fails 


moleus:
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

### Description

In our CI we setup buildx builder with 2 nodes - arm64 and amd64 to build multi-arch docker images. CI runs `docker inspect --bootstrap` to check that both of them are up and ready. When one node fails (because of a timeout) the inspect command ignores it, prints the summary and completes with `exit code 0`.

### Expected behaviour

`docker inspect --bootstrap` should exit with non-zero exit-code when one of the nodes fails to start. When command fails it's possible to catch a non-zero status code and retry bootstraping the nodes. 

### Actual behaviour

When single node fails the `driver.Boot` returns an error which is stored in the channel `errCh`. `printer.Wait()` never returns an error. So `err` is `nil` and `len(errCh) == 1`. So the last `return` statement is executed with `true, nil`. For more details see my gist with minimal reproducible example. 
https://github.com/docker/buildx/blob/65c475647313b3edc502dcdd48b8ce4128b60dee/builder/builder.go#L178-L206

`printer.Wait()`:
https://github.com/docker/buildx/blob/65c475647313b3edc502dcdd48b8ce4128b60dee/util/progress/printer.go#L39-L45

`pw.err` is assigned here
https://github.com/docker/buildx/blob/65c475647313b3edc502dcdd48b8ce4128b60dee/util/progress/printer.go#L133

but `UpdateFrom` never sets the error, except the `ctx.Done` case.
https://github.com/docker/buildx/blob/171fcbeb69d67c90ba7f44f41a9e418f6a6ec1da/vendor/github.com/moby/buildkit/util/progress/progressui/display.go#L80-L80

### Buildx version

v0.14.0

### Builders list

-

### Configuration

Here is the link to this gist with code to reproduce bug in the `(b *Builder) Boot()` function:
https://gist.github.com/moleus/6bf1ad37f47dd86ed12f4c4345601e22

### Build logs

```
+ docker buildx inspect --bootstrap
#1 [arm64-demo-jenkins-job-name internal] booting buildkit
#1 waiting for 1 pods to be ready
#1 ...
[...]

PR #3062, opened 2025-03-12
builder: return error if a node fails to boot

crazy-max:
fixes https://github.com/docker/buildx/issues/2730

With https://github.com/docker/buildx/pull/1869 we return an error only if all nodes fail to boot but that's not a correct pattern and we should return an error if any node fails.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -200,7 +200,7 @@ func (b *Builder) Boot(ctx context.Context) (bool, error) {
 		err = err1
 	}
 
-	if err == nil && len(errCh) == len(toBoot) {
+	if err == nil && len(errCh) > 0 {
 		return false, <-errCh
 	}
 	return true, err


tonistiigi (commented):
I think this was intentional to make it so that if you have a multi-node builder and one of the nodes (temporarily) goes down then you can still build with rest of the builders. Eg. you could even have multiple nodes with the same arch for some redundancy.

I think it would be ok if `inspect --bootstrap` fails but I think then it would need some more specific condition as this `Boot()` function is also used in other cases.

crazy-max:
> I think this was intentional to make it so that if you have a multi-node builder and one of the nodes (temporarily) goes down then you can still build with rest of the builders. Eg. you could even have multiple nodes with the same arch for some redundancy.

Ah yes that makes sense

> I think it would be ok if `inspect --bootstrap` fails but I think then it would need some more specific condition as this `Boot()` function is also used in other cases.

Looks good I'm taking a look

crazy-max:
> > I think it would be ok if `inspect --bootstrap` fails but I think then it would need some more specific condition as this `Boot()` function is also used in other cases.
> 
> Looks good I'm taking a look

[`Boot` func in builder package](https://github.com/docker/buildx/blob/0d708c0bc2422efb7bad7026c7dbfcb84ded51fe/builder/builder.go#L160) is only used by `create` and `inspect` command. On build we directly call `driver.Boot` https://github.com/docker/buildx/blob/0d708c0bc2422efb7bad7026c7dbfcb84ded51fe/build/driver.go#L267

So I don't think we need anything special here.

tonistiigi (approved):

PR #3073, opened 2025-03-19
history: add export command

tonistiigi:
part of https://github.com/docker/buildx/issues/2711

Allow builds to be exported into .dockerbuild bundles that can be shared and imported into Docker Desktop.

### DIFF
--- a/commands/history/export.go
+++ b/commands/history/export.go
@@ -1,0 +1,137 @@
+package history
+
+import (
+	"context"
+	"io"
+	"os"
+	"slices"
+
+	"github.com/containerd/console"
+	"github.com/containerd/platforms"
+	"github.com/docker/buildx/builder"
+	"github.com/docker/buildx/localstate"
+	"github.com/docker/buildx/util/cobrautil/completion"
+	"github.com/docker/buildx/util/confutil"
+	"github.com/docker/buildx/util/desktop/bundle"
+	"github.com/docker/cli/cli/command"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+type exportOptions struct {
+	builder string
+	ref     string
+	output  string
+}
+
+func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) error {
+	b, err := builder.New(dockerCli, builder.WithName(opts.builder))
+	if err != nil {
+		return err
+	}
+
+	nodes, err := b.LoadNodes(ctx, builder.WithData())
+	if err != nil {
+		return err
+	}
+	for _, node := range nodes {
+		if node.Err != nil {
+			return node.Err
+		}
+	}
+
+	recs, err := queryRecords(ctx, opts.ref, nodes, &queryOptions{
+		CompletedOnly: true,
+	})
+	if err != nil {
+		return err
+	}
+
+	if len(recs) == 0 {
+		if opts.ref == "" {
+			return errors.New("no records found")
+		}
+		return errors.Errorf("no record found for ref %q", opts.ref)
+	}
+
+	if opts.ref == "" {
+		slices.SortFunc(recs, func(a, b historyRecord) int {
+			return b.CreatedAt.AsTime().Compare(a.CreatedAt.AsTime())
+		})
+	}
+
+	recs = recs[:1]
+
+	ls, err := localstate.New(confutil.NewConfig(dockerCli))
+	if err != nil {
+		return err
+	}
+
+	c, err := recs[0].node.Driver.Client(ctx)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


commands/history/export.go
```diff
@@ -0,0 +1,137 @@
...
+	return bundle.Export(ctx, c, w, toExport)
+}
+
+func exportCmd(dockerCli command.Cli, rootOpts RootOptions) *cobra.Command {
+	var options exportOptions
+
+	cmd := &cobra.Command{
+		Use:   "export [OPTIONS] [REF]",
+		Short: "Export a build into Docker Desktop bundle",
+		Args:  cobra.MaximumNArgs(1),
```
crazy-max, line 122: If we want feature parity with export-build tool, I think we should allow multiple refs to be exported. This is needed for GHA case when exporting bake builds.

Also the export-build exports all records for the builder being used if no refs are set but I think to be consistent with other history commands this is fine to export the first one. So maybe we could have an `--all` flag for such case?

commands/history/export.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return bundle.Export(ctx, c, w, toExport)
+}
+
+func exportCmd(dockerCli command.Cli, rootOpts RootOptions) *cobra.Command {
+	var options exportOptions
+
+	cmd := &cobra.Command{
+		Use:   "export [OPTIONS] [REF]",
+		Short: "Export a build into Docker Desktop bundle",
```
crazy-max, line 121: Not sure if "Docker Desktop" should be stipulated here. I think the following is enough:

```suggestion
		Short: "Export build records",
```

or

```suggestion
		Short: "Export build records bundle",
```

crazy-max: Also there are some places where we talk about build and other build record. I think we should be consistent:

https://github.com/docker/buildx/blob/18ccba072076ddbfb0aeedd6746d7719b0729b58/commands/history/import.go#L123

`Import build records into Docker Desktop`

https://github.com/docker/buildx/blob/18ccba072076ddbfb0aeedd6746d7719b0729b58/commands/history/logs.go#L101

`Print the logs of a build record`

https://github.com/docker/buildx/blob/18ccba072076ddbfb0aeedd6746d7719b0729b58/commands/history/open.go#L60

`Open a build record in Docker Desktop`

crazy-max (commented):
When exporting a build record and importing it back, the name is not the same:

image

I think this is because we don't check the local state for imported builds in desktop-build backend on Docker Desktop. ~Maybe we could set the builder and node name as annotation in the export bundle so we can look if a local state exists for this ref? WDYT?~

commands/history/export.go
```diff
@@ -0,0 +1,137 @@
...
+			if err != nil {
+				return err
+			}
+		}
+
+		toExport = append(toExport, &bundle.Record{
+			BuildHistoryRecord: rec.BuildHistoryRecord,
+			DefaultPlatform:    defaultPlatform,
+			LocalState:         st,
+			StateGroup:         stg,
```
crazy-max, line 95: About my comment https://github.com/docker/buildx/pull/3073#pullrequestreview-2702042410 I see we export the localstate so maybe that's something else

crazy-max: Okay after looking at our export-build tool implementation, the localstate is used for generating the summary payload for GHA and was never part of the exported bundle.

Do you think we could have a descriptor with the localstate?

commands/history/export.go
tonistiigi: > Not sure if "Docker Desktop" should be stipulated here. I think the following is enough:

It kind of shows what you can do with the bundle file.

commands/history/export.go
tonistiigi: > but doesn't seem exported in the bundle.

I'm not sure what you mean by that.

tonistiigi:
Added support for `--all` and multiple refs.

commands/history/export.go
crazy-max: Nevermind it's actually exported in https://github.com/docker/buildx/pull/3073/files#diff-e7b1db3ff4b89bf38887f616f47ee9c553b2a823446ef699641cd844ae87b1c2R185

crazy-max (approved):

crazy-max (approved):
Merging this one.

We need a follow-up for consistency in command descriptions related to https://github.com/docker/buildx/pull/3073#discussion_r2005272837

I will open an issue to track work that needs to be done on GHA to use this command instead of export-build tool.

And also needs documentation for all history commands and probably new manual in https://docs.docker.com/build/? Can someone from your team take care of it @usha-mandya?

PR #3087, opened 2025-04-04
cmd: support cli environment variables in standalone mode

### LINKED ISSUE #3081: [BUG] Building with Compose on remote Docker server via TLS TCP by setting `DOCKER_HOST` is broken
### Description

After upgrading Docker to version `28.0.4`, building with Compose on remote Docker server via TLS encrypted TCP socket is broken if the connection is defined by setting `DOCKER_HOST`, `DOCKER_CERT_PATH` and `DOCKER_TLS_VERIFY` environment variables.

Output:
```
Error response from daemon: Client sent an HTTP request to an HTTPS server.
```

Using Docker context with the same values still works.

### Steps To Reproduce

1. Create `compose.yaml` with the following content:
   ```
   name: compose
   services:
     foo:
       build:
         dockerfile_inline: FROM alpine
   ```
2. Run a DinD container in the background acting as remote Docker server:  
   `docker run -d --rm --name dind --privileged --security-opt='no-new-privileges=false' -p 127.0.0.1:2376:2376/tcp -v ./certs:/certs/client docker:dind`
3. Create a Docker context that points to the DinD container:  
   `docker context create dind --docker "host=tcp://localhost:2376,ca=$(pwd)/certs/ca.pem,cert=$(pwd)/certs/cert.pem,key=$(pwd)/certs/key.pem"`
4. Verify that building on remote Docker server by using context works:  
   `DOCKER_CONTEXT=dind docker build -t foo - <<< 'FROM alpine' && docker exec dind docker image ls`
5. Verify that building with Compose on remote Docker server by using context works:  
   `DOCKER_CONTEXT=dind docker compose build && docker exec dind docker image ls`
6. Clean up previously built images in the DinD container:  
   `docker exec dind docker image rm foo compose-foo`
7. Verify that building on remote Docker server by setting `DOCKER_HOST` works:  
   `DOCKER_CERT_PATH="$(pwd)/certs" DOCKER_TLS_VERIFY=1 DOCKER_HOST=tcp://localhost:2376 docker build -t foo - <<< 'FROM alpine' && docker exec dind docker image ls`
8. Building with Compose on remote Docker server by setting `DOCKER_HOST` exhibits the bug:  
   `DOCKER_CERT_PATH="$(pwd)/certs" DOCKER_TLS_VERIFY=1 DOCKER_HOST=tcp://localhost:2376 docker compose build`
9. Clean up:  
   `docker context rm dind && docker container stop dind`

### Compose Version

```Text
Docker Compose version v2.34.0
```

### Docker Environment

```Text
 Version:    28.0.4
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.22.0
    Version:  v2.34.0

Server:
 Containers: 0
  Running: 0
  Paused: 0
  Stopped: 0
 Images: 13
 Server Version: 28.0.4
 Storage Driver: overlayfs
  driver-type: io.containerd.snapshotter.v1
 Logging Driver: local
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 753481ec61c7c8955a23d6ff7bc8e4daed455734
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
  no-new-privileges
 Kernel Version: 5.15.167.4-microsoft-standard-WSL2
 Operating System: Ubuntu 24.04.2 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 8
 Total Memory: 15.59GiB
 Name: geth-desktop
 ID: 5814c356-970f-4d80-8dfa-4bbe4da79f33
 Debug Mode: false
 Username: gethdeeo
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Default Address Pools:
   Base: 172.25.0.0/16, Size: 24
   Base: fd4f:edd1:2b21:1011::/64, Size: 80
```

### Anything else?

_No response_

crazy-max:
fixes #3081


thaJeztah:
Slightly wondering what the most correct approach would be here. Ultimately this is one of the issues we discussed at some point when the cli plugins were created (i.e., ideally, the CLI ('host') would be able to handle the connection, and plugins to be able to use that connection.

The tricky bit here is that we start to (more, and more) cli-plugins to take over env-vars (and options) that are basically owned by the CLI, which also means that the CLI won't be able to change meaning for its own env-vars, without all plugins that exist to _also_ update accordingly.

Does buildx, when running standalone, have top-level flags to specify tis-config etc? i.e., would compose be able to;

- load the context, or other env-vars
- set the correct, buildx-standalone-specific command-line options to make it set the right connection?

crazy-max:
Thanks for your feedback, I was keeping this one in draft because I got questions related to these changes so I'm glad you came up here :sweat_smile:

> Does buildx, when running standalone, have top-level flags to specify tis-config etc? i.e., would compose be able to;

No, buildx does not have any top-level flags to set tls-config but I would assume that env vars like `DOCKER_TLS`, `DOCKER_CERT_PATH` would still be handled by docker/cli when running in standalone mode similar to plugin one. Maybe that's something that `cli.Initialize` should support in https://github.com/docker/cli/blob/58fba25b0935586eb14aa6b20b936cfc26df7a19/cli/command/cli.go#L234?

commands/root.go
```diff
@@ -40,7 +41,17 @@ func NewRootCmd(name string, isPlugin bool, dockerCli command.Cli) *cobra.Comman
...
-				return nil
+				// InstallFlags and SetDefaultOptions are necessary to match
+				// the plugin mode behavior to handle env vars such as
+				// DOCKER_TLS, DOCKER_TLS_VERIFY, ... and we also need to use a
+				// new flagset to avoid conflict with the global debug flag
+				// that we already handle in the root command otherwise it
+				// would panic.
+				nflags := pflag.NewFlagSet(cmd.DisplayName(), pflag.ContinueOnError)
+				options := cliflags.NewClientOptions()
+				options.InstallFlags(nflags)
```
tonistiigi, line 52: I'm confused what is docker-specific in this `InstallFlags` call? And if there is nothing docker-specific and just needs to be called then why isn't `NewClientOptions()` handling it automatically.

crazy-max: > Or is the difference from calling this function in `PersistentPreRunE` instead of before `NewRootCmd()`?

No there isn't bu I was hoping to be able to use `options.InstallFlags(cmd.Flags())` here but it panics because we already declare a global `debug` flag.

PR #3092, opened 2025-04-08
avoid import to testing helpers outside of tests

tonistiigi:
These utilities should not be included in the main build.

This only has a minimal 2.5KB binary size difference atm as there is still import to "testing" via containerd libs atm https://github.com/containerd/containerd/pull/11666 . Removing that one releases more.

### DIFF
--- a/build/git_test.go
+++ b/build/git_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/docker/buildx/util/gitutil"
+	"github.com/docker/buildx/util/gitutil/gittestutil"
 	"github.com/moby/buildkit/client"
 	specs "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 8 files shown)


tonistiigi:
This is much more complicated than I would have hoped as some of these helpers are used in integration tests. Tried to separate these into `gittestutil` subpackage, but it is really hard to understand the scoping difference of these packages. Needs a much bigger refactor and analysis of the intent of a given "gitutils" package.

crazy-max (approved):
LGTM

> Needs a much bigger refactor and analysis of the intent of a given "gitutils" package.

This also reminds of https://github.com/docker/buildx/pull/2005

thaJeztah:
Sometimes I wish Go would allow importing `_test.go` code from other packages' `_test.go` files.

PR #3094, opened 2025-04-08
command: runBuild, runBake: remove uses of DockerCLI.MeterProvider

thaJeztah:
- relates to https://github.com/docker/cli/commit/02537eac596a6e1bab435c8d2445603990445678 / https://github.com/docker/cli/pull/5067

The [DockerCli.MeterProvider] (and [DockerCli.TracerProvider]) methods have no direct connection with the CLI itself and are wrappers for `otel.GetMeterProvider()` and `otel.GetTracerProvider()`.

Remove the use of the CLI method, and use the otel functions directly.

[DockerCli.MeterProvider]: https://github.com/docker/cli/blob/91cbde67c509c8702417ee1cd64cd5d845ccd254/cli/command/telemetry.go#L62-L64
[DockerCli.TracerProvider]: https://github.com/docker/cli/blob/91cbde67c509c8702417ee1cd64cd5d845ccd254/cli/command/telemetry.go#L58-L60

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -38,6 +38,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/tonistiigi/go-csvvalue"
+	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 )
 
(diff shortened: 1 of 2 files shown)


jsternberg:
I think the original intention of these methods was that they would construct the `TraceProvider` and the `MeterProvider` if they didn't already exist so I don't think this is as simple as just "use the otel function" for these. I also think it's still good if there's a CLI exposed way of accessing the providers outside of the otel library. This is partly because we know the otel library implementations can create memory leaks if they are used without being set.

Issue #2613, opened 2024-07-18
docker-bake.json applying HCL string templates


amartani:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

When using `docker bake` with a JSON docker bake config, strings are being parsed as if they were HCL strings (applying HCL string template interpolation / directives).

### Expected behaviour

Expected the JSON file to be parsed as a JSON file, such that strings on it are not further processed as HCL templates. In the example below, I expect the image to be build and labeled with the `foo` label getting the literal value `%{bar}`.

### Actual behaviour

On the example below, parsing the docker-bake.json fails, as it tries to parse as HCL template.

### Buildx version

github.com/docker/buildx v0.13.0 0de5f1c

### Docker info

```text
docker info          
Client: Docker Engine - Community
 Version:    25.0.4
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.13.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.24.7
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 14
  Running: 0
  Paused: 0
  Stopped: 14
 Images: 55
 Server Version: 25.0.4
 Storage Driver: overlay2
  Backing Filesystem: btrfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: d2d58213f83a351ca8f528a95fbd145f5654e957
 runc version: v1.1.12-0-g51d5e94
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
[...]

PR #3097, opened 2025-04-08
bake: keep escaped interpolation in print output

crazy-max:
fixes https://github.com/docker/buildx/issues/2613

The interpolation and directive introductions are escaped by doubling their leading characters. The `${` sequence is escaped as `$${` and the `%{` sequence is escaped as `%%{`.

When the definition is parsed, the escape sequence is removed but we need to keep it as we expect a valid input when printing the definition with `--print`.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -3,6 +3,7 @@ package bake
 import (
 	"context"
 	"encoding"
+	"encoding/json"
 	"io"
 	"maps"
 	"os"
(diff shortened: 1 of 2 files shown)


bake/bake.go
```diff
@@ -733,6 +734,41 @@ type Target struct {
...
+	tgt.Args = make(map[string]*string, len(t.Args))
+	for k, v := range t.Args {
+		if v != nil {
+			escaped := esc(*v)
+			tgt.Args[k] = &escaped
+		}
+	}
+
+	return json.Marshal(tgt)
+}
```
crazy-max, line 770: I don't really like it but didn't find a way to achieve that during context evaluation with hcl.

tonistiigi (approved):

PR #3106, opened 2025-04-10
build: print frontend inline message

crazy-max:


commands/bake.go
```diff
@@ -424,6 +424,14 @@ func runBake(ctx context.Context, dockerCli command.Cli, targets []string, in ba
 		fmt.Fprintln(dockerCli.Out(), string(dt))
 	}
 
+	for _, name := range names {
+		if sp, ok := resp[name]; ok {
+			if v, ok := sp.ExporterResponse["frontend.result.inlinemessage"]; ok {
+				fmt.Fprintf(dockerCli.Out(), "\n%s:\n> %s\n", name, v)
```
tonistiigi, line 430: Why the `>` in the output here? I would rather put something in front of the target name if you want to distinguish it from the rest of the output.

Especially if output is multi-line then the `>` looks out of place.

tonistiigi (approved):
^ see comment

commands/bake.go
crazy-max: Opted for prefix

PR #3120, opened 2025-04-16
ci: assign author on pull request

crazy-max:
* https://github.com/crazy-max/.github?tab=readme-ov-file#pr-assign-author
* https://github.com/crazy-max/.github/blob/3e5689984859136d8c931e9cbd294412b88baa21/.github/workflows/pr-assign-author.yml

### DIFF
--- a/.github/workflows/pr-assign-author.yml
+++ b/.github/workflows/pr-assign-author.yml
@@ -1,0 +1,17 @@
+name: pr-assign-author
+
+permissions:
+  contents: read
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+
+jobs:
+  run:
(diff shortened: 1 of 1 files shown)


thompson-shaun (approved):

thaJeztah:
Curious; I recall you can only assign people with write access and/or part of the org; does it ignore contributors who don't have that?

crazy-max:
> Curious; I recall you can only assign people with write access and/or part of the org; does it ignore contributors who don't have that?

Yes there is a check for that using the GitHub API: https://github.com/crazy-max/.github/blob/3e5689984859136d8c931e9cbd294412b88baa21/.github/workflows/pr-assign-author.yml#L37-L41

More info: https://docs.github.com/en/rest/issues/assignees?apiVersion=2022-11-28#check-if-a-user-can-be-assigned-to-a-issue

PR #3122, opened 2025-04-17
Output correct image ID when using Docker with the containerd snapshotter

ctalledo:
Prior to this change, the following command emits the wrong image ID when using the "docker" driver and Docker is configured with the containerd snapshotter.

```
$ docker buildx build --load --iidfile=img.txt

$ docker run --rm "$(cat img.txt)" echo hello
docker: Error response from daemon: No such image: sha256:4ac37e81e00f242010e42f3251094e47de6100e01d25e9bd0feac6b8906976df. See 'docker run --help'.
```

The problem is that buildx is outputing the incorrect image ID in this scenario (it's outputing the container image config digest, instead of the container image digest, and the latter is needed to run the image with containerd).

This change fixes this.

See https://github.com/moby/moby/issues/45458.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -33,6 +33,7 @@ import (
 	"github.com/docker/buildx/util/cobrautil"
 	"github.com/docker/buildx/util/confutil"
 	"github.com/docker/buildx/util/desktop"
+	"github.com/docker/buildx/util/dockerutil"
 	"github.com/docker/buildx/util/ioset"
 	"github.com/docker/buildx/util/metricutil"
 	"github.com/docker/buildx/util/osutil"
@@ -321,11 +322,17 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions) 
 	if err != nil {
 		return err
 	}
+
 	_, err = b.LoadNodes(ctx)
 	if err != nil {
 		return err
 	}
 	driverType := b.Driver
+
+	dockerUsingContainerdSnapshotter := false
+	if driverType == "docker" || driverType == "docker-container" {
+		dockerUsingContainerdSnapshotter = isDockerUsingContainerdSnapshotter(ctx, dockerCli)
+	}
 
 	var term bool
 	if _, err := console.ConsoleFromFile(os.Stderr); err == nil {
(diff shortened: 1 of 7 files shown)


commands/build.go
```diff
@@ -321,12 +321,23 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
...
-	_, err = b.LoadNodes(ctx)
+	nodes, err := b.LoadNodes(ctx)
 	if err != nil {
 		return err
 	}
 	driverType := b.Driver
 
+	driverUsesContainerdSnapshotter := false
+	if driverType == "docker" {
+		// XXX: Each builder has multiple nodes, each with a potentially different
```
ctalledo, line 332: This does not seem totally right, need some advice on best approach.

ctalledo: I changed the solution and this comment is no longer applicable.

driver/docker/driver.go
```diff
@@ -144,6 +135,20 @@ func (d *Driver) IsMobyDriver() bool {
 	return true
 }
 
+func (d *Driver) UsesContainerdSnapshotter(ctx context.Context) bool {
```
tonistiigi, line 138: Can you just define a new `Feature` for this. Seems lot of duplication and extra unnecessary requests.

commands/build.go
```diff
@@ -321,12 +322,18 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
...
 	}
+
 	_, err = b.LoadNodes(ctx)
 	if err != nil {
 		return err
 	}
 	driverType := b.Driver
 
+	dockerUsingContainerdSnapshotter := false
+	if driverType == "docker" || driverType == "docker-container" {
```
tonistiigi, line 333: I don't think there is anything special for these drivers. For "docker" containerd support can be detected via feature flag like it is for other things not supported by graphdrivers.

For other drivers, I guess they are only affected if `--load` is set. If there is no `--load` then there is no reason to check a specific docker instance at build time, as it would be completely unused. If this feature has problems with other drivers as well on `--load` then I think we need to capture the image ID from moby side when we are doing `docker load`. So we load image into Moby, and it reports to us what is the ID of the image that was just imported (as it would be different in different Moby versions). It's quite weird though that the value of `--iidfile` depends on in `--load` is set or not.

ctalledo:
After a conversation with @tonistiigi and @jsternberg, the solution being proposed by this PR is not ideal, because this PR selects the image ID returned by buildkit (i.e., image digest or image config digest) based on the build driver config (particularly whether the driver is "docker" or "docker-container" and Docker is configured with the containerd-snapshotter), but there's no guarantee that this image ID can be used to run the image since the run may occur under a different host config.

The better solution is to have Moby run the image with either the image digest or image config digest. 

Closing this PR.

PR #3126, opened 2025-04-17
controller: remove controller grpc service

### LINKED ISSUE #3124: Proposal: Remove build controller
The build controller was created as a way to abstract away the command line interface with the idea it would be useful for creating a remote version of buildx that could be persisted and used for debugging. The CLI would create a background buildx process and would communicate through it.

The reality of the implementation is that it was never completed and currently just lives hidden behind the `BUILDX_EXPERIMENTAL` tag. It introduces additional complexity in the CLI code paths that are difficult to follow and have subtle differences that result in bugs when `BUILDX_EXPERIMENTAL` is used. There are currently no plans to finish the implementation and we're not really sure if this is the best way to complete the work on debuggable build environments.

We'd like to move the debugging builds conversation to creating a DAP interface for builds so we are only building the debugger and are not also building a frontend for the debugger. This alternative method would give us compatibility with a wider array of common editors: particularly vscode and neovim which have either DAP support builtin or prominent dap plugins.

At this point, the build controller was a useful experiment but we think it would not be worth the time and effort to finish the implementation.

jsternberg:
Remove the controller grpc service along with associated code related to
sessions or remote controllers.

Data types that are still used with complicated dependency chains have
been kept in the same package for a future refactor.

Fixes #3124.


tonistiigi (approved):
This seems like a good first step towards https://github.com/docker/buildx/issues/3124. I tested the `debug` command/monitor manually and all local commands still seemed to work as expected.

Looks like errdefs proto could have also been removed directly with this, but that can be the next PR.

tonistiigi:
BSD error is unrelated. Looks like another change in the vagrant vm.

PR #3136, opened 2025-04-21
Output correct image ID when using Docker with the containerd-snapshotter

ctalledo:
Prior to this change, the following command emits the wrong image ID when buildx uses the "docker-container" driver and Docker is configured with the containerd-snapshotter.

```
$ docker buildx build --load --iidfile=img.txt

$ docker run --rm "$(cat img.txt)" echo hello
docker: Error response from daemon: No such image: sha256:4ac37e81e00f242010e42f3251094e47de6100e01d25e9bd0feac6b8906976df. See 'docker run --help'.
```

The problem is that buildx is outputing the incorrect image ID in this scenario (it's outputing the container image config digest, instead of the container image digest used by the containerd-snapshotter).

This commit fixes this. See https://github.com/moby/moby/issues/45458.

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -237,6 +237,11 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt *O
 					opt.Exports[i].Output = func(_ map[string]string) (io.WriteCloser, error) {
 						return w, nil
 					}
+					// if docker is using the containerd snapshotter, prefer to export the image digest
+					// (rather than the image config digest). See https://github.com/moby/moby/issues/45458.
+					if features[dockerutil.OCIImporter] {
+						opt.Exports[i].Attrs["prefer-image-digest"] = "true"
+					}
 				}
 			} else if !nodeDriver.Features(ctx)[driver.DockerExporter] {
 				return nil, nil, notSupported(driver.DockerExporter, nodeDriver, "https://docs.docker.com/go/build-exporters/")
--- a/build/build.go
+++ b/build/build.go
@@ -525,7 +525,6 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
 							}
 						}
 					}
-
 					node := dp.Node().Driver
 					if node.IsMobyDriver() {
 						for _, e := range so.Exports {
@@ -559,6 +558,14 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
 									}
 								}
 							}
+						}
+					}
+					// if prefer-image-digest is set in the solver options, remove the image
+					// config digest from the exporter's response
+					for _, e := range so.Exports {
+						if e.Attrs["prefer-image-digest"] == "true" {
+							delete(rr.ExporterResponse, exptypes.ExporterImageConfigDigestKey)
+							break
 						}
 					}
 					return nil
--- a/tests/build.go
+++ b/tests/build.go
@@ -399,18 +399,25 @@ func testImageIDOutput(t *testing.T, sb integration.Sandbox) {
 
 	require.Equal(t, dgst.String(), strings.TrimSpace(stdout.String()))
 
+	// read the md.json file
 	dt, err = os.ReadFile(filepath.Join(targetDir, "md.json"))
 	require.NoError(t, err)
 
 	type mdT struct {
+		Digest       string `json:"containerimage.digest"`
 		ConfigDigest string `json:"containerimage.config.digest"`
 	}
+
 	var md mdT
 	err = json.Unmarshal(dt, &md)
 	require.NoError(t, err)
 
 	require.NotEmpty(t, md.ConfigDigest)
-	require.Equal(t, dgst, digest.Digest(md.ConfigDigest))
+	require.NotEmpty(t, md.Digest)
+
+	// verify the image ID output is correct
+	// XXX: improve this by checking that it's one of the two expected digests depending on the scenario.
+	require.Contains(t, []digest.Digest{digest.Digest(md.ConfigDigest), digest.Digest(md.Digest)}, dgst)
 }
 
 func testBuildMobyFromLocalImage(t *testing.T, sb integration.Sandbox) {


commands/build.go
```diff
@@ -412,12 +416,40 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
...
+			}
+		}
 	}
-	return dgst
+
+	docker := dockerutil.NewClient(dockerCli)
+	features := docker.Features(ctx, dockerCtx)
+	if features[dockerutil.OCIImporter] {
+		usingContainerdSnapshotter = true
+	}
```
crazy-max, line 450: It should check features from driver, not docker client imo: https://github.com/docker/buildx/blob/9730a20f6b0b1f15a3b917e90c83f8b97a62dfd2/driver/driver.go#L74

crazy-max, line 450: I think this is best to check worker info using buildkit client for this entitlement similar to https://github.com/docker/buildx/blob/9730a20f6b0b1f15a3b917e90c83f8b97a62dfd2/driver/docker/driver.go#L81-L90

commands/build.go
```diff
@@ -321,12 +322,15 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
...
 		return err
 	}
+
 	_, err = b.LoadNodes(ctx)
 	if err != nil {
 		return err
 	}
 	driverType := b.Driver
 
+	dockerUsingContainerdSnapshotter := isDockerUsingContainerdSnapshotter(ctx, dockerCli, opts.Exports)
```
tonistiigi, line 332: This is doing (multiple) extra requests, so should be avoided if it is unnecessary. Could be avoided completely I think when connected with "load".

commands/build.go
```diff
@@ -412,12 +416,40 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
...
+
+func isDockerUsingContainerdSnapshotter(ctx context.Context, dockerCli command.Cli, exports []*controllerapi.ExportEntry) bool {
+	usingContainerdSnapshotter := false
+
+	// If the build command has "-o type=docker,context=<mycontext>", then pass that
+	// context to docker.Features() below, so we can find out if the associated
+	// Docker engine is using the containerd snapshotter.
+	dockerCtx := ""
+	for _, e := range exports {
+		if e.Type == "docker" {
```
tonistiigi, line 437: this doesn't work with 

```
else if nodeDriver.Features(ctx)[driver.DefaultLoad] {
```
in build/opt.go

commands/build.go
```diff
@@ -412,12 +416,40 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
 }
 
 // getImageID returns the image ID - the digest of the image config
-func getImageID(resp map[string]string) string {
-	dgst := resp[exptypes.ExporterImageDigestKey]
-	if v, ok := resp[exptypes.ExporterImageConfigDigestKey]; ok {
-		dgst = v
+func getImageID(resp map[string]string, dockerUsingContainerdSnapshotter bool) string {
```
tonistiigi, line 419: `dockerUsingContainerdSnapshotter` should be `preferImageDigest`.

commands/build.go
```diff
@@ -412,12 +416,40 @@ func runBuild(ctx context.Context, dockerCli command.Cli, options buildOptions)
...
+	if dockerUsingContainerdSnapshotter {
+		// If Docker is using the containerd snapshotter, use the image digest, not
+		// the image config digest. See https://github.com/moby/moby/issues/45458.
+		return resp[exptypes.ExporterImageDigestKey]
+	}
+
+	return resp[exptypes.ExporterImageConfigDigestKey]
```
tonistiigi, line 426: This is not compatible with previous code there image digest is returned if config digest does not exist.

ctalledo: Thanks, I've changed it to be compatible with how it was (when `preferImageDigest` is false).

commands/build.go
ctalledo: Done, thanks.

commands/build.go
ctalledo: Thanks; can you expand a bit more on what "connected with `load`" means?

commands/build.go
ctalledo: Why is that? in `build/opt.go` I see:

```
 166 │          } else if nodeDriver.Features(ctx)[driver.DefaultLoad] {                                                                                                                                                                                                        
 167 │             opt.Exports = []client.ExportEntry{{Type: "docker", Attrs: map[string]string{}}}                                                                                                                                                                             
 168 │          }
```

In L167 `Attrs` is an empty map, so the code I added in `commands/build.go` should simply ignore and set `dockerCtx` to an empty context, which in turn means the "current Docker context" in `util/dockerUtil/client.go`:

```
  26 │ // API returns a new docker API client.                                                                                                                                                                                                                                  
  27 │ func (c *Client) API(name string) (dockerclient.APIClient, error) {                                                                                                                                                                                                      
  28 │    if name == "" {                                                                                                                                                                                                                                                       
  29 │       name = c.cli.CurrentContext()                    <<< HERE                                                                                                                     
  30 │    }                                                                                                                                                                                                                                                                     
  31 │    return NewClientAPI(c.cli, name)                                                                                                                                                                                                                                      
  32 │ } 
```

ctalledo:
Thanks @tonistiigi for the review and comments, I've answered them, PTAL when you get a chance. Thanks again!

commands/build.go
tonistiigi: > code I added in commands/build.go should simply ignore

But it shouldn't ignore it if the build actually performs a load to docker daemon, right?

vendor/github.com/moby/buildkit/client/graph.go
```diff
@@ -56,5 +56,6 @@ type SolveStatus struct {
 
 type SolveResponse struct {
 	// ExporterResponse is also used for CacheExporter
-	ExporterResponse map[string]string
+	ExporterResponse  map[string]string
+	PreferImageDigest bool
```
tonistiigi, line 60: Can't change the files in here. You can delete the config digest in resp (or add "prefer" key there) in build/build.go after build has completed.

build/opt.go
```diff
@@ -237,6 +237,11 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt *O
 					opt.Exports[i].Output = func(_ map[string]string) (io.WriteCloser, error) {
 						return w, nil
 					}
+					// if docker is using the containerd snapshotter, prefer to export the image digest
+					// (rather than the image config digest). See https://github.com/moby/moby/issues/45458.
+					if features[dockerutil.OCIImporter] {
```
tonistiigi, line 242: You shouldn't look at the driver features here but the `dapi` inside the `docker.LoadImage`.

ctalledo: It seems strange to put that inside `docker.LoadImage()` because that function has a very specific task (i.e., to return a function that will perform the image load). 

The way it's now seems cleaner and addresses the points we discussed yesterday: the code only executes when we know we will load the image, and it reuses the existing Docker client API.

tonistiigi: Was again confused by this not being driver features but a variable from `dockerutil`. This can be cleaned up in follow-up to avoid creating separate clients for `LoadImage` and `Features` happening sequentially. I already left a comment in [here](https://docker.slack.com/archives/C02JGHEPZJ5/p1745348065164869) that this whole `Features` call is broken an incorrectly cached.

Logically I don't think there is anything wrong with this being in the `LoadImage` caller as what you are asking is the image format of the image loader. But if the `docker info` has already happened and saved to `features` var then should of course avoid it going to waste and doing another call.

vendor/github.com/moby/buildkit/client/graph.go
ctalledo: Got it, thanks; I opted for removing the config digest in that case. PTAL when you can, thanks!

build/opt.go
ctalledo: OK thanks, I'll leave it as is then.

tonistiigi (approved):

PR #3141, opened 2025-04-22
use filepath.IsAbs to support windows paths

ndeloof:
`path.IsAbs` assumes Unix paths, prefer filepath so an absolute windows paths is well identified

fixes https://github.com/docker/compose/issues/12669


glours (approved):

crazy-max (approved):
LGTM thanks

PR #3152, opened 2025-04-23
history: make sure build record is finalized before exporting

### LINKED ISSUE #3149: Finalize records before exporting
follow-up https://github.com/docker/actions-toolkit/pull/652#discussion_r2055080629

We should make sure records are finalized before exporting them similar to `history trace`: https://github.com/docker/buildx/blob/9ba3f77219b15cff4cea37262d270083dec334ad/commands/history/trace.go#L57-L70

@tonistiigi I wonder if we should have a `Finalize` attribute in https://github.com/docker/buildx/blob/9ba3f77219b15cff4cea37262d270083dec334ad/commands/history/utils.go#L116-L119

To share logic for both `export` and `trace` in `queryRecords`?

crazy-max:
fixes #3149

### DIFF
--- a/commands/history/export.go
+++ b/commands/history/export.go
@@ -50,6 +50,7 @@ func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) e
 	for _, ref := range opts.refs {
 		recs, err := queryRecords(ctx, ref, nodes, &queryOptions{
 			CompletedOnly: true,
+			Finalize:      true,
 		})
 		if err != nil {
 			return err
--- a/commands/history/utils.go
+++ b/commands/history/utils.go
@@ -116,6 +116,7 @@ type historyRecord struct {
 type queryOptions struct {
 	CompletedOnly bool
 	Filters       []string
+	Finalize      bool
 }
 
 func queryRecords(ctx context.Context, ref string, nodes []builder.Node, opts *queryOptions) ([]historyRecord, error) {
@@ -161,6 +162,15 @@ func queryRecords(ctx context.Context, ref string, nodes []builder.Node, opts *q
 				w.Write(filters)
 				w.Flush()
 				filters = []string{strings.TrimSuffix(sb.String(), "\n")}
+			}
+
+			if opts.Finalize {
+				if _, err := c.ControlClient().UpdateBuildHistory(ctx, &controlapi.UpdateBuildHistoryRequest{
+					Ref:      ref,
+					Finalize: true,
+				}); err != nil {
+					return err
+				}
 			}
 
 			serv, err := c.ControlClient().ListenBuildHistory(ctx, &controlapi.BuildHistoryRequest{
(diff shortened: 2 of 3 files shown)


tonistiigi (commented):
As discussed offline:
- Finalize should only happen if `Trace==nil`
- For export, we can leave some opt-in mechanism to force finalize, so we don't need a sleep to wait for the client traces to finish up if the user can already confirm it.

commands/history/export.go
```diff
@@ -62,6 +63,25 @@ func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) e
...
+		if opts.finalize {
+			var finalize bool
+			for _, rec := range recs {
+				if rec.Trace == nil {
+					finalize = true
+					break
+				}
+			}
+			if finalize {
+				recs, err = queryRecords(ctx, ref, nodes, &queryOptions{
```
tonistiigi, line 75: This would only work if ref is set. Might need to call finalize directly for all the refs in here where `Trace==nil` and then to full `queryRecords()` again.

commands/history/utils.go
```diff
@@ -248,6 +248,28 @@ func queryRecords(ctx context.Context, ref string, nodes []builder.Node, opts *q
...
+			if err != nil {
+				return err
+			}
+			_, err = c.ControlClient().UpdateBuildHistory(ctx, &controlapi.UpdateBuildHistoryRequest{
+				Ref:      ref,
+				Finalize: true,
+			})
+			return err
+		})
+	}
```
crazy-max, line 269: @tonistiigi For multi-node case I think this is safe even if the ref doesn't exist on unrelated node: https://github.com/moby/buildkit/blob/b46daef219fd318bd0f9c8f630472af771b2bb4a/solver/llbsolver/history.go#L676-L686

commands/history/export.go
```diff
@@ -62,6 +63,22 @@ func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) e
...
 
+		if opts.finalize {
+			for _, rec := range recs {
+				if rec.Trace == nil {
+					if err := finalizeRecord(ctx, rec.Ref, nodes); err != nil {
+						return err
+					}
+				}
+			}
+			recs, err = queryRecords(ctx, ref, nodes, &queryOptions{
```
tonistiigi, line 74: only do this if some records were finalized

docs/reference/buildx_history_export.md
```diff
@@ -49,6 +50,16 @@ docker buildx history export --builder builder0 ^1 -o builder0-build.dockerbuild
 docker buildx history export --debug qu2gsuo8ejqrwdfii23xkkckt -o debug-build.dockerbuild
 ```
 
+### <a name="finalize"></a> Ensure build records are finalized before exporting (--finalize)
+
+Clients can report their own traces concurrently and not all traces may be
+finalized when a build is completed. Use the `--finalize` flag to ensure all
```
tonistiigi, line 56: "not all traces may be
saved yet by the time of the export"

commands/history/export.go
crazy-max: :see_no_evil:

tonistiigi (approved):

PR #3157, opened 2025-04-24
Add BUILDKIT_SYNTAX option handling

wnonnemaker:
This fix allows building with a remote builder where frontend.dockerfile.v0 enabled = false in the buildkitd yaml file.

Note that this change only allows the usage of BUILDKIT_SYNTAX with a custom frontend image, and using the #syntax directive in this case will still fail.

Resolves: docker#3077

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -109,6 +109,11 @@ func toSolveOpt(ctx context.Context, node builder.Node, multiDriver bool, opt *O
 		so.FrontendAttrs["cgroup-parent"] = opt.CgroupParent
 	}
 
+	if v, ok := opt.BuildArgs["BUILDKIT_SYNTAX"]; ok {
+		so.Frontend = "gateway.v0"
+		so.FrontendAttrs["source"] = v
+	}
+
 	if v, ok := opt.BuildArgs["BUILDKIT_MULTI_PLATFORM"]; ok {
 		if v, _ := strconv.ParseBool(v); v {
 			so.FrontendAttrs["multi-platform"] = "true"


wnonnemaker:
This is my first pull request on this project, so any tips/guidance is much appreciated.

I didn't write any unit tests for this feature, so if you want me to do that, let me know. 

While looking into this, I noticed a related bug where using the #syntax directive in a dockerfile will not be properly recognized if the BuildKit instance has 
dockerfile.v0 disabled. If you want me to look more into this bug I am happy to do that as well.

tonistiigi (commented):
Look at the `forwardGateway` in https://github.com/moby/buildkit/blob/master/frontend/dockerfile/builder/build.go#L55-L56 . There is another parameter `cmdline` as well that allows passing extra args to the external frontend.

tonistiigi (changes requested):
^

wnonnemaker:
> ^

on it chief

marxarelli:
> Look at the `forwardGateway` in https://github.com/moby/buildkit/blob/master/frontend/dockerfile/builder/build.go#L55-L56 . There is another parameter `cmdline` as well that allows passing extra args to the external frontend.

@tonistiigi IIUC I don't think the `cmdline` parameter is relevant in the case where the `dockerfile.v0` frontend is disabled since the solve request never makes it to the forwarder.

What @wnonnemaker has here seems right from my perspective.

(We also have `dockerfile.v0` disabled on some prod image builders, and I was looking to make this change so we can start using `buildx` everywhere.)

tonistiigi:
@marxarelli This PR is not directly related to dockerfile syntax being disabled. Currently if you set `BUILDKIT_SYNTAX` (and it is handled by Dockerfile frontend) then the `cmdline` component is handled by the value. If now we set it directly in buildx without Dockerfile frontend then the behavior of the value should remain the same as it was before.

marxarelli:
> @marxarelli This PR is not directly related to dockerfile syntax being disabled.

I'm a bit confused because @wnonnemaker said:
> This fix allows building with a remote builder where frontend.dockerfile.v0 enabled = false in the buildkitd yaml file.

But maybe you don't mean the motivation of the PR but rather the implementation? That makes sense to me.

> Currently if you set `BUILDKIT_SYNTAX` (and it is handled by Dockerfile frontend) then the `cmdline` component is handled by the value. If now we set it directly in buildx without Dockerfile frontend then the behavior of the value should remain the same as it was before.

Ah, I see. Is the correct behavior to set the `cmdline` frontend option to the same value as `source`?

marxarelli:
> Ah, I see. Is the correct behavior to set the `cmdline` frontend option to the same value as `source`?

Sorry, I think I see it now. Set `cmdline` to the raw value of `BUILDKIT_SYNTAX` and `source` to the first part...

marxarelli:
@tonistiigi I went ahead and created a new PR, so I can contribute the changes you've requested. @wnonnemaker I added your commit there so you don't lose credit for your work.

crazy-max:
Fixed in https://github.com/docker/buildx/pull/3385

PR #3159, opened 2025-04-26
Consistent timeout parameter

CamJN:
I rebased https://github.com/docker/buildx/pull/2586 off master, and am willing to try to fix any issues that remain if they are mentioned here.

### DIFF
--- a/builder/node.go
+++ b/builder/node.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/containerd/platforms"
 	"github.com/docker/buildx/driver"
@@ -38,6 +39,8 @@ type Node struct {
 	Labels     map[string]string
 	CDIDevices []client.CDIDevice
 }
+
+const defaultDriverTimeout = 120 * time.Second
 
 // Nodes returns nodes for this builder.
 func (b *Builder) Nodes() []Node {
@@ -130,6 +133,7 @@ func (b *Builder) LoadNodes(ctx context.Context, opts ...LoadNodesOption) (_ []N
 					Platforms:       n.Platforms,
 					ContextPathHash: b.opts.contextPathHash,
 					DialMeta:        lno.dialMeta,
+					Timeout:         defaultDriverTimeout,
 				})
 				if err != nil {
 					node.Err = err
--- a/commands/create.go
+++ b/commands/create.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/docker/buildx/builder"
 	"github.com/docker/buildx/driver"
@@ -27,6 +28,7 @@ type createOptions struct {
 	buildkitdFlags      string
 	buildkitdConfigFile string
 	bootstrap           bool
+	timeout             time.Duration
 	// upgrade      bool // perform upgrade of the driver
 }
 
@@ -61,6 +63,7 @@ func runCreate(ctx context.Context, dockerCli command.Cli, in createOptions, arg
 		Use:                 in.use,
 		Endpoint:            ep,
 		Append:              in.actionAppend,
+		Timeout:             in.timeout,
 	})
 	if err != nil {
 		return err
@@ -80,7 +83,7 @@ func runCreate(ctx context.Context, dockerCli command.Cli, in createOptions, arg
 	return nil
 }
 
-func createCmd(dockerCli command.Cli) *cobra.Command {
+func createCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	var options createOptions
 
 	var drivers bytes.Buffer
@@ -96,6 +99,7 @@ func createCmd(dockerCli command.Cli) *cobra.Command {
 		Short: "Create a new builder instance",
 		Args:  cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			options.timeout = rootOpts.timeout
 			return runCreate(cmd.Context(), dockerCli, options, args)
 		},
 		ValidArgsFunction: completion.Disable,
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -217,7 +217,7 @@ func (d *Driver) wait(ctx context.Context, l progress.SubLogger) error {
 			select {
 			case <-ctx.Done():
 				return context.Cause(ctx)
-			case <-time.After(time.Duration(try*120) * time.Millisecond):
+			case <-time.After(d.Timeout / time.Microsecond):
 				try++
 				continue
 			}
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"math"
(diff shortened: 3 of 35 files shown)


driver/docker-container/driver.go
```diff
@@ -217,7 +217,7 @@ func (d *Driver) wait(ctx context.Context, l progress.SubLogger) error {
 			select {
 			case <-ctx.Done():
 				return context.Cause(ctx)
-			case <-time.After(time.Duration(try*120) * time.Millisecond):
+			case <-time.After(d.Timeout / time.Microsecond):
```
tonistiigi, line 220: see https://github.com/docker/buildx/pull/2586/files#diff-403916c02204229a19313f586f2fffa2536dde693a6e625bc4525f25be068ec9

commands/create.go
```diff
@@ -61,6 +63,7 @@ func runCreate(ctx context.Context, dockerCli command.Cli, in createOptions, arg
 		Use:                 in.use,
 		Endpoint:            ep,
 		Append:              in.actionAppend,
+		Timeout:             in.timeout,
```
tonistiigi, line 66: Is it intentional that this is passed as a value rather than added to `ctx`? So this `Create()` call can still take more time than timeout.

builder/node.go
```diff
@@ -130,6 +133,7 @@ func (b *Builder) LoadNodes(ctx context.Context, opts ...LoadNodesOption) (_ []N
 					Platforms:       n.Platforms,
 					ContextPathHash: b.opts.contextPathHash,
 					DialMeta:        lno.dialMeta,
+					Timeout:         defaultDriverTimeout,
```
tonistiigi, line 136: If the driver timeout is still constant 120s and can't be changed, wouldn't it be safer if this would be added in the `driver` pkg. Atm I think you could still theoretically call driver pkg forgetting to pass in `InitConfig.Timeout` and then possibly bad things can happen as it doesn't seem to be validated.

driver/remote/factory.go
```diff
@@ -87,6 +88,12 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
 				return nil, err
 			}
 			d.defaultLoad = parsed
+		case "timeout":
```
tonistiigi, line 91: Why is this specific for remote driver?

driver/docker-container/driver.go
CamJN: So it looks like before it did a linear backoff with step=120ms:

try 1: After 120ms
try2: After 240ms
…
try 15: After 1,800ms

total: 14,400ms or 14.4s


if the new Timeout is the total time and we want to keep the linear backoff then I think the math should be:

total = 15*(1*step+15*step)/2
(total*2)/(15*16) = step

I implemented somethng like that.

driver/remote/factory.go
CamJN: In the `driver/kubernetes/factory.go` New function the option parsing is done in `processDriverOpts`, the `driver/docker-container/factory.go` New function missing this option parsing seems like an oversight possibly, I added the timeout option parsing there too, the `driver/docker/factory.go` New function doesn't seem to do any option parsing at all so not doing any timeout parsing there is consistent with the existing code but I don't know if it is okay.

commands/create.go
CamJN: Addressing how the timeout is used:
In `builder.Create` that this `builder.CreateOpts` struct is passed to, the `context.WithTimeoutCause` is created using the `opts.Timeout` and the `opts` there is this `builder.CreateOpts` we are passing to `builder.Create` in the last argument here. So this Timeout is the one that is used.

Addressing where the Timeout comes from:
The `in` on this line comes from the `createOptions` struct passed to the `runCreate` function we are in, and that is populated in the `createCmd` function that calls us here. The options variable passed into `runCreate` has its timeout field populated from the `rootOpts` struct's `timeout` field before being passed into the function. The rootOpts var seems to be populated in the rootFlags function, including the timeout.

As for passing the timeout as a field in CreateOpts vs adding it to ctx, I don't know for sure why it was done that way, but I'd guess because nothing else in runCreate modifies the ctx. Not being much of a go user, I will take your word that the Create call can take more time than timeout but the slow part of runCreate seems bounded by the timeout just like before?

builder/node.go
CamJN: I added a check in GetDriver that the Timeout is set to something > 0.

driver/docker-container/driver.go
```diff
@@ -199,12 +199,14 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 }
 
 func (d *Driver) wait(ctx context.Context, l progress.SubLogger) error {
-	try := 1
+	var try uint64 = 1
+	var maxTries uint64 = 15
+	var step uint64 = max((uint64(d.Timeout.Abs().Milliseconds()) * 2) / (maxTries * (1 + maxTries)), 1)
```
tonistiigi, line 204: I don't understand this construction. Why `Abs(), why `Milliseconds()` , why `maxTries ** 2`?

I don't think this really makes much difference what the step is. If you set the timeout too short it will error anyway and checking more frequently than 120ms doesn't make it faster but might just increase overhead. Maybe better is to just remember how much time we have spent waiting on each iteration and if the last `time.After(time.Duration(try * step)` goes over Timeout then just clip it to smth like `Timeout - 1 second` so that we give the waiter the best chance to succeed before the timeout is reached.

My original comment was for division by `time.Microsecond` that didn't make any logical sense to me. I still see that in k8s driver.

commands/create.go
tonistiigi: I'd just add it to the context as with the other commands if we don't know the reason why this one should be different.

tonistiigi (commented):
After you are done and CI is green, you can also squash the commits.

driver/docker-container/driver.go
CamJN: > Why Abs(), why Milliseconds() , why maxTries ** 2?

`Abs()` because duration is signed and the math doesn't work if the duration is negative, it's just a defensive coding thing.

`Milliseconds()` because I am working in milliseconds, like the original code before the PR did anything.

`maxtries*(maxtries+1)` is just part of the math for getting the sum of a number of elements in an arithmetic sequence, like the old timeouts (120ms, 240ms, 360ms, etc).

I just solved for the step size given a total duration, where before the step size was fixed and the total duration was implied (14,400ms). I could keep the step size the same and rather cap the total duration and not control for the number of tries, that's definitely an option. 

I replaced the use of time.Microsecond in the k8s driver with `/1000`.

commands/create.go
CamJN: Actually, I'm a bit confused what you mean. Upon closer inspection the rm, ls, inspect, and create commands all do it this way, vs the build command that's different. The build command gets the timeout from the control.ControlOptions inclusion in the buildOptions struct, but doesn't seem to use it anywhere, possibly because builds tend to take absolute ages?

As a non go user what does, "add it to the context" mean, and in what way is that different between the create command and the rm, ls, or inspect commands?

driver/docker-container/driver.go
CamJN: I switched to altering the number of tries instead of the step size

avoidik:
if you don't mind, I will try again ;)

PR #3161, opened 2025-04-29
bake: allow override of list variables via environment

### LINKED ISSUE #3160: Bake: allow override of list variables via environment
### Description

Given a bake file with
```hcl
variable "FOO" {
  default = ["one", "two"]
}
```

```hcl
variable "FOO" {
}
```

rrjjvv:
Roughly speaking, if a bake variable is a list with uniform type, it can be overridden via environment variable.  The  environment variable must be a CSV string whose fields can be coerced to that of the list.

Resolves #3160 

I decided to proactively create a PR (despite my original comments in the issue) since I thought it unlikely to be outright rejected (given it was marked as a TODO), and any feedback or requested changes might be easier and/or more productive alongside an actual proposed implementation with test cases.

The diff looks artificially large due to extracting the conversion logic into its own function.  There's no obvious reason the primitives in a list should be processed differently than a primitive on its own, so I thought it'd be preferred to do the refactor rather than copy/pasting that logic into the new tuple/list branches.  The actual functionality for the three primitives remains unchanged.

While not the longest in the file, my new test method is among the longest.  I thought it more important to match the style of existing tests, using comments where I'd typically use a subtest.  And speaking of tests, there was one integration test that did not pass for me on a pristine checkout.  After my change, all existing test continue to pass except for that same one that failed prior to my change.  It's definitely unrelated to my change and has the feel of something that a core maintainer might not see due to having a unique setup (or maybe using different procedures to build).  Specifically, https://github.com/docker/buildx/blob/2eaea647d8aba61686f88ecbb5148ae91e3ad6da/tests/version.go#L48
If this isn't actually expected/anticipated, I'd be happy to dig and/or provide any other details you need.

### DIFF
--- a/bake/hcl_test.go
+++ b/bake/hcl_test.go
@@ -1603,13 +1603,99 @@ func TestFunctionNoResult(t *testing.T) {
 func TestVarUnsupportedType(t *testing.T) {
 	dt := []byte(`
 		variable "FOO" {
-			default = []
+			default = {}
 		}
 		target "default" {}`)
 
 	t.Setenv("FOO", "bar")
 	_, err := ParseFile(dt, "docker-bake.hcl")
 	require.Error(t, err)
+}
+
+func TestListVarOverride(t *testing.T) {
+	// no empty lists; don't know what to convert overrides to
+	dt := []byte(`
+		variable "FOO" {
+			default = []
+		}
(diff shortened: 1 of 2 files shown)


rrjjvv:
Closing in favor of #3167

PR #3164, opened 2025-05-01
controller: remove controller/errdefs protobuf files

jsternberg:
Remove the protobuf files associated with controller/errdefs.

This doesn't completely remove the type as the monitor still uses it as
a signal to start the monitor.

Related to #3124.


controller/errdefs/build.go
```diff
@@ -1,48 +1,20 @@
...
-	"github.com/containerd/typeurl/v2"
-	"github.com/docker/buildx/util/desktop"
-	"github.com/moby/buildkit/util/grpcerrors"
-)
-
-func init() {
-	typeurl.Register((*Build)(nil), "github.com/docker/buildx", "errdefs.Build+json")
-}
-
 type BuildError struct {
```
tonistiigi, line 3: This doesn't seem to be doing anything atm. As it doesn't have any extra properties and just forwards to internal error.

cmd/buildx/main.go
```diff
@@ -112,11 +111,6 @@ func main() {
 	var ebr *desktop.ErrorWithBuildRef
 	if errors.As(err, &ebr) {
 		ebr.Print(cmd.Err())
-	} else {
-		var be *controllererrors.BuildError
-		if errors.As(err, &be) {
-			be.PrintBuildDetails(cmd.Err())
```
tonistiigi, line 118: Can you explain bit what is going on here and commands/build and why this needed anymore. Where is the replacement `PrintBuildDetails`.

jsternberg: This section was previously needed because `desktop.ErrorWithBuildRef` wasn't serializable. So this type was used as a wrapper for `desktop.ErrorWithBuildRef`. `PrintBuildDetails` will just create a `desktop.ErrorWithBuildRef` and then call `Print` so it's equivalent to the above.

controller/errdefs/build.go
jsternberg: Yea it's current only use is to detect a specific method of failure [here](https://github.com/docker/buildx/blob/6df3a0928478a9584e469f8527ed2c23a6f10c33/monitor/commands/reload.go#L50-L55). I think that we can fully remove this once we do a bit of a refactor to how the debug adapter works with the DAP work. For now, it was easier to keep this as a sentinel error for reload.

tonistiigi (approved):

PR #3167, opened 2025-05-04
Allow variables to be explicitly typed (and enforced)

### LINKED ISSUE #3160: Bake: allow override of list variables via environment

rrjjvv:
Resolves #3160 and likely replaces #3161

In https://github.com/docker/buildx/issues/3160#issuecomment-2840204139 it was suggested that prior to allowing variable lists to be overridden via environment, proper typing capabilities should be implemented (and used).  This implements that typing, which results in the actual override code being superior to #3161 in every way, but touches more code.

This allows (but does not require) variables to have explicit types, similar to Terraform variables.  It uses HCL's `typeexpr` extension for the specification. For conversion of overrides to complex types (when explicit typing is provided), HCL's native JSON-based unmarshalling is used.

Typing is independent of any default value, but if a default is provided, it will be validated.  Similarly, if an override is provided, it will be converted to that type.  If typing is not provided, previous behavior is used.

For complex types, the happy path is lists of primitives, but in theory any complex/composite type can be used provided they are expressed correctly in JSON.  In the interest of simplicity and correctness, there are no shortcuts for lists (i.e., `VAR='["foo","bar"]'` and not `VAR=foo,bar`).  There *is* a shortcut for strings as users don't provide them for untyped variables and would be unintuitive (i.e. `VAR=foo` is valid for a string, though not valid JSON).

---

There should be no changes in behavior (for current or future un-typed usage), with one theoretical exception.  Today, extraneous attributes (for variables at least) are accepted, but ignored with no warnings.  It is theoretical possibly that users may have become reliant on this, e.g.
```hcl
variable "server" {
  default = "some.host"
  type = "linux"
}
```
Anyone using that attribute for their own purposes *will* encounter breaking behavior with this implementation (unless it happens to be a legitimate type expression).  I don't know if that is a real concern.

Also, this PR only contains code changes.  I assumed that 1) this change was unlikely to go through on the first submission and may require changes (or, of course, rejected), and 2) you may have dedicated people/processes for updating user docs.  Let me know if you want/need me to do anything in this regard.

### DIFF
--- a/bake/hcl_test.go
+++ b/bake/hcl_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"runtime"
 	"testing"
 
 	hcl "github.com/hashicorp/hcl/v2"
@@ -2306,6 +2307,9 @@ func TestJSONOverridePriority(t *testing.T) {
 	})
 
 	t.Run("JSON override works with lowercase vars", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
+			t.Skip("Windows case-insensitivity")
+		}
 		dt := []byte(`
             variable "foo" {
                 type = number
@@ -2316,7 +2320,7 @@ func TestJSONOverridePriority(t *testing.T) {
                     bar = foo
                 }
             }`)
-		// may seem reasonable, but not supported
+		// may seem reasonable, but not supported (on case-sensitive systems)
 		t.Setenv("foo_json", "9000")
 		c, err := ParseFile(dt, "docker-bake.hcl")
 		require.NoError(t, err)
--- a/bake/hcl_test.go
+++ b/bake/hcl_test.go
@@ -1840,6 +1840,8 @@ func TestDefaultVarTypeWithAttrValuesEnforcement(t *testing.T) {
 
 func TestTypedVarOverrides(t *testing.T) {
 	const unsuitableValueType = "Unsuitable value type"
+	const unsupportedType = "unsupported type"
+	const failedToParseElement = "failed to parse element"
 	tests := []struct {
 		name         string
 		varType      string
@@ -1860,17 +1862,14 @@ func TestTypedVarOverrides(t *testing.T) {
 			override:  "99",
 			wantValue: "99",
 		},
-		// this breaks the rule about needing proper JSON as it would violate
-		// the principle of least surprise and hinder usability
 		{
-			name:      "enquoted string accepted",
+			name:      "unquoted string accepted",
 			varType:   "string",
 			override:  "hello",
 			wantValue: "hello",
 		},
-		// similar to above, an environment variable with a quoted string would
-		// most likely be intended to be a string whose first and last characters
-		// are quotes
+		// an environment variable with a quoted string would most likely be intended
+		// to be a string whose first and last characters are quotes
 		{
 			name:      "quoted string keeps quotes in value",
 			varType:   "string",
@@ -1891,27 +1890,11 @@ func TestTypedVarOverrides(t *testing.T) {
 			wantErrorMsg: "collection must be a list",
 		},
 		{
-			name:      "proper JSON list of strings",
-			varType:   "list(string)",
-			override:  `["hi","there"]`,
-			argValue:  `join("-", FOO)`,
-			wantValue: "hi-there",
-		},
-		{
 			name:      "proper CSV list of strings",
 			varType:   "list(string)",
 			override:  "hi,there",
 			argValue:  `join("-", FOO)`,
 			wantValue: "hi-there",
-		},
-		// pseudo-documentation that this is a scenario that might be expected to work,
-		// but will parse as a valid CSV in (usually) an undesirable way
-		{
-			name:      "pseudo-JSON list of unquoted strings",
-			varType:   "list(string)",
-			override:  `[hi,there]`,
-			argValue:  `join("-", FOO)`,
-			wantValue: "[hi-there]",
 		},
 		{
 			name:      "CSV of unquoted strings okay",
@@ -1921,13 +1904,6 @@ func TestTypedVarOverrides(t *testing.T) {
 			wantValue: "hi-there",
 		},
 		{
-			name:      "JSON list of numbers",
-			varType:   "list(number)",
-			override:  "[3, 1, 4]",
-			argValue:  `join("-", [for v in FOO: v + 1])`,
-			wantValue: "4-2-5",
-		},
-		{
 			name:      "CSV list of numbers",
 			varType:   "list(number)",
 			override:  "3,1,4",
@@ -1935,11 +1911,12 @@ func TestTypedVarOverrides(t *testing.T) {
 			wantValue: "4-2-5",
 		},
 		{
-			name:      "JSON map of numbers",
-			varType:   "map(number)",
-			override:  `{"foo": 1, "bar": 2}`,
-			argValue:  `join("-", sort(values(FOO)))`,
-			wantValue: "1-2",
+			name:     "CSV set of numbers",
+			varType:  "set(number)",
+			override: "3,1,4",
+			// anecdotally sets are sorted but may not be guaranteed
+			argValue:  `join("-", [for v in sort(FOO): v + 1])`,
+			wantValue: "2-4-5",
 		},
 		{
 			name:      "CSV map of numbers",
@@ -1947,22 +1924,6 @@ func TestTypedVarOverrides(t *testing.T) {
 			override:  "foo:1,bar:2",
 			argValue:  `join("-", sort(values(FOO)))`,
 			wantValue: "1-2",
-		},
-		// though a JSON payload, any failure (types in this case) defers to
-		// CSV parsing and its error; not ideal and could be improved
-		{
-			name:     "invalid JSON map of numbers",
-			varType:  "map(number)",
-			override: `{"foo": "oops", "bar": 2}`,
-			// in lieu of something like ErrorMatches, this is the best single phrase
-			wantErrorMsg: "as CSV",
-		},
-		{
-			name:      "JSON tuple",
-			varType:   "tuple([number,string])",
-			override:  `[99, "bottles"]`,
-			argValue:  `format("%d %s", FOO[0], FOO[1])`,
-			wantValue: "99 bottles",
 		},
 		{
 			name:      "CSV tuple",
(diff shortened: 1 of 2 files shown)


bake/hcl_test.go
```diff
@@ -1645,6 +1646,328 @@ func TestHCLIndexOfFunc(t *testing.T) {
...
+		varType   string
+		wantError bool
+	}{
+		{
+			name:      "attribute literal which matches var type",
+			attrValue: `"hello"`,
+			varType:   "string",
+		},
+		{
+			name:      "attribute literal which coerces to var type",
```
tonistiigi, line 1791: This was surprising to me. Could we add error cases in here so that `99foo` and `split("1,2,3foo")` return errors.

bake/hcl_test.go
```diff
@@ -1645,6 +1646,328 @@ func TestHCLIndexOfFunc(t *testing.T) {
...
+		{
+			name:         "CSV of unquoted strings not okay",
+			varType:      "list(string)",
+			override:     `hi,there`,
+			wantErrorMsg: convertFailure,
+		},
+		{
+			name:      "JSON list of numbers",
+			varType:   "list(number)",
+			override:  "[3, 1, 4]",
```
tonistiigi, line 1914: 🤔 I though this would just be "3,1,4" in env (not in HCL itself).

tonistiigi (commented):
This looks good, but I would expect just csv values to work for env instead of `[`. Eg. `PLATFORMS=linux/amd64,linux/arm64 bake` instead of the JSON form. This looks like something that we can't change later so should be done as same PR.

bake/hcl_test.go
rrjjvv: Yes, in my first PR (esp. based on the code TODO), I had `3,1,4` as well.  For interactive use, that is certainly easier (and probably more intuitive).  The rationale I had for abandoning it was:
1. Though other complex types were not the goal, it was trivial to support them (and no reason to prohibit them), but I thought the inconsistency between lists (non-JSON) and other types (JSON only and no shortcuts) might be undesirable.
2. In my actual use case, it's (trivially) more work to produce a CSV; my overrides are actually data from a JSON API that's getting transformed via `jq`.  So the CSV is easier for interactive use, but could be clunkier for scripted use.
3. Ideally, even if CSV was allowed, it wouldn't prohibit the use of actual JSON, and wasn't sure if the additional complexity to support both would be welcome.

Do you agree that both should be supported?  And with CSV, are you good with rules like "every comma is a separator" (e.g., not dealing with escaped commas that are intended to be part of a string value)?

rrjjvv: > I don't see how both could be supported. 

Could be as simple as trying to parse it as JSON (accepting it as such as if it does), and if it fails, fall back to CSV.  Given `V='["hello","world"]'`, it's far more likely they intend a list containing `hello` and `world`, and not `["hello"` and `"world"]`.  If you don't object in principle, I'd like to take a stab at supporting both (and of course can be removed if you don't like the result).  But if you'd prefer to only handle CSV, I can do that.

> We have a csvvalue package

You mean `github.com/tonistiigi/go-csvvalue`?  I'll give it a go.  (I had previously looked at the parser within go-cty, but wouldn't have worked.)

bake/hcl_test.go
rrjjvv: I think the `99foo` scenario was actually covered in the existing "missing" test a few lines down, unless you were stressing the other combinations of coercion-friendly types.  Hopefully the other two tests capture the intent of your second request.

bake/hcl_test.go
rrjjvv: Let me know if you're good with the dual JSON/CSV support.  I basically duplicated the existing JSON tests and made CSV equivalents (and obviously altered the existing tests which previously failed with CSV).  I also added support for maps, since your library made it trivial (but can be ripped out if you think it's too much), and re-added tuples.

After running the tests with coverage, there's a few paths not covered that probably should (mostly errors), but didn't want to waste the time doing so if you asked for the JSON (and/or map/tuple) stuff to be removed.  But if you're good with the approach, I can add more negative tests to exercise the important error paths.

(Also, I kept this as a separate commit assuming you'd find it easier to review and that you/I would squash it later, but if you prefer any change to be squashed as if it's the final product, let me know.)

tonistiigi: I'm not sure about this. The string case at least is ambiguous, even if unlikely. `list(number)` doesn't have that issue though so that seems fine.

Random thought. What if there is a convention for `export VARIABLE_JSON=[]` that can be used to load the value from JSON form, assuming no `VARIABLE` env exists.

rrjjvv: My first reaction is "clunky", but very attractive in that it
1.  removes all ambiguity 
2. this code path is probably not considered performance critical, but would save on the double-parsing for CSV usage
3. similarly, the funkiness around error messages goes away (user intended to use JSON, messed up, and gets a CSV error)

I originally thought `VARIABLE_JSON` was a list of variables to be processed as JSON, but now see that `_JSON` was intended to be a suffix; it took too long for that last sentence to click.  Correct?  I've removed most of my original response that thought it was a list.
 
The only real issue I see would be in the off-chance *both* (`VARIABLE` and `VARIABLE_JSON`) were defined in the file.  The obvious behavior would be to use the environment variable as a CSV value for the respective bake variable, but only the author would know whether that's what they wanted.  Most of my other ideas in that same vein have problems of their own.  The best I can come up with is something like `VARIABLE_json`; it's less likely to be a defined name, and posix says
> The name space of environment variable names containing lowercase letters is reserved for applications.

but I don't know if mixed casing has implications for Windows users.

I would think that if a user is using JSON, they're likely using it everywhere.  I wouldn't think it unreasonable to say that JSON is all-or-nothing, i.e. `export BUILDX_BAKE_VARIABLES_AS_JSON=1`.  It may even be preferred.  (Of course, both could be supported as well.)  Taking it a step further, if it's a single flag, it could maybe be promoted to a switch (e.g., `--json-variables`), and the concept extended to the values used in `--set`, if that was ever useful/desired?

At this point, the CSV stuff is expressive enough that it serves my current and anticipated needs.  Though JSON is still my personal preference, it's a "nice to have" and nobody else is asking for it yet.  If JSON processing is an explicit opt-in, it can be tacked on later without compatibility worries, if you want to hold off making any UX decisions.

I think a dedicated configuration (whether through environment variable or CLI flag) would be cleaner, but I'm good with implementing your suggestion if you want to limit the footprint.  The limitations are hypothetical, the rules straightforward, and should be "good enough" for the vast majority of users.  Let me know.

tonistiigi: > The only real issue I see would be in the off-chance both (VARIABLE and VARIABLE_JSON) were defined in the file. 

I think if this happens, then they should be there to define the same logical value anyway. And in that case, if the user needs this capability, we would even make their code simpler because of this built-in feature. If you like, we could also make this case just error or warn if we can detect that two such variables are defined. We can also make it longer to reduce the chance of accidental collision, eg. `VARIABLE_AS_JSON`, `VARIABLE_FROM_JSON`.

> export BUILDX_BAKE_VARIABLES_AS_JSON=1. It may even be preferred. (Of course, both could be supported as well.) Taking it a step further, if it's a single flag, it could maybe be promoted to a switch (e.g., --json-variables),

I don't think the global flag makes sense as bake can work with multiple inputs, possibly written by separate people with different styles. `--json-variables` I think is too much extra logic for a rarely used addon of the variables feature.

I'm ok with either special suffix per variable for loading JSON, removing JSON support completely (for now) or only supporting JSON when the input is not ambiguous.

rrjjvv: > can work with multiple inputs, possibly written by separate people

Great point; I was too tunnel-visioned on my own use-case.

> I'm ok with either special suffix per variable for loading JSON

We'll go with that, with the suffix you suggested.  It's straightforward, explicit, and can always be augmented in the future if needed.

tonistiigi (changes requested):
https://github.com/docker/buildx/pull/3167#discussion_r2076424973

rrjjvv:
I have the new behavior all wired in and it's looking good.  I'll push them tonight after revisiting the tests once more to ensure I didn't miss anything.

bake/hcl_test.go
```diff
@@ -1645,6 +1646,692 @@ func TestHCLIndexOfFunc(t *testing.T) {
...
+				require.ErrorContains(t, err, "FOO default value")
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestTypedVarOverrides(t *testing.T) {
+	const unsuitableValueType = "Unsuitable value type"
```
rrjjvv, line 1842: The number of negative tests has increased substantially.  Most of the existing tests didn't 'care' about the kind of error, just that it happened.  I originally followed that pattern until I found many of my tests were passing, but for the wrong reasons.  I haven't seen much use of typed errors, so I didn't either., even though the tests would likely be clearer and have less churn.  Let me know if you're okay or prefer I've been doing it (string-based matching).

bake/hcl_test.go
```diff
@@ -1645,6 +1646,692 @@ func TestHCLIndexOfFunc(t *testing.T) {
...
+            }`)
+		// may seem reasonable, but not supported
+		t.Setenv("foo_json", "9000")
+		c, err := ParseFile(dt, "docker-bake.hcl")
+		require.NoError(t, err)
+		require.Equal(t, 1, len(c.Targets))
+		// a variable with no value has always resulted in an empty string
+		require.Equal(t, "", *c.Targets[0].Args["bar"])
+
+		t.Setenv("foo_JSON", "42")
```
rrjjvv, line 2327: Are there implications for Windows here (whether the test itself, or the functionality)?

bake/hcl_test.go
rrjjvv: > What if there is a convention for export VARIABLE_JSON=[] that can be used to load the value from JSON form, assuming no VARIABLE env exists.

At the time I didn't notice that was ambiguous (whether `VARIABLE` meant an environment variable or bake variable).  I wrote some tests specific to each, but TL;DR `VARIABLE_JSON` will be not be used as an override if that bake variable exists, but _will_ be used if that environment variable exists (and no bake variable).  My rationale for the latter is that JSON is only used if the variable carries typing, which is new.  Today, if someone happens to have `VARIABLE_JSON` in their environment, it is not used by bake.  After this functionality is released, that will still hold true, unless they explicitly add the `type` attribute to their bake file.

Hopefully I'm in tune with how you're thinking about this.  That said, there's still a sliver of risk related to what I mentioned in my PR description (about people that may be using `type` for their own purposes since it's not disallowed).  At least one person (https://github.com/docker/buildx/discussions/2398#discussioncomment-9779041) has attempted to use typing, expecting/hoping it would work like Terraform.  So *if* they kept the typing there, even though it had no effect, and *if* they have the relevant `_JSON` environment variables populated, they could experience a change in behavior.

That's a lot of 'ifs', but not impossible.  Let me know if you're onboard with my thought process or if you'd like it limited further.

bake/hcl_test.go
tonistiigi: Windows env are case insensitive. It is fine if `_json` works on windows but doesn't in unix.

tonistiigi (commented):
Apart from test failure this looks good to me.

crazy-max (commented):
Looks good thanks!

We should update the documentation to reflect this new type constraint.

In reference docs https://github.com/docker/buildx/blob/master/docs/bake-reference.md#variable we could have something similar to https://developer.hashicorp.com/terraform/language/values/variables#type-constraints

And in manual docs https://github.com/docker/docs/blob/main/content/manuals/build/bake/variables.md we could have some examples like https://developer.hashicorp.com/terraform/language/expressions/types#types

rrjjvv:
> We should update the documentation to reflect this new type constraint.

Does "we" mean me?  I can give it a shot.  (I didn't know whether you folks had dedicated technical writers and/or if docs followed a separate release process, so figured I'd wait for clarity.)

tonistiigi (approved):

PR #3179, opened 2025-05-13
ls: make sure current builder is available in JSON output

### LINKED ISSUE #3175: `buildx ls --format=json` missing "selected" field
### Contributing guidelines

### Description

In cli output of `docker buildx ls` an asterisk (*) next to a builder name indicates the selected builder.
In a json output, `docker buildx ls --format=json`, there is no such field, and there is no way to determine the selected one.

### Expected behaviour

Json output has a field indicating if a builder is a selected one (see line 3).
```
{
  "Name": "default",
  "Selected": true,
  "Driver": "docker",
  "LastActivity": "2025-01-01T00:00:00Z",
  "Dynamic": false,
  "Nodes": [ ... ]
}
```

### Actual behaviour

Json output does not have a "Selected" field in any way.
```
{
  "Name": "default",
  "Driver": "docker",
  "LastActivity": "2025-01-01T00:00:00Z",
  "Dynamic": false,
  "Nodes": [ ... ]
}
```

### Buildx version

github.com/docker/buildx v0.23.0 28c90ea

### Docker info

```text

```

### Builders list

```text
NAME/NODE        DRIVER/ENDPOINT                   STATUS    BUILDKIT   PLATFORMS
multiarch        docker-container
 \_ multiarch0    \_ ssh://user@192.168.1.1        running   v0.20.2    linux/amd64 (+2), linux/386
 \_ multiarch1    \_ unix:///var/run/docker.sock   running   v0.20.2    linux/arm64
default*         docker
 \_ default       \_ default                       running   v0.21.0    linux/arm64
```

### Configuration

```dockerfile
FROM scratch
```

### Build logs

```text

```

### Additional info

_No response_

tonistiigi:
fixes #3175

While lsBuilder has a field called Current that gets lost because the embedded struct implements custom MarshalJSON method.

This was the best workaround hack I could come up with. Didn't know about this behavior, but here is the sample case https://go.dev/play/p/XH8oRG_wGEU

### DIFF
--- a/commands/ls.go
+++ b/commands/ls.go
@@ -213,7 +213,17 @@ type lsContext struct {
 }
 
 func (c *lsContext) MarshalJSON() ([]byte, error) {
-	return json.Marshal(c.Builder)
+	// can't marshal c.Builder directly because Builder type has custom MarshalJSON
+	dt, err := json.Marshal(c.Builder.Builder)
+	if err != nil {
+		return nil, err
+	}
+	var m map[string]any
+	if err := json.Unmarshal(dt, &m); err != nil {
+		return nil, err
+	}
+	m["Current"] = c.Builder.Current
+	return json.Marshal(m)
 }
 
 func (c *lsContext) Name() string {


commands/ls.go
```diff
@@ -213,7 +213,17 @@ type lsContext struct {
...
+	dt, err := json.Marshal(c.Builder.Builder)
+	if err != nil {
+		return nil, err
+	}
+	var m map[string]any
+	if err := json.Unmarshal(dt, &m); err != nil {
+		return nil, err
+	}
+	m["Current"] = c.Builder.Current
+	return json.Marshal(m)
```
crazy-max, line 226: Oh interesting, this happens when marshaling structs that embed pointers. What do you think of this:

```go
func (c *lsContext) MarshalJSON() ([]byte, error) {
	type Alias builder.Builder
	return json.Marshal(&struct {
		*Alias
		Current bool
	}{
		Alias:   (*Alias)(c.Builder.Builder),
		Current: c.Builder.Current,
	})
}
```

tonistiigi: This doesn't work though. If you the original `Builder.MarshalJSON` does not get called anymore. https://go.dev/play/p/eJgDvB1mVAh

crazy-max: Yes indeed

crazy-max (approved):

PR #3189, opened 2025-05-19
Add variable typing to reference docs

rrjjvv:
This documents the variable typing introduced in #3167.

### DIFF
--- a/docs/bake-reference.md
+++ b/docs/bake-reference.md
@@ -1081,6 +1081,7 @@ or interpolate them in attribute values in your Bake file.
 
 ```hcl
 variable "TAG" {
+  type = string
   default = "latest"
 }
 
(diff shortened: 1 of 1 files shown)


docs/bake-reference.md
```diff
@@ -1102,6 +1103,202 @@ overriding the default `latest` value shown in the previous example.
...
+To use JSON, simply append `_JSON` to the variable name.
+In this contrived example, CSV cannot handle the second value; despite being a supported CSV type, JSON must be used:
+```hcl
+variable "VALS" {
+  type = list(string)
+  default = ["some", "list"]
+}
+```
+```console
+$ VALS_JSON=$(jo -a hello with,comma 'with"quote') docker buildx bake
```
crazy-max, line 1254: can we use the canonical form to make it more explicit?

```suggestion
$ VALS_JSON='["hello","with,comma","withquote"]' docker buildx bake
```

docs/bake-reference.md
```diff
@@ -1102,6 +1103,202 @@ overriding the default `latest` value shown in the previous example.
...
+
+# Sets *only* variable FOO_JSON; FOO is untouched
+$ FOO_JSON=bar docker buildx bake <...>
+
+# This also sets FOO_JSON, but will fail due to not being valid JSON
+$ FOO_JSON_JSON=bar docker buildx bake <...>
+
+# These are all equivalent
+$ FOO_JSON_JSON='"bar"' docker buildx bake <...>
+$ FOO_JSON_JSON=$(jo -a bar) docker buildx bake <...>
```
crazy-max, line 1292: Same here

```suggestion
$ FOO_JSON_JSON='["bar"]' docker buildx bake <...>
```

docs/bake-reference.md
rrjjvv: How about if I did something like
```sh
$ cat data.json
["hello","with,comma","with\"quote"]
$ VALS_JSON=$(< data.json) docker buildx bake
```

In the previous section for CSV I had suggested that CSV was (generally) better for interactive use, while JSON better for machine use.  And also, that JSON was the answer for unusual data.  I know `jo` may not be as familiar to readers as `jq`, so not immediately clear what I was trying to convey.

If I have that data coming from a file, that's a little more in the spirit of why this processing mode exists.  Whether this file came from an API, generated from another tool, etc. doesn't really matter.

If you still prefer it the way you suggested, that's fine.  But usually one wouldn't handcraft JSON like that, which is why I avoided it as well.

rrjjvv:
Even though this is a reference doc, it seemed like the amount of documentation devoted to it should be proportional to the anticipated frequency of use and its importance to building stuff.  That resulted in total omission of some things (like the `any` type), trying to pack many multiple concepts/illustrations into some examples, and not showing real execution outputs like is done elsewhere.  Let me know if that was a bad assumption or if I made the wrong compromises.

Questions/issues that have come up between the initial implementation and these docs:
1. I don't remember seeing it officially documented, but I stumbled on a feature where `--set` had the capability to append values (via `+=`).  It would be a logical feature for these new overrides as well.  Do you see that as a "wait and see if anybody asks"?
2. The issue of typed variables with no override and no default values, previously unconsidered but brought up in https://github.com/docker/buildx/pull/2530#issuecomment-2892803029.  Personally, this feels like a bug (due to oversight).  Should the docs call out this (probably) non-intuitive behavior?  Should it be 'fixed'?  Or left as undefined behavior (probably not a good idea).
3. That discussion led to whether wondering if `null` makes sense as an override value.  There are neither tests nor docs around whether that's supported or what the behavior is.  So very similar questions... should it defined?  Explicitly documented?  Fixed (if applicable)?

docs/bake-reference.md
crazy-max: > while JSON better for machine use

Oh ok makes sense then we can keep it

crazy-max:
> Even though this is a reference doc, it seemed like the amount of documentation devoted to it should be proportional to the anticipated frequency of use and its importance to building stuff. That resulted in total omission of some things (like the `any` type), trying to pack many multiple concepts/illustrations into some examples, and not showing real execution outputs like is done elsewhere. Let me know if that was a bad assumption or if I made the wrong compromises.

I think current changes you made in reference docs are good. If we want more specific use cases that are out of scope for reference docs, we can define them in https://github.com/docker/docs/blob/main/content/manuals/build/bake/variables.md

> 1\. I don't remember seeing it officially documented, but I stumbled on a feature where `--set` had the capability to append values (via `+=`).  It would be a logical feature for these new overrides as well.  Do you see that as a "wait and see if anybody asks"?

As `--set` is a bake flag it's defined in the cli reference docs: https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md#set

> 2\. The issue of typed variables with no override and no default values, previously unconsidered but brought up in https://github.com/docker/buildx/pull/2530#issuecomment-2892803029.  Personally, this feels like a bug (due to oversight).  Should the docs call out this (probably) non-intuitive behavior?  Should it be 'fixed'?  Or left as undefined behavior (probably not a good idea).

> 3\. That discussion led to whether wondering if `null` makes sense as an override value.  There are neither tests nor docs around whether that's supported or what the behavior is.  So very similar questions... should it defined?  Explicitly documented?  Fixed (if applicable)?

Yes agree this feels like a bug. As discussed offline with @tonistiigi, we can have a either a quick patch if possible or just document it if you feel it needs more thoughts.

rrjjvv:
>  we can have a either a quick patch if possible

Here's my proposal:
1. In *all* cases where typing is specified, `default = null`, and no override is provided, the value is `null`.  This is mostly current behavior, with one exception.
2. In *all* cases where typing is specified, and neither a default nor override is provided, the value is `null`.  Using a zero-value might be more intuitive and convenient, but `null` is probably more correct and is a simple implementation.  I'm fine either way, but this is what I currently have.
3. Leave the behavior of the `any` type as undefined, or even disallowed.

The current exception for (1) is the `any` type.  Currently, `type = any; default = null` yields an empty string.  That's because the type of something that was not defined is exactly the same as one defined as `any`, but the compatibility issue makes "not defined" return the empty string.  I've added that distinction, which allows (1) to be strengthened.  Might be clearer with this snippet:
```go
// before
typeSpecified = !varType.Equals(cty.DynamicPseudoType)
// after
typeSpecified = !varType.Equals(cty.DynamicPseudoType) || hcl.ExprAsKeyword(vr.Type) == "any"
```
That change broke existing tests asserting the 'missing'/'any' equivalence.  That's why it might make sense to simply ignore `any` (as I did in the docs), or actively disallow it (until there are use-cases to help guide how it should behave).  While I can see value in using `any` to type function arguments, it's hard to imagine being used for standard variables.

Based on previous comments, I'm guessing you'll be good with (1) and (3), and at least not disagree with (2), so I'll proactively proceed to create a separate PR (but obviously let me know if you disagree or have other proposals).

I'll sidestep the "null as an override" question for the time being; if it even makes sense, and if it doesn't behave as expected, I'd think that would be a simple bug/request; the issue around (2) is much bigger than that.

crazy-max (approved):

PR #3198, opened 2025-05-20
Consider typed, value-less variables to have `null` value

### LINKED ISSUE #3160: Bake: allow override of list variables via environment

rrjjvv:
Resolves #3160
Addresses issues discovered in https://github.com/docker/buildx/pull/2530#issuecomment-2892803029 and further discussed in #3189 

A variable with a type but no default value or override resulted in an empty string.  This matches the legacy behavior of untyped variables, but does not make sense when using types (an empty string is itself a type violation for everything except `string`).  All variables defined with a type but with no value are now a typed `null`.

A variable explicitly typed `any` was previously treated as if the typing was omitted; with no defined value or override, that resulted in an empty string.  The `any` type is now distinguished from an omitted type; these variables, with no default or override, are also `null`.

In other respects, the behavior of `any` is unchanged and largely behaves as if the type was omitted.  It's not clear whether it should be supported, let alone how it should behave, so these tests were removed.
It's being treated as undefined behavior.

### DIFF
--- a/bake/hcl_test.go
+++ b/bake/hcl_test.go
@@ -423,6 +423,63 @@ func TestHCLNullVariables(t *testing.T) {
 	require.Equal(t, ptrstr("bar"), c.Targets[0].Args["foo"])
 }
 
+func TestHCLTypedNullVariables(t *testing.T) {
+	types := []string{
+		"any",
+		"string", "number", "bool",
+		"list(string)", "set(string)", "map(string)",
+		"tuple([string])", "object({val: string})",
+	}
+	for _, varType := range types {
+		tName := fmt.Sprintf("variable typed %q with null default remains null", varType)
+		t.Run(tName, func(t *testing.T) {
+			dt := fmt.Sprintf(`
+	            variable "FOO" {
+	                type = %s
+	                default = null
+	            }
+	
+	            target "default" {
+	                args = {
+	                    foo = equal(FOO, null)
+	                }
+	            }`, varType)
+			c, err := ParseFile([]byte(dt), "docker-bake.hcl")
+			require.NoError(t, err)
+			require.Equal(t, 1, len(c.Targets))
+			require.Equal(t, "true", *c.Targets[0].Args["foo"])
+		})
+	}
+}
+
+func TestHCLTypedValuelessVariables(t *testing.T) {
+	types := []string{
+		"any",
+		"string", "number", "bool",
+		"list(string)", "set(string)", "map(string)",
+		"tuple([string])", "object({val: string})",
+	}
+	for _, varType := range types {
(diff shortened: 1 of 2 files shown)


bake/hcl_test.go
```diff
@@ -423,6 +423,63 @@ func TestHCLNullVariables(t *testing.T) {
...
+		tName := fmt.Sprintf("variable typed %q with no default is null", varType)
+		t.Run(tName, func(t *testing.T) {
+			dt := fmt.Sprintf(`
+                variable "FOO" {
+                    type = %s
+                }
+
+                target "default" {
+                    args = {
+                        foo = equal(FOO, null)
```
tonistiigi, line 472: just making sure that we do have a test that checks that without the `type` the value is empty string?

tonistiigi (approved):
I was considering if `type=bool` should default to false, but I think just keeping in `null` for all makes it most consistent.

bake/hcl_test.go
rrjjvv: It doesn't appear so. 
There's one that verifies it's not an error condition (ignoring the value): https://github.com/docker/buildx/blob/ea2b7020a4645bff395eb49e4e87ef08ba24eb93/bake/hcl_test.go#L1568-L1576

But I actually think this one indirectly does so: https://github.com/docker/buildx/blob/ea2b7020a4645bff395eb49e4e87ef08ba24eb93/bake/bake_test.go#L1989-L2003
I'm fairly sure that equality checks will not invoke coercion.  I know the validation does (one of its defining features), but don't know whether it would coerce a `null` to an empty string.  Seems unlikely, but can't say for sure.

Do you want an explicit one just to be safe?

rrjjvv:
> I was considering if type=bool should default to false

Yeah, I struggled with that.  A case could be made that primitives should get zero values (empty string, false, zero), but it really feels that it should be all or nothing... all null, or all zero values.  However, all of our attention and tests are around build args, where a `null` isn't valid (and why this PR isn't super friendly, even if correct).  But are there other bake attributes that legitimately accept a `null`?  That answer might solidify whether the `null` approach will be a usability issue or be useful in its own right.

bake/hcl_test.go
crazy-max: > Do you want an explicit one just to be safe?

Yes would be good

crazy-max: Pushed extra commit to add extra test

crazy-max (approved):

tonistiigi (approved):

Issue #3172, opened 2025-05-08
Allow passing env vars from host using `--env` like podman


ion-elgreco:
### Description

Podman provides a very nice way to pass environment variables of the host system to the build step, by doing --env "VERY_NICE_VAR", this is particularly useful where you don't want to hardcode the required env vars in the build stage. See podman docs:

```
Add a value (e.g. env=value) to the built image. Can be used multiple times. If neither = nor a value are specified, but env is set in the current environment, the value from the current environment is added to the image. To remove an environment variable from the built image, use the --unsetenv option
```
https://docs.podman.io/en/stable/markdown/podman-build.1.html#env-env-value

crazy-max:
This is already supported with the [`--build-arg` flag](https://docs.docker.com/reference/cli/docker/buildx/build/#build-arg):

```
docker buildx build --build-arg VERY_NICE_VAR .
```

PR #3203, opened 2025-05-23
3172 - Allow passing env vars with set for docker compose cases

solvingj:
Rough draft.
Adds support for using existing `set` flag to pass individual env vars through to docker-compose.yaml files:
```
docker buildx bake  --set "env.FOO=BAR"
```
- Related: 
- #3172 
- #1135

Note: Does not add `--env-file` flag which was original plan, only because not sure if acceptable to add a compose-specific flag to the general command as in:

```
docker buildx bake  --env-file=path`
```

If maintainers say thats ok, it's easy to add.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -181,7 +181,7 @@ func readWithProgress(r io.Reader, setStatus func(st *client.VertexStatus)) (dt 
 }
 
 func ListTargets(files []File) ([]string, error) {
-	c, _, err := ParseFiles(files, nil)
+	c, _, err := ParseFiles(files, nil, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -196,7 +196,7 @@ func ListTargets(files []File) ([]string, error) {
 }
 
 func ReadTargets(ctx context.Context, files []File, targets, overrides []string, defaults map[string]string, ent *EntitlementConf) (map[string]*Target, map[string]*Group, error) {
-	c, _, err := ParseFiles(files, defaults)
+	c, _, err := ParseFiles(files, overrides, defaults)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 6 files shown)


crazy-max:
> Adds support for using existing `set` flag to pass individual env vars through to docker-compose.yaml files:
> 
> ```
> docker buildx bake  --set "env.FOO=BAR"
> ```

`env` is not a defined attribute in a bake definition. Also you can already set/override an env var or [HCL variable](https://docs.docker.com/build/bake/reference/#variable) with just:

```
FOO=BAR docker buildx bake
```

So not sure why we would need this.

> - Related: 
> - #3172 
> - #1135

It does not seems to relate to what is suggested in these issues.
* #3172 is for loading an env during `build` which is already supported with [`--build-arg` flag](https://docs.docker.com/reference/cli/docker/buildx/build/#build-arg)
* #1135 is to load additional env files for compose.

> Note: Does not add `--env-file` flag which was original plan, only because not sure if acceptable to add a compose-specific flag to the general command as in:
> 
> ```
> docker buildx bake  --env-file=path`
> ```

Yes this should be for another PR anyway. We support loading `.env`: https://github.com/docker/buildx/blob/e3c6618db27d052ef46a9442a39bb46dc27ffb02/bake/compose.go#L252-L255 but not additional ones. I think in such case we could support env files with `-f` flag.

solvingj:
Indeed I did not actually expect this PR to be viable as-is. At the end, it was clear it was an un-intended and non-conforming use of `set`, and the function signature changes were unreasonable.  Nonetheless, I wanted submit it as a show of effort and try something to move the conversation forward. 

>  I think in such case we could support env files with -f flag.

Thank you, I think this is mission accomplished. 

But, since `-f` is already used for `Build definition file`, would `--env-file` not be preferable? 

> FOO=BAR...

Indeed this handles the common case for posix, but there are a few ancillary reasons various tools (`docker run` for example) add the explicit flag for environment variables.  One is to support uniform API (CLI syntax) between posix and windows, another is to provide a fully controlled and better testable path for all core behaviors through the CLI. If we can get --env-file in it solves our problems, but it's food for thought for future support of a `-e` flag.

PR #3207, opened 2025-05-27
Show types during variable list operation

### LINKED ISSUE #3202: bake: Show variable types during variable list operation
### Description

The new typing feature introduced in #3167 should be part of the output of `--list variables`.

I assume this is desirable and already have an initial implementation, but there are enough questions that I thought I should open an issue first to hash it out.
1. For the (default) table view, it's generally the case that the structure/layout isn't guaranteed, but I didn't want to assume.  Are you okay with adding it to the table?  If so, can it go in the column that 'makes sense', or should it be last?  (I assumed that modifying the table layout was fine and that the JSON format is a non-issue)
2. I noticed that the 'value' field will *only* populate for values that are `string` or `bool` (with `number` being omitted).  Was this oversight or intentional?  (I assumed the former and now show numbers)
3. Is there any reason complex values should not be displayed?  (Since I assumed lack of `number` was not intentional, I assumed showing both numbers and complex types would be desirable).
4. Currently string values are not shown with quotes.  If we start showing numbers, it will be ambiguous whether a value that looks like a number was intended to be a string or an actual number..  Is that a concern?  (I've left strings unquoted, assuming strings are far more prevalent thus quotes would be unwelcome noise)
5. Same question, but for complex types (e.g., values for `list(string)` vs. `list(number)`).  (I've kept quotes in this case.)
6. I assumed a 'type' field should only be populated when an explicit type was provided, leaving it empty otherwise (i.e., not populating an inferred type).

For a visual example of all these put together, this is the output from a slightly doctored project using `v0.24.0`:
```
VARIABLE	VALUE	DESCRIPTION
CI		<null>	do not set; will always be 'true' in Jenkins
EXPORT_DIR	./out	for local file/tar exports
EXPORT_TARS		whether to export images as tarballs (for tests)
JAVA_VERSIONS		java LTS versions to build for
MAVEN_VERSIONS		maven versions to build for
PREFIX			
REGISTRY		registry to push to (for CI)
REPOSITORY		repository name (for CI)
TAG		latest	default tag to apply to all generated images
docker_gid		
flavor		debian	OS flavor
jenkins_gid		
jenkins_uid		
untyped_number		
untyped_string	99
```
and the output from my proposed/current changes:
```
VARIABLE	TYPE		VALUE		DESCRIPTION
CI		string		<null>		do not set; will always be 'true' in Jenkins
EXPORT_DIR	string		./out		for local file/tar exports
EXPORT_TARS	number		0		whether to export images as tarballs (for tests)
JAVA_VERSIONS	set of string	["17","21"]	java LTS versions to build for
MAVEN_VERSIONS	set of string	["3.9.8"]	maven versions to build for
PREFIX		string				
REGISTRY	string				registry to push to (for CI)
REPOSITORY	string				repository name (for CI)
TAG		string		latest		default tag to apply to all generated images
docker_gid	number		5004		
flavor		string		debian		OS flavor
jenkins_gid	number		597		
jenkins_uid	number		596		
untyped_number			99		
untyped_string			99
```

Let me know what you think, or if you'd rather just work off of an actual PR, I can do that as well.

rrjjvv:
If a type was explicitly provided, it will be displayed in the variable listing. Inferred type names are not displayed, as they likely would not match the user's intent.

Previously only `string` and `bool` default values were displayed in the listing.  All default values, regardless of type, are now displayed.

Fixes #3202

### DIFF
--- a/bake/hclparser/hclparser.go
+++ b/bake/hclparser/hclparser.go
@@ -38,6 +38,9 @@ type variable struct {
 	Validations []*variableValidation `json:"validation,omitempty" hcl:"validation,block"`
 	Body        hcl.Body              `json:"-" hcl:",body"`
 	Remain      hcl.Body              `json:"-" hcl:",remain"`
+
+	// the type described by Type if it was specified
+	constraint *cty.Type
 }
 
 type variableValidation struct {
@@ -296,6 +299,9 @@ func (p *parser) resolveValue(ectx *hcl.EvalContext, name string) (err error) {
 			return diags
 		}
 		typeSpecified = !varType.Equals(cty.DynamicPseudoType) || hcl.ExprAsKeyword(vr.Type) == "any"
+		if typeSpecified {
+			vr.constraint = &varType
+		}
 	}
 
 	if def == nil {
@@ -674,6 +680,7 @@ func (p *parser) validateVariables(vars map[string]*variable, ectx *hcl.EvalCont
 type Variable struct {
 	Name        string  `json:"name"`
 	Description string  `json:"description,omitempty"`
+	Type        string  `json:"type,omitempty"`
 	Value       *string `json:"value,omitempty"`
 }
 
@@ -803,13 +810,31 @@ func Parse(b hcl.Body, opt Opt, val any) (*ParseMeta, hcl.Diagnostics) {
 			Name:        p.vars[k].Name,
 			Description: p.vars[k].Description,
 		}
+		tc := p.vars[k].constraint
+		if tc != nil {
+			v.Type = tc.FriendlyNameForConstraint()
+		}
 		if vv := p.ectx.Variables[k]; !vv.IsNull() {
 			var s string
-			switch vv.Type() {
-			case cty.String:
-				s = vv.AsString()
-			case cty.Bool:
-				s = strconv.FormatBool(vv.True())
+			switch {
+			case tc != nil:
+				if bs, err := ctyjson.Marshal(vv, *tc); err == nil {
+					s = string(bs)
+					// untyped strings were always unquoted, so be consistent with typed strings as well
+					if tc.Equals(cty.String) {
+						s = strings.Trim(s, "\"")
+					}
+				}
+			case vv.Type().IsPrimitiveType():
+				// all primitives can convert to string, so error should never occur
+				if val, err := convert.Convert(vv, cty.String); err == nil {
+					s = val.AsString()
+				}
+			default:
+				// must be an (inferred) tuple or object
+				if bs, err := ctyjson.Marshal(vv, vv.Type()); err == nil {
+					s = string(bs)
+				}
 			}
 			v.Value = &s
 		}
--- a/tests/bake.go
+++ b/tests/bake.go
@@ -74,6 +74,7 @@ var bakeTests = []func(t *testing.T, sb integration.Sandbox){
 	testBakeLoadPush,
 	testListTargets,
 	testListVariables,
+	testListTypedVariables,
 	testBakeCallCheck,
 	testBakeCallCheckFlag,
 	testBakeCallMetadata,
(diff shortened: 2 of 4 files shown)


bake/hclparser/hclparser.go
```diff
@@ -803,13 +810,31 @@ func Parse(b hcl.Body, opt Opt, val any) (*ParseMeta, hcl.Diagnostics) {
...
+			case tc != nil:
+				if bs, err := ctyjson.Marshal(vv, *tc); err == nil {
+					s = string(bs)
+					// untyped strings were always unquoted, so be consistent with typed strings as well
+					if tc.Equals(cty.String) {
+						s = strings.Trim(s, "\"")
+					}
+				}
+			case vv.Type().IsPrimitiveType():
+				// all primitive's can convert to string, so error should ever occur anyway
```
tonistiigi, line 829: nit: no `'`, `/s/ever/never/`

tests/bake.go
```diff
@@ -1737,7 +1738,93 @@ target "default" {
...
+    type = map(string)
+    default = {"foo": "bar"}
+}
+variable "vwx" {
+    type = set(bool)
+    default = null
+    description = "collection, null default"
+}
+
+// untyped, but previously didn't have its value output
```
tonistiigi, line 1783: seem to be missing a case with just variable block without type and default, and implied string cases with default.

I think it would also be ok to show the type even if it is inferred from default, as long as it is not a string. But if it complicated things then we can skip that.

tonistiigi: Why is it detected as "tuple" in this case? I would guess that an array is more generic and a tuple is a subset of array.

rrjjvv: > Why is it detected as "tuple" in this case? I would guess that an array is more generic and a tuple is a subset of array.

The rationale is probably "since I can only see two elements both of which are strings, I can't guarantee you can safely add/remove elements at will".  But more explicitly from [the docs](https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md#collection-values):
> Only tuple and object values can be directly constructed via native syntax. Tuple and object values can in turn be converted to list, set and map values with other operations, which behaves as defined by the syntax-agnostic HCL information model.

tonistiigi: The way I'm looking at it is that if it is really meant as "VERSION_PAIR", then user should explicitly define it as a tuple, to make it clear that only (for example) 2 elements are possible and no less and no more. If it is just defined as two values in a list, then I wouldn't assume that this is the only option.

tonistiigi: I think let's leave it with 1 then, unless there is some simple trick to make 4 possible

rrjjvv: In short, I think we'll have to go with (1); I'm not sure we'll find any truly simple solution for (4).

In hindsight, I should have never started batting around "inferred type".  Hopefully I'm not responsible for any confusion.  If we were truly dealing with inferred types (types inferred by how/where they're used), we probably wouldn't be having this discussion (to this extent at least).  An *actual* inferred type would be the solution you're hoping for.  We're just dealing with the simplest type that can be guaranteed to represent a specific value.

I'll take another stroll through the docs and code to see if there's a capability there that I missed, but I'm not hopeful.  Probably the best we'll get is some insight or tools to make our own heuristics simpler/safer to implement.  I'll let you know what I find.

rrjjvv: From what I can see, there is no general solution for (4) that is also simple.  But if you're willing to sacrifice correctness for the 1% intending to use tuples for the sake of the 99% using lists (an earlier comment suggests you might), there could be a 'simple' (in terms of bake code) solution.

Via the cty "unification" functions, I was basically able to say "I have `["hi", "there"]`; can that be made into a `list` of some type?"  And it will (not just whether it can be done, but exactly what type).

But... given `["hello", 99]`, it says "it can be made into a `list(string)` (via coercion)".  Under normal circumstances, something that superficially looks like a list but contains elements of various types would usually be considered a tuple.  So increasing accuracy for list users may come at the expense of decreasing accuracy for tuple users.

There are two main variables in play specific to this functionality... 'safe' vs. 'unsafe', and whether `cty.DynamicPseudoType` is being utilized; I have not experimented with all combinations.


Let me know what you think... if you think this deserves more exploration, or if it's fine to leave as-is.

tonistiigi: Let's just leave it to 1. No need to overthink it.

rrjjvv: k.  I guess there's another flavor of (4)... rather than try guessing the complex types, we don't even bother trying and simply fill the column with "other", "unknown", "\<missing>", or similar.

That wouldn't provide any real value (though could be a subtle hint to the user that they _can_ provide it), but would make the output look a little nicer (no gaps), would be easy, and would not be inaccurate.

If you like that option and give me an exact term to use, I can get that implemented later.  Otherwise, I guess we're done?

PR #3208, opened 2025-05-27
Fix RegistryToken support

sztwiorok:
This Fix is related to issue [#2749](https://github.com/docker/buildx/issues/2749.)
I have tested this change and confirmed it resolves the issue


crazy-max (changes requested):
Needs to rebase, first commit should not be there.

PR #3209, opened 2025-05-27
Fix support for RegistryToken

sztwiorok:
This Fix is related to issue [#2749](https://github.com/docker/buildx/issues/2749)
I have tested this change and confirmed it resolves the issue

### DIFF
--- a/util/imagetools/inspect.go
+++ b/util/imagetools/inspect.go
@@ -134,6 +134,9 @@ func toCredentialsFunc(a Auth) func(string) (string, string, error) {
 		if ac.IdentityToken != "" {
 			return "", ac.IdentityToken, nil
 		}
+		if ac.RegistryToken != "" {
+			return "", ac.RegistryToken, nil
+		}
 		return ac.Username, ac.Password, nil
 	}
 }


tonistiigi (commented):
I don't think this is correct. iiuc `RegistryToken` means the token sent directly to registry, so if it would be supported then it would be used in the implementation of `FetchToken`. As these tokens would already be specific to repository and scope (for oauth) then it is unlikely that they would be very useful for build as builder pulls lots of images and would need separate tokens for different images and separate tokens for pull and push. If the current PR works for you then you should be defining your token as `IdentityToken` instead of `RegistryToken`.

tonistiigi:
Looking at it more, we do seem to support `RegistryToken` in buildkit side https://github.com/moby/buildkit/blob/master/session/auth/authprovider/authprovider.go#L109-L112 inside FetchToken as I suggested. But in here (imagetools) we use containerd authorizer directly and that doesn't seem to have support for it afaics. https://github.com/moby/buildkit/blob/master/session/auth/authprovider/authprovider.go#L109-L112 . If you do need `RegistryToken` and not `IdentityToken` then looks like https://github.com/containerd/containerd/blob/main/core/remotes/docker/authorizer.go#L277 needs to have a way to return that token directly without trying to fetch it.

tonistiigi (changes requested):
^

PR #3213, opened 2025-05-30
build: refactor some of the build functions into smaller utility functions

jsternberg:


build/build.go
```diff
@@ -138,75 +138,57 @@ func filterAvailableNodes(nodes []builder.Node) ([]builder.Node, error) {
...
+	} else {
+		warnNoOutputBuf.WriteString(fmt.Sprintf("for %s target(s) with %s driver", strings.Join(noOutputTargets, ", "), noMobyDriver.Factory().Name()))
 	}
+	logrus.Warnf("%s. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load", warnNoOutputBuf.String())
+}
 
+func newBuildRequests(ctx context.Context, docker *dockerutil.Client, cfg *confutil.Config, drivers map[string][]*resolvedNode, w progress.Writer, opts map[string]Options) (map[string][]*reqForNode, func(), error) {
 	reqForNodes := make(map[string][]*reqForNode)
-	eg, ctx := errgroup.WithContext(ctx)
+	var releasers []func()
```
tonistiigi, line 192: these releasers need to be released in defer on error

monitor/monitor.go
```diff
@@ -332,6 +344,27 @@ func (m *monitor) invoke(ctx context.Context, pid string, cfg *controllerapi.Inv
 	return err
 }
 
+func (m *monitor) Close() error {
+	m.cancelRunningProcesses()
+	// if m.buildConfig.resultCtx != nil {
```
tonistiigi: comment should be removed?

tonistiigi (commented):
The last commit "[build: add build.Handler type](https://github.com/docker/buildx/pull/3213/commits/ff1f68c6d324ea24c5de39f5ee3da1e6bff03692)" seems to not add anything to the current PR by just adding additional wrapping, and therefore outside the scope of the PR description. I think that one can be removed for now.

monitor/monitor.go
jsternberg: Done.

build/build.go
jsternberg: Done.

jsternberg:
I removed the last commit. I had just been using this PR to stack changes and it was related to something I haven't finished yet.

build/build.go
tonistiigi: That last change is under a wrong commit atm. Can you move it to correct one.

jsternberg: Moved this change to the correct commit.

tonistiigi (approved):

controller/local/controller.go
```diff
@@ -103,15 +88,6 @@ func (b *localController) Invoke(ctx context.Context, pid string, cfg *controlle
 	}
 }
 
-func (b *localController) Close() error {
-	b.cancelRunningProcesses()
-	if b.buildConfig.resultCtx != nil {
-		b.buildConfig.resultCtx.Done()
```
jsternberg: I think this section might actually be important so I'm going to add it back.

jsternberg: Yea, this is important to ensure the result context is properly released. It does appear like there might be an existing bug though. The `Build` method in the controller doesn't close the result context when it tries to rebuild. I'll have a future PR that refactors this so it's not worth fixing since that code will be deleted soon anyway.

tonistiigi (approved):

Issue #3174, opened 2025-05-09
Expand buildx exit codes to include buildkit-specific error classifications


thompson-shaun:
## Motivation
Currently, buildx doesn't provide granular exit codes that correspond to specific error types from buildkit. This makes it difficult to programmatically determine the exact nature of a failure, especially in CI/CD environments where automatic retry or fallback logic might depend on the specific type of error.

## Proposal
Expand the buildx exit code system to map buildkit's error classifications to unique exit codes. As an initial implementation, focus on the "internal" errors defined in the [moby/buildkit errdef package](https://github.com/moby/buildkit/tree/f2f3719ec0c5d4d9b7aba0e8bd9bd9bdce7b4024/errdefs).

## Follow-on work
- Expand codes to cover additional error categories/cases
- GHA handling

tonistiigi:
> As an initial implementation, focus on the "internal" errors defined in the [moby/buildkit errdef package](https://github.com/moby/buildkit/tree/f2f3719ec0c5d4d9b7aba0e8bd9bd9bdce7b4024/errdefs).

What exit code do you want?

> GHA handling

wdym?

thaJeztah:
Some similar ticket; 

- https://github.com/docker/cli/issues/1683

I'm not sure if it'll be very useful / easy to make such exit-codes actionable though; there's a limited number of exit-codes that _could_ be used, but without context, it may be hard to programmatically decide whether they can be retried (or otherwise), as many things could depend on context (if a command in a build-step fails, can it be retried? _which command failed_?).

thompson-shaun:
> What exit code do you want?

Likely a block that we could expand with that's not in conventional/reserved ranges so smth in the 80-99 range or 100-126 (whatever you feel is most appropriate) [ref](https://pkg.go.dev/github.com/square/exit#pkg-types). @thaJeztah has a nice breakdown of ranges in https://github.com/docker/buildx/issues/3174#issuecomment-2867490072:~:text=%5BRFC%5D%20standardise%20exit%2Dcodes%C2%A0cli%231683 as well. 

> 
> > GHA handling
> 
> wdym?

Scrapped this

> I'm not sure if it'll be very useful / easy to make such exit-codes actionable though; there's a limited number of exit-codes that could be used, but without context, it may be hard to programmatically decide whether they can be retried (or otherwise), as many things could depend on context (if a command in a build-step fails, can it be retried? which command failed?).

[...]

PR #3214, opened 2025-05-30
cmd: custom exit codes for internal, resource and canceled errors

tonistiigi:
fix https://github.com/docker/buildx/issues/3174

### DIFF
--- a/cmd/buildx/main.go
+++ b/cmd/buildx/main.go
@@ -14,10 +14,12 @@ import (
 	"github.com/docker/cli/cli-plugins/plugin"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/debug"
-	"github.com/moby/buildkit/solver/errdefs"
+	solvererrdefs "github.com/moby/buildkit/solver/errdefs"
+	"github.com/moby/buildkit/util/grpcerrors"
 	"github.com/moby/buildkit/util/stack"
 	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel"
+	"google.golang.org/grpc/codes"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 
@@ -99,7 +101,7 @@ func main() {
 		os.Exit(sterr.StatusCode)
 	}
 
-	for _, s := range errdefs.Sources(err) {
+	for _, s := range solvererrdefs.Sources(err) {
 		s.Print(cmd.Err())
 	}
 	if debug.IsEnabled() {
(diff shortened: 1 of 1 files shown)


crazy-max (approved):

cmd/buildx/main.go
```diff
@@ -113,5 +115,19 @@ func main() {
...
 
-	os.Exit(1)
+	exitCode := 1
+	switch grpcerrors.Code(err) {
+	case codes.Internal:
+		exitCode = 100 // https://github.com/square/exit/blob/v1.3.0/exit.go#L70
+	case codes.ResourceExhausted:
+		exitCode = 102
+	case codes.Canceled:
+		exitCode = 130
```
thaJeztah, line 125: ISTR 130 may also be used for containers, but perhaps thats `1 + 129` :thinking: https://docs.docker.com/reference/cli/docker/container/attach/#attach-to-and-detach-from-a-running-container


Looking at that square project linked above, that also looks to take 128+ into account for signals; is that a concern?
https://github.com/square/exit/blob/v1.3.0/exit.go#L92-L102

cmd/buildx/main.go
```diff
@@ -113,5 +115,19 @@ func main() {
 		ebr.Print(cmd.Err())
 	}
 
-	os.Exit(1)
+	exitCode := 1
+	switch grpcerrors.Code(err) {
+	case codes.Internal:
+		exitCode = 100 // https://github.com/square/exit/blob/v1.3.0/exit.go#L70
```
thaJeztah, line 121: The only thing with `Internal` errors that I wanted to comment on in the ticket (but didn't get round to) is that, at least in moby/moby, but not unlikely "elsewhere", we consider anything that's not explicitly assigned an `errdefs` type to be an internal server error;
https://github.com/moby/moby/blob/a8d36459572b6d50ab34341bd0c3d62ffa8c3e7b/api/server/httpstatus/status.go#L72


Will it be a problem if errors fall through that were not explicitly given a type to be `1` instead of `100` (or vice-versa)?

cmd/buildx/main.go
tonistiigi: 130 is 128 + SIGINT . Same as you get when you cancel a command in shell.

cmd/buildx/main.go
tonistiigi: In buildkit most errors are from executing user code or params/config somehow derived from user provided information. So only specific errors can be guaranteed to be internal if they actually represent some algorithm/storage/machine etc. error. What is internal and what is not is defined by BuildKit side so not directly related to this PR.

PR #3219, opened 2025-06-03
monitor: refactor how reload works

jsternberg:
The build now happens in a loop and the monitor is run after every
build. The monitor can return `ErrReload` to signal to the main thread
that it should reload the build result.

This will be used in the future to move the monitor into a callback
rather than as a separate existence. It allows the monitor to not
control the build itself which now makes it possible to completely
remove the controller.

### DIFF
--- a/monitor/monitor.go
+++ b/monitor/monitor.go
@@ -10,8 +10,6 @@ import (
 	"text/tabwriter"
 
 	"github.com/containerd/console"
-	"github.com/docker/buildx/build"
-	cbuild "github.com/docker/buildx/controller/build"
 	"github.com/docker/buildx/controller/local"
 	controllerapi "github.com/docker/buildx/controller/pb"
 	"github.com/docker/buildx/controller/processes"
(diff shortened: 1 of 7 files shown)


util/ioset/mux.go
```diff
@@ -255,3 +259,7 @@ type readerWithClose struct {
 func (r *readerWithClose) Close() error {
 	return r.closeFunc()
 }
+
+func isEOF(err error) bool {
```
tonistiigi: nit: `isReaderClosed()` or smtm

monitor/monitor.go
```diff
@@ -20,25 +19,23 @@ import (
...
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/term"
 )
 
-type MonitorBuildResult struct {
-	Resp *client.SolveResponse
-	Err  error
-}
+var ErrReload = stderrors.New("monitor: reload")
```
tonistiigi: why is this `stderrors`?

tonistiigi (commented):
two nits, but otherwise looks good and seems to work correctly

monitor/monitor.go
jsternberg: It's so it can be used as a sentinel value. I don't think the stacktrace included with `github.com/pkg/errors` is something that's very helpful for a sentinel value.

tonistiigi (approved):

PR #3220, opened 2025-06-03
monitor: move remaining controller functionality into monitor

jsternberg:
This creates a `Monitor` type that keeps the global state between
monitor invocations and allows the monitor to exist during the build so
it can be utilized for callbacks.

The result handler is now registered with the monitor during the build
and `Run` will use the result if it is present and the configuration
intends the monitor to be invoked with the given result.


commands/debug/root.go
```diff
@@ -38,27 +28,6 @@ func RootCmd(dockerCli command.Cli, children ...DebuggableCmd) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "debug",
 		Short: "Start debugger",
-		Args:  cobra.NoArgs,
```
tonistiigi, line 41: I guess without the remote controller this command doesn't make sense anymore (for now) without subcommands

controller/build/build.go
```diff
@@ -180,56 +180,40 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in *Options, inStream
...
-			mu.Lock()
-			defer mu.Unlock()
-			if res == nil || driverIndex < idx {
-				idx, res = driverIndex, gotRes
-			}
-		})
-	} else {
-		resp, err = build.Build(ctx, nodes, opts, dockerutil.NewClient(dockerCli), confutil.NewConfig(dockerCli), progress)
-	}
+func buildTargets(ctx context.Context, dockerCli command.Cli, nodes []builder.Node, opts map[string]build.Options, progress progress.Writer, bh *build.Handler) (*client.SolveResponse, error) {
```
tonistiigi, line 211: note for future: this function is weird as it takes map of options per target and then returns just the `defaultTarget` of that map.

PR #3222, opened 2025-06-04
controller: remove remaining parts of the controller

### LINKED ISSUE #3124: Proposal: Remove build controller

jsternberg:
Removes all references to the controller and moves the remaining
sections of code to other packages.

Processes has been moved to monitor where it is used and the data
structs have been removed so buildflags is used directly. The controller
build function has been moved to the commands package.

Fixes #3124.


bake/bake.go
```diff
@@ -1439,20 +1438,19 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 			})
 		}
 	}
-	secrets = secrets.Normalize()
-	bo.SecretSpecs = secrets.ToPB()
-	secretAttachment, err := controllerapi.CreateSecrets(bo.SecretSpecs)
+	bo.SecretSpecs = secrets.Normalize()
+	secretAttachment, err := build.CreateSecrets(bo.SecretSpecs)
```
tonistiigi, line 1442: I think eventually these two sets of types will probably also not be needed. We can just create the correct type from the start.

commands/build.go
```diff
@@ -1057,3 +1030,241 @@ func otelErrorType(err error) string {
...
+	if err != nil {
+		return nil, nil, WrapBuild(err)
+	}
+	if i, ok := buildOptions[defaultTargetName]; ok {
+		inputs = &i.Inputs
+	}
+	return resp, inputs, nil
+}
+
+// buildTargets runs the specified build and returns the result.
```
tonistiigi: This function seems not needed. It seems to have only one caller and just calls another function internally.

build/invoke.go
```diff
@@ -8,12 +8,41 @@ import (
 	"sync/atomic"
 	"syscall"
 
-	controllerapi "github.com/docker/buildx/controller/pb"
 	gateway "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
+type InvokeConfig struct {
```
tonistiigi, line 16: This should eventually move out of build pkg. InvokeConfig should be referred only by the specific `Evaluate` hook implementation and controlling the coordination between that hook and the monitor.

commands/build.go
jsternberg: Removed.

tonistiigi (approved):

PR #3224, opened 2025-06-05
build: change build handler to evaluate instead of onresult

jsternberg:
This changes the build handler to customize the behavior of evaluate
rather than onresult and also simplifies the `ResultHandle`. The
`ResultHandle` is now only valid within the gateway callback and can be
used to start containers from the handler.

`Evaluate` now executes inside of the gateway callback rather than
having a separate implementation that executes or re-invokes the build.
This keeps the gateway callback session open until the debugger has
returned.

The `ErrReload` for monitor has now been moved into the `build` package
and been renamed to `ErrRestart`. This is because it restarts the build
so the name makes a bit more sense. The actual use of this functionality
is still tied to the monitor reload.


build/build.go
```diff
@@ -504,24 +507,28 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
 						results.Set(rKey, res)
 
 						if children := childTargets[rKey]; len(children) > 0 {
-							if err := waitForChildren(ctx, res, results, children); err != nil {
+							if err := waitForChildren(ctx, bh, c, res, results, children); err != nil {
```
tonistiigi, line 514: Why is this connected to `waitForChildren`. Wouldn't it make more sense to have errgroup with one part going to `waitForChildren` and other to eval callback. Or can't the hook just run before?

build/build.go
```diff
@@ -504,24 +507,28 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
...
+								return nil, err
+							}
+						} else if bh != nil && bh.Evaluate != nil {
+							if err := bh.Evaluate(ctx, c, res); err != nil {
```
tonistiigi, line 518: In case of `ErrRestart` shouldn't we return `canceled` from the callback. Or maybe canceled wrapped with "build restarted by client".

jsternberg:
Added an additional commit to refactor the pause/unpause functionality for the printer. It seems the new evaluation method is causing some conflicts because it's possible for statuses to be sent while the printer is paused. This change should make that more reliable.

monitor/monitor.go
```diff
@@ -54,28 +51,32 @@ func New(cfg *build.InvokeConfig, stdin io.ReadCloser, stdout, stderr io.WriteCl
 
 func (m *Monitor) Handler() build.Handler {
 	return build.Handler{
-		OnResult: func(driverIndex int, gotRes *build.ResultHandle) {
-			m.mu.Lock()
-			defer m.mu.Unlock()
+		Evaluate: func(ctx context.Context, c gateway.Client, res *gateway.Result) error {
```
tonistiigi: As this doesn't use any local variables in the closure, move this to actual private method. Or you can just add `Evaluate` to `Monitor` and pass it as handler.

build/build.go
jsternberg: Is this for the build history?

build/build.go
jsternberg: The main reason is because `waitForChildren` calls `Evaluate` on the result so it needs the custom handler to invoke it. If there are children, we unconditionally invoke evaluate but if there are no children we only call evaluate if a handler has been given.

We talked about this offline yesterday, but we may want to look at logic in `waitForChildren` that was refactored from the previous code. It seems to call `results.Get` outside of the goroutine (blocking call) and then it creates a goroutine to call evaluate and `results.Get` again inside of the goroutine. It's unclear to me if the second `results.Get` is necessary. If it isn't, we can probably just remove this function entirely, remove the errgroup, and call evaluate outside of the errgroup.

One other piece of logic. Do we need to call evaluate in this code path or is that optional?

build/build.go
tonistiigi: Yes

jsternberg: Done.

monitor/monitor.go
jsternberg: Done.

PR #3233, opened 2025-06-11
imagetools: support registrytoken auth in docker config

tonistiigi:
replaces and closes #3209

@crazy-max @dmcgowan

### DIFF
--- a/util/imagetools/auth.go
+++ b/util/imagetools/auth.go
@@ -1,26 +1,68 @@
 package imagetools
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
+	"net/http"
+	"sync"
+	"time"
 
+	"github.com/containerd/containerd/v2/core/remotes/docker"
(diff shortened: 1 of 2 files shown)


util/imagetools/auth.go
```diff
@@ -0,0 +1,103 @@
...
+		return "", "", err
+	}
+	if ac.IdentityToken != "" {
+		return "", ac.IdentityToken, nil
+	}
+	return ac.Username, ac.Password, nil
+}
+
+func (a *authConfig) authConfig(host string) (types.AuthConfig, error) {
+	const defaultExpiration = 2 * time.Minute
```
crazy-max, line 46: Is it arbitrary?

tonistiigi: Yes, just needs to smaller than 5 min, because some credential helpers expire.

crazy-max (approved):

Issue #494, opened 2020-12-24
Support custom host-to-IP mappings in `bake` the same way it's done in `build --add-host`


unshare:
See subject.
Ideally, it's better to be implemented with `--set` on a per-target basis.

johnbizokk:
This feature could benefit virtually any virtual testing environment (from load testing to E2E testing), primarily if the remote host is behind the proxy and does not support transparent TCP proxying to the host. Services such as Cloudflare provide transparent TCP proxying, but that can come at the cost of a big buck. Therefore we would like to have an opportunity to remap some hosts during the building process and drop silly workarounds like semi-manual modification of hosts file. Moreover, this feature is an old citizen of the legacy builder.

PR #3234, opened 2025-06-11
bake: extra-hosts support

crazy-max:
fixes https://github.com/docker/buildx/issues/494

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding"
 	"encoding/json"
+	"fmt"
 	"io"
 	"maps"
 	"os"
@@ -726,6 +727,7 @@ type Target struct {
 	Ulimits          []string                `json:"ulimits,omitempty" hcl:"ulimits,optional" cty:"ulimits"`
 	Call             *string                 `json:"call,omitempty" hcl:"call,optional" cty:"call"`
 	Entitlements     []string                `json:"entitlements,omitempty" hcl:"entitlements,optional" cty:"entitlements"`
+	ExtraHosts       map[string]*string      `json:"extra-hosts,omitempty" hcl:"extra-hosts,optional" cty:"extra-hosts"`
 	// IMPORTANT: if you add more fields here, do not forget to update newOverrides/AddOverrides and docs/bake-reference.md.
 
 	// linked is a private field to mark a target used as a linked one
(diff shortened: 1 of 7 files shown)


bake/bake.go
```diff
@@ -726,6 +726,7 @@ type Target struct {
 	Ulimits          []string                `json:"ulimits,omitempty" hcl:"ulimits,optional" cty:"ulimits"`
 	Call             *string                 `json:"call,omitempty" hcl:"call,optional" cty:"call"`
 	Entitlements     []string                `json:"entitlements,omitempty" hcl:"entitlements,optional" cty:"entitlements"`
+	ExtraHosts       []string                `json:"extra-hosts,omitempty" hcl:"extra-hosts,optional" cty:"extra-hosts"`
```
crazy-max: I opted for `extra-hosts` instead of `add-hosts` in Bake. Compose has also the same name.

tests/bake.go
```diff
@@ -2162,6 +2163,29 @@ target "third" {
...
 }
 
+func testBakeExtraHosts(t *testing.T, sb integration.Sandbox) {
+	dockerfile := []byte(`
+FROM busybox
+RUN cat /etc/hosts | grep myhost | grep 1.2.3.4
+	`)
+	bakefile := []byte(`
+target "default" {
+  extra-hosts = ["myhost=1.2.3.4"]
```
tonistiigi: shouldn't this be map?

crazy-max: changed to a map

bake/bake.go
```diff
@@ -726,6 +727,7 @@ type Target struct {
 	Ulimits          []string                `json:"ulimits,omitempty" hcl:"ulimits,optional" cty:"ulimits"`
 	Call             *string                 `json:"call,omitempty" hcl:"call,optional" cty:"call"`
 	Entitlements     []string                `json:"entitlements,omitempty" hcl:"entitlements,optional" cty:"entitlements"`
+	ExtraHosts       map[string]*string      `json:"extra-hosts,omitempty" hcl:"extra-hosts,optional" cty:"extra-hosts"`
```
ndeloof, line 730: a single host name can actually have multiple IPs assigned (typically, to support IPv4 and IPv6)
For this purpose, Compose's `HostsList` is a `map[string][]string`. Maybe simpler to keep using a plain `[]string` to mimic --add-host build flag

crazy-max: Ah indeed this would not work with current logic. I guess with `map[string]*string` it would be smth like `myhost = "8.8.8.8,::1"` to have both v4/v6 ?

On LLB side I don't think we support that format though :thinking:

tonistiigi (approved):

bake/bake.go
ndeloof: Better align with LLB model then, to avoid yet another undocumented format that will require additional parsing.
IMHO `[]string` is fine as an API, as this naturally matches with `--add-host` on the CLI

tonistiigi:
Let's check the multiple-ip case for rc2.

ndeloof:
@tonistiigi doing so, there's a risk json model changes, isn't it?  => compatibility break with Docker Compose to generate bake.json

PR #3235, opened 2025-06-11
dap: add debug adapter implementation

jsternberg:
Adds a simple implementation of the debug adapter that supports the very
basics of a debug adapter.

It supports the launch request, the configuration done request, the
creation of threads, stopping, resuming, and disconnecting from server.

It does not support custom breakpoints, stack traces, or variable
inspection yet. These are planned to be added in the future.


dap/adapter.go
```diff
@@ -0,0 +1,395 @@
...
+	err := d.eg.Wait()
+	d.eg = nil
+	return err
+}
+
+func (d *Adapter) Initialize(c Context, req *dap.InitializeRequest, resp *dap.InitializeResponse) error {
+	close(d.initialized)
+
+	// Set capabilities.
+	resp.Body.SupportsConfigurationDoneRequest = true
```
ktock, line 96: `SupportsTerminateRequest` and `SupportsRestartRequest` seem to be needed to allow the terminate request and the restart request (c.f. https://microsoft.github.io/debug-adapter-protocol//specification.html )

docs/reference/buildx_dap_build.md
```diff
@@ -0,0 +1,52 @@
+# docker buildx dap build
+
+<!---MARKER_GEN_START-->
+Start a build
+
+### Aliases
+
+`docker build`, `docker builder build`, `docker image build`, `docker buildx b`
+
+### Options
```
ktock, line 10: I'm not sure how these options can be used by editors. Maybe these flags will be need to be moved to the launch config when this DAP plugin allows the editor to invoke the build through the launch request.

dap/adapter.go
```diff
@@ -0,0 +1,395 @@
...
+	return nil
+}
+
+type LaunchConfig struct {
```
ktock: Will this be used for build flags as well? (e.g. the user can write the build flags to launch.json instead of to the global editor config file) If so, the Launch handler will somehow need to be able to trigger the build.

dap/adapter.go
jsternberg: I haven't added the functionality for `Terminate` and `Restart` yet. We're capable of them but the DAP adapter implementation isn't quite there yet.

dap/adapter.go
jsternberg: I haven't quite determined how that is. For neovim, the adapter is the one that reads the `launch.json` file and it can then invoke the tool with the correct arguments. My initial thought was to have the editor translate the JSON launch arguments to the appropriate CLI arguments. I have to see how the VSCode side works though to see what is better.

docs/reference/buildx_dap_build.md
jsternberg: The intention is the debug adapter implementation would translate these to command line arguments. It might be worth putting them into the launch config though just to keep things consistent.

Here's an example: https://github.com/docker/nvim-dap-docker/blob/27f73dae22a4be31caaacc37ce4ba8566a1fbcc5/lua/dap-docker.lua#L19-L29

ktock: Thanks for the example. 

> The intention is the debug adapter implementation would translate these to command line arguments. It might be worth putting them into the launch config though just to keep things consistent.

Sure. launch.json support can be following up.

dap/adapter.go
ktock: > My initial thought was to have the editor translate the JSON launch arguments to the appropriate CLI arguments.

:+1:

> I have to see how the VSCode side works though to see what is better.

IIUC, in all of nvim, VSCode([`startDebugging API`](https://code.visualstudio.com/api/references/vscode-api)), and emacs ([dap-mode](https://github.com/emacs-lsp/dap-mode/blob/b97756665709bea37b9ffe262c5fa9196f1b4577/docs/page/features.md#launchjson-support)), DAP plugin can receive launch.json contents from LaunchRequest.

ktock (approved):

jsternberg:
After merging this, here are some proposed follow up issues.

- Implement stack traces
- Add breakpoint support
- Add support for the restart request
- Add support for the terminate request
- Add build args and environment variables to stack traces
- Add support for invoke through evaluate and a reverse request

dap/adapter.go
jsternberg: I've changed the method for how this works and it now supports a combination. Any arguments provided on the command line will be used (to facilitate direct usage or arbitrary args from a plugin/extension) but certain JSON arguments can be used to override any command line arguments. Options that are considered very common will go into JSON properties while the "args" can be a catch-all for everything else.

So `launch.json` would end up looking like this:

```
{
    "type": "dockerfile",
    "request": "launch",
    "dockerfile": "Dockerfile",
    "contextPath": "${workspacePath}",
    "target": "mytarget",
    "args": "--build-arg FOO=1"
}
```

ktock (approved):
Thanks.

commands/build.go
```diff
@@ -453,7 +469,9 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugConfig *debugOp
 			"aliases": "docker build, docker builder build, docker image build, docker buildx b",
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			options.contextPath = args[0]
+			if len(args) > 0 {
+				options.contextPath = args[0]
+			}
```
crazy-max, line 474: Did you encounter an exception without this change?

jsternberg: This was mostly to allow the `dap build` command to reuse `build` but not require the required argument. Another portion of the code will catch that this is missing later but it allows calling `buildx dap build` and letting the context path to be sent later by the launch arguments.

It doesn't impact the normal build command because cobra will never let the code get there.

PR #3240, opened 2025-06-12
commands: remove debug package in commands

jsternberg:
The package just causes the entire flow to be more complicated as build
has to pretend it doesn't know about debug options and the debugger has
to pretend it doesn't know about the build.

This abstraction has been difficult when integrating a DAP command into
this same workflow so I don't think this abstraction has much of a
value.


jsternberg:
The copilot comment does not really seem to be correct. There's no legacy debug command or implementation. I just copied it to the same package as the other commands. I didn't want someone to get the wrong impression from it.

tonistiigi (changes requested):
Docs needs regeneration

jsternberg:
The docs change was caused by removing the `--progress` from debug. `--progress` is on the `build` command itself and this version of progress was used by the monitor CLI that got removed. We just didn't notice that this line of code did nothing.

PR #3241, opened 2025-06-13
remove "experimental" hint from help / usage output

thaJeztah:
I saw https://github.com/docker/buildx/pull/3240 which reminded me that I wanted to open this patch as suggestion.

This was added in 02c2073feb3681e5c13e944d5f1cd2246cc22013, which changed the buildx CLI to match the behavior of the docker CLI, which conditionally hid some options that were marked experimental. The CLI now shows has client-side experimental options enabled by default (but labeled as experimental where suitable).

Before this patch, every command would print this message unless experimental was enabled;

    docker buildx --help
    ...
    Run 'docker buildx COMMAND --help' for more information on a command.

    Experimental commands and flags are hidden. Set BUILDX_EXPERIMENTAL=1 to show them.

With this patch applied, that message is no longer shown.

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -85,10 +85,6 @@ func NewRootCmd(name string, isPlugin bool, dockerCli *command.DockerCli) *cobra
 		"using default config store",
 	))
 
-	if !confutil.IsExperimental() {
-		cmd.SetHelpTemplate(cmd.HelpTemplate() + "\nExperimental commands and flags are hidden. Set BUILDX_EXPERIMENTAL=1 to show them.\n")
-	}
-
 	addCommands(cmd, &opt, dockerCli)
 	return cmd
 }


commands/root.go
```diff
@@ -84,10 +84,6 @@ func NewRootCmd(name string, isPlugin bool, dockerCli *command.DockerCli) *cobra
 		"using default config store",
 	))
 
-	if !confutil.IsExperimental() {
```
thaJeztah, line 87: Also considering if we should just (as we did for the CLI) always enable the `debug` command;

https://github.com/docker/buildx/blob/08dd378b5943ef48a4258ca7c640c309d1a7fdb9/commands/root.go#L121-L123


The output already shows it's experimental;

```console
BUILDX_EXPERIMENTAL=1 docker buildx debug --help
Usage:  docker buildx debug [OPTIONS] COMMAND

Start debugger (EXPERIMENTAL)

EXPERIMENTAL:
  docker buildx debug is an experimental feature.
  Experimental features provide early access to product functionality. These
  features may change between releases without warning, or can be removed from a
  future release. Learn more about experimental features in our documentation:
  https://docs.docker.com/go/experimental/
```

crazy-max: > Also considering if we should just (as we did for the CLI) always enable the `debug` command

With on-going work with `debug` command related to new DAP implementation I think we should keep it experimental for now:
* https://github.com/docker/buildx/pull/3235
* https://github.com/docker/buildx/pull/3257
* https://github.com/docker/buildx/pull/3279

thaJeztah: Yeah, I was considering that we would have it visible, but call out that it's experimental? It may help getting more people to try it if we don't require config changes, as long as we explicitly call out that it's experimental?

thaJeztah:
Did a quick rebase to get a fresh run of CI and moved it out of draft

commands/root.go
crazy-max: In the future I agree that we should deprecate `BUILDX_EXPERIMENTAL` as the CLI already marks commands as experimental anyway.

thaJeztah:
All green again; PTAL

crazy-max (approved):

commands/root.go
jsternberg: For what it's worth, the debug adapter stuff is going under another command in `dap`. Right now, `debug` controls the monitor. We do potentially have plans to completely remove the `debug` command and integrate it directly with build.

jsternberg (approved):

commands/root.go
crazy-max: Discussed with @tonistiigi, we would need to always enable the `debug` command if we are removing this hint from help so users are aware of it.

thaJeztah:
Updated to now enable experimental commands by default;

- commands: enable experimental commands by default
- remove uses of BUILDX_EXPERIMENTAL for generating docs
- util/confutil: remove IsExperimental() utility
- remove "experimental" integration tests
- docs: remove use of BUILDX_EXPERIMENTAL as example

tests/integration.go
```diff
@@ -172,9 +160,6 @@ func buildkitVersion(t *testing.T, sb integration.Sandbox) string {
 			require.NoError(t, err, "undock not found")
 
 			destDir := t.TempDir()
-			t.Cleanup(func() {
-				os.RemoveAll(destDir)
-			})
```
thaJeztah, line 177: Not really related, but my IDE was complaining about the unhandled error, and it was redundant, so added a commit to remove it 😂

crazy-max:
@thaJeztah For a follow-up I was wondering if we still need to alter `Usage` and `Short` for flags and commands in: https://github.com/docker/buildx/blob/55698253a52d4b09a538fc6888a5eacae6c6266b/util/cobrautil/cobrautil.go#L22-L53

Maybe docker/cli has this kind of helpers already?

thaJeztah:
> Maybe docker/cli has this kind of helpers already?

docker/cli adds an experimental warning to the "usage" / "help" output of the command itself, but does not add a ` (EXPERIMENTAL)` to flags and/or commands in overviews.

We should probably look what we want to do there; to some extend the "ALL CAPS" format seems a bit aggressive, but maybe there's some middle ground.

PR #3242, opened 2025-06-16
Allow bake files to be specified via environment variable

### LINKED ISSUE #3218: New environment variable to specify bake file(s)
### Description

It would be nice to have an environment variable like `BAKE_FILE`, analogous to [`COMPOSE_FILE`](https://docs.docker.com/compose/how-tos/environment-variables/envvars/#compose_file) in compose, to allow specifying the bake file(s) to execute.  For example, I'd like to replace
```shell
$ docker buildx bake -f docker-bake.hcl -f docker-bake.test.hcl <target>
```
with
```shell
$ export BAKE_FILE='docker-bake.hcl:docker-bake.test.hcl'
$ docker buildx bake <target>
```

My actual problem (annoyance) is having to continually specify two `-f` arguments for every command.  For readability, I chose to split up my primary targets from my targets dedicated to testing.  I know I could technically solve this simply renaming my second file to `docker-bake.override.hcl`.  But in reality, it's not an override file; it's 100% additive with no overriding or merging.  Even though the second file is additive, it's not independent; it uses functions defined in the primary file, which is why I have to specify both files in order to build a target in the second.

Something like `import: "docker-bake.test.hcl"` would also solve my problem.  Obviously that also has precedent in compose (and would actually be even cleaner/friendlier in my case), but I assumed the environment variable would be more likely to get a thumbs up, and of course, much simpler to implement.

rrjjvv:
Fixes #3218 

The environment variable `BUILDX_BAKE_FILE` (and optional variable `BUILDX_BAKE_FILE_SEPARATOR`) can be used to specify one or more bake files (similar to `compose`).  This is mutually exclusive with`--file` (which takes precedence).

This is done very early to ensure the values are treated just like `--file`, e.g., participate in telemetry.  This includes leaving relative paths as-is, which deviates from `compose` (which makes them absolute).

This was intended to functionally be just like compose, but there was no opportunity to reuse code and the patterns/flows used were different enough that I went with something simple rather than trying to make them look the same.

When I created the feature request, it didn't occur to me that there could be potential confusion/ambiguity between `COMPOSE_FILE` (not respected by bake to my knowledge), `--file` (which can be used to specify compose files), and whether the new `BUILDX_BAKE_FILE` would/should allow compose files.  The issue is ignored/unaddressed in this initial implementation, as it wasn't immediately obvious whether `COMPOSE_FILE` _should_ be supported and what its relationship to `BUILDX_BAKE_FILE` would be.  The focus is solely on `BUILDX_BAKE_FILE` is equivalent to `--file`.

### DIFF
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -143,6 +143,11 @@ Use the `-f` / `--file` option to specify the build definition file to use.
 The file can be an HCL, JSON or Compose file. If multiple files are specified,
 all are read and the build configurations are combined.
 
+Alternatively, the environment variable `BUILDX_BAKE_FILE` can be used to specify the build definition to use.
+This is mutually exclusive with `-f` / `--file`; if both are specified, the environment variable is ignored.
+Multiple definitions can be specified by separating them with the system's path separator
+(typically `;` on Windows and `:` elsewhere), but can be changed with `BUILDX_BAKE_PATH_SEPARATOR`.
+
 You can pass the names of the targets to build, to build only specific target(s).
 The following example builds the `db` and `webapp-release` targets that are
 defined in the `docker-bake.dev.hcl` file:
--- a/tests/bake.go
+++ b/tests/bake.go
@@ -81,6 +81,7 @@ var bakeTests = []func(t *testing.T, sb integration.Sandbox){
 	testBakeMultiPlatform,
 	testBakeCheckCallOutput,
 	testBakeExtraHosts,
+	testBakeFileFromEnvironment,
 }
 
 func testBakePrint(t *testing.T, sb integration.Sandbox) {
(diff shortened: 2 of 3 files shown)


docs/reference/buildx_bake.md
```diff
@@ -143,6 +143,11 @@ Use the `-f` / `--file` option to specify the build definition file to use.
 The file can be an HCL, JSON or Compose file. If multiple files are specified,
 all are read and the build configurations are combined.
 
+Alternatively, the environment variable `BUILDX_BAKE_FILE` can be used to specify the build definition to use.
+This is mutually exclusive with `-f` / `--file`; if both are specified, the environment variable is ignored.
```
crazy-max, line 147: Ok seems it follows compose behavior https://docs.docker.com/reference/cli/docker/compose/#set-up-environment-variables but I would tend to error out in such case. Although with bake we print the loaded definitions during build so at least user is aware of performed definitions.

rrjjvv: > I would tend to error out in such case

That would be pretty unusual and unexpected I would think.  If it contradicted with another variable/flag that would be one thing.  Otherwise I wouldn't see it as any different than `BUILDKIT_PROGRESS` vs. `--progress`, `BUILDX_BUILDER` vs. `--builder`, etc., where a flag always silently wins over the environment.  I guess in those cases it's a _little_ different in that those are single value options, whereas these files are multi-value, and maybe you're concerned about user expectations regarding whether flags and environment are additive in this case?

Let me know if I misunderstood or misinterpreted your comment (or if I did read it correctly and you still feel it should be an error).

crazy-max: > Otherwise I wouldn't see it as any different than `BUILDKIT_PROGRESS` vs. `--progress`, `BUILDX_BUILDER` vs. `--builder`, etc., where a flag always silently wins over the environment.

Yes indeed, hence why we print builder used and bake definitions loaded in the logs as "hints". So agreed to stay consistent.

rrjjvv: > we print builder used and bake definitions loaded in the logs as "hints"

Incidentally, that is one of the main reasons I deviated from compose's behavior of converting environment-provided files to absolute paths (which I noted in the commit comment).  An absolute path made the output very ugly.  And after seeing the input files were added as trace attributes, I figured not forcing an absolute path would simply avoid the question of whether it would be (or needed to be) scrubbed or not.

Just wanted to call that out in case it triggers any thoughts.  I thought/hoped by putting the processing where I did, the questions/rules about where relative paths are anchored would be no different than for `--file`.

crazy-max: > And after seeing the input files were added as trace attributes, I figured not forcing an absolute path would simply avoid the question of whether it would be (or needed to be) scrubbed or not.

You mean through the traced command in https://github.com/docker/buildx/blob/d09eb752a5be19ce942a1031223d5b1eba69d992/commands/bake.go#L68-L72? If so that's good this is being traced here :+1:

What I meant by "hence why we print builder used and bake definitions loaded in the logs" is when invoking the command you have the loaded bake definitions displayed in progress output:

```
$ docker buildx bake --print
#1 reading docker-bake.hcl 8.03kB / 8.03kB done
#1 DONE 0.0s
```

I was thinking that maybe we could mark files loaded from an env var with another message like:

```
$ docker buildx bake --print
#1 [internal] load bake definitions from BUILDX_BAKE_FILE env
#1 reading docker-bake.hcl 8.03kB / 8.03kB done
#1 DONE 0.0s
```

So user is not confused.

rrjjvv: > You mean through the traced command in

Yes, exactly; I didn't follow the entire code path, but that is exactly the code I saw as I was trying to figure out where best to add my new code.  I know many folks consider an absolute path leakage (e.g., seeing a user's directory structure and username).  I don't know if a full path would have actually been passed through or if this community would have cared even if it was.  Of course if a user specifies absolute paths that's a different story.  But I saw no technical reason/advantage to converting relative paths to absolute, and a few downsides.

> What I meant by [...]
I was thinking that maybe we could mark files loaded from an env var with another message like

Yup, I knew what you meant (the UI portion), but I didn't realize your concern was around communicating *which* mechanism was used to specify the file; I thought it was around *that* a file was used (vs. ignored).  Sorry I missed that!

So yes, I can do that.  Quickly looking at the code, I think there's going to be a choice between doing it:
* "simply"; doing a second lookup in `readBakeFiles` where that printing occurs to see if that environment variable exists, and if does, assume it was respected.  This is obviously the case today, but not really future-safe.
* "correctly"; tracking the fact it was specified via environment, as opposed to the simple sleight-of-hand I'm currently doing (directly populating `in.files` from the environment as if they had been specified via `-f`) 

I'll try to find an unobtrusive way to do this "correctly", but if you actually prefer "simply" in this case, let me know.

rrjjvv: Let me know what you think of what I pushed.  Most of the 'good' solutions I could see entailed adding code disproportionate to the feature itself or changing the flow (which you already said you liked).  I left it as a separate commit anticipating you might want (possibly substantial) changes or even opt to drop it.

tests/bake.go
```diff
@@ -2188,6 +2189,165 @@ target "default" {
...
+			fstest.CreateFile("first", []byte("first"), 0600),
+		)
+		cmd := buildxCmd(sb,
+			withDir(dir),
+			withArgs("bake", "--progress=plain", "first"),
+			withEnv("BUILDX_BAKE_FILE=first.hcl"))
+
+		dt, err := cmd.CombinedOutput()
+		require.NoError(t, err, string(dt))
+		require.Contains(t, string(dt), `#1 [internal] load local bake definitions from BUILDX_BAKE_FILE`)
```
crazy-max:
```suggestion
		require.Contains(t, string(dt), `#1 [internal] load local bake definitions from BUILDX_BAKE_FILE env`)
```

docs/reference/buildx_bake.md
rrjjvv: I assumed that your suggestion to to specify the full message on the one test was intended to cover all the tests, so I did just that.

crazy-max (approved):
LGTM thanks!

PTAL @tonistiigi

crazy-max:
> The environment variable `BUILDX_BAKE_FILE` (and optional variable `BUILDX_BAKE_FILE_SEPARATOR`)

@ArthurFlag Would need additional docs follow-up in https://github.com/docker/docs/blob/main/content/manuals/build/building/variables.md#build-tool-configuration-variables

tonistiigi (approved):

PR #3243, opened 2025-06-16
bake: rename BAKE_ALLOW_REMOTE_FS_ACCESS env var

crazy-max:
follow-up https://github.com/docker/buildx/issues/3218#issuecomment-2975500596


tonistiigi:
I believe this env is effectively deprecated anyway and replaced with entitlements so we don't need to update/document it anymore.

PR #3246, opened 2025-06-16
progress: invoke onclose every time display is stopped

jsternberg:
During the refactor, it was changed to only call `onclose` when the
printer itself was closed. The original version invoked this when the
display was stopped in anyway.

Changing this back to see if it fixes a display issue that's popped up
with the refactor.

Restores the metric writer which accidentally got left off in the
refactor and wasn't noticed.


jsternberg:
This wasn't the cause of the issue. We found the real cause and it's fixed here: https://github.com/docker/buildx/pull/3248

The extra stuff in this PR such as the metric writer have been moved to that PR instead.

PR #3247, opened 2025-06-16
build: slsa v1 support for metadata

crazy-max:
Adds support for SLSA v1 when outputting to metadata file. This also adds a new field `buildx.build.provenance.predicate-type`.


tonistiigi (commented):
We may need to think about this case bit more. Thinking it bit more I don't think it should really matter what provenance version was used in history for the output of `--metadata-file`. Might instead do a conversion like in https://github.com/docker/buildx/pull/3245

crazy-max:
> Might instead do a conversion like in #3245

Yes sounds good

crazy-max:
needs docs follow-up in https://docs.docker.com/build/metadata/attestations/slsa-provenance/#provenance-attestation-example

PR #3253, opened 2025-06-17
driver kubernetes: allow to work in a Memory mount to speed up things

samifruit514:
Depending on the size of the layers, it can increase build speed by a lot.
our use case is an docker buildx on a c7i.4xlarge. For a big COPY of about 1GB, we gained 1min of build time!

### DIFF
--- a/driver/kubernetes/factory.go
+++ b/driver/kubernetes/factory.go
@@ -286,6 +286,10 @@ func (f *factory) processDriverOpts(deploymentName string, namespace string, cfg
 			if v != "" {
 				deploymentOpt.Qemu.Image = v
 			}
+		case "buildkit-root-volume-memory":
+			if v != "" {
+				deploymentOpt.BuildKitRootVolumeMemory = v
+			}
 		case "default-load":
 			defaultLoad, err = strconv.ParseBool(v)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


driver/kubernetes/factory.go
```diff
@@ -286,6 +286,10 @@ func (f *factory) processDriverOpts(deploymentName string, namespace string, cfg
 			if v != "" {
 				deploymentOpt.Qemu.Image = v
 			}
+		case "buildkit-root-folder-memory":
```
AkihiroSuda:
```suggestion
		case "buildkit-root-volume-memory":
```

Please update the docs too

driver/kubernetes/manifest/manifest.go
```diff
@@ -50,6 +51,7 @@ const (
 	containerName      = "buildkitd"
 	AnnotationPlatform = "buildx.docker.com/platform"
 	LabelApp           = "app"
+	memoryVolumeName   = "buildkit-memory"
```
AkihiroSuda:
```suggestion
	rootVolumeName   = "buildkit-root"
```

Because the volume may potentially have a non-memory backend in future

samifruit514:
Thanks @AkihiroSuda , I created a PR for doc: https://github.com/docker/docs/pull/22865

AkihiroSuda:
LGTM, but please squash the commits

crazy-max:
Needs docs follow-up in https://docs.docker.com/build/builders/drivers/kubernetes/#synopsis, cc @ArthurFlag

PR #3257, opened 2025-06-17
dap: support evaluate request to invoke a container

jsternberg:
Supports using the `evaluate` request in REPL mode to start a container
with the `exec` command. Presently doesn't support any arguments.

This improves the dap server so it is capable of sending reverse
requests and receiving the response. It also adds a hidden command
`dap attach` that attaches to the socket created by `evaluate`.

This requires the client to support `runInTerminal`.

Likely needs some additional work to make sure resources are cleaned up
cleanly especially when the build is unpaused or terminated, but it
should work as a decent base.

### DIFF
--- a/commands/dap.go
+++ b/commands/dap.go
@@ -2,12 +2,17 @@ package commands
 
 import (
 	"context"
+	"io"
+	"net"
+	"os"
 
+	"github.com/containerd/console"
 	"github.com/docker/buildx/dap"
 	"github.com/docker/buildx/dap/common"
 	"github.com/docker/buildx/util/cobrautil"
 	"github.com/docker/buildx/util/ioset"
 	"github.com/docker/buildx/util/progress"
+	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -24,6 +29,8 @@ func dapCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	dapBuildCmd := buildCmd(dockerCli, rootOpts, &options)
 	dapBuildCmd.Args = cobra.RangeArgs(0, 1)
 	cmd.AddCommand(dapBuildCmd)
+
+	cmd.AddCommand(dapAttachCmd())
 	return cmd
 }
 
@@ -70,4 +77,40 @@ func (d *adapterProtocolDebugger) Start(printer *progress.Printer, opts *BuildOp
 func (d *adapterProtocolDebugger) Stop() error {
 	defer d.conn.Close()
 	return d.Adapter.Stop()
+}
+
+func dapAttachCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "attach PATH",
+		Short:  "Attach to a container created by the dap evaluate request",
+		Args:   cli.ExactArgs(1),
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c, err := console.ConsoleFromFile(os.Stdout)
+			if err != nil {
+				return err
+			}
+
+			if err := c.SetRaw(); err != nil {
+				return err
+			}
+
+			conn, err := net.Dial("unix", args[0])
+			if err != nil {
+				return err
+			}
+
+			fwd := ioset.NewSingleForwarder()
+			fwd.SetReader(os.Stdin)
+			fwd.SetWriter(conn, func() io.WriteCloser {
+				return conn
+			})
+
+			if _, err := io.Copy(os.Stdout, conn); err != nil && !errors.Is(err, io.EOF) {
+				return err
+			}
+			return nil
+		},
+	}
+	return cmd
 }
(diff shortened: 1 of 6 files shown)


commands/dap.go
```diff
@@ -0,0 +1,134 @@
...
+func dapCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
+	var options dapOptions
+	cmd := &cobra.Command{
+		Use:   "dap",
+		Short: "Start debug adapter",
+	}
+	cobrautil.MarkCommandExperimental(cmd)
+
+	flags := cmd.Flags()
+	flags.StringVar(&options.OnFlag, "on", "error", "When to pause the adapter ([always, error])")
```
tonistiigi: Isn't this `ExceptionBreakMode` sent via DAP protocol itself?

commands/dap.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (d *adapterProtocolDebugger) Start(printer *progress.Printer) error {
+	if err := d.Adapter.Start(context.Background(), d.conn); err != nil {
+		return errors.Wrap(err, "debug adapter did not start")
+	}
+	return nil
+}
+
+func (d *adapterProtocolDebugger) Out() console.File {
```
tonistiigi: Why is this returning `console.File`? I assume detecting the progress mode can be done outside without leaking this into dap layer.

build/build.go
```diff
@@ -313,9 +313,12 @@ func toRepoOnly(in string) (string, error) {
 	return strings.Join(out, ","), nil
 }
 
-type Handler struct {
-	Evaluate func(ctx context.Context, c gateway.Client, res *gateway.Result) error
-}
+type (
+	EvaluateFunc func(ctx context.Context, name string, c gateway.Client, res *gateway.Result) error
+	Handler      struct {
+		Evaluate EvaluateFunc
```
tonistiigi: `EvaluateFunc` seems unused.

commands/build.go
```diff
@@ -1210,3 +1126,24 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in *BuildOptions, inSt
 	}
 	return resp[defaultTargetName], inputs, nil
 }
+
+// debuggerOptions will start a debuggerOptions instance.
```
tonistiigi: Looks like this should be in `commands/debug`.

dap/adapter.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+}
+
+type thread struct {
+	d    *Adapter
+	id   int
+	name string
+
+	paused chan struct{}
+	rCtx   *build.ResultHandle
```
tonistiigi: follow-up: this `ResultHandle` seems incorrectly named. I'm not sure if we need it at all but if we do then it should be called something else and should not be defined in the `build` pkg, as it is basically the transformation of the `Evaluate()` parameters into struct form and therefore duplicate in the context of build pkg.

`Ctx` variable naming also makes it confusing as it looks like some form of `context.Context`.

commands/dap.go
```diff
@@ -0,0 +1,134 @@
...
+		Short: "Start debug adapter",
+	}
+	cobrautil.MarkCommandExperimental(cmd)
+
+	flags := cmd.Flags()
+	flags.StringVar(&options.OnFlag, "on", "error", "When to pause the adapter ([always, error])")
+
+	cobrautil.MarkFlagsExperimental(flags, "on")
+
+	cmd.AddCommand(buildCmd(dockerCli, rootOpts, &options))
```
tonistiigi: What's the use case of the `build` subcommand?

jsternberg: The command is invoked with `docker buildx dap build` and the bake version will be `docker buildx dap bake`.

dap/adapter.go
jsternberg: Agreed. It's mostly been kept around because the code exists and it was useful to reuse it. It can be renamed and probably should be renamed.

commands/dap.go
jsternberg: I think this is cited somewhere else in the PR but it's mostly because the `NewPrinter` takes a `console.File` even though it shouldn't. `progressui.NewDisplay` takes an `io.Writer` and creates a `console.File` from that so `NewPrinter` is just going `console.File -> io.Writer -> console.File`.

We can change `NewPrinter` to take an `io.Writer` and then this can change to an `io.Writer`.

commands/dap.go
jsternberg: Not currently implemented but yes ideally we would remove this and use `ExceptionBreakMode` to set the behavior.

tonistiigi (commented):
As discussed offline, we probably want a different variant of this where the `exec` is not per container, but there is one terminal exec that is associated with the `dap` session. When the status of dap session changes between running and paused, the terminal should internally switch to new container without the exec process itself shutting down. Later we can add extra capabilities that would keep the state from the previous container by carrying over things like bash history and cwd.

I think we can still merge this and test both flows during the experimental period. Eg. with a config switch in the vscode extension side.

ktock:
Sorry for the slow review. This patch looks good to me, too.

PR #3259, opened 2025-06-18
build: fix buildx.build.provenance metadata

crazy-max:
alternative to #3247
needs https://github.com/moby/buildkit/pull/6034

Keep SLSA v0.2 format for `buildx.build.provenance` field in metadata output.


tonistiigi (approved):
We can still have a follow-up issue about what we want to do about this long-term but this should avoid breaking changes for now.

crazy-max:
> We can still have a follow-up issue about what we want to do about this long-term

I would suggest to deprecate `buildx.build.provenance` property in favor of `history inspect attachment` command. WDYT?

PR #3260, opened 2025-06-18
dap: add stack traces to stopped threads

jsternberg:
Stack traces show the first parent of each operation that was run to
get to the current location. It shows either their custom name from the
LLB or it will show the command name.

stack-trace-example


jsternberg:
Parts of this have been integrated with https://github.com/docker/buildx/pull/3279. I'm going to close this as that other PR is more complete.

Issue #2802, opened 2024-11-20
fix cases or early exit with skipping defer


tonistiigi:
There are couple of cases (eg. when using `--call check`) that do an early return that can skip running defers and (for example reporting metrics).

After fixing https://github.com/docker/buildx/blob/master/.golangci.yml#L38 linter rule should be enabled.

crazy-max:
> After fixing https://github.com/docker/buildx/blob/master/.golangci.yml#L38 linter rule should be enabled.

https://github.com/docker/buildx/blob/6deb9ff38467d32ca823f67f91622650b7719168/.golangci.yml#L52

PR #3278, opened 2025-06-30
cmd: fix possible skipped defers for build and bake

crazy-max:
fixes https://github.com/docker/buildx/issues/2802


util/cobrautil/error.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+type ExitCodeError int
+
+func (e ExitCodeError) Error() string {
+	return fmt.Sprintf("exiting with code %d", int(e))
+}
+
+func (e ExitCodeError) Unwrap() error {
+	return errors.New(e.Error())
```
tonistiigi: What's this? Should return `nil`.

crazy-max: :man_facepalming:

PR #3279, opened 2025-06-30
dap: add stack traces with next and continue functionality

jsternberg:
It is now possible to send next and continue as separate signals. When
executing a build, the debug adapter will divide the LLB graph into
regions. Each region corresponds to an uninterrupted chain of
instructions. It will also record which regions depend on which other
ones.

This determines the execution order and it also determines what the
stack traces look like.

When continue is used, we will attempt to evaluate the last leaf node
(the head). If we push next, we will determine which digest would be the
next one to be processed.

In the future, this will be used to also support breakpoints.

### DIFF
--- a/dap/adapter.go
+++ b/dap/adapter.go
@@ -1,11 +1,14 @@
 package dap
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io"
+	"path"
 	"sync"
+	"sync/atomic"
 
 	"github.com/docker/buildx/build"
 	"github.com/google/go-dap"
@@ -28,6 +31,9 @@ type Adapter[T any] struct {
 	threads      map[int]*thread
 	threadsMu    sync.RWMutex
 	nextThreadID int
+
+	sourceMap sourceMap
+	idPool    *idPool
 }
 
 func New[T any](cfg *build.InvokeConfig) *Adapter[T] {
@@ -38,6 +44,7 @@ func New[T any](cfg *build.InvokeConfig) *Adapter[T] {
 		evaluateReqCh: make(chan *evaluateRequest),
 		threads:       make(map[int]*thread),
 		nextThreadID:  1,
+		idPool:        new(idPool),
 	}
 	if cfg != nil {
 		d.cfg = *cfg
@@ -131,7 +138,16 @@ func (d *Adapter[T]) Continue(c Context, req *dap.ContinueRequest, resp *dap.Con
 	t := d.threads[req.Arguments.ThreadId]
 	d.threadsMu.RUnlock()
 
-	t.Resume(c)
+	t.Continue()
+	return nil
+}
+
+func (d *Adapter[T]) Next(c Context, req *dap.NextRequest, resp *dap.NextResponse) error {
+	d.threadsMu.RLock()
+	t := d.threads[req.Arguments.ThreadId]
+	d.threadsMu.RUnlock()
+
+	t.Next()
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


dap/thread.go
```diff
@@ -0,0 +1,535 @@
...
+			region--
+
+			r = t.regions[region]
+			i = 0
+			continue
+		}
+
+		next := r.digests[i]
+		if loc, ok := t.def.Source.Locations[string(next)]; !ok || len(loc.Locations) == 0 {
+			// Skip this digest because ti has no location in the source file.
```
ktock: `s/ti has/it has`?

dap/thread.go
```diff
@@ -0,0 +1,535 @@
...
+func (t *thread) init(ctx Context, c gateway.Client, ref gateway.Reference, meta map[string][]byte, inputs build.Inputs) error {
+	t.c = c
+	t.ref = ref
+	t.meta = meta
+	t.sourcePath = inputs.ContextPath
+
+	if err := t.createRegions(ctx); err != nil {
+		return err
+	}
+	return nil
```
ktock: nit:
```suggestion
	return t.createRegions(ctx)
```

dap/thread.go
```diff
@@ -0,0 +1,535 @@
...
+		if frame.Column == 0 || int(r.Start.Character) < frame.Column {
+			frame.Column = int(r.Start.Character)
+		}
+		if frame.EndLine == 0 || int(r.End.Line) > frame.EndLine {
+			frame.EndLine = int(r.End.Line)
+		}
+		if frame.EndColumn == 0 || int(r.End.Character) > frame.EndColumn {
+			frame.EndColumn = int(r.End.Character)
+		}
+	}
```
ktock: If an op shared by 2 distant lines (e.g. L3 and L10), the editor will highlight a large region containing unrelated lines, right? Maybe returning the first region is easier to understand.

dap/adapter.go
```diff
@@ -454,3 +409,63 @@ func (d *adapterWriter[T]) Write(p []byte) (n int, err error) {
...
+type idPool struct {
+	next atomic.Int64
+}
+
+func (p *idPool) Get() int64 {
+	return p.next.Add(1)
+}
+
+func (p *idPool) Put(x int64) {
+	// noop
```
ktock, line 422: nit: It's unlikely happen but it's safe to return an error or panic if the id is exhausted.

jsternberg: I was mostly just adding this struct in case we ever wanted to reuse ids for some reason. That's the main reason why this function exists. I can also just remove this entirely and we can add that later if it's ever used.

dap/thread.go
```diff
@@ -0,0 +1,520 @@
...
+
+	t.def, err = st.Marshal(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, src := range t.def.Source.Infos {
+		fname := filepath.Join(t.sourcePath, src.Filename)
+		t.sourceMap.Put(ctx, fname, src.Data)
```
tonistiigi, line 212: We might need to check for duplicates here in the future as `Filename` is not guaranteed to be unique.

dap/thread.go
```diff
@@ -0,0 +1,520 @@
...
+	head := &pb.Op{
+		Inputs: []*pb.Input{{Digest: string(target)}},
+	}
+	dt, err := head.MarshalVT()
+	if err != nil {
+		return nil, err
+	}
+
+	def := t.def.ToPB()
+	def.Def[len(def.Def)-1] = dt
```
tonistiigi, line 447: This looks wasteful. We should only send the ops that actually participate in the build.

tonistiigi (approved):

dap/thread.go
jsternberg: Source map has some code in it that will deduplicate things. It checks the filename and the data and doesn't emit an event when the data doesn't change.

PR #3280, opened 2025-07-01
bake: add pattern matching for targets input

crazy-max:


crazy-max:
Needs docs follow-up at https://docs.docker.com/build/bake/targets/

Issue #3197, opened 2025-05-20
buildx bake: provide access to SOURCE_DATE_EPOCH as an RFC3339 timestamp string


sdavids:
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

### Description

I want to be able to set the `org.opencontainers.image.created` label via `SOURCE_DATE_EPOCH`.

### Expected behaviour

There should be a way to access `SOURCE_DATE_EPOCH` as a  RFC3339 timestamp string.

### Actual behaviour

- `SOURCE_DATE_EPOCH` cannot be accessed as a RFC3339 timestamp string

- [provider::time::unix_timestamp_parse](https://registry.terraform.io/providers/hashicorp/time/latest/docs/functions/unix_timestamp_parse) cannot be used

### Buildx version

github.com/docker/buildx v0.23.0-desktop.1 503f948aadbddb6de3ec5581f766e1d27f6975a1

### Docker info

```text
Client:
 Version:    28.1.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
  ai: Docker AI Agent - Ask Gordon (Docker Inc.)
    Version:  v1.1.7
    Path:     /Users/redacted/.docker/cli-plugins/docker-ai
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.23.0-desktop.1
    Path:     /Users/redacted/.docker/cli-plugins/docker-buildx
  cloud: Docker Cloud (Docker Inc.)
    Version:  v0.3.0
    Path:     /Users/redacted/.docker/cli-plugins/docker-cloud
  compose: Docker Compose (Docker Inc.)
    Version:  v2.35.1-desktop.1
    Path:     /Users/redacted/.docker/cli-plugins/docker-compose
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.38
    Path:     /Users/redacted/.docker/cli-plugins/docker-debug
  desktop: Docker Desktop commands (Docker Inc.)
    Version:  v0.1.8
    Path:     /Users/redacted/.docker/cli-plugins/docker-desktop
  dev: Docker Dev Environments (Docker Inc.)
    Version:  v0.1.2
[...]

PR #3286, opened 2025-07-02
bake: add rfc339parse function

crazy-max:
fixes https://github.com/docker/buildx/issues/3197

Adds `rfc339parse` similar to terraform [`rfc339_parse` function](https://registry.terraform.io/providers/hashicorp/time/latest/docs/functions/rfc3339_parse)

```hcl
variable "BUILD_TIME" {
  type     = string
  default  = timestamp()
}
variable "SOURCE_DATE_EPOCH" {
  type     = number
  default  = rfc339parse(timestamp()).unix
}
group "default" {
  targets = ["test"]
}
target "test" {
  tags = ["baked"]
  labels = {
    "org.opencontainers.image.created" = "${BUILD_TIME}"
    "epoch" = "${SOURCE_DATE_EPOCH}"
  }
}
```

```
$ docker buildx bake --print
{
  "group": {
    "default": {
      "targets": [
        "test"
      ]
    }
  },
  "target": {
    "test": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "labels": {
        "epoch": "1751463884",
        "org.opencontainers.image.created": "2025-07-02T13:44:44Z"
      },
      "tags": [
        "baked"
      ]
    }
  }
}
```

cc @sdavids


bake/hclparser/stdlib.go
```diff
@@ -248,6 +249,61 @@ func timestampFunc() function.Function {
 	})
 }
 
+// rfc339ParseFunc, given an RFC3339 timestamp string, will parse and return an
```
sdavids: Sometimes you use `339` … sometimes `3339` — please check all occurrences of `339`.

crazy-max: Thanks

sdavids:
One uses `SOURCE_DATE_EPOCH` as follows:

```console
$ SOURCE_DATE_EPOCH=0 docker buildx bake
$ SOURCE_DATE_EPOCH=42 docker buildx bake
```

Does this PR support this use case?

`SOURCE_DATE_EPOCH` is used to pass a fixed timestamp from the outside into the build.

—

Having this `rfc3339parse` function is appreciated as well …

bake/hclparser/stdlib.go
```diff
@@ -248,6 +249,61 @@ func timestampFunc() function.Function {
 	})
 }
 
+// rfc3339ParseFunc, given an RFC3339 timestamp string, will parse and return an
+// object representation of that date and time.
+//
+// This function is similar to the `rfc3339_parse` function in Terraform: https://registry.terraform.io/providers/hashicorp/time/latest/docs/functions/rfc3339_parse
+// See also https://github.com/hashicorp/terraform-provider-time/blob/69f9f399b3ba82a6cf8e03052b00aeeddb0b74e1/internal/provider/function_rfc3339_parse.go
```
sdavids: I am not sure if these links should be part of the documentation. Perhaps it is better to “copy” the relevant passages (if the licenses of the two projects are compatible).

A blob link in official documentation does not feel right to me.

crazy-max:
> One uses `SOURCE_DATE_EPOCH` as follows:
> 
> ```
> $ SOURCE_DATE_EPOCH=0 docker buildx bake
> $ SOURCE_DATE_EPOCH=42 docker buildx bake
> ```
> 
> Does this PR support this use case?
> 
> `SOURCE_DATE_EPOCH` is used to pass a fixed timestamp from the outside into the build.

This is already supported as a [standardized environment variable](https://reproducible-builds.org/docs/source-date-epoch/)

Edit: We only have docs for it in https://docs.docker.com/build/ci/github-actions/reproducible-builds/ but I think we should also have some for build and bake. Maybe in https://docs.docker.com/build/building/variables/#build-tool-configuration-variables. cc @ArthurFlag

sdavids:
Maybe I was not clear enough in #3197:

```console
$ SOURCE_DATE_EPOCH=0 docker buildx bake
```

```hcl
labels = {
    "org.opencontainers.image.created" = "how do I get SOURCE_DATE_EPOCH as an RFC3339-formatted string here"
}
```


The label should look like `"org.opencontainers.image.created"="1970-01-01T00:00:00.000Z"` and not `"org.opencontainers.image.created"="0"`

crazy-max:
> Maybe I was not clear enough in #3197:
> 
> ```
> $ SOURCE_DATE_EPOCH=0 docker buildx bake
> ```
> 
> ```hcl
> labels = {
>     "org.opencontainers.image.created" = "how do I get SOURCE_DATE_EPOCH as an RFC3339-formatted string here"
> }
> ```
> 
> The label should look like `"org.opencontainers.image.created"="1970-01-01T00:00:00.000Z"` and not `"org.opencontainers.image.created"="0"`

Pushed extra commit to add `unixtimestampparse` similar to terraform [`unix_timestamp_parse` function](https://registry.terraform.io/providers/hashicorp/time/latest/docs/functions/unix_timestamp_parse) and update examples in docs: https://github.com/crazy-max/buildx/blob/bake-rdc339parse/docs/bake-stdlib.md#-rfc3339parse

docs/bake-stdlib.md
```diff
@@ -118,11 +120,136 @@ title: Bake standard library functions
...
+```
+
+### <a name="rfc3339parse"></a> `rfc3339parse`
+
+The returned object has the following attributes:
+* `year` (Number) The year for the timestamp.
+* `year_day` (Number) The day of the year for the timestamp, in the range 1-365 for non-leap years, and 1-366 in leap years.
+* `day` (Number) The day of the month for the timestamp.
+* `month` (Number) The month of the year for the timestamp.
+* `month_name` (String) The name of the month for the timestamp (ex. "January").
```
sdavids: How does this work in different locales? 

january/Januar/janvier/Январь/มกราคม

Maybe not expose this attribute even if the underlying implementation supports it?

---

This comment also applies for `weekday_name` (and both in `unixtimestampparse`).

bake/hclparser/stdlib.go
```diff
@@ -254,6 +256,115 @@ func timestampFunc() function.Function {
...
+			"weekday_name": cty.String,
+			"hour":         cty.Number,
+			"minute":       cty.Number,
+			"second":       cty.Number,
+			"unix":         cty.Number,
+			"iso_year":     cty.Number,
+			"iso_week":     cty.Number,
+		})),
+		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+			ts := args[0].AsString()
```
sdavids: Is it possible for `args` to be empty?

bake/hclparser/stdlib.go
```diff
@@ -254,6 +256,115 @@ func timestampFunc() function.Function {
...
+			"weekday_name": cty.String,
+			"hour":         cty.Number,
+			"minute":       cty.Number,
+			"second":       cty.Number,
+			"rfc3339":      cty.String,
+			"iso_year":     cty.Number,
+			"iso_week":     cty.Number,
+		})),
+		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+			ts, _ := args[0].AsBigFloat().Int64()
```
sdavids: see above

bake/hclparser/stdlib_test.go
```diff
@@ -197,3 +197,58 @@ func TestSanitize(t *testing.T) {
...
+	}
+	for k, v := range expected {
+		require.True(t, got.GetAttr(k).RawEquals(v), "field %s: got %v, want %v", k, got.GetAttr(k), v)
+	}
+
+	_, err = fn.Call([]cty.Value{cty.StringVal("not-a-date")})
+	require.Error(t, err)
+}
+
+func TestUnixTimestampParseFunc(t *testing.T) {
```
sdavids, line 262: You might want to add tests for `-1`, `0`, `1.2`, and `"0"`.

bake/hclparser/stdlib.go
sdavids: You might want to also handle the error.

bake/hclparser/stdlib_test.go
```diff
@@ -197,3 +197,58 @@ func TestSanitize(t *testing.T) {
 		})
 	}
 }
+
+func TestRfc3339ParseFunc(t *testing.T) {
+	fn := rfc3339ParseFunc()
+	input := cty.StringVal("2024-06-01T15:04:05Z")
```
sdavids: You might want to add tests for `""` and an syntactically incorrect timestamp.

docs/bake-stdlib.md
```diff
@@ -118,11 +120,136 @@ title: Bake standard library functions
...
+```hcl
+# docker-bake.hcl
+variable "BUILD_TIME" {
+  type     = string
+  default  = timestamp()
+}
+target "default" {
+  args = {
+    BUILD_TIME = BUILD_TIME
+    SOURCE_DATE_EPOCH = rfc3339parse(BUILD_TIME).unix
```
sdavids, line 177: I would remove `SOURCE_DATE_EPOCH` from this example.

bake/hclparser/stdlib.go
crazy-max: No it would fail earlier because an arg is required for this func

bake/hclparser/stdlib.go
crazy-max: > You might want to also handle the error.

`.Int64()` does not return an error as second value but accuracy: https://pkg.go.dev/math/big#Float.Int64

sdavids: See my note for the tests … someone could conceivably supply the following values `-1`, `0`, `1.2`, and `"0"`

docs/bake-stdlib.md
```diff
@@ -118,11 +120,136 @@ title: Bake standard library functions
...
+* `hour` (Number) The hour within the day for the timestamp, in the range 0-23.
+* `minute` (Number) The minute offset within the hour for the timestamp, in the range 0-59.
+* `second` (Number) The second offset within the minute for the timestamp, in the range 0-59.
+* `unix` (Number) The number of seconds elapsed since January 1, 1970 UTC.
+* `iso_year` (Number) The ISO 8601 year number.
+* `iso_week` (Number) The ISO 8601 week number.
+
+```hcl
+# docker-bake.hcl
+variable "BUILD_TIME" {
```
tonistiigi: I don't think this is a convention that we would want to promote. Users should set `SOURCE_DATE_EPOCH` in the env as UNIX timestamp, not `BUILD_TIME` as string, so that `bake` uses the same convention as other apps.

docs/bake-stdlib.md
```diff
@@ -118,11 +120,136 @@ title: Bake standard library functions
...
+* `weekday` (Number) The day of the week for the unix timestamp.
+* `weekday_name` (String) The name of the day for the unix timestamp (ex. "Sunday").
+* `hour` (Number) The hour within the day for the unix timestamp, in the range 0-23.
+* `minute` (Number) The minute offset within the hour for the unix timestamp, in the range 0-59.
+* `second` (Number) The second offset within the minute for the unix timestamp, in the range 0-59.
+* `rfc3339` (String) The RFC3339 format string.
+* `iso_year` (Number) The ISO 8601 year number.
+* `iso_week` (Number) The ISO 8601 week number.
+
+```hcl
```
tonistiigi, line 1458: how about

```
variable "SOURCE_DATE_EPOCH" {
  type    = number
  default = formattimestamp("X", "2015-10-21T00:00:00Z")
}

target "default" {
  args = {
    SOURCE_DATE_EPOCH = SOURCE_DATE_EPOCH
    BUILD_TIME = formattimestamp("YYYY-MM-DD'T'00:00:00Z", SOURCE_DATE_EPOCH)
  }
}
```

I'm also ok for special formatting for rfc3399 for less verbosity.

tonistiigi (commented):
_(looks like github skipped my main review message, reposing in case it doesn't appear)_

Still somewhat confused about this.
- what is the use case for `rfc4449parse`? We would want to the `SOURCE_DATE_EPOCH` to be the source of truth that is picked up from env as that is the convention and that is already unix.
- Is returning a big struct like these new functions do some kind of convention. Can't find anything similar from current stdlib. The current convention seems to be `formatdate`. If we don't want to modify `formatdate` to have the capability to take date as UNIX timestamp (and maybe to return it) then would it make sense to add a new function, eg. `formattimestamp` that is the superset of the existing function (and then hide/deprecate the previous `formatdate`).

docs/bake-stdlib.md
crazy-max: Indeed, this was just for example purpose, renamed `SOURCE_DATE_EPOCH` to `BUILD_TIMESTAMP`.

docs/bake-stdlib.md
crazy-max: I looked at forking [`FormatDateFunc` func](https://github.com/zclconf/go-cty/blob/3c2b6a0e21d8e541c2ff5b3825b27483fad0bd1b/cty/function/stdlib/datetime.go#L14) as they are not willing to accept timestamp format: https://github.com/zclconf/go-cty/issues/130#issuecomment-1215903779 but there's quite a lot to fork and it might be better to be aligned so users are not confused. So that's why I opted for the terraform [`unix_timestamp_parse` function](https://registry.terraform.io/providers/hashicorp/time/latest/docs/functions/unix_timestamp_parse) func instead.

crazy-max:
> _(looks like github skipped my main review message, reposing in case it doesn't appear)_
> 
> Still somewhat confused about this.
> 
> * what is the use case for `rfc4449parse`? We would want to the `SOURCE_DATE_EPOCH` to be the source of truth that is picked up from env as that is the convention and that is already unix.

Removed `rfc4449parse` func and just keep `unixtimestampparse` one to be consistent with Terraform [`unix_timestamp_parse` function](https://registry.terraform.io/providers/hashicorp/time/latest/docs/functions/unix_timestamp_parse) func.

> 
> * Is returning a big struct like these new functions do some kind of convention. Can't find anything similar from current stdlib. The current convention seems to be `formatdate`. If we don't want to modify `formatdate` to have the capability to take date as UNIX timestamp (and maybe to return it) then would it make sense to add a new function, eg. `formattimestamp` that is the superset of the existing function (and then hide/deprecate the previous `formatdate`).

Yes I think as follow-up we could have a `formattimestamp` func.

docs/bake-stdlib.md
crazy-max: Updated with new `formattimestamp` func

crazy-max:
> It looks like the ability to return unix timestamp from `formattimestamp` [#3286 (comment)](https://github.com/docker/buildx/pull/3286#discussion_r2205693153) is missing. I think if we are adding a new function, we should add that capability as well.

Done, see last commit

PR #3292, opened 2025-07-03
bake: ignore unrelated fields when parsing compose files

### LINKED ISSUE #3288: bake: only validate build property in compose file services
relates to:
* https://github.com/docker/buildx/issues/612
* https://github.com/docker/buildx/issues/718
* https://github.com/docker/buildx/issues/1106
* https://github.com/docker/buildx/issues/1187
* https://github.com/docker/buildx/issues/2241
* https://github.com/docker/buildx/issues/2548

When new properties are added to the compose spec, Bake might fail to build them as the validation is not yet aware of them unless the compose-go dependency is updated and a new release of Buildx made. We should just validate the `build` property that is the only part we are using with bake to prevent such issues.

@ndeloof @glours Would it be possible to have an option in the loader to just validate specific properties?: https://github.com/docker/buildx/blob/758ea75f60d86ef67289c3b70e0f27f2b1569a69/bake/compose.go#L44-L52

Maybe smth like `ValidateBuildOnly` or `SkipUnknown`?

crazy-max:
fixes #3288


bake/compose.go
```diff
@@ -208,6 +195,67 @@ func ParseCompose(cfgs []composetypes.ConfigFile, envs map[string]string) (*Conf
...
+	for _, key := range []string{"services", "secrets"} {
+		if key == "services" {
+			if services, ok := raw["services"].(map[string]any); ok {
+				filteredServices := make(map[string]any)
+				for svcName, svc := range services {
+					if svc == nil {
+						filteredServices[svcName] = map[string]any{}
+					} else if svcMap, ok := svc.(map[string]any); ok {
+						filteredService := make(map[string]any)
+						for _, svcField := range []string{"image", "build", "environment", "env_file"} {
```
ndeloof, line 231: AFAICT you don't need `environment` or `env_file` as those won't have any impact on the build context (they only apply on runtime environment)

bake/compose.go
```diff
@@ -208,6 +195,67 @@ func ParseCompose(cfgs []composetypes.ConfigFile, envs map[string]string) (*Conf
...
+		}
+	}
+
+	if err := composevalidation.Validate(filtered); err != nil {
+		return nil, err
+	}
+
+	return loader.ModelToProject(filtered, loader.ToOptions(&cfgDetails, append([]func(*loader.Options){func(options *loader.Options) {
+		options.SkipNormalization = true
+		options.Profiles = []string{"*"}
```
ndeloof, line 252: building with all profiles enabled is not consistent with Compose. Is this expected ?

bake/compose.go
crazy-max: We need them to evaluate build arguments: https://github.com/docker/buildx/blob/a00736872905b6204b5ab56584a7f4cb5cc12259/bake/compose_test.go#L396-L424

If not set then this test fails:

```
        	Error:      	Not equal: 
        	            	expected: map[string]*string{"CT_ECR":(*string)(0xc00060a510), "FOO":(*string)(0xc00060a520), "NODE_ENV":(*string)(0xc00060a530)}
        	            	actual  : map[string]*string{"CT_ECR":(*string)(0xc00060a4a0)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,5 +1,3 @@
        	            	-(map[string]*string) (len=3) {
        	            	- (string) (len=6) "CT_ECR": (*string)((len=3) "foo"),
        	            	- (string) (len=3) "FOO": (*string)((len=10) "bsdf -csdf"),
        	            	- (string) (len=8) "NODE_ENV": (*string)((len=4) "test")
        	            	+(map[string]*string) (len=1) {
        	            	+ (string) (len=6) "CT_ECR": (*string)((len=3) "foo")
        	            	 }
        	Test:       	TestEnv
--- FAIL: TestEnv (0.00s)
```

ndeloof: There's a bug then. Runtime environment MUST not impact build environment; Sounds there's a confusion here with client environment (which is defined by `.env` file and `--env-file` flag in Compose) ... which is unfortunately a pretty common source of confusion 😭

bake/compose.go
crazy-max: We allow all profiles because Bake doesn't support them: https://github.com/docker/buildx/pull/1903. If we add something similar in our spec I think we could reconsider that. cc @colinhemmings

bake/compose.go
crazy-max: Argh yes seems it was a mistake, this was introduced in https://github.com/docker/buildx/pull/905. What about `.env` then?

ndeloof: For parity with Compose, `.env` should be loaded (from compose file's parent folder), and interpolated using os.Env

bake/compose_test.go
```diff
@@ -888,6 +878,39 @@ services:
...
+services:
+  webapp:
+    build:
+     context: .
+     foo: bar
+`)
+
+	chdir(t, tmpdir)
+	_, err := ParseComposeFiles([]File{{Name: "compose.yml", Data: dt}})
+	require.Error(t, err)
```
tonistiigi, line 912: Can we check something specific about the error, eg. should contain the name of the unknown key.

bake/compose.go
```diff
@@ -208,6 +195,67 @@ func ParseCompose(cfgs []composetypes.ConfigFile, envs map[string]string) (*Conf
...
+	if envs == nil {
+		envs = make(map[string]string)
+	}
+
+	cfgDetails := composetypes.ConfigDetails{
+		ConfigFiles: cfgs,
+		Environment: envs,
+	}
+
+	raw, err := loader.LoadModelWithContext(context.Background(), cfgDetails, append([]func(*loader.Options){func(opts *loader.Options) {
```
tonistiigi, line 208: We can do this as follow-up as not directly related but seems fine to pass in the right context from the command.

bake/compose.go
tonistiigi: In follow-up we could consider mapping compose profiles to bake groups and maybe defining a default profile that would be used if one is defined.

PR #3294, opened 2025-07-03
dap: support stopOnEntry to configure behavior when starting the debugger

jsternberg:
This will configure the default behavior when beginning to evaluate a
build target. When `stopOnEntry` is used, it will default to `stepNext`.
Otherwise, `stepContinue` will be used.


ktock:
Needs rebase. Changes look good to me.

PR #3302, opened 2025-07-08
dap: implement first pass at breakpoints

jsternberg:
Implement the first iteration of breakpoints. When a breakpoint is set,
it starts unverified. When a thread begins evaluation, it tries to see
if a breakpoint corresponds to one of the parsed instructions and will
verify it.

Breakpoints work when continue is used.

At the current moment, setting breakpoints while a thread is currently
running doesn't work. Breakpoints are rechecked each time execution is
about to restart.

### DIFF
--- a/dap/adapter.go
+++ b/dap/adapter.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"io"
 	"path"
+	"path/filepath"
+	"slices"
 	"sync"
 	"sync/atomic"
 
@@ -14,6 +16,8 @@ import (
 	"github.com/docker/buildx/dap/common"
 	"github.com/google/go-dap"
 	gateway "github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/moby/buildkit/solver/pb"
+	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
 )
@@ -33,8 +37,9 @@ type Adapter[C LaunchConfig] struct {
 	threadsMu    sync.RWMutex
 	nextThreadID int
 
-	sourceMap sourceMap
-	idPool    *idPool
+	breakpointMap *breakpointMap
+	sourceMap     sourceMap
+	idPool        *idPool
 }
 
 func New[C LaunchConfig]() *Adapter[C] {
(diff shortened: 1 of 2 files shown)


jsternberg:
It turned out to be way easier to reset the breakpoints when they changed during execution than I thought it would be. I thought it would need to be a different PR since I thought it would involve some message passing but it turns out it was just easy to check the breakpoints before each time continue was used.

dap/thread.go
```diff
@@ -110,8 +119,13 @@ func (t *thread) needsDebug(target digest.Digest, step stepType, err error) (rea
 	if err != nil {
 		reason = "exception"
 		desc = "Encountered an error during result evaluation"
-	} else if target != "" && step == stepNext {
+	} else if step == stepNext && target != "" {
 		reason = "step"
+	} else if step == stepContinue {
+		if _, ok := t.bps[target]; ok {
```
ktock: Doesn't the breakpoint id need to be passed to the editor via `StoppedEvent.hitBreakpointIds`?

jsternberg: Fixed.

ktock (approved):
Thanks

dap/adapter.go
```diff
@@ -468,3 +468,100 @@ func (s *sourceMap) Get(fname string) ([]byte, bool) {
...
+	return breakpoints
+}
+
+func (b *breakpointMap) Intersect(ctx Context, src *pb.Source, ws string) map[digest.Digest]int {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	digests := make(map[digest.Digest]int)
+
+NEXT_DIGEST:
```
tonistiigi: style: no caps-only

dap/thread.go
```diff
@@ -426,15 +457,26 @@ func (t *thread) nextDigest() digest.Digest {
...
 		return next
 	}
 }
 
+func (t *thread) continueDigest() digest.Digest {
+	if len(t.bps) == 0 {
+		return t.head
+	}
+
+	fn := func(dgst digest.Digest) bool {
```
tonistiigi: style: inline this function, or call it `isValid`

dap/adapter.go
jsternberg: I just refactored this to put the inner for loop into another function to avoid the need for a labeled continue.

PR #3307, opened 2025-07-10
dap: implement variable references

jsternberg:
Implement variable references to inspect the state of a stack frame.

Variable reference ids are composed of two sections. A thread mask that
is the first 8 bytes and the remainder is an increasing number that gets
reset each time a thread is resumed. This allows the adapter to know
which thread to delegate the variables request to and allows the
variable references to still remain confined to each thread. An int32 is
used for this because variable references need to be in the range of
(0, 2^32).

At the moment, only the platform variables and some of the exec
operations for an operation. These are labeled as "arguments" to the
stack frame.

### DIFF
--- a/dap/adapter.go
+++ b/dap/adapter.go
@@ -216,6 +216,7 @@ func (d *Adapter[C]) newThread(ctx Context, name string) (t *thread) {
 		sourceMap:     &d.sourceMap,
 		breakpointMap: d.breakpointMap,
 		idPool:        d.idPool,
+		variables:     newVariableReferences(),
 	}
 	d.threads[t.id] = t
 	d.nextThreadID++
@@ -240,6 +241,11 @@ func (d *Adapter[C]) getThread(id int) (t *thread) {
 
 func (d *Adapter[C]) deleteThread(ctx Context, t *thread) {
 	d.threadsMu.Lock()
+	if t := d.threads[t.id]; t != nil {
+		if t.variables != nil {
+			t.variables.Reset()
+		}
+	}
 	delete(d.threads, t.id)
 	d.threadsMu.Unlock()
 
@@ -250,6 +256,18 @@ func (d *Adapter[C]) deleteThread(ctx Context, t *thread) {
 			ThreadId: t.id,
 		},
 	}
+}
+
+func (d *Adapter[T]) getThreadByFrameID(id int) (t *thread) {
+	d.threadsMu.RLock()
+	defer d.threadsMu.RUnlock()
+
+	for _, t := range d.threads {
+		if t.hasFrame(id) {
+			return t
+		}
+	}
+	return nil
 }
 
 type evaluateRequest struct {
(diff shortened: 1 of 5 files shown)


dap/variables.go
```diff
@@ -0,0 +1,208 @@
...
+		},
+	}
+}
+
+func platformVars(platform *pb.Platform, refs *variableReferences) dap.Variable {
+	return dap.Variable{
+		Name:  "platform",
+		Value: fmt.Sprintf("%s/%s", platform.OS, platform.Architecture),
+		VariablesReference: refs.New(func() []dap.Variable {
+			vars := []dap.Variable{
```
tonistiigi, line 73: Why split these apart? They are also part of the same field when configuring build.

dap/variables.go
```diff
@@ -0,0 +1,208 @@
...
+	return dap.Variable{
+		Name:  "exec",
+		Value: strings.Join(exec.Meta.Args, " "),
+		VariablesReference: refs.New(func() []dap.Variable {
+			vars := []dap.Variable{
+				{
+					Name:  "args",
+					Value: brief(strings.Join(exec.Meta.Args, " ")),
+					VariablesReference: refs.New(func() []dap.Variable {
+						vars := make([]dap.Variable, 0, len(exec.Meta.Args))
```
tonistiigi, line 112: Maybe missing something but why are args split into separate vars?

dap/adapter.go
```diff
@@ -330,6 +348,35 @@ func (d *Adapter[C]) StackTrace(c Context, req *dap.StackTraceRequest, resp *dap
...
+func (d *Adapter[C]) Variables(c Context, req *dap.VariablesRequest, resp *dap.VariablesResponse) error {
+	tid := req.Arguments.VariablesReference >> 24
+
+	t := d.getThread(tid)
+	if t == nil {
+		return errors.Errorf("no such thread: %d", tid)
+	}
+
+	varRef := req.Arguments.VariablesReference & ((1 << 24) - 1)
+	resp.Body.Variables = t.Variables(varRef)
```
ktock, line 373: I think it's safe to set an empty slice (`make([]dap.Variable, 0)`) as the default value to avoid being marshalled to null (related discussion: https://github.com/mfussenegger/nvim-dap/issues/225 )

dap/adapter.go
```diff
@@ -330,6 +348,35 @@ func (d *Adapter[C]) StackTrace(c Context, req *dap.StackTraceRequest, resp *dap
...
 
+func (d *Adapter[C]) Scopes(c Context, req *dap.ScopesRequest, resp *dap.ScopesResponse) error {
+	t := d.getThreadByFrameID(req.Arguments.FrameId)
+	if t == nil {
+		return errors.Errorf("no such frame id: %d", req.Arguments.FrameId)
+	}
+
+	resp.Body.Scopes = t.Scopes(req.Arguments.FrameId)
+	for i, s := range resp.Body.Scopes {
+		resp.Body.Scopes[i].VariablesReference = (t.id << 24) | s.VariablesReference
```
ktock, line 359: Shouldn't we check the overflow?

jsternberg: I guess we can but I figure this is probably pretty low risk and it would just create some weird UI artifacts if there was an accidental conflict.

The main reason for this is because I haven't created a way where the threads will synchronize on pause and I didn't want to micromanage the variable references in a global container. This isn't currently an issue since, with build, there will only ever be one thread. But, since we want to add support for bake debugging too, we'll either need to allow multiple threads to exist or to synchronize them.

I figure the only way this overflows is if there are over 256 targets (only possible with bake) or over 2^24 variables (16,777,216). I figure if you're debugging something with these numbers you'll probably break the UI first.

dap/variables.go
jsternberg: It's so they show up in the debugger like this:

Screenshot 2025-07-14 at 9 42 08 AM

That way you can directly inspect the specific indexes of the array.

dap/variables.go
jsternberg: It ends up looking like this. So the first value is the "summary value" and then you can click further in and get individual variables.

Screenshot 2025-07-14 at 9 43 29 AM

tonistiigi (approved):

PR #3308, opened 2025-07-11
docs: add docs related to dap

jsternberg:
Adds some entry-level and developer-friendly docs for the debug adapter.
The one in `docs/dap.md` is meant for someone trying to use the debugger
while the one in `docs/reference/buildx_dap_build.md` is more focused on
documenting the command to be integrated in a debugger extension.


docs/dap.md
crazy-max, line 1: I think we want tthis page on docs repo instead. Probably under debugging section: 

image

Similar to OpenTelemetry: https://docs.docker.com/build/debug/opentelemetry/

docs/reference/buildx_dap_build.md
```diff
@@ -50,3 +46,42 @@ Start a build
...
+Arguments are the same as the `build`
+
+> [!NOTE]
+> `buildx dap build` command may receive backwards incompatible features in the future
+> if needed. We are looking for feedback on improving the command and extending
+> the functionality further.
+
+## Examples
+
+### <a name="launch-config"></a> Launch request arguments
```
crazy-max, line 62: We set anchor manually only if they match a flag in `Options` section so they are taken into account during docs generation. So this is not necessary here.

docs/reference/buildx_dap_build.md
```diff
@@ -50,3 +46,42 @@ Start a build
...
+The following [launch request arguments](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Launch) are supported. These are sent as a JSON body as part of the launch request.
+
+| Name                | Type          | Default      | Description                                                                  |
+|:--------------------|:--------------|:-------------|:-----------------------------------------------------------------------------|
+| `dockerfile`        | `string`      | `Dockerfile` | Name of the Dockerfile                                                       |
+| `contextPath`       | `string`      | `.`          | Set the context path for the build (normally the first positional argument)  |
+| `target`            | `string`      |              | Set the target build stage to build                                          |
+| `stopOnEntry`       | `boolean`     | `false`      | Stop on the first instruction                                                |
+
+### <a name="additional-args"></a>
```
crazy-max: Same here and it's missing the title

docs/reference/buildx_dap_build.md
```diff
@@ -50,3 +46,42 @@ Start a build
...
+Start a debug session using the [debug adapter protocol](https://microsoft.github.io/debug-adapter-protocol/overview) to communicate with the debugger UI.
+
+Arguments are the same as the `build`
+
+> [!NOTE]
+> `buildx dap build` command may receive backwards incompatible features in the future
+> if needed. We are looking for feedback on improving the command and extending
+> the functionality further.
+
+## Examples
```
crazy-max, line 60: Examples are usually attached to flags but I think this is fine for dap command

commands/dap.go
```diff
@@ -23,6 +23,10 @@ func dapCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 	dapBuildCmd := buildCmd(dockerCli, rootOpts, &options)
 	dapBuildCmd.Args = cobra.RangeArgs(0, 1)
+
+	// Remove aliases and annotations for documentation.
+	dapBuildCmd.Aliases = nil
+	dapBuildCmd.Annotations = nil
```
crazy-max: We should not unset annotations otherwise command will not be marked as experimental through `cobrautil.MarkCommandExperimental(cmd)` line 22, see https://github.com/docker/buildx/blob/d5b9564c04a40a55848c31412bf2ad329306ef62/util/cobrautil/cobrautil.go#L44-L53

crazy-max (approved):

tonistiigi (approved):
Let's get this in for the repo side. For the GA release we should have something more user focused with vscode flows.

PR #3309, opened 2025-07-11
dap: alias step into to next and step out to continue

jsternberg:
Step into and step out are required by the UI for DAP. We don't have a
way to implement these in a logical manner but they need to exist. We'll
discuss in further iterations how these might differ from next and
continue, but for now, we just need some implementation for the UI.

### DIFF
--- a/dap/adapter.go
+++ b/dap/adapter.go
@@ -156,6 +156,29 @@ func (d *Adapter[C]) Next(c Context, req *dap.NextRequest, resp *dap.NextRespons
 	return nil
 }
 
+func (d *Adapter[C]) StepIn(c Context, req *dap.StepInRequest, resp *dap.StepInResponse) error {
+	var (
+		subReq  dap.NextRequest
+		subResp dap.NextResponse
+	)
+
+	subReq.Arguments.ThreadId = req.Arguments.ThreadId
+	subReq.Arguments.SingleThread = req.Arguments.SingleThread
+	subReq.Arguments.Granularity = req.Arguments.Granularity
+	return d.Next(c, &subReq, &subResp)
+}
+
+func (d *Adapter[C]) StepOut(c Context, req *dap.StepOutRequest, resp *dap.StepOutResponse) error {
+	var (
+		subReq  dap.ContinueRequest
+		subResp dap.ContinueResponse
+	)
+
+	subReq.Arguments.ThreadId = req.Arguments.ThreadId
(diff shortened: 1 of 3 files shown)


dap/adapter.go
```diff
@@ -156,6 +156,37 @@ func (d *Adapter[C]) Next(c Context, req *dap.NextRequest, resp *dap.NextRespons
...
+func (d *Adapter[C]) StepOut(c Context, req *dap.StepOutRequest, resp *dap.StepOutResponse) error {
+	var (
+		subReq  dap.ContinueRequest
+		subResp dap.ContinueResponse
+	)
+
+	subReq.Arguments.ThreadId = req.Arguments.ThreadId
+	subReq.Arguments.SingleThread = req.Arguments.SingleThread
+
+	if err := d.Continue(c, &subReq, &subResp); err != nil {
```
ktock: Implementation looks good to me. 
But if there is no concept like step-in and step-out in dockerfile, maybe leaving them as nop is enough ?

jsternberg: They show up in the visual studio code UI and don't seem to be behind a capability so I thought aliasing them was probably easiest for now. If also like to get done feedback from users about how they feel these buttons should work.

If it doesn't make sense we can leave it out but I figured that step in is the equivalent of next anyway if there's no function and step out is the equivalent of continue if you consider llb to be one big main function.

ktock: Thanks for the explanation, making them as aliases sounds good to me.

dap/adapter.go
```diff
@@ -156,6 +156,37 @@ func (d *Adapter[C]) Next(c Context, req *dap.NextRequest, resp *dap.NextRespons
...
+	)
+
+	subReq.Arguments.ThreadId = req.Arguments.ThreadId
+	subReq.Arguments.SingleThread = req.Arguments.SingleThread
+	subReq.Arguments.Granularity = req.Arguments.Granularity
+
+	if err := d.Next(c, &subReq, &subResp); err != nil {
+		return err
+	}
+	return nil
```
ktock: nit: This can be simplified as:

```golang
return d.Next(c, &subReq, &subResp)
````

dap/adapter.go
ktock: ditto

ktock (approved):

tonistiigi (approved):

PR #3314, opened 2025-07-14
dap: always return the error from execution if we paused on an error and resume

jsternberg:


rcjsuen (commented):
I made a build off of this branch and it worked for me. 👍

tonistiigi (approved):

PR #3318, opened 2025-07-15
vendor: update hcl dependencies

crazy-max:
* https://github.com/hashicorp/go-cty-funcs/compare/dda779884a9f...6aab671309282ceb3432b07f94658c3f61d83a46
* https://github.com/hashicorp/hcl/compare/v2.23.0...v2.24.0
* https://github.com/zclconf/go-cty/compare/v1.16.2...v1.17.0


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/docker/buildx
 
-go 1.23.0
+go 1.24.0
```
crazy-max: Related to https://github.com/hashicorp/go-cty-funcs/blob/af775f281b148e30893d65a3712b9d362e711260/go.mod#L3

@thaJeztah I know we want to keep compat with n-1 Go release. Does it also applies to cli plugins?

crazy-max: With Go 1.24 set in `go.mod` we now have some gopls issues revealed: https://github.com/docker/buildx/actions/runs/16295155457/job/46014741150?pr=3318#step:4:507

tonistiigi (approved):

PR #3322, opened 2025-07-21
do not assume input is a compose file on .env parsing error

ndeloof:
close https://github.com/docker/compose/issues/13070


crazy-max (approved):
LGTM thanks!

PR #3324, opened 2025-07-21
driver/kubernetes: remove uses of pkg/homedir

thaJeztah:
Create a local fork to keep the existing behavior.


thaJeztah:
Removed or "moved" - see https://github.com/moby/moby/pull/50446

We need to look closely how to handle that.

crazy-max:
> Removed or "moved" - see [moby/moby#50446](https://github.com/moby/moby/pull/50446)
> 
> We need to look closely how to handle that.

Ah yes I see https://github.com/moby/moby/pull/50446#discussion_r2218965213

PR #3325, opened 2025-07-21
dap: refactor how step in/step out works

jsternberg:
Change how breakpoints and stepping works. These now work more how you
would expect another programming language to work. Breakpoints happen
before the step has been invoked rather than after which means you can
inspect the state before the command runs.

This has the advantage of being more intuitive for someone familiar with
other debuggers. The negative is that you can't run to after a certain
step as easily as you could before. Instead, you would run to that stage
and then use next to go to the step directly afterwards.

Step in and out also now have different behaviors. When a step has
multiple inputs, the inputs of non-zero index are considered like
"function calls". The most common cause of this is to use `COPY --from`
or a bind mount. Stepping into these will cause it to jump to the
beginning of the call chain for that branch. Using step out will exit
back to the location where step in was used.

This change also makes it so some steps may be invoked multiple times in
the callgraph if multiple steps depend on them. The reused steps will
still be cached, but you may end up stepping through more lines than the
previous implementation.

Stack traces now represent where these step in and step out areas
happen rather than the previous steps. This can help you know from where
a certain step is being used.

### DIFF
--- a/dap/thread.go
+++ b/dap/thread.go
@@ -3,7 +3,6 @@ package dap
 import (
 	"context"
 	"path/filepath"
-	"slices"
 	"sync"
 
 	"github.com/docker/buildx/build"
@@ -40,9 +39,11 @@ type thread struct {
 	head digest.Digest
 	bps  map[digest.Digest]int
 
+	frames         map[int32]*frame
+	framesByDigest map[digest.Digest]*frame
+
 	// Runtime state for the evaluate call.
-	regions         []*region
-	regionsByDigest map[digest.Digest]int
+	entrypoint *step
 
 	// Controls pause.
 	paused chan stepType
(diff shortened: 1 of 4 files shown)


dap/thread.go
```diff
@@ -261,233 +376,119 @@ func (t *thread) setBreakpoints(ctx Context) {
...
+				// Iterate from the first step to find the one
+				// we failed on.
+				result = t.entrypoint
+				for result != nil {
+					next := result.in
+					if next != nil && next.dgst == errDgst {
+						break
+					}
+					result = next
+				}
```
ktock: nit: no need to return an error if not found?

jsternberg: I don't think so. The main reason for returning the step is to be able to include the correct location information in the stack trace. I don't think there's really an effective way to return an error here. It might be useful to add some more debugging statements and a way to enable debugging for the dap adapter in the future.

PR #3327, opened 2025-07-21
dap: filesystem inspection when paused on a digest

jsternberg:
Add a file explorer to the debugger that allows exploring the filesystem
of the current container. It will show directory contents, file
contents, and symlink destinations. It will also show the file mode
associated with a file.

The file explorer defaults to marking itself as an expensive operation
so the debugger doesn't automatically retrieve the information.

Screenshot 2025-07-21 at 1 49 21 PM
Screenshot 2025-07-25 at 2 44 49 PM

### DIFF
--- a/dap/variables.go
+++ b/dap/variables.go
@@ -1,16 +1,22 @@
 package dap
 
 import (
+	"context"
 	"fmt"
+	"io/fs"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
+	"unicode/utf8"
 
 	"github.com/google/go-dap"
 	"github.com/moby/buildkit/client/llb"
+	gateway "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/solver/pb"
+	"github.com/tonistiigi/fsutil/types"
 )
 
 type frame struct {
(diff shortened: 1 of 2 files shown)


jsternberg:
I'll have to try and fix that. The reason is because it's trying to reevaluate the build that just failed to read the filesystem which, predictably, fails. Need to grab the filesystem from right before the failure happens.

rcjsuen:
> I'll have to try and fix that. The reason is because it's trying to reevaluate the build that just failed to read the filesystem which, predictably, fails. Need to grab the filesystem from right before the failure happens.

If we can't show anything that's fine too. I think either a) the tree node shouldn't even appear or b) clearly state why the tree node is empty with some appropriately worded "error message" reason.

jsternberg:
I think we can but it might be easier if I build this on top of the change in #3325 since the method I would use to get the state before the error is already done in that PR but isn't done on master so I'd have to redo it here in a different way.

dap/variables.go
```diff
@@ -154,7 +169,59 @@ func execOpVars(exec *pb.ExecOp, refs *variableReferences) dap.Variable {
...
+		paths[i] = fs
+	}
+	return paths
+}
+
+func statf(st *types.Stat) string {
+	mode := fs.FileMode(st.Mode)
+	modTime := time.Unix(0, st.ModTime).UTC()
+	return fmt.Sprintf("%s\t%d\t%d\t%s", mode, st.Uid, st.Gid, modTime.Format("Jan 2 15:04:05 2006"))
+}
```
rcjsuen: The tabs might be hit-or-miss depending on the client... 🤔

image

jsternberg: Hm... I don't think the tabs are likely a good idea. I'll look into another way of representing this.

jsternberg:
I've updated the PR. Files will now contain their content. The file reading doesn't currently have any restrictions, but it uses a basic heuristic that mimics Perl to determine whether something is a binary file and avoids reading those. Symlinks will show where they link to. File nodes show up as a tab to open. Directories will have their contents and the mode and other stat information for the directory entry are attached to a fake `.` entry.

I think this should end up being both easy to use and also provide enough information to limit the need for using `exec` to check files.

dap/variables.go
```diff
@@ -159,6 +170,137 @@ func execOpVars(exec *pb.ExecOp, refs *variableReferences) dap.Variable {
...
+						return statVars(file)
+					}),
+				}
+				return append([]dap.Variable{dvar}, fsVars(ctx, ref, fullpath, vars)...)
+			})
+			fv.Value = ""
+		} else {
+			fv.Value = stat
+			fv.VariablesReference = vars.New(func() (dvars []dap.Variable) {
+				if fs.FileMode(file.Mode)&fs.ModeType == 0 {
```
ktock: nit: [`IsRegular()` ](https://pkg.go.dev/io/fs#FileMode.IsRegular) can be used here

dap/variables.go
```diff
@@ -159,6 +170,137 @@ func execOpVars(exec *pb.ExecOp, refs *variableReferences) dap.Variable {
...
+	} else if isBinaryData(b) {
+		data = "binary data"
+	} else {
+		if len(b) == 512 {
+			// Get the remainder of the file.
+			remaining, err := ref.ReadFile(ctx, gateway.ReadRequest{
+				Filename: fullpath,
+				Range:    &gateway.FileRange{Offset: 512},
+			})
+			if err == nil {
```
ktock: Maybe the error should be printed on the editor via using an additional field?

dap/variables.go
jsternberg: Done.

dap/variables.go
jsternberg: I added the error as an additional field.

rcjsuen:
Thanks for the updates, Jonathan. It looks like this now.

image

crazy-max (approved):

Issue #3321, opened 2025-07-20
Add an option to override the `BuildName`


BarelElbaz:
### Description

The assumption that the contextPath / dockerfile paths should be the docker build name is not always true.

In some rare cases we have generic dockerfiles with wide context so their context folder means nothing (it's called `src` or `software`)

jsternberg:
Can you give an example of what you mean? Like a command that you run with a dockerfile, what it does, and then what you would want it to do? Are you referring to the build history?

crazy-max:
@jsternberg I think he's referring to the Builds view in Docker Desktop: https://docs.docker.com/desktop/use-desktop/builds/

<img width="818" height="707" alt="Image" src="https://github.com/user-attachments/assets/c0d1dbee-ab5a-48dc-b6d2-4d111150096a" />

tonistiigi:
@crazy-max I think we can define a build-arg that is picked up for the name.

BarelElbaz:
That's exactly what I meant! Thank you very much for this.

PR #3330, opened 2025-07-22
history: use built-in build-arg to override the build name

crazy-max:
fixes https://github.com/docker/buildx/issues/3321


crazy-max:
Needs docs follow-up in https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#buildkit-built-in-build-args

PR #3334, opened 2025-07-23
bump compose-go to v2.8.0

glours:
This release add the support of `sbom` and `provenance` attributes
Added tests checking correct parsing of attestations

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/aws/aws-sdk-go-v2/config v1.27.27
-	github.com/compose-spec/compose-go/v2 v2.7.2-0.20250703132301-891fce532a51 // main
+	github.com/compose-spec/compose-go/v2 v2.8.0
 	github.com/containerd/console v1.0.5
 	github.com/containerd/containerd/v2 v2.1.3
 	github.com/containerd/continuity v0.4.5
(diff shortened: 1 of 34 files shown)


go.mod
```diff
@@ -158,6 +158,7 @@ require (
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.35.0 // indirect
 	go.opentelemetry.io/otel/sdk/metric v1.35.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.5.0 // indirect
+	go.yaml.in/yaml/v3 v3.0.4 // indirect
```
crazy-max: Ah looks like you are using this fork now but we are still importing `gopkg.in/yaml.v3` in https://github.com/docker/buildx/blob/7d0efdc50e218f3ffa97ba0d77a4c0d4fcf6dd57/bake/compose.go#L20

Can you change it there?

Also cc @thaJeztah @tonistiigi I see we have an indirect dep on https://github.com/moby/buildkit/blob/0396f3eda2c756696cfc773dc2cc15845891176b/go.mod#L185 so we have two sets for yaml module now. Didn't look yet which package imports it.

glours: Done!

crazy-max (approved):
LGTM thanks!

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/aws/aws-sdk-go-v2/config v1.27.27
-	github.com/compose-spec/compose-go/v2 v2.7.2-0.20250703132301-891fce532a51 // main
+	github.com/compose-spec/compose-go/v2 v2.8.0 // main
```
crazy-max: Ah can you remove `// main`, we put that because we used commit sha from main branch previously.

```suggestion
	github.com/compose-spec/compose-go/v2 v2.8.0
```

glours: Done

PR #3338, opened 2025-07-24
bake: Support preservation of `.git` directory for remote context

marxarelli:
(updated implementation)

When a remote Git URL is used with `buildx bake`, `.git` directories are
always omitted (see `bake.ReadRemoteFiles`).

Targets that rely on the default bake context should still be able to
change this behavior by passing the `BUILDKIT_CONTEXT_KEEP_GIT_DIR=1`
build argument. Check for this argument and skip the use of the bake
context state if the value is set to true.

---

(original implementation)

When a remote URL is used with `buildx bake`, the resulting remote context is reused as the default main context of bake targets. Git remote contexts are handled using `dockerui.DetectGitContext` which accepts a `keepGitDir` argument and uses that for the `llb.KeepGitDir` Git source option.

Prior to this change, the value for `keepGitDir` was hardcoded as `false`, resulting in a main context for targets that never retains the `.git` directory from the cloned repo even where
`BUILDKIT_CONTEXT_KEEP_GIT_DIR` is defined as a build argument in a bake target definition.

Support preservation of the `.git` directory in main target contexts when using `buildx bake` with remote Git URLs via a new environment variable `BUILDX_BAKE_KEEP_GIT_DIR`. This mirrors the behavior of `BUILDKIT_CONTEXT_KEEP_GIT_DIR` for build targets and is keeping with the convention of other environment variables that effect remote bake contexts such as `BUILD_BAKE_GIT_SSH`.

Note that for a target to retain the `.git` directory, both the `BUILDX_BAKE_KEEP_GIT_DIR` environment variable and `BUILDKIT_CONTEXT_KEEP_GIT_DIR` build argument must be set to a true value, the latter being specific to each target that inherits the remote Git context.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -1274,7 +1274,7 @@ func TargetsToBuildOpt(m map[string]*Target, inp *Input) (map[string]build.Optio
 	return m2, nil
 }
 
-func updateContext(t *build.Inputs, inp *Input) {
+func updateContext(t *build.Inputs, inp *Input, target *Target) {
 	if inp == nil || inp.State == nil {
 		return
 	}
@@ -1282,6 +1282,7 @@ func updateContext(t *build.Inputs, inp *Input) {
 	for k, v := range t.NamedContexts {
 		if v.Path == "." {
 			t.NamedContexts[k] = build.NamedContext{Path: inp.URL}
+			continue
 		}
 		if strings.HasPrefix(v.Path, "cwd://") || strings.HasPrefix(v.Path, "target:") || strings.HasPrefix(v.Path, "docker-image:") {
 			continue
@@ -1303,6 +1304,16 @@ func updateContext(t *build.Inputs, inp *Input) {
 	if build.IsRemoteURL(t.ContextPath) {
 		return
 	}
+
+	// If the target includes a BUILDKIT_CONTEXT_KEEP_GIT_DIR=1 build argument,
+	// do not reuse the bake context state as it never keeps the .git directory
+	// (see ReadRemoteFiles).
+	if v, ok := target.Args["BUILDKIT_CONTEXT_KEEP_GIT_DIR"]; ok && v != nil {
+		if keepGitDir, err := strconv.ParseBool(*v); err == nil && keepGitDir {
+			return
+		}
+	}
+
 	st := llb.Scratch().File(
 		llb.Copy(*inp.State, t.ContextPath, "/", &llb.CopyInfo{
 			CopyDirContentsOnly: true,
@@ -1383,7 +1394,7 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 	if t.DockerfileInline != nil {
 		bi.DockerfileInline = *t.DockerfileInline
 	}
-	updateContext(&bi, inp)
+	updateContext(&bi, inp, t)
 	if v, ok := strings.CutPrefix(bi.DockerfilePath, "cwd://"); ok {
 		// If Dockerfile is local for a remote invocation, we first check if
 		// it's not outside the working directory and then resolve it to an
--- a/tests/bake.go
+++ b/tests/bake.go
@@ -46,6 +46,7 @@ var bakeTests = []func(t *testing.T, sb integration.Sandbox){
 	testBakeRemote,
 	testBakeRemoteAuth,
 	testBakeRemoteCmdContext,
+	testBakeRemoteKeepGitDir,
 	testBakeRemoteLocalOverride,
 	testBakeLocalCwdOverride,
 	testBakeRemoteCmdContextOverride,
@@ -489,6 +490,42 @@ EOT
 	require.NoError(t, err, out)
 
 	require.FileExists(t, filepath.Join(dirDest, "foo"))
+	require.NoDirExists(t, filepath.Join(dirDest, ".git"))
+}
+
+func testBakeRemoteKeepGitDir(t *testing.T, sb integration.Sandbox) {
+	bakefile := []byte(`
+target "default" {
+	args = {
+		BUILDKIT_CONTEXT_KEEP_GIT_DIR = "1"
+	}
+	dockerfile-inline = <<EOT
+FROM scratch
+COPY . .
+EOT
+}
+`)
+	dir := tmpdir(
+		t,
+		fstest.CreateFile("docker-bake.hcl", bakefile, 0600),
+	)
+	dirDest := t.TempDir()
+
+	git, err := gitutil.New(gitutil.WithWorkingDir(dir))
+	require.NoError(t, err)
+
+	gittestutil.GitInit(git, t)
+	gittestutil.GitAdd(git, t, "docker-bake.hcl")
+	gittestutil.GitCommit(git, t, "initial commit")
+	require.DirExists(t, filepath.Join(dir, ".git"))
+	addr := gittestutil.GitServeHTTP(git, t)
+
+	out, err := bakeCmd(sb, withDir(dir),
+		withArgs(addr, "--set", "*.output=type=local,dest="+dirDest),
+	)
+	require.NoError(t, err, out)
+
+	require.DirExists(t, filepath.Join(dirDest, ".git"))
 }
 
 func testBakeRemoteAuth(t *testing.T, sb integration.Sandbox) {


marxarelli:
@tonistiigi for your review.

Note that this still requires folks to know about the additional environment variable. An alternative approach that might be less surprising to users but less efficient by default, would be to always keep git directories on the remote bake context. In that case, the `BUILDKIT_CONTEXT_KEEP_GIT_DIR` build argument for individual targets would work as (I originally) expected without having to know about the additional environment variable.

I'm happy to refactor if you think that's a better approach.

tonistiigi (commented):
Isn't this supposed to come from the bake definition itself. `ReadRemoteFiles` itself should never need git dir, but after it has read the bake definition and it has been parsed it could detect that the target set `BUILDKIT_CONTEXT_KEEP_GIT_DIR` build arg and in that case switch the `inp.State` over to the one that has git dir enabled?

marxarelli:
> Isn't this supposed to come from the bake definition itself. `ReadRemoteFiles` itself should never need git dir, but after it has read the bake definition and it has been parsed it could detect that the target set `BUILDKIT_CONTEXT_KEEP_GIT_DIR` build arg and in that case switch the `inp.State` over to the one that has git dir enabled?

I like this approach in that it would obviate the need for an additional environment variable. I also like that it comes from the bake definition, like you mentioned.

The only downside I see is that we can't reuse the bake input state and thus there will be duplicate clone operations (two distinct source identifiers, one where `KeepGitDir` is always false, and one where it might be true). However, this inefficiency also arises with the change I made here, where `BUILDX_BAKE_KEEP_GIT_DIR=1` is set but the targets don't specify `BUILDKIT_CONTEXT_KEEP_GIT_DIR=1`.

Maybe we do both?

For correctness: detect targets that set `BUILDKIT_CONTEXT_KEEP_GIT_DIR=1` and use their context state instead of the bake input state.

For optimization: allow users to set `BUILDX_BAKE_KEEP_GIT_DIR=1` and avoid duplicate clones because although the input state will not be reused, at least the cache keys should match.

tonistiigi:
> The only downside I see is that we can't reuse the bake input state and thus there will be duplicate clone operations (two distinct source identifiers, one where KeepGitDir is always false,

This shouldn't be a big problem. `llb.Git` implementation uses a shared repository for all reference lookups, so you would still only get one remote clone that is then checked out two different ways. I have a draft of a sparse checkout implementation as well. If that one ever lands, we would probably always use that for `ReadRemoteFiles`.

marxarelli:
> This shouldn't be a big problem. `llb.Git` implementation uses a shared repository for all reference lookups, so you would still only get one remote clone

Ah! In that case, I will attempt to implement the build arg detection.

marxarelli:
@tonistiigi I've amended the commit and updated the PR description to reflect the new approach.

bake/bake.go
```diff
@@ -1274,14 +1274,15 @@ func TargetsToBuildOpt(m map[string]*Target, inp *Input) (map[string]build.Optio
...
-func updateContext(t *build.Inputs, inp *Input) {
+func updateContext(t *build.Inputs, inp *Input, target *Target) {
 	if inp == nil || inp.State == nil {
 		return
 	}
 
 	for k, v := range t.NamedContexts {
 		if v.Path == "." {
 			t.NamedContexts[k] = build.NamedContext{Path: inp.URL}
+			continue
```
marxarelli, line 1285: @tonistiigi I saw this was (seemingly) missing while implementing my change. Without it, the context is simply reassigned below. Let me know if my assumption isn't correct and I'll remove it.

bake/bake.go
```diff
@@ -1303,6 +1304,16 @@ func updateContext(t *build.Inputs, inp *Input) {
 	if build.IsRemoteURL(t.ContextPath) {
 		return
 	}
+
+	// If the target includes a BUILDKIT_CONTEXT_KEEP_GIT_DIR=1 build argument,
+	// do not reuse the bake context state as it never keeps the .git directory
+	// (see ReadRemoteFiles).
+	if v, ok := target.Args["BUILDKIT_CONTEXT_KEEP_GIT_DIR"]; ok && v != nil {
```
tonistiigi, line 1311: Not sure I get this. Shouldn't this only run in the `t.ContextPath == "." ` case above?

tests/bake.go
```diff
@@ -489,6 +490,42 @@ EOT
...
+	git, err := gitutil.New(gitutil.WithWorkingDir(dir))
+	require.NoError(t, err)
+
+	gittestutil.GitInit(git, t)
+	gittestutil.GitAdd(git, t, "docker-bake.hcl")
+	gittestutil.GitCommit(git, t, "initial commit")
+	require.DirExists(t, filepath.Join(dir, ".git"))
+	addr := gittestutil.GitServeHTTP(git, t)
+
+	out, err := bakeCmd(sb, withDir(dir),
```
tonistiigi, line 523: Iiuc this is building from a local directory (that contains .git), not the git address as first parameter.

crazy-max: Seems to be the git address as first arg (see next line). `withDir` is used to set working dir when running command.

tests/bake.go
```diff
@@ -489,6 +490,42 @@ EOT
...
+	bakefile := []byte(`
+target "default" {
+	args = {
+		BUILDKIT_CONTEXT_KEEP_GIT_DIR = "1"
+	}
+	dockerfile-inline = <<EOT
+FROM scratch
+COPY . .
+EOT
+}
```
crazy-max, line 506: > When a remote Git URL is used with `buildx bake`, `.git` directories are
> always omitted (see `bake.ReadRemoteFiles`).

Made some tests with latest v0.26.1 release and it seems to work already:

```
$ docker buildx version
github.com/docker/buildx v0.26.1-desktop.1 532a478c2ea39e2d0eb40ad2e3f6bec57df4c8af
```

```
$ docker buildx bake --no-cache --set *.output=./bin https://github.com/crazy-max/buildx-3338.git#e405059143a75f7ea3835f09d78f0d14d4469379 --print
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": "https://github.com/crazy-max/buildx-3338.git#e405059143a75f7ea3835f09d78f0d14d4469379",
      "dockerfile": "Dockerfile",
      "dockerfile-inline": "FROM scratch\nCOPY . .\n",
      "args": {
      },
      "output": [
        {
          "dest": "./bin",
          "type": "local"
        }
      ],
      "no-cache": true
    }
  }
}
```

```
$ docker buildx bake --no-cache --set *.output=./bin https://github.com/crazy-max/buildx-3338.git#e405059143a75f7ea3835f09d78f0d14d4469379

#1 [internal] load git source https://github.com/crazy-max/buildx-3338.git#e405059143a75f7ea3835f09d78f0d14d4469379
#1 CACHED

#2 transferring dockerfile: 59B done
#2 DONE 0.0s

#3 [internal] load git source https://github.com/crazy-max/buildx-3338.git#e405059143a75f7ea3835f09d78f0d14d4469379
#3 0.069 Initialized empty Git repository in /var/lib/desktop-containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/9459/fs/.git/
#3 0.073 commit
#3 0.106 From file:///var/lib/desktop-containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/9452/fs
#3 0.106  * branch            refs/buildkit/7f9ratgj0wa89tbui7dje1k6z -> FETCH_HEAD
#3 0.113 Note: switching to 'FETCH_HEAD'.
#3 0.113
#3 0.113 You are in 'detached HEAD' state. You can look around, make experimental
#3 0.113 changes and commit them, and you can discard any commits you make in this
#3 0.113 state without impacting any branches by switching back to a branch.
#3 0.113
#3 0.113 If you want to create a new branch to retain commits you create, you may
#3 0.113 do so (now or later) by using -c with the switch command. Example:
#3 0.113
#3 0.113   git switch -c <new-branch-name>
#3 0.113
#3 0.113 Or undo this operation with:
#3 0.113
#3 0.113   git switch -
#3 0.113
#3 0.113 Turn off this advice by setting config variable advice.detachedHead to false
#3 0.113
#3 0.113 HEAD is now at e405059 keep git dir
#3 DONE 0.2s

#4 [1/1] COPY . .
#4 DONE 0.1s

#5 exporting to client directory
#5 copying files 19.22kB 0.1s
#5 copying files 27.44kB 0.3s done
#5 DONE 0.3s
```

```
$ ls -a bin/
.  ..  .git  docker-bake.hcl  foo
```

marxarelli: :exploding_head: I must have really messed something up when trying to repro this. Even the test I added here passes when I remove my changes from `bake/bake.go`.

Sorry for wasting your time and thank you!

PR #3341, opened 2025-07-29
dap: make exec shell persistent across the build

jsternberg:
Invoking the shell will cause it to persist across the entire build and
to re-execute whenever the builder pauses at another location again.

This still requires using `exec` to launch the shell. Launching by frame
id is also removed since it no longer applies to this version.


ktock (approved):
Thanks

dap/debug_shell.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	defer s.sem.Release(1)
+
+	ctr, err := build.NewContainer(ctx, rCtx, cfg)
+	if err != nil {
+		return err
+	}
+	defer ctr.Cancel()
+
+	fmt.Fprintf(in.Stdout, "Running %s in build container.\n", strings.Join(append(cfg.Entrypoint, cfg.Cmd...), " "))
```
rcjsuen:
```
Running /bin/sh in build container.
/ # 
Running /bin/sh in build container.
/ # 
Running /bin/sh in build container.
/ # 
```
Is there a way for us to include the line number or some other additional context information here I wonder? 🤔 For someone new to the feature they may think it is misbehaving if they see the same log message being sent multiple times.

rcjsuen: Preferably some information that _changes_ with every step. That will help clearly indicate to the user that _something_ has changed.

crazy-max: Was wondering the same thing when testing on my side:

image

For this Dockerfile:

```dockerfile
FROM alpine
RUN echo hello
RUN echo foo
RUN echo bar
```

Maybe smth like this?:

```
Running /bin/sh in build container from "RUN echo foo"
/ # 
Running /bin/sh in build container from "RUN echo bar"
/ # 
```

crazy-max: And if it's the default shell then just the following would be fine I think:

```
Running shell from "RUN echo foo"
/ # 
Running shell from "RUN echo bar"
/ # 
```

crazy-max: > to include the line number

Or maybe line number is better as a similar command could exist in multiple places or could be very long in some Dockerfiles.

rcjsuen: We also need to make it clear that the directory structure has been reset. If I edit a file or run `apk add` or whatever, it all gets wiped when I advance to the next step in the debugger.

jsternberg: I've included the line number and an additional message to say the changes will be reset when the build continues.

crazy-max:
Looks good on Linux and macOS but I don't have the prompt on Windows:

```
2025-07-31T14:31:44.161Z [debug] {"message":"DAP message will be received from the editor","dapMessage":{"command":"evaluate","arguments":{"expression":"exec","frameId":3,"context":"repl"},"type":"request","seq":11},"id":"3d979d6e-9864-444a-b86b-5121aeeb166e","name":"Docker: Build"}
2025-07-31T14:31:44.181Z [debug] {"message":"DAP message has been sent to the editor","dapMessage":{"seq":34,"type":"request","command":"runInTerminal","arguments":{"kind":"integrated","cwd":"","args":["docker","buildx","dap","attach","C:\\Users\\crazy\\AppData\\Local\\Temp\\buildx-dap-exec4118734859\\s.sock"],"env":{"BUILDX_EXPERIMENTAL":"1"}}},"id":"3d979d6e-9864-444a-b86b-5121aeeb166e","name":"Docker: Build"}
2025-07-31T14:31:44.191Z [debug] {"message":"DAP message will be received from the editor","dapMessage":{"type":"response","seq":12,"command":"runInTerminal","request_seq":34,"success":true,"body":{"shellProcessId":30624}},"id":"3d979d6e-9864-444a-b86b-5121aeeb166e","name":"Docker: Build"}
2025-07-31T14:31:44.192Z [debug] {"message":"DAP message has been sent to the editor","dapMessage":{"seq":35,"type":"response","request_seq":11,"success":true,"command":"evaluate","body":{"result":"Started process attached to C:\\Users\\crazy\\AppData\\Local\\Temp\\buildx-dap-exec4118734859\\s.sock.","variablesReference":0}},"id":"3d979d6e-9864-444a-b86b-5121aeeb166e","name":"Docker: Build"}
2025-07-31T14:31:44.192Z [debug] {"message":"DAP message will be received from the editor","dapMessage":{"command":"scopes","arguments":{"frameId":3},"type":"request","seq":13},"id":"3d979d6e-9864-444a-b86b-5121aeeb166e","name":"Docker: Build"}
2025-07-31T14:31:44.192Z [debug] {"message":"DAP message has been sent to the editor","dapMessage":{"seq":36,"type":"response","request_seq":13,"success":true,"command":"scopes","body":{"scopes":[{"name":"Arguments","presentationHint":"arguments","variablesReference":16777217,"expensive":false}]}},"id":"3d979d6e-9864-444a-b86b-5121aeeb166e","name":"Docker: Build"}
2025-07-31T14:31:44.193Z [debug] {"message":"DAP message will be received from the editor","dapMessage":{"command":"variables","arguments":{"variablesReference":16777217},"type":"request","seq":14},"id":"3d979d6e-9864-444a-b86b-5121aeeb166e","name":"Docker: Build"}
2025-07-31T14:31:44.193Z [debug] {"message":"DAP message has been sent to the editor","dapMessage":{"seq":37,"type":"response","request_seq":14,"success":true,"command":"variables","body":{"variables":[{"name":"platform","value":"linux/amd64","variablesReference":16777218},{"name":"exec","value":"/bin/sh -c echo foo","variablesReference":16777219}]}},"id":"3d979d6e-9864-444a-b86b-5121aeeb166e","name":"Docker: Build"}
```

https://github.com/user-attachments/assets/f46730a8-06da-408f-b3af-9ef0c912c88e

Works fine in Buildx 0.26.1:

image

jsternberg:
There's a bit of a weird thing that happens with this PR that didn't for the original version and I have an idea to fix it, but it might be difficult.

At the current moment, it's not possible to cancel the exec without terminating the build. And if you exit the container process from within the terminal, it only closes the container and doesn't close the attachment which will just hang until a new process comes up.

I think it might be good to differentiate these. If the container is the one who exits before the program is resumed, I think we should terminate the connection to the client so the attach command exits. If the program is resumed and the reason for the terminal dying is because the container was canceled, we should restore it on the next breakpoint. Basically, try to differentiate from a user initiated exit and a debugger initiated exit.

Might be difficult to do but I'll give it a try. I think it'll likely make a better user experience.

jsternberg:
I've made some changes to this. If the exec exits with a zero status, the exec is considered "complete" and it will release control of the attached shell. If the exec exits with a non-zero status or exits because it was canceled (such as resuming the build), it will restore itself after the next pause.

I've also changed it so the exec can be done again. If you try and exec while a shell is still attached, it will give the same error. If the previous shell exited and you do exec again, it will re-use the socket and connect again.

I also moved some of the logging messages to somewhere more appropriate. I added a "reason" to the message when it doesn't invoke a build container for a paused step. This should be an improved workflow.

jsternberg:
Testing this out more and it might not work properly when execing on an error state. Moving this to draft for now.

jsternberg:
This is ready for review. The pausing on exception was broken on something unrelated to this so I'll fix that separately.

crazy-max:
When exiting the shell and doing another `exec`:

image

I can't get the prompt:

image

jsternberg:
Fixed. This is because a paused a build step was only attempting to attach to the shell once. So when you exit and try to re-enter it didn't get the shell again.

I've now modified it so it attempts to reattach on a clean exit and it won't reattach on an exit with an error which includes the context being canceled.

dap/debug_shell.go
```diff
@@ -0,0 +1,308 @@
...
+
+	var f dap.StackFrame
+	if len(t.stackTrace) > 0 {
+		f = t.frames[t.stackTrace[0]].StackFrame
+	}
+
+	cfg := &build.InvokeConfig{Tty: true}
+	if len(cfg.Entrypoint) == 0 && len(cfg.Cmd) == 0 {
+		cfg.Entrypoint = []string{"/bin/sh"} // launch shell by default
+		cfg.Cmd = []string{}
```
crazy-max, line 150: If `/bin/sh` is not there (such as `scratch`), I think we should not attach:

image

If we `exec` it returns instantly atm:

image

rcjsuen: That would presumably happen for the WCOW case also I suppose.

jsternberg: It's supposed to wait. I must have broken that with the last change.

jsternberg: @crazy-max this is the output I get when trying to exec on a scratch container.

```
Attached to build process.
Build container is not executable. (reason: stat error: lstat bin/sh: no such file or directory)
```

This should prevent trying to run a container when the shell doesn't exist.

crazy-max: > `Build container is not executable. (reason: stat error: lstat bin/sh: no such file or directory)`

@jsternberg Hum I don't have this message when testing on my side:

https://github.com/user-attachments/assets/895a5020-a78d-4f28-ae8a-1c0d5612960b

jsternberg: This should now be fixed. There was an error in `StatFile` where it would attempt to use the ref from the scratch volume and it resulted in a nil reference. Now it checks for that so the `canInvoke` method works properly without crashing.

PR #3348, opened 2025-07-31
dap: utilize the `dap.hint.noparent` from LLB if it exists

jsternberg:
This annotation signals that the LLB node doesn't have a direct parent
and the inputs are disconnected. This prevents us from accidentally
including them in the same stack frame.

Requires https://github.com/moby/buildkit/pull/6117.

### DIFF
--- a/dap/thread.go
+++ b/dap/thread.go
@@ -3,6 +3,7 @@ package dap
 import (
 	"context"
 	"path/filepath"
+	"strconv"
 	"sync"
 
 	"github.com/docker/buildx/build"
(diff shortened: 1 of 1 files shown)


dap/thread.go
```diff
@@ -163,8 +164,22 @@ func (t *thread) createBranch(last *step) (first *step) {
 		}
 
 		op := t.ops[first.dgst]
+
+		hasParent := len(op.Inputs) > 0
+		if m, ok := t.def.Metadata[first.dgst]; ok {
+			if s, ok := m.Description["com.docker.dap.hint.v1.noparent"]; ok {
```
crazy-max: Was wondering if we could use a fallback frontend image like we have done for build checks and outline requests so everyone can benefit from this feature but would need first a frontend capability I suppose which I don't think is necessary.

jsternberg:
Not needed anymore.

PR #3356, opened 2025-08-06
dockerfile: update docker to 28.3.3

crazy-max:


crazy-max:
> LGTM
> 
> we could drop the minor or patch version perhaps (as in: latest 28.x or latest 28.3.x)

Yes good point, I think we have tags for minor now

PR #3366, opened 2025-08-13
dap: improve determination of the proper parent for certain ops

jsternberg:
Improves the determination of the proper parent for exec and file ops.
With file ops, it will only consider inputs and ignore secondary inputs.
This prevents the following case:

```
FROM busybox AS build1
RUN echo foo > /hello

FROM scratch
COPY --from=build1 /hello .
```

Previously, `build1` would be considered the parent of the copy
instruction. Now, copy properly does not have a parent.

If there are multiple file ops and the operations disagree on the
canonical "parent", we give up on trying to find a canonical parent and
assume there is none.

For exec operations, whichever input is associated with the root mount
is considered the primary parent.

For all other operations, the first parent is considered the primary
parent if it exists.


dap/thread.go
```diff
@@ -217,6 +224,46 @@ func (t *thread) getStackFrame(dgst digest.Digest) *frame {
...
+	case *pb.Op_File:
+		// Determine possible candidate parents from the actions.
+		// All actions need to be consistent in which input is the
+		// direct parent.
+		candidates := make(map[int]struct{})
+		for _, action := range op.File.Actions {
+			candidates[int(action.Input)] = struct{}{}
+		}
+
+		if len(candidates) != 1 {
```
tonistiigi: Multiple actions are possible, even in Dockerfile. For example `COPY` with multiple source paths. Non-negative input of first action should be ok.

jsternberg: I figured it was best to be cautious when it came to the inputs. Is it possible for the multiple source paths in dockerfile to also have multiple different input indices?

tonistiigi: The way it works is that input keys in action point first to the input array of the op and after that is exhausted then then index is result of another action https://github.com/moby/buildkit/blob/master/solver/pb/ops.proto#L310. So if you have two copy actions after each other it is smth like

```
fileop {
  inputs = [ A , B ]
  actions = [
     copy{input: 0, secondaryInput: 1},
     copy{input: 2, secondaryInput: 1}
```

where the 2 means the result of the first action `len(inputs)+0`

jsternberg: I see what you mean. While unlikely, I'm still concerned about what to do if there are multiple copies with inputs that are below index 2 in your example. Maybe I can just update this so it filters out any input index that references something within the action?

jsternberg: I've updated this to just use the first suitable input from the list of file actions.

dap/thread.go
```diff
@@ -217,6 +224,40 @@ func (t *thread) getStackFrame(dgst digest.Digest) *frame {
...
+		for _, m := range op.Exec.Mounts {
+			if m.Dest == "/" {
+				return int(m.Input)
+			}
+		}
+		return -1
+	case *pb.Op_File:
+		// Use the first input where the index is from one of the inputs.
+		for _, action := range op.File.Actions {
+			if input := int(action.Input); input < n {
```
tonistiigi: Input can be `-1` as well and should be ignored then in case there is another action with suitable one.

tonistiigi: Input can be `-1` as well and should be ignored then in case there is another action with a suitable one.

jsternberg: Done.

Issue #3367, opened 2025-08-14
Support `--format=json` for `buildx du --verbose`


fenollp:
### Description

While looking into https://github.com/docker/buildx/discussions/3363#discussioncomment-14092542 I realized I need more than one field per entry, thus to parse this data. And so a `--format=json` as supported by other commands would be great.

Also, please please document the possible filters this command supports with ample examples AND maybe add a way to sort the results (createdat or last used or usage counts, ...) but JSON formatting gives sorting...

fenollp:
WRT filters:
* `type=regular` for exact matches on `type` field
* `until=24h` to filter on creation date
* `description~=cached` to fuzzy match

crazy-max:
I recall my comment about available filters https://github.com/docker/buildx/issues/1065#issuecomment-3103195569 and agree we should document them and explain variations with examples.

PR #3377, opened 2025-08-22
cmd: multiple formats output support for du command

crazy-max:
fixes https://github.com/docker/buildx/issues/3367

Adds `format` flag to support multiple formats output using docker/cli formatter. `verbose` flag is now a shorthand for `--format=pretty`.

### DIFF
--- a/tests/diskusage.go
+++ b/tests/diskusage.go
@@ -1,0 +1,51 @@
+package tests
+
+import (
+	"testing"
+
+	"github.com/moby/buildkit/util/testutil/integration"
+	"github.com/stretchr/testify/require"
+)
+
+var diskusageTests = []func(t *testing.T, sb integration.Sandbox){
+	testDiskusage,
+	testDiskusageVerbose,
+	testDiskusageVerboseFormatError,
+	testDiskusageFormatJSON,
+	testDiskusageFormatGoTemplate,
+}
+
+func testDiskusage(t *testing.T, sb integration.Sandbox) {
+	buildTestProject(t, sb)
+	cmd := buildxCmd(sb, withArgs("du"))
+	out, err := cmd.Output()
+	require.NoError(t, err, string(out))
+}
+
+func testDiskusageVerbose(t *testing.T, sb integration.Sandbox) {
+	buildTestProject(t, sb)
+	cmd := buildxCmd(sb, withArgs("du", "--verbose"))
+	out, err := cmd.Output()
+	require.NoError(t, err, string(out))
+}
+
+func testDiskusageVerboseFormatError(t *testing.T, sb integration.Sandbox) {
+	buildTestProject(t, sb)
+	cmd := buildxCmd(sb, withArgs("du", "--verbose", "--format=json"))
+	out, err := cmd.Output()
+	require.Error(t, err, string(out))
+}
+
+func testDiskusageFormatJSON(t *testing.T, sb integration.Sandbox) {
+	buildTestProject(t, sb)
+	cmd := buildxCmd(sb, withArgs("du", "--format=json"))
+	out, err := cmd.Output()
+	require.NoError(t, err, string(out))
+}
+
+func testDiskusageFormatGoTemplate(t *testing.T, sb integration.Sandbox) {
+	buildTestProject(t, sb)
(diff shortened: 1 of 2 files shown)


docs/reference/buildx_du.md
```diff
@@ -117,3 +118,41 @@ uj802yxtvkcjysnjb4kgwvn2v         true           11.68MB       45 hours ago
 Reclaimable:    2.627GB
 Total:          2.627GB
 ```
+
+### <a name="format"></a> Format the output in verbose mode (--format)
+
+The formatting options (`--format`) can be used with [`--verbose` flag](#verbose)
+to print the output in `raw` (default) or `json`:
```
tonistiigi: I don't think we have `raw` in other commands. I think we have `pretty` (at least in some places).

crazy-max: Ah yes that's `pretty`, I'm always confused by both.

crazy-max:
> So it should work with template but also the json shorthand.

Yes you right and also shouldn't be an array, I will change that.

docs/reference/buildx_du.md
```diff
@@ -117,3 +118,41 @@ uj802yxtvkcjysnjb4kgwvn2v         true           11.68MB       45 hours ago
...
+
+```console
+$ docker buildx du --verbose --format=json
+[
+  ...
+  {
```
thaJeztah: FWIW, as this is brand new, I'd be tempted to not pretty-print the json and just 1 entry per line; users can still `jq` to pretty-print the JSON

docs/reference/buildx_du.md
```diff
@@ -61,33 +62,146 @@ The asterisks (\*) in the default output indicate the following:
...
+  "Shared": false,
+  "Size": "829.9MB",
+  "Type": "regular",
+  "UsageCount": 1
+}
+{
+  "CreatedAt": "2025-08-05 09:24:09.15742275 +0000 UTC",
+  "Description": "pulled from docker.io/library/node:22@sha256:3218f0d1b9e4b63def322e9ae362d581fbeac1ef21b51fc502ef91386667ce92",
+  "ID": "jsw7fx09l5zsda3bri1z4mwk5",
+  "LastUsedAt": "2 weeks ago",
```
crazy-max, line 168: Hum in JSON output we should not have this formatted, I will change that.

commands/prune.go
```diff
@@ -241,3 +242,55 @@ func toBuildkitPruneInfo(f filters.Args) (*client.PruneInfo, error) {
...
+	size := units.HumanSize(float64(di.Size))
+	if di.Shared {
+		size += "*"
+	}
+	lastAccessed := ""
+	if di.LastUsedAt != nil {
+		lastAccessed = units.HumanDuration(time.Since(*di.LastUsedAt)) + " ago"
+	}
+	fmt.Fprintf(tw, "%-40s\t%-5v\t%-10s\t%s\n", id, !di.InUse, size, lastAccessed)
+}
```
crazy-max, line 296: Moved logic not used anymore to prune command for now but we should also consider using docker/cli formatter for this command.

tonistiigi (commented):
If I do `--format=table` then I get empty rows, so it seems to know that this input is special but not handle it correctly. This could be follow-up.

commands/diskusage.go
```diff
@@ -4,29 +4,84 @@ import (
...
+		if opts.verbose {
+			opts.format = duDefaultPrettyTemplate
+		} else {
+			opts.format = duDefaultTableFormat
+		}
+	} else if opts.format == formatter.PrettyFormatKey {
+		opts.format = duDefaultPrettyTemplate
+	} else if opts.format == formatter.TableFormatKey {
+		opts.format = duDefaultTableFormat
+	}
```
crazy-max, line 83: > If I do `--format=table` then I get empty rows, so it seems to know that this input is special but not handle it correctly. This could be follow-up.

@tonistiigi fixed here

thaJeztah:
> we should also consider using docker/cli formatter for this command.

Before you do, please check; we're trying to move many parts in the CLI internal, as there were too many things public that caused us not being able to make changes; for some we could find a new place though.

Also considering if the CLI should execute buildx for performing prune instead of using the moby REST API (as part of `docker system prune` at least).

tonistiigi (approved):

PR #3387, opened 2025-08-27
vendor: update buildkit to v0.24.0-rc1

tonistiigi:


tonistiigi:
Had to skip DAP test `TestLaunch`. It panics via leaked goroutine but previous errgroup swallowed the panic via `recover()` https://github.com/docker/buildx/issues/3388

Issue #1922, opened 2023-06-29
[Feature] Ability to inspect specific cache mounts


anthonyalayo:
### Description

Following up from: https://github.com/docker/buildx/issues/931

1. Why can't we have something like, `docker buildx du --filter /var/cache/apt`? It would allow direct inspecting/clearing of cache mounts. I'm writing some logic to allow deleting of a particular build cache mount if desired. I'm not able to have a clear script for this because of the current support.

2. Is it also possible to add more details with `docker buildx du --verbose --filter type=exec.cachemount` in order to know which image that particular cache mount is associated with? I don't even understand why there is an "id" field configurable if you can't key off of it in any way. I just changed one of the mounts for a java program's .m2 directory and all I get is 3 entries here and none of them discernable.

```
~ ❯ docker buildx du --verbose --filter type=exec.cachemount | grep -B 6 -F "/root/.m2"     06:28:51 PM
ID:		qd98bx4cmfe7de8fln5nx65ut
Created at:	2023-06-29 21:09:39.551389639 +0000 UTC
Mutable:	true
Reclaimable:	true
Shared:		false
Size:		19.31MB
Description:	cached mount /root/.m2 from exec /bin/sh -c ./mvnw test
--
ID:		glydhkz1vo25rado1zoq7ajbs
Created at:	2023-06-29 21:35:45.00107996 +0000 UTC
Mutable:	true
Reclaimable:	true
Shared:		false
Size:		1.897MB
Description:	cached mount /root/.m2 from exec /bin/sh -c ./mvnw dependency:resolve-plugins
--
ID:		2iq3u28y28xlsagy1esqkpfu6
Created at:	2023-06-29 22:30:07.203308554 +0000 UTC
Mutable:	true
Reclaimable:	false
Shared:		false
Size:		0B
Description:	cached mount /root/.m2 from exec /bin/sh -c ./mvnw dependency:resolve-plugins
```

anthonyalayo:
Thanks @Azhovan ! This would still be fantastic, it's a limitation that has been frustrating.

dixyes:
Shall we have some command like `docker buildx cache ls`, `docker buildx cache inspect qd98bx4cmfe7de8fln5nx65ut` `docker buildx cache rm qd98bx4cmfe7de8fln5nx65ut` things?

I have a build host build images all the day, `docker buildx du` will take eternal

Sometimes a cache being corrupt, but I cannot just remove that single cache unless build a dockerfile with `rm -rf`

[...]

Issue #1065, opened 2022-04-06
Feature request: Pruning of layers of specific image


kasvtv:

tonistiigi:
> This is inconvenient, as one will lose cached layers from all the containers of all projects on the machine.

`docker builder prune --keep-storage 2GB` (or whatever size you want) is the best default when doing manual pruning. This will delete the oldest and least used cache first, leaving the most useful cache behind.

I think the solution we could consider would be to add some kind of tagging mechanism. So when you run your project you can tag the build cache it creates with a certain value. And then you can use this tag value as a filter for removing cache associated with a specific project.

There are already prune filters by ID/Type etc but they are not very convenient to use.

wojtrawi:
> This is already possible, and is what docker image prune does; docker image prune looks for images that are no longer used (and no longer associated with a tag / name (imagename:tag), and removes those.

This seems not to be true. After doing that I can see that build cache has increased by the amount of layers that existed in pruned images (checking using `docker system df`)

thaJeztah:
> This seems not to be true. After doing that I can see that build cache has increased by the amount of layers that existed in pruned images (checking using `docker system df`)

Hmm.. not sure how that would be related; the BuildKit build-cache does not use the image-store, and `docker image prune` only deletes images; code is in the `ImagesPrune()` function in the docker daemon; relevant part of that here; https://github.com/moby/moby/blob/ce550fa9c2e0f7b85a77e72a9596b2f92f1d0e32/daemon/images/image_prune.go#L93-L139

wojtrawi:
What I experience is that after I prune image, the build cache increases and if I happen to build the same image, I can see that cache has been used to do that. However, I keep a separate image with dependencies so I do not mind doing `docker builder prune -f `after each CI run. And I do not know how it is related, but I can see empty build cache (storage is 0 but items is not 0) and I can experience using cached layers from the kept image with dependencies. As if the build cache had some pointer to layers contained in an image. I'm not a DevOps, just FE developer so I do not know if it's desired or not, but that's how it is.

bdrnglm:
[...]

Issue #931, opened 2022-01-28
[Feature] Ability to inspect cache mounts


Aposhian:
While `docker buildx prune` can be used to delete cache mounts with `docker buildx prune --filter type=exec.cachemount`, there is no way to look at the active list of cache mounts from the `buildx` CLI, or a good way to manage and delete cache mounts individually.

Yes, builds should not be dependent on the contents of the cache mounts, but in practice when debugging and creating a new build, mistakes can introduce bad data into the cache mounts.

tonistiigi:
`docker buildx du --verbose`, you can delete individually by filtering by ID

Aposhian:
Ok, that is helpful, thank you, although all the cache mount ids are random strings. Shouldn't they default to a machine readable version of the target directory, as is suggested in [the docs](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypecache)?
> Optional ID to identify separate/different caches. Defaults to value of target.

tonistiigi:
> Shouldn't they default to a machine readable version of the target directory, as is suggested in the docs?

That's not the same thing. `id` key in mount doesn't actually need to be unique. For example, with `sharing=private` you can have multiple mounts with the same `id` if they are used in parallel.

Aposhian:
Ok. So how do I correlate the cache mounts in my `Dockerfile` with the output of `docker buildx du --verbose`? How do I get from `--mount=type=cache,target=/var/cache/apt,sharing=locked` to something like
```
ID:             0em48gkezupcayt83f9c9pzmr
Parent:         qpyv30sgntv1k4vr1b2xm6nyz
Created at:     2022-01-31 17:21:44.833173217 +0000 UTC
Mutable:        false
Reclaimable:    true
Shared:         true
Size:           0B
Description:    fileop target
Usage count:    1
Last used:      2022-01-31 17:22:35.445416281 +0000 UTC
Type:           regular
```

tonistiigi:
"Type:           regular" and "Description:    fileop target" should tell you that this is not the cache mount you are looking for.

Aposhian:
Ok thanks. So I can do `docker buildx du --verbose | grep -B 9 exec.cachemount` so see all my cache mounts. And then the target name is buried in there. I suppose that works.
```
ID:		oh5dzk0n897717mllhegbo78y
Created at:	2022-01-28 23:49:10.035242944 +0000 UTC
Mutable:	true
Reclaimable:	true
[...]

PR #3392, opened 2025-08-28
docs: list available filters for du and prune commands

crazy-max:
follow-up:
* https://github.com/docker/buildx/issues/1065#issuecomment-3102936155
* https://github.com/docker/buildx/issues/3367#issue-3321034339
* https://github.com/docker/buildx/issues/931

fixes https://github.com/docker/buildx/issues/1922

Currently we have filters for the following keys: https://github.com/moby/buildkit/blob/3d43066f2e814d4d10383ba871256245edf08a25/cache/manager.go#L1641-L1659

Key `type` can be one of: https://github.com/moby/buildkit/blob/3d43066f2e814d4d10383ba871256245edf08a25/client/diskusage.go#L81-L88

We also have the `until` key that is used on client side to set `KeepDuration` for prune info request: https://github.com/docker/buildx/blob/3f4bf829d8c5b92a8f670609417d9aa4c0b6be98/commands/prune.go#L190-L243

We are using containerd module to parse filters: https://github.com/containerd/containerd/blob/7d8765d440e4991a7d005db19e7873651ce3551d/pkg/filters/parser.go#L26-L49

### DIFF
--- a/commands/prune.go
+++ b/commands/prune.go
@@ -175,7 +175,7 @@ func pruneCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.BoolVarP(&options.all, "all", "a", false, "Include internal/frontend images")
-	flags.Var(&options.filter, "filter", `Provide filter values (e.g., "until=24h")`)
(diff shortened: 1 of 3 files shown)


tonistiigi:
@fenollp

PR #3393, opened 2025-08-28
commands: NewRootCmd: use local impl for DisableFlagsInUseLine

thaJeztah:
- relates to https://github.com/docker/cli/pull/6273
- alternative to, closes https://github.com/docker/buildx/pull/3394

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -79,7 +79,9 @@ func NewRootCmd(name string, isPlugin bool, dockerCli *command.DockerCli) *cobra
 		cmd.SilenceErrors = true
 		cmd.TraverseChildren = true
 		cmd.DisableFlagsInUseLine = true
-		cli.DisableFlagsInUseLine(cmd)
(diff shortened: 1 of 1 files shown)


commands/root.go
```diff
@@ -79,7 +79,9 @@ func NewRootCmd(name string, isPlugin bool, dockerCli *command.DockerCli) *cobra
 		cmd.SilenceErrors = true
 		cmd.TraverseChildren = true
 		cmd.DisableFlagsInUseLine = true
-		cli.DisableFlagsInUseLine(cmd)
+		visitAll(cmd, func(c *cobra.Command) {
+			c.DisableFlagsInUseLine = true
```
thaJeztah, line 83: Ideally, we would not be doing this at runtime, but perhaps instead a test that validates all commands have the option set (I  was considering this for the CLI itself as well).

thaJeztah: - ☝️ opened https://github.com/docker/buildx/pull/3394 as alternative

PR #3395, opened 2025-08-29
install/uninstall: unhide

AkihiroSuda:
The install/uninstall commands were introduced in #34 (Apr 2019) as hidden commands.

These commands should no longer be hidden, as they have been explicitly documented since docker/docs#21641 (Dec 2024).

https://docs.docker.com/build/builders/#difference-between-docker-build-and-docker-buildx-build


crazy-max:
> These commands should no longer be hidden

I think we want to keep them hidden since `docker build` is an alias of `docker buildx build` and `docker builder` an alias of `docker buildx` and therefore install/uninstall does not have any effect anymore, see https://github.com/docker/cli/pull/3314. As there are hidden commands, we also explicitly removed the documentation in https://github.com/docker/buildx/pull/1918.

We should remove them in the future imo but first deprecate them. Same in our GHA https://github.com/docker/setup-buildx-action/blob/1583c0f09d26c58c59d25b0eef29792b7ce99d9a/action.yml#L29-L32. WDYT @tonistiigi @colinhemmings?

AkihiroSuda:
> I think we want to keep them hidden since `docker build` is an alias of `docker buildx build` and `docker builder` an alias of `docker buildx` and therefore install/uninstall does not have any effect anymore, see [docker/cli#3314](https://github.com/docker/cli/pull/3314). 

Seems to still have an effect.
The difference is that `docker buildx install` makes `docker build` respect the builder specified via `docker buildx use`.

https://github.com/docker/docs/blob/3d174fdbea6765c86344beab0f90067f640c937b/content/manuals/build/builders/_index.md?plain=1#L68-L108

crazy-max:
> > I think we want to keep them hidden since `docker build` is an alias of `docker buildx build` and `docker builder` an alias of `docker buildx` and therefore install/uninstall does not have any effect anymore, see [docker/cli#3314](https://github.com/docker/cli/pull/3314).
> 
> Seems to still have an effect. The difference is that `docker buildx install` makes `docker build` respect the builder specified via `docker buildx use`.
> 
> https://github.com/docker/docs/blob/3d174fdbea6765c86344beab0f90067f640c937b/content/manuals/build/builders/_index.md?plain=1#L68-L108

Ah indeed forgot about this case: https://github.com/docker/cli/pull/3676#issuecomment-1290100701

AkihiroSuda:
Alternative
- https://github.com/docker/buildx/pull/3472

PR #3399, opened 2025-09-03
test: git query string

crazy-max:


tonistiigi (commented):
These tests are accessing hub without mirror? Is this happening in some other tests as well?

crazy-max:
> These tests are accessing hub without mirror? Is this happening in some other tests as well?

ah yes need to address that

PR #3400, opened 2025-09-03
[v0.28] cherry-picks for v0.28.0

crazy-max:
backport of:
* https://github.com/docker/buildx/pull/3401
* https://github.com/docker/buildx/pull/3398
* https://github.com/docker/buildx/pull/3399

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -692,6 +692,11 @@ func wrapBuildError(err error, bake bool) error {
 			msg += " Named contexts are supported since Dockerfile v1.4. Use #syntax directive in Dockerfile or update to latest BuildKit."
 			return &wrapped{err, msg}
 		}
+		if st.Code() == codes.Unimplemented && strings.Contains(st.Message(), "unsupported frontend capability moby.buildkit.frontend.gitquerystring") {
+			msg := "current frontend does not support Git URLs with query string components."
+			msg += " Git URLs with query string are supported since Dockerfile v1.18 and BuildKit v0.24. Use BUILDKIT_SYNTAX build-arg, #syntax directive in Dockerfile or update to latest BuildKit."
+			return &wrapped{err, msg}
+		}
 	}
 	return err
 }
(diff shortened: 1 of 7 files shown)


commands/build.go
```diff
@@ -692,6 +692,11 @@ func wrapBuildError(err error, bake bool) error {
 			msg += " Named contexts are supported since Dockerfile v1.4. Use #syntax directive in Dockerfile or update to latest BuildKit."
 			return &wrapped{err, msg}
 		}
+		if st.Code() == codes.Unimplemented && strings.Contains(st.Message(), "unsupported frontend capability moby.buildkit.frontend.gitquerystring") {
+			msg := "current frontend does not support Git URLs with query string components."
+			msg += " Git URLs with query string are supported since Dockerfile v1.18 and BuildKit v0.24. Use BUILDKIT_SYNTAX build-arg, #syntax directive in Dockerfile or update to latest BuildKit."
```
thaJeztah, line 697: Does this require _both_ the updated frontend _and_ latest BuildKit, or is updating the frontend sufficient? The error is a bit ambiguous (using an older frontend with BuildKit v0.24 would still fail I guess? But not sure if just updating the syntax would fix that).

Recommending users to update to latest BuildKit also won't work if this error is returned from the default builder in dockerd; perhaps we should have a `/go/` link to the Dockerfile syntax directive (which could link to alternatives such as the `BUILDKIT_SYNTAX`.

crazy-max: > perhaps we should have a `/go/` link to the Dockerfile syntax directive (which could link to alternatives such as the `BUILDKIT_SYNTAX`.

Do you mean a `/go` link in the error message with more information? Was thinking of putting some of these use cases https://github.com/crazy-max/buildx-buildkit-tests/tree/main/buildkit-6178#readme along https://github.com/docker/docs/pull/23322

thaJeztah: Yes, was considering a `https://docs.docker.com/go/....` URL in the error; mostly because it's probably too much to describe all possible options in the error itself, and I could see `# syntax=` could be useful to reference in some places (not everyone is familiar with them yet).

crazy-max: Yeah that makes sense. I discussed with @tonistiigi, we will create a `/go` page on docs website but will do after this release as we can't publish the docs page before DD release. Fyi, made some changes in https://github.com/docker/docs/pull/23322.

PR #3409, opened 2025-09-05
bake: handle tilde expansion in filepaths

### LINKED ISSUE #1739: bake doesn't handle tilde expansion
### Contributing guidelines

### Description

Maybe intentional, but Bake is not able to resolve tilde expansion to mean the user's home directory when used in a bake attribute value.

For example:

```
$ stat -F ~/token
-rw-r--r-- 1 david david 5 Apr 15 21:51:35 2023 /Users/david/token
$ docker buildx bake
[+] Building 0.0s (0/0)                                                                                                                                                                                
ERROR: failed to stat ~/token: stat ~/token: no such file or directory
```

### Expected behaviour

I expected tilde expansion to resolve to the home directory

### Actual behaviour

Tilde is not expanded

### Buildx version

github.com/docker/buildx v0.10.4 c513d34049e499c53468deac6c4267ee72948f02

### Docker info

_No response_

### Builders list

```text
NAME/NODE       DRIVER/ENDPOINT STATUS  BUILDKIT                         PLATFORMS
default *       docker                                                   
  default       default         running 22.06.0-beta.0-926-g914b02ebaf.m linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/mips64le, linux/mips64
desktop-linux   docker                                                   
  desktop-linux desktop-linux   running 22.06.0-beta.0-926-g914b02ebaf.m linux/arm64, linux/amd64, linux/amd64/v2, linux/riscv64, linux/ppc64le, linux/s390x, linux/mips64le, linux/mips64
```

### Configuration

```hcl
target "default" {
  secret = [
    "type=file,id=token,src=~/token"
  ]
}
```

```dockerfile
FROM alpine
RUN --mount=type=secret,id=token \
    cat /run/secrets/token > /token
```

### Build logs

_No response_

### Additional info

_No response_

dvdksn:
## Description

Adds tilde expansion for filepath attributes in bake files

## Related issues

- Closes #1739


dvdksn:
cc @crazy-max ptal 🙂

thaJeztah:
Slightly concerned if this will start users expecting `~` to work everywhere; I know we had similar cases in the docker cli, where (by design), paths were expected to be absolute in various flags; also having only support on Linux can mean that the bakefile is no longer portable. But now could become a sliding slope, where (e.g.) `%APPDIR%` and other similar vars are also expected to work.

dvdksn:
@thaJeztah that's a valid concern. For things like `%APPDIR%`, that's already supported via variables. There's just on syntactic sugar in the bake file format - you'd have to do `$APPDIR`, and then it should work (if set, as it would be on win)

I quite like the `~username` syntax for things like builds though because it lets you map cache, output, etc without explicitly hardcoding the paths.

jsternberg:
I agree with @thaJeztah that adding support for a `tilde` expansion is a bit of a slippery slope. Tilde expansions are generally done as part of the shell program and it isn't supported by most command line tools. This shows up in cases where `--path=~/foo` doesn't work but `--path ~/foo` will work because the first gets passed with `~` verbatim.

Recently, @crazy-max added a `homedir()` function to bake. I think it would be better and more explicit if we suggested something like this:

```
variable "HOME" {
    default = homedir()
}
```

Then `HOME` can be expanded consistently with everything else and we don't have to add any extra code or wade into trying to expand tilde symbols everywhere.

jsternberg (changes requested):
Putting a request changes until we have a further discussion on if we want to merge this feature or not.

jsternberg:
I'm going to close this since we haven't had a further discussion on it and I think the original comments still stand for why to avoid doing this.

PR #3415, opened 2025-09-09
build: allow client side querystring resolution

tonistiigi:
For the cases where frontend/buildkit doesn't support new Git Querystring format, allow resolving the URL into LLB on client side and then build from input.

Note that this produces slightly different provenance where context is not set as string, so added opt-in via environment variable for now.

We will likely use it in Docker Github Actions so that we can use checksum and ref together in all the builds by default.

Example builds:

```
BUILDX_BUILDER=bk24 docker buildx build -f ./frontend/dockerfile/cmd/dockerfile-frontend/Dockerfile -t tonistiigi/dockerfile:1.18.0-bk24  --platform linux/amd64,linux/arm64  --push "https://github.com/moby/buildkit.git?tag=dockerfile/1.18.0&checksum=b772c318368090fb2ffc9c0fed92e0a35bf82389&keep-git-dir=true" --progress=plain

BUILDX_SEND_GIT_QUERY_AS_INPUT=true BUILDX_BUILDER=bk23 docker buildx build -f ./frontend/dockerfile/cmd/dockerfile-frontend/Dockerfile -t tonistiigi/dockerfile:1.18.0-bk23  --platform linux/amd64,linux/arm64  --push "https://github.com/moby/buildkit.git?tag=dockerfile/1.18.0&checksum=b772c318368090fb2ffc9c0fed92e0a35bf82389&keep-git-dir=true" --progress=plain
```

https://oci.dag.dev/?blob=docker.io/tonistiigi/dockerfile@sha256:12461cdf7db06fb3130cee0fb963333712d63dd849929f194e358609b05e98d0&mt=application%2Fvnd.in-toto%2Bjson&size=1744

https://oci.dag.dev/?blob=docker.io/tonistiigi/dockerfile@sha256:29d88a9d397e45c98e475225bd510622b4eef65bdf9274729ddb166ce75ee91e&mt=application%2Fvnd.in-toto%2Bjson&size=1544

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -10,6 +10,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
@@ -29,6 +30,7 @@ import (
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/client/ociindex"
 	"github.com/moby/buildkit/exporter/containerimage/exptypes"
+	"github.com/moby/buildkit/frontend/dockerui"
 	gateway "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/identity"
 	"github.com/moby/buildkit/session"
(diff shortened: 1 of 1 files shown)


build/opt.go
```diff
@@ -672,6 +683,55 @@ func createTempDockerfile(r io.Reader, multiReader *SyncMultiReader) (string, er
...
+	if !ok {
+		return nil
+	}
+
+	if target.FrontendInputs == nil {
+		target.FrontendInputs = make(map[string]llb.State)
+	}
+	if name == "context" {
+		target.FrontendInputs["context"] = *st
+		delete(target.FrontendAttrs, "context")
```
crazy-max, line 725: Wonder if we could add a new frontend attribute like `context-input` that would be set with git URL so we can set it in `configSource.uri` of provenance?

```suggestion
		delete(target.FrontendAttrs, "context")
		target.FrontendAttrs["context-input"] = url
```

crazy-max (approved):

Issue #3230, opened 2025-06-10
--load and other cli flags regarding output behave different than expected


tg-freigmbh:
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

### Description

`docker buildx bake --help `
`
 --load                   Shorthand for "--set=*.output=type=docker"`

The help would indicate that its just a shorthand and nothing magical is going on

### Expected behaviour

It just a shorthand and nothing magical is going on

### Actual behaviour

```
docker buildx bake --load --print
[+] Building 0.0s (1/1) FINISHED                                                                                                                                                                                                                                                                                        
 => [internal] load local bake definitions                                                                                                                                                                                                                                                                         0.0s 
 => => reading docker-bake.hcl 59B / 59B                                                                                                                                                                                                                                                                           0.0s 
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        {
          "dest": "hi.tar",
          "type": "tar"
        }
      ]
    }
  }
}

```
It is not possible to override the ouptut using --load
```
docker buildx bake  "--set=*.output=type=docker"  --print 
[+] Building 0.0s (1/1) FINISHED                                                                                                                                                                                                                                                                                        
 => [internal] load local bake definitions                                                                                                                                                                                                                                                                         0.0s 
 => => reading docker-bake.hcl 59B / 59B                                                                                                                                                                                                                                                                           0.0s 
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        {
          "type": "docker"
        }
      ]
    }
  }
}
```

Using the longhand syntax its suddenly possible.

If it is intended that --load cant override the values speciefied in the buildfile this should be clearly documented, because this is highly confusing.
I think it would be preferable if --load would behave exactly the same as the --set syntax



### Buildx version

github.com/docker/buildx v0.22.0-desktop.1 b45d7a53aa26a5e76aae051acdb15775c0e178da

### Docker info

```text

```

### Builders list

```text
[...]

PR #3421, opened 2025-09-11
docs: --load flag clarity

aevesdocker:
Fixes https://github.com/docker/buildx/issues/3230

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -499,8 +499,8 @@ func bakeCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags := cmd.Flags()
 
 	flags.StringArrayVarP(&options.files, "file", "f", []string{}, "Build definition file")
-	flags.BoolVar(&options.exportLoad, "load", false, `Shorthand for "--set=*.output=type=docker"`)
-	flags.BoolVar(&options.exportPush, "push", false, `Shorthand for "--set=*.output=type=registry"`)
+	flags.BoolVar(&options.exportLoad, "load", false, `Shorthand for "--set=*.output=type=docker". Conditional.`)
+	flags.BoolVar(&options.exportPush, "push", false, `Shorthand for "--set=*.output=type=registry". Conditional.`)
 	flags.StringVar(&options.sbom, "sbom", "", `Shorthand for "--set=*.attest=type=sbom"`)
 	flags.StringVar(&options.provenance, "provenance", "", `Shorthand for "--set=*.attest=type=provenance"`)
 	flags.StringArrayVar(&options.overrides, "set", nil, `Override target value (e.g., "targetpattern.key=value")`)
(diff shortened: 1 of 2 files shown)


commands/bake.go
```diff
@@ -499,7 +499,7 @@ func bakeCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags := cmd.Flags()
 
 	flags.StringArrayVarP(&options.files, "file", "f", []string{}, "Build definition file")
-	flags.BoolVar(&options.exportLoad, "load", false, `Shorthand for "--set=*.output=type=docker"`)
+	flags.BoolVar(&options.exportLoad, "load", false, `Shorthand for "--set=*.output=type=docker" if no output is already defined. Unlike "--set", it won't override outputs already defined in build files.`)
```
crazy-max: Description is a bit long, can we have instead have a new section in `docs/reference/buildx_bake.md` for `--load` flag with this explanation and example maybe?

Also we prefer build definition over build files.

I looked at the implementation in https://github.com/docker/buildx/blob/f4e02addf14b5cc299aeac89df9fff0a0eccffff/bake/bake.go#L1649-L1668

And description is not entirely correct. It enables load if output is not set in the build definition or if output in the build definition is not a local one (`local` or `tar`).

docs/reference/buildx_bake.md
```diff
@@ -13,25 +13,25 @@ Build from a file
...
+| [`--print`](#print)                 | `bool`        |         | Print the options without building                                                                                                                    |
+| [`--progress`](#progress)           | `string`      | `auto`  | Set type of progress output (`auto`, `quiet`, `plain`, `tty`, `rawjson`). Use plain to show container output                                          |
+| [`--provenance`](#provenance)       | `string`      |         | Shorthand for `--set=*.attest=type=provenance`                                                                                                        |
+| [`--pull`](#pull)                   | `bool`        |         | Always attempt to pull all referenced images                                                                                                          |
+| `--push`                            | `bool`        |         | Shorthand for `--set=*.output=type=registry`                                                                                                          |
```
crazy-max: Seems the same applies to `--push`:

```hcl
target "default" {
  output= ["type=tar,dest=hi.tar"]
}
```

```
$ docker buildx bake --push --print
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "output": [
        {
          "dest": "hi.tar",
          "type": "tar"
        }
      ]
    }
  }
}
```

Looking at the implementation: https://github.com/docker/buildx/blob/f4e02addf14b5cc299aeac89df9fff0a0eccffff/bake/bake.go#L1618-L1642

It will enable push either if output is empty in the build definition or if current output is not a local output.

commands/bake.go
```diff
@@ -499,8 +499,8 @@ func bakeCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags := cmd.Flags()
 
 	flags.StringArrayVarP(&options.files, "file", "f", []string{}, "Build definition file")
-	flags.BoolVar(&options.exportLoad, "load", false, `Shorthand for "--set=*.output=type=docker"`)
-	flags.BoolVar(&options.exportPush, "push", false, `Shorthand for "--set=*.output=type=registry"`)
+	flags.BoolVar(&options.exportLoad, "load", false, `Shorthand for "--set=*.output=type=docker". Conditional.`)
```
aevesdocker, line 502: This was the shortest way I could think of doing it

commands/bake.go
aevesdocker: Addressed feedback! Think I caught everything

PR #3425, opened 2025-09-15
driver/docker-container: remove conditional UsernsMode=host for userns

thaJeztah:
relates to:

- https://github.com/docker/buildx/pull/887
    - to address https://github.com/docker/buildx/issues/561
- https://github.com/moby/moby/pull/42736
- https://github.com/moby/moby/pull/43084
- https://github.com/moby/moby/pull/50825

-----

This special handling was added in 5f8600f09865ece328641f965e1e8af65dd37ba1, to work around an issue in the daemon. That issue was fixed in [moby@a826ca3], which was backported to docker v20.10.13 in [moby@660b996].

Given that docker 20.10 reached EOL, and any currently supported version of docker would have the fix from [moby@a826ca3], we can remove this special handling.

[moby@a826ca3]: https://github.com/moby/moby/commit/a826ca3aefbd4d29344d851723731e3809f2a4ad
[moby@660b996]: https://github.com/moby/moby/commit/660b9962e4d2d4ad68898b00571a43127b84f6c3

### DIFF
--- a/driver/docker-container/driver.go
+++ b/driver/docker-container/driver.go
@@ -25,7 +25,6 @@ import (
 	"github.com/docker/docker/api/types/image"
 	"github.com/docker/docker/api/types/mount"
 	"github.com/docker/docker/api/types/network"
-	"github.com/docker/docker/api/types/system"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/moby/buildkit/client"
@@ -205,17 +204,6 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 				hc.CgroupParent = "/docker/buildx"
 				if d.cgroupParent != "" {
 					hc.CgroupParent = d.cgroupParent
-				}
-			}
-
-			secOpts, err := system.DecodeSecurityOptions(info.SecurityOptions)
-			if err != nil {
-				return err
-			}
-			for _, f := range secOpts {
-				if f.Name == "userns" {
-					hc.UsernsMode = "host"
-					break
 				}
 			}
 		}


thaJeztah:
I see there also was a nestybox / sysbox issue linked to the original PR; in case we need to verify behavior with that somehow;

- https://github.com/nestybox/sysbox/issues/384

thaJeztah:
cc @crazy-max @AkihiroSuda

driver/docker-container/driver.go
```diff
@@ -194,17 +193,6 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
...
-			secOpts, err := system.DecodeSecurityOptions(info.SecurityOptions)
-			if err != nil {
-				return err
-			}
-			for _, f := range secOpts {
-				if f.Name == "userns" {
-					hc.UsernsMode = "host"
-					break
-				}
-			}
```
AkihiroSuda, line 220: Prior to https://github.com/docker/buildx/pull/887 , `UsernsMode` was rather unconditionally set to "host".

thaJeztah: Wah, sorry, looks like I missed your comment @AkihiroSuda

So, IIUC, before https://github.com/docker/buildx/issues/561, we unconditionally set it to `UsernsMode=host`, which was a workaround for https://github.com/moby/moby/pull/43084, because the daemon didn't correctly detect user-ns. And  https://github.com/docker/buildx/issues/561 made it a bit more granular so that we wouldn't set `UsernsMode=host` unconditionally, and only when needed.

But (again, IIUC), with https://github.com/moby/moby/pull/43084 now included in the daemon, that handling wouldn't be needed, as it would no longer be needed to set `UsernsMode=host` ?

But maybe I mis-interpreted the intent here; what's the best way to verify (other than CI in this repo?)

AkihiroSuda: > But maybe I mis-interpreted the intent here; what's the best way to verify (other than CI in this repo?)

Probably we should have a CI with userns-remap mode.
If the CI passes, I have no objection to merge this PR.

Issue #3424, opened 2025-09-15
`docker build` quiet `-q` with output `-o` specified prints empty line


jonasbadstuebner:
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

### Description

I ran `docker build -q -t myimage:local -o output-path/ .`

I'm aware that `-q` explicitly states that it prints the Image ID on success.

### Expected behaviour

It should print nothing and just put all files to the output directory.

### Actual behaviour

I get an empty line in my output.

### Buildx version

github.com/docker/buildx 0.25.0 faaea65da4ba0e58a13cd9cadcb950c51cf3b3c9

### Docker info

```text

```

### Builders list

```text
not relevant
```

### Configuration

I could provide, if necessary - but this is probably not related.

### Build logs

```text

```

### Additional info

I know it is a nitpick, but I run this command in a `make` command and it just looks as if there is something missing. I could `sed` or `awk` it away, but this just works around the root cause.

I'm happy to open a PR for this small inconvenience, I just need some input, if the output should be skipped if:
1. the image id is empty anyway? (could break scripts that check the last line of the output)
2. only if a special flag is set? (would be explicit, non-breaking and fit my use-case just fine)
3. only local and tar exports are specified? (is the image id always empty then?)

jonasbadstuebner:
I found an argument against option 1: https://github.com/docker/buildx/pull/740/files#r702071754

I would prefer option 2 and call the flag `--suppress-iid-output`.
But that would not make it clear, that it is only in quiet mode...

I wish `quiet` would just mean "no-output on success"...

tonistiigi:
I think we should just add `--progress=none`. `-q` is an old flag with weird behavior, where it controls multiple things and is only meant for build with special conditions. Controlling progress should be done with `--progress` and getting image ID/metadata with `--iidfile` or more powerful `--metadata-file`.

I was actually surprised we didn't support this already. We have `--progress=quiet`, but unfortunately, that copies the behavior of printing image ID.

jsternberg:
Closing this since `--progress=none` was added in #3431.

PR #3431, opened 2025-09-22
add progress=none option

tonistiigi:
related #3424

Currently there is no way to completely suppress the progress output. We have --progress=quiet but it has additional effect of printing image ID as a progress for historical reasons. --progress=none behaves like quiet but without this extra output printing depending on build output.

It might be worth adding this to buildkit as well to clean it up a bit. I didn't do it because in buildkit `quiet` does work correctly and this image ID printing is something specific to buildx.

In other commands except `build` there is no functional change but added option in there as well for consistency.


commands/build.go
```diff
@@ -624,7 +633,7 @@ type commonFlags struct {
 
 func commonBuildFlags(options *commonFlags, flags *pflag.FlagSet) {
 	options.noCache = flags.Bool("no-cache", false, "Do not use cache when building the image")
-	flags.StringVar(&options.progress, "progress", "auto", `Set type of progress output ("auto", "quiet", "plain", "tty", "rawjson"). Use plain to show container output`)
+	flags.StringVar(&options.progress, "progress", "auto", `Set type of progress output ("auto", "none",  "plain", "quiet", "rawjson", "tty"). Use plain to show container output`)
```
jsternberg, line 636: I think it might be more straightforward to modify the `--quiet` flag.

If you change `quiet` to be an integer, you can use `pflag.CountVarP` so `-qq` will be "very" quiet. Then we can also add `--progress=none` and have it set `o.quiet = 2`. If `o.quiet <= 1` then we'll print the image id.

I think that will keep backwards compatibility and will be an easier command line experience.

tonistiigi: I don't get it. `--quiet` is a boolean atm. Afaics you can't change its type in a backwards-compatible way.

This isn't about extending the capabilites of `--quiet`. For me it is flag that you should never use and only provided for historic reasons. Ideally we should have left the printing on image ID only to `--quiet` and make `--progress=quiet` actually return empty progress but can't change that anymore.

PR #3444, opened 2025-10-01
driver/kubernetes/context: don't use ResolveDefaultContext in test

thaJeztah:
- relates to https://github.com/docker/cli/pull/6528

The ResolveDefaultContext function is only used internally by the CLI, and has no known external users, except for this test in buildx. It was exported in [cli@f820766] to allow (unit) testing, but did not document that it was only exported for this purpose.

This patch rewrites the test to allow deprecating / removing the function in the CLI.

[cli@f820766]: https://github.com/docker/cli/commit/f820766f6ac57188d96c9ca377f2b4627e90da28

### DIFF
--- a/driver/kubernetes/context/load_test.go
+++ b/driver/kubernetes/context/load_test.go
@@ -13,10 +13,20 @@ import (
 
 func TestDefaultContextInitializer(t *testing.T) {
 	t.Setenv("KUBECONFIG", "./fixtures/test-kubeconfig")
-	ctx, err := command.ResolveDefaultContext(&cliflags.ClientOptions{}, command.DefaultContextStoreConfig())
+	dockerCLI, err := command.NewDockerCli()
 	require.NoError(t, err)
-	assert.Equal(t, "default", ctx.Meta.Name)
-	assert.Equal(t, "zoinx", ctx.Meta.Endpoints[KubernetesEndpoint].(EndpointMeta).DefaultNamespace)
+
+	// FIXME(thaJeztah): the context-store is not initialized until "Initialize" is called.
+	err = dockerCLI.Initialize(cliflags.NewClientOptions())
+	require.NoError(t, err)
+
+	cs := dockerCLI.ContextStore()
+	assert.NotNil(t, cs)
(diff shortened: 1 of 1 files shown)


crazy-max (approved):

crazy-max:
Not related but macos-14 runner looks broken: https://github.com/docker/buildx/actions/runs/18158118072/job/51682545037?pr=3444

image

thaJeztah:
Yeah; github is having an incident with macOS runners.

Also worth noting (but not sure if used here) that macOS 13 is being deprecated, and they added `macos-15-intel` runners now; https://github.com/docker/cli/pull/6530

driver/kubernetes/context/load_test.go
```diff
@@ -13,10 +13,20 @@ import (
...
 	t.Setenv("KUBECONFIG", "./fixtures/test-kubeconfig")
-	ctx, err := command.ResolveDefaultContext(&cliflags.ClientOptions{}, command.DefaultContextStoreConfig())
+	dockerCLI, err := command.NewDockerCli()
 	require.NoError(t, err)
-	assert.Equal(t, "default", ctx.Meta.Name)
-	assert.Equal(t, "zoinx", ctx.Meta.Endpoints[KubernetesEndpoint].(EndpointMeta).DefaultNamespace)
+
+	// FIXME(thaJeztah): the context-store is not initialized until "Initialize" is called.
+	err = dockerCLI.Initialize(cliflags.NewClientOptions())
+	require.NoError(t, err)
```
thaJeztah, line 21: I want to have a look at this; the whole CLI (and CLI-plugin) bootstrapping is really involved. IMO, the CLI should be in a usable state after `NewDockerCLI`, but need to check if everything _CAN_ be initialised in that constructor due to some of the oddities with flag parsing and top-level flags; 

- https://github.com/docker/cli/pull/1722


If we can't initialise in the `NewDockerCli` constructor, perhaps we can add some lazy constructor for the `ContextStore()` to at least prevent it returning `nil` (and panic in this test 😂)

thaJeztah:
🎉 OK, all happy now; should be ready to merge ❤️

PR #3449, opened 2025-10-03
[v0.29] cherry-picks for v0.29.1

crazy-max:
backport of:
* https://github.com/docker/buildx/pull/3447
* https://github.com/docker/buildx/pull/3448

### DIFF
--- a/util/imagetools/create.go
+++ b/util/imagetools/create.go
@@ -222,6 +222,7 @@ func (r *Resolver) Push(ctx context.Context, ref reference.Named, desc ocispecs.
 
 func (r *Resolver) Copy(ctx context.Context, src *Source, dest reference.Named) error {
 	ctx = remotes.WithMediaTypeKeyPrefix(ctx, "application/vnd.in-toto+json", "intoto")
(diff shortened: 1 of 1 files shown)


util/imagetools/create.go
```diff
@@ -222,6 +222,7 @@ func (r *Resolver) Push(ctx context.Context, ref reference.Named, desc ocispecs.
 
 func (r *Resolver) Copy(ctx context.Context, src *Source, dest reference.Named) error {
 	ctx = remotes.WithMediaTypeKeyPrefix(ctx, "application/vnd.in-toto+json", "intoto")
+	ctx = remotes.WithMediaTypeKeyPrefix(ctx, "application/vnd.oci.empty.v1+json", "empty")
```
thaJeztah, line 225: Just to raise awareness, and in case it's relevant here and/or at some point; compose also defines a couple of media types for its artifacts, including an "empty" one; see https://github.com/docker/compose/blob/d9423f6872586c5cd9e1dd761bd5a8f02113a93c/internal/oci/push.go#L38-L60

thaJeztah (approved):
"LGTM" in case you need to be unblocked 👍

PR #3450, opened 2025-10-03
dap: utilize the container filesystem for the file explorer

jsternberg:
This utilizes the container filesystem for the file explorer instead of
using the reference directly. For the error condition, this allows us to
gain access to the filesystem when the error happened with its current
state at the time of the error rather than before the step even ran.

A better solution for resolving #3410. Relies on functionality in #6262.

### DIFF
--- a/dap/thread.go
+++ b/dap/thread.go
@@ -294,13 +294,13 @@ func (t *thread) pause(c Context, ref gateway.Reference, err error, mounts map[s
 	}
 	t.paused = make(chan stepType, 1)
 
-	ctx, cancel := context.WithCancelCause(c)
-	t.collectStackTrace(ctx, pos, mounts)
-	t.cancel = cancel
-
 	if ref != nil || err != nil {
 		t.prepareResultHandle(c, ref, err)
 	}
+
+	ctx, cancel := context.WithCancelCause(c)
+	t.collectStackTrace(ctx, pos)
+	t.cancel = cancel
 
 	event.ThreadId = t.id
 	c.C() <- &dap.StoppedEvent{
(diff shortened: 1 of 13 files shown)


jsternberg:
Finally had time to come back to this and update it for what got merged. It's tested and works with the capability detection. When the capability isn't present, it will not attempt to use the result handle which causes it not to include cache mounts. When the capability is there, cache mounts are present.

dap/thread.go
```diff
@@ -678,3 +709,52 @@ func (t *thread) rewind(ctx Context, inErr error) (k string, result *step, mount
...
 	return k, result, mounts, inErr
 }
+
+type mountReference struct {
+	getContainer func() (gateway.Container, error)
+	index        int
+}
+
+func (r *mountReference) ToState() (llb.State, error) {
+	return llb.State{}, errors.New("unimplemented")
```
tonistiigi: nit: add method name to unimplemented message

dap/thread.go
```diff
@@ -353,6 +359,31 @@ func (t *thread) prepareResultHandle(c Context, ref gateway.Reference, err error
...
+		var solveErr *errdefs.SolveError
+		// If we had a solve error and the exec filesystem capability, we can
+		// get the filesystem mounts used in the actual build rather than only the input
+		// mounts.
+		if gwcaps.Supports(gwpb.CapGatewayExecFilesystem) == nil && errors.As(err, &solveErr) {
+			if exec, ok := solveErr.Op.Op.(*pb.Op_Exec); ok {
+				rCtx := t.rCtx
+
+				getContainer := sync.OnceValues(func() (gateway.Container, error) {
+					return rCtx.NewContainer(ctx, &build.InvokeConfig{})
```
tonistiigi: Is there a way this calls `ReleaseContainer`.

The resources should be released anyway when the gateway session expires, but I am wondering if we can be explicit.

jsternberg: Discussed this offline. This was intended to use `build.NewContainer` which registers this cleanup with the result context and executes it when `Done()` is called. I mistakenly called the function directly instead of through `build.NewContainer`. I've modified this to use `build.NewContainer` and added for methods to forward the `ReadFile`/`ReadDir`/`StatFile` calls to the container directly to `*build.Container` so we can just reuse that code here.

PR #3453, opened 2025-10-06
docker-container: write github actions payload to container for provenance

crazy-max:
alternative to and closes #3452 

follow-up https://github.com/moby/buildkit/pull/6210

needs docs follow-up at https://docs.docker.com/build/builders/drivers/docker-container/#synopsis

example: https://oci.dag.dev/?blob=public.ecr.aws/q3b5f1u4/test-docker-action@sha256:176f66ab52141487a3e5b06058f391d3e44c1f86511ec54a692eccd10989853d&mt=application%2Fvnd.in-toto%2Bjson&size=17247


driver/docker-container/driver.go
```diff
@@ -564,3 +579,33 @@ func isSocket(addr string) bool {
...
+		var evt map[string]any
+		if err := json.Unmarshal(dt, &evt); err == nil {
+			m["github_event_payload"] = evt
+		}
+	}
+	if len(m) == 0 {
+		return "", nil
+	}
+	dt, _ := json.Marshal(m)
+	d, err := os.MkdirTemp("", "gha-event")
```
crazy-max: Maybe temp dir is not a good location and should instead be in our config dir?

driver/docker-container/factory.go
```diff
@@ -111,6 +111,11 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
 				return nil, errors.Errorf("invalid env option %q, expecting env.FOO=bar", k)
 			}
 			d.env = append(d.env, fmt.Sprintf("%s=%s", envName, v))
+		case k == "no-gha-event":
```
tonistiigi: Something more descriptive like "provenance-add-github-event" (provenance-add-github-actions-context)

driver/docker-container/driver.go
```diff
@@ -564,3 +577,25 @@ func isSocket(addr string) bool {
 		return false
 	}
 }
+
+func githubActionsEvent() ([]byte, error) {
+	m := make(map[string]any)
+	if v := os.Getenv("GITHUB_EVENT_NAME"); v != "" {
```
tonistiigi: I think we should only add the event scope if all the envs we expect to be present are actually present.

driver/docker-container/driver.go
```diff
@@ -137,6 +139,17 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
...
 		}
 
+		if !d.noGHAEvent {
+			if ghaedt, err := githubActionsEvent(); err != nil {
+				return err
+			} else if ghaedt != nil {
+				if d.Files == nil {
+					d.Files = make(map[string][]byte)
+				}
+				d.Files["provenance.d/github_event.json"] = ghaedt
```
tonistiigi: `github_actions_context.json`

driver/docker-container/driver.go
```diff
@@ -564,3 +578,139 @@ func isSocket(addr string) bool {
 		return false
 	}
 }
+
+func githubActionsContext() ([]byte, error) {
```
tonistiigi: Even though this is only used by container driver atm. I think it is better placed in generic pkg as not really driver specific in any way.

crazy-max: Moved to `util/ghutil` and added some tests.

Issue #872, opened 2021-12-02
Unable to set multiple tags in bake cmd


zachary-povey:
## Steps to reproduce
- Create an empty dir
- Add a `Dockerfile`:
```
FROM centos:8
```
- Add a `docker-bake.hcl`:
```
group "default" {
  targets = ["image"]
}
target "image" {
  context = "."
}
```
- Run docker `buildx bake --set "*.tags=image_name_1" image --load` - this passes as expected and loads `image_1:latest` into the local docker instance
- Run `docker buildx bake --set "*.tags=image_name_1,image_name_2" image --load` - this failes with:
```
[+] Building 0.0s (0/0)                                                                                                                                              
error: invalid tag "image_name_1,image_name_2": invalid reference format
```

## Expected behaviour
Based on the plural name, and looking at the code, I believe the intent is for `target.tags` to be an array - am I using the wrong separator or is this a bug?

marcellodesales:
I particularly had a hard time to get around it as well... It would be great if this is well documented with examples...

pongnguy:
Similar comment regarding docker buildx bake -f, where -f takes a stringArray.  I want to pass in multiple docker-bake.json files.

berney:
In the a docker-bake.hcl file, we would set the tags like this:

```hcl
# docker-bake.hcl
target "webapp-dev" {
  dockerfile = "Dockerfile.webapp"
  tags = ["docker.io/username/webapp:latest"]
}
```

Where `tags` takes an array. So, I expected the arg to `--set` to accept an array, e.g. `docker buildx bake --set 'webapp-dev.tags=[foo/bar:latest,foo/bar:v1.2.3]'`, that didn't work and searching I found this issue.

I would have expected repeating  args, that the last arg would override the previous args. I think it makes more sense that the arg is an array, like the HCL syntax.

neersighted:
It turns out shell brace expansion is a pretty good fit here, so maybe we can document that? e.g. `docker buildx bake all --set=\*.tags=foo/bar:{baz,qux}`.

tsouza-squid:
Not sure if documentation was updated recently, but this is buried down in the [docs](https://docs.docker.com/reference/cli/docker/buildx/bake/#set):
```bash
docker buildx bake --set target.platform+=linux/arm64 # appends 'linux/arm64' to the platform list
```

crazy-max:
[...]

PR #3456, opened 2025-10-07
docs: update --set

aevesdocker:
fixes https://github.com/docker/buildx/issues/872

Also has an equivalent docs PR https://github.com/docker/docs/pull/23512

### DIFF
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -410,6 +410,12 @@ $ docker buildx bake --set foo*.no-cache                # bypass caching only fo
 $ docker buildx bake --set target.platform+=linux/arm64 # appends 'linux/arm64' to the platform list
 ```
 
+> [!NOTE]
+>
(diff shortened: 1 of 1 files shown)


docs/reference/buildx_bake.md
```diff
@@ -410,6 +410,12 @@ $ docker buildx bake --set foo*.no-cache                # bypass caching only fo
 $ docker buildx bake --set target.platform+=linux/arm64 # appends 'linux/arm64' to the platform list
 ```
 
+`--set` is a repeatable flag. For array fields such as `tags`, repeat `--set` to provide multiple values or use the `+=` operator to append without replacing.
+
+Array literal syntax like `--set target.tags=[a,b]` is not supported.
+
+The `platform` field is a special case. it also accepts a comma-separated list.
```
crazy-max: Can you make a note callout similar to the one on docs website?: https://github.com/docker/docs/pull/23512/files#diff-9b3c11d573a8ffdfe530ac6a8f3a60d7755057e11fda394947212bfe4e8f4775R215-R219

PR #3459, opened 2025-10-08
history: input:context attribute support

crazy-max:
follow-up https://github.com/docker/buildx/pull/3415


commands/history/utils.go
```diff
@@ -25,7 +25,7 @@ import (
 
 const recordsLimit = 50
 
-func buildName(fattrs map[string]string, ls *localstate.State) string {
+func BuildName(fattrs map[string]string, ls *localstate.State) string {
```
crazy-max, line 28: expose build name for desktop build backend

commands/history/utils.go
```diff
@@ -40,9 +40,13 @@ func buildName(fattrs map[string]string, ls *localstate.State) string {
 		contextPath = filepath.ToSlash(v)
 	} else if v, ok := fattrs["vcs:localdir:context"]; ok && v != "." {
 		contextPath = filepath.ToSlash(v)
+	} else if v, ok := fattrs["input:context"]; ok && !build.IsRemoteURL(v) {
+		contextPath = filepath.ToSlash(v)
```
tonistiigi: What case is this?

crazy-max: Ah yes I was thinking we might have a future case where it's a path but not necessary here. Removed.

commands/history/utils.go
```diff
@@ -43,6 +43,8 @@ func buildName(fattrs map[string]string, ls *localstate.State) string {
 	}
 	if v, ok := fattrs["vcs:source"]; ok {
 		vcsSource = v
+	} else if v, ok := fattrs["input:context"]; ok && build.IsRemoteURL(v) {
+		vcsSource = v
```
tonistiigi: Did you want to update this to the fragment-style URL?

crazy-max: Yes will do, I'm looking to make a public func on BuildKit side as it will be used on both side

tonistiigi (approved):

PR #3460, opened 2025-10-08
build loong64 arch

crazy-max:
~needs https://github.com/tonistiigi/xx/pull/188~

cc @wojiushixiaobai


thompson-shaun:
Unfortunately, we don't have plans to support this architecture, as we won't be able to provide ongoing maintenance and testing for it. Closing for now.

PR #3474, opened 2025-10-17
docs: fix bake set flag note

crazy-max:


docs/reference/buildx_bake.md
```diff
@@ -412,9 +412,10 @@ $ docker buildx bake --set target.platform+=linux/arm64 # appends 'linux/arm64'
 
 > [!NOTE]
 >
-> `--set` is a repeatable flag. For array fields such as `tags`, repeat `--set` to provide multiple values or use the `+=` operator to append without replacing.
-> Array literal syntax like `--set target.tags=[a,b]` is not supported.
-> The `platform` field is a special case. It also accepts a comma-separated list.
```
crazy-max, line 417: `platform` is actually not a special case:

```hcl
# docker-bake.hcl
target "default" {}
```

```
$ docker buildx bake --set "*.platform=linux/amd64,linux/arm64" --print
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "platforms": [
        "linux/amd64,linux/arm64"
      ]
    }
  }
}
```

It needs to be repeated:

```
$ docker buildx bake --set "*.platform=linux/amd64" --set "*.platform=linux/arm64" --print
{
  "group": {
    "default": {
      "targets": [
        "default"
      ]
    }
  },
  "target": {
    "default": {
      "context": ".",
      "dockerfile": "Dockerfile",
      "platforms": [
        "linux/amd64",
        "linux/arm64"
      ]
    }
  }
}
```

PR #3498, opened 2025-11-03
dockerfile: update to docker v29.0.0

crazy-max:
closes #3477

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,6 +76,16 @@ jobs:
           - worker: docker+containerd  # same as docker, but with containerd snapshotter
             pkg: ./tests
             mode: experimental
+          - worker: "docker@28.5"
+            pkg: ./tests
+          - worker: "docker+containerd@28.5"  # same as docker, but with containerd snapshotter
+            pkg: ./tests
+          - worker: "docker@28.5"
+            pkg: ./tests
+            mode: experimental
+          - worker: "docker+containerd@28.5"  # same as docker, but with containerd snapshotter
+            pkg: ./tests
+            mode: experimental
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -76,24 +76,24 @@ jobs:
...
             mode: experimental
-          - worker: "docker+containerd@27.5"  # same as docker, but with containerd snapshotter
+          - worker: "docker+containerd@28.5"  # same as docker, but with containerd snapshotter
             pkg: ./tests
             mode: experimental
-          - worker: "docker@26.1"
+          - worker: "docker@27.5"
             pkg: ./tests
-          - worker: "docker+containerd@26.1"  # same as docker, but with containerd snapshotter
+          - worker: "docker+containerd@27.5"  # same as docker, but with containerd snapshotter
```
thaJeztah, line 91: Wondering if we should include a 25.0 somewhere; 26.x and 27.x are EOL, but 25.0 is maintained by mirantis and AWS

thaJeztah (approved):
LGTM, but left a suggestion

.github/workflows/build.yml
crazy-max: We chose to keep n-2 for our integration tests to check for regressions with past stables like we do for BuildKit: https://github.com/docker/buildx/blob/45731b890203e93b049f83670df1455175cc408e/.github/workflows/build.yml#L57-L59

But also to check for issues with default Docker version used on GitHub Hosted runners: https://github.com/crazy-max/ghaction-dump-context/actions/runs/19068023635/job/54463406479#step:3:638

Maybe we could add 25 if you think this can be useful.

PR #3509, opened 2025-11-11
docs: document using build contexts to override syntax directive

jedevc:
This functionality was added in moby/buildkit#3633. It's quite useful, especially for developers of new frontends. I couldn't find any docs for the feature anywhere, but figured that this would be a good place to include it (it's the first place I looked).

### DIFF
--- a/docs/reference/buildx_build.md
+++ b/docs/reference/buildx_build.md
@@ -241,8 +241,13 @@ Learn more about the built-in build arguments in the [Dockerfile reference docs]
 --build-context=name=VALUE
 ```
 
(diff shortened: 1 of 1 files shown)


crazy-max (approved):

crazy-max:
@dvdksn Might be worth a change on docs repo as well? https://docs.docker.com/build/concepts/context/#named-contexts

dvdksn (approved):
Nice one. I didn't know about this either.

We don't document custom syntaxes a lot. I'm sure we could do more of that. It's kind of a niche market though.

Issue #3508, opened 2025-11-11
segmentation fault on multi-image build with bake


orgads:
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

### Description

Building a multiarch image with bake where passing contexts causes segmentation fault.

### Expected behaviour

It should work

### Actual behaviour

Segmentation fault.

### Buildx version

github.com/docker/buildx 0.19.3+ds1 0.19.3+ds1-4

### Docker info

```text
Client:
 Version:    27.5.1+dfsg3
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.19.3+ds1
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.32.4-3
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 25
  Running: 18
  Paused: 0
  Stopped: 7
 Images: 79
 Server Version: 27.5.1+dfsg3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 1.7.24~ds1-10
 runc version: 1.3.2+ds1-1
 init version: 
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.6.87.2-microsoft-standard-WSL2
 Operating System: Debian GNU/Linux forky/sid
 OSType: linux
 Architecture: x86_64
 CPUs: 28
 Total Memory: 31.16GiB
 Name: IL-ORGADS-NEW
 ID: ade50a3a-44af-40af-b718-5dfc3eb3bdd5
 Docker Root Dir: /var/lib/docker
[...]

PR #3511, opened 2025-11-11
build: avoid panic on linking to nil target

tonistiigi:
improves #3508

Note that nil targets still do not work as buildkit does not currently support build inputs with nil values.

But this gives cleaner error from buildkit instead of panic on the client side.


orgads:
Thank you!

PR #3519, opened 2025-11-13
build: Don't unpack by default when pushing

vvoland:
- alternative to: https://github.com/moby/moby/pull/51494
- fixes https://github.com/moby/moby/issues/51442

Automatically set `unpack=false` for registry exports unless explicitly overridden by the user.

This applies to:

- `registry` exporter type (converted to `image` exporter with `push=true`)
- `--push` flag usage with image exporters

Users can still explicitly set `unpack=true` if they need local image storage alongside registry push.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -542,7 +542,7 @@ func buildCmd(dockerCli command.Cli, rootOpts *rootOptions, debugger debuggerOpt
 
 	flags.StringArrayVar(&options.platforms, "platform", platformsDefault, "Set target platform for build")
 
-	flags.BoolVar(&options.exportPush, "push", false, `Shorthand for "--output=type=registry"`)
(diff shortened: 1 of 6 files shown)


commands/build.go
```diff
@@ -1047,15 +1047,22 @@ func RunBuild(ctx context.Context, dockerCli command.Cli, in *BuildOptions, inSt
...
+			}
+			// Skip unpacking when only pushing to registry
+			attrs["unpack"] = "false"
 			outputs = append(outputs, client.ExportEntry{
-				Type: client.ExporterImage,
-				Attrs: map[string]string{
-					"push": "true",
-				},
+				Type:  client.ExporterImage,
+				Attrs: attrs,
```
crazy-max: Should be enough:

```suggestion
				Attrs: map[string]string{
					"push":   "true",
					"unpack": "false",
				},
```

vvoland: Done

PR #3005, opened 2025-02-18
ci: distribute multi-platform bin-image build

crazy-max:
Testing https://github.com/docker/bake-action/pull/302

PR #3040, opened 2025-03-04
ci: fix faulty bin-image job

crazy-max:
relates to https://github.com/docker/buildx/actions/runs/13643604134/job/38138677210

```
System.IO.IOException: No space left on device : '/home/runner/runners/2.322.0/_diag/Worker_20250304-005208-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.Strategies.OSFileStreamStrategy.Write(ReadOnlySpan`1 buffer)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
System.IO.IOException: No space left on device : '/home/runner/runners/2.322.0/_diag/Worker_20250304-005208-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.Strategies.OSFileStreamStrategy.Write(ReadOnlySpan`1 buffer)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Common.Tracing.Error(Exception exception)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
[...]

PR #3520, opened 2025-11-13
ci: use docker github builder to build bin image and binaries

crazy-max:
https://github.com/docker/github-builder-experimental/issues/21
closes #3005
closes https://github.com/docker/buildx/pull/3441

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -347,19 +347,20 @@ jobs:
           sarif_file: ${{ env.DESTDIR }}/govulncheck.out
 
   binaries:
-    uses: docker/github-builder-experimental/.github/workflows/bake.yml@8fc70909404a502fd0eca6601b99b32fa7192b03
+    uses: docker/github-builder-experimental/.github/workflows/bake.yml@5876e8deef3c899c298ec80b07c43dd9e89d37f6
     permissions:
       contents: read # same as global permission
       id-token: write # for signing attestation(s) with GitHub OIDC Token
     with:
       runner: amd64
+      artifact-name: buildx
+      artifact-upload: true
+      cache: true
+      cache-scope: bin-image
       target: release
       output: local
-      push: ${{ github.event_name != 'pull_request' }}
-      artifact-name: buildx
-      cache: true
-      cache-scope: binaries
-      bake-sbom: true
+      sbom: true
+      sign: ${{ github.event_name != 'pull_request' }}
 
   binaries-finalize:
     runs-on: ubuntu-24.04
@@ -371,8 +372,7 @@ jobs:
         uses: actions/download-artifact@v6
         with:
           path: /tmp/buildx-output
-          pattern: ${{ needs.binaries.outputs.artifact-name }}*
-          merge-multiple: true
+          name: ${{ needs.binaries.outputs.artifact-name }}
       -
         name: Rename provenance and sbom
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -408,67 +408,34 @@ jobs:
...
+      id-token: write # for signing attestation manifests with GitHub OIDC Token
+      packages: write # needed to push images to GitHub Container Registry
+    with:
+      target: image-cross
+      output: ${{ github.event_name != 'pull_request' && 'registry' || 'cacheonly' }}
+      cache: true
+      cache-scope: bin-image
+      set-meta-labels: true
+      meta-images: |
+        ${{ env.REPO_SLUG }}
```
crazy-max: Can't use `env` with reusable workflows :disappointed:

https://github.com/docker/buildx/actions/runs/19329818722

image

Seems we need an extra job just to set it as an output: https://github.com/orgs/community/discussions/26671 :expressionless:

:upside_down_face: https://github.com/orgs/community/discussions/26671#discussioncomment-4295807

crazy-max:
GitHub hosted runner doesn't have enough space to build our image: https://github.com/docker/buildx/actions/runs/19330419315/job/55292703625#step:8:521

```
#28 [linux/amd64->darwin/amd64 buildx-build 1/1] RUN --mount=type=bind,target=.   --mount=type=cache,target=/root/.cache   --mount=type=cache,target=/go/pkg/mod   --mount=type=bind,from=buildx-version,source=/buildx-version,target=/buildx-version <<EOT (set -e...)
#28 1051.0 k8s.io/api/autoscaling/v2beta1: mkdir /tmp/go-build135885827/b986/: no space left on device
```

Hence why we free up disk space currently (see https://github.com/docker/buildx/pull/3040)

For our reusable workflow, one way to fix it would be to distribute build across runners: https://github.com/docker/github-builder-experimental/issues/2

crazy-max:
Switch reusable workflow ref to `bake-distrib` https://github.com/docker/github-builder-experimental/pull/22 so builds are distributed across runners.

crazy-max:
Pushed extra commit to also build our binaries with our reusable workflow.

.github/workflows/build.yml
```diff
@@ -346,129 +346,115 @@ jobs:
...
+  binaries:
+    uses: docker/github-builder-experimental/.github/workflows/bake.yml@8fc70909404a502fd0eca6601b99b32fa7192b03
+    permissions:
+      contents: read # same as global permission
+      id-token: write # for signing attestation(s) with GitHub OIDC Token
+    with:
+      runner: amd64
+      target: release
+      output: local
+      push: ${{ github.event_name != 'pull_request' }}
```
dvdksn: This one shouldn't be here right?
```suggestion
```

dvdksn: Oh I see:

https://github.com/docker/github-builder-experimental/blob/36714dbaf49f1f3518cd865748c85b863cdbcbe4/.github/workflows/bake.yml#L36C69-L36C117

Something like `sign-artifacts: ${{ github.event_name != 'pull_request' }}` might be more intuitive, splitting push and manifest signing for local artifacts.

crazy-max: Yes I'm working on refactoring the UX, I don't really like `push` as well, that's misleading. I will open an issue on github-builder repo to address this :pray:

PR #3521, opened 2025-11-13
docs: add a link from the cli reference to the bake file reference for variables

jsternberg:
I spent some time looking how to override a variable in a bakefile. My
assumption was that there would be a CLI option to do that and was
confused when I couldn't find it. This adds a reference to point someone
in the correct direction of using environment variables which is present
in the bake file reference but isn't in the CLI documentation.

### DIFF
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -458,3 +458,7 @@ You can append using `+=` operator for the following fields:
 
 > [!NOTE]
 > ¹ These fields already append by default.
+
+### <a name="variables"></a> Override variables in a Bake file
+
+See the [Variable](buildx_bake.md#variable) section in the Bake file reference.


docs/reference/buildx_bake.md
```diff
@@ -458,3 +458,7 @@ You can append using `+=` operator for the following fields:
 
 > [!NOTE]
 > ¹ These fields already append by default.
+
+### <a name="variables"></a> Override variables in a Bake file
```
dvdksn, line 462: We typically only include an anchor if the section corresponds to a flag/option on the command. That isn't the case here so we can remove this.
```suggestion
### Override variables in a Bake file
```

jsternberg:
A formal `--var` option is added to `bake` now so this documentation isn't needed anymore.

PR #3534, opened 2025-11-24
dap: Fix breakpoint overlap check

cpuguy83:
The old logic wasn't quite right, and a little hard to follow. Before this, because breakpoints sent from setBreakpoints don't have an end line or column they are just a single point the comparison would fail the check due to the column check.

E.g. if I have a breakpoint on line 30.
The content lines range from 29-31.
Now lets say the end column is at postion 10, but my breakpoint is at L30 column 11 (still within range).
The overlap check would fail because 11 > 10 but these shouldn't be compared at all in this case.


cpuguy83:
Added a test.

PR #3539, opened 2025-11-26
add rego integration to source policies

tonistiigi:
This is an early version of [rego](https://www.openpolicyagent.org/docs/policy-language) format support for defining source policies. Not ready for merge but suitable for early testing and feedback.

Currently many untested parts and unimplemented fields/builtins. I also discovered a bug with git/http metadata resolve in BuildKit that needs to be fixed for support of some fields.

For input schema see `policy/types.go` atm until we expose it better.

For `app.Dockerfile`, matching `app.Dockerfile.rego` is loaded. Only local files supported atm. This would be extended with manual control via flags and helper commands for testing.

Set `export BUILDX_POLICY_DEBUG=1` and use plain progress mode to see the internal input data and policy decisions. This is a temporary debug until there is better progressbar integration.

Some examples in ~https://gist.github.com/tonistiigi/a8a1fdf39796ba484a31af18afb04bfc~ https://github.com/tonistiigi/buildx-rego-examples . BuildKit v0.26 is needed.

@crazy-max @dvdksn @cpuguy83 @colinhemmings


dvdksn:
Encountered this:

```dockerfile
FROM alpine
```

```rego
package docker

default allow := false

allow if input.local

allow if {
  input.image.hasProvenance
}

decision := {
    "allow": allow 
}
```

```text
ERROR: failed to build: failed to solve: alpine: failed to resolve source metadata for docker.io/library/alpine:latest: could not resolve image due to policy: too many policy requests
```

tonistiigi:
@dvdksn should be fixed as part of https://github.com/moby/buildkit/pull/6383 https://gist.github.com/tonistiigi/a8a1fdf39796ba484a31af18afb04bfc#file-9-provenance-dockerfile . Although it looks like there is another issue where if the image is not an OCI index, then you get a different error when parsing instead of a policy deny error. Signature keys have not been implemented yet in this PR.

policy/types.go
```diff
@@ -0,0 +1,137 @@
...
+
+type HTTP struct {
+	URL     string              `json:"url,omitempty"`
+	Schema  string              `json:"schema,omitempty"`
+	Host    string              `json:"host,omitempty"`
+	Path    string              `json:"path,omitempty"`
+	Query   map[string][]string `json:"query,omitempty"`
+	Perm    int                 `json:"perm,omitempty"`
+	UID     int                 `json:"uid,omitempty"`
+	GID     int                 `json:"gid,omitempty"`
```
dvdksn: Perm would always be 600 I guess - and not sure what UID/GID would show?

tonistiigi: Maybe we should remove them. These are options you can set in LLB that set the values for the downloaded file. But these are not exposed in Dockerfile, and if you use smth like `COPY --chmod` then this is actually different mode that is applied during the copy.

build/opt.go
```diff
@@ -535,6 +552,19 @@ func loadInputs(ctx context.Context, d *driver.DriverHandle, inp *Inputs, pw pro
 			return nil, err
 		}
 		dockerfileName = handleLowercaseDockerfile(dockerfileDir, dockerfileName)
+
+		if fi, err := os.Lstat(filepath.Join(dockerfileDir, dockerfileName+".rego")); err == nil {
```
AkihiroSuda, line 707: nit: err should be returned if `err != nil && !errors.Is(err, os.ErrNotExist)`
e.g., EACCES

build/opt.go
```diff
@@ -535,6 +552,19 @@ func loadInputs(ctx context.Context, d *driver.DriverHandle, inp *Inputs, pw pro
...
+			if fi.Mode().IsRegular() {
```
AkihiroSuda, line 708: Any reason to reject symlink?

tonistiigi: Just wanted to make sure there is no breakout here. Normally this shouldn't matter as Dockerfile path symlinks are already resolved and default policy is relative to that resolved path. But I'm ok with symlinks only for policy as well, as long as breakout cases are protected.

tonistiigi:
Added new features:
- `print()` for debugging
- import support for loading shared policies (only local context atm)
- `load_json()` for importing JSON files (only local context atm)
- git signature fields and verification with local key

New examples in https://github.com/tonistiigi/buildx-rego-examples . I needed to move away from gist as it doesn't allow pushing directories.

tonistiigi:
Added image signature verification example.

tianon:
> For `app.Dockerfile`, matching `app.Dockerfile.rego` is loaded.

As a `Dockerfile` author / content creator, I'm honestly struggling to understand the purpose of this -- why would I, as the content author, want to specify policies for the build of my `Dockerfile`?  As-is, this feels like a fully out-of-band way to add extra constraints to the `Dockerfile` itself, and I'm not really understanding what the goals are (or why they're all so fully out-of-band).

Is there a design document somewhere that I could read that explains in more detail what problems are being solved?

From the DOI perspective, _some_ kind of policy engine is interesting, since I'd like to apply polices to the `Dockerfile`s written by third parties / control the "remote content fetching" behavior of BuildKit more explicitly, but not if it's picking up policies written by the content authors. :thinking:

tonistiigi:
> As a Dockerfile author / content creator, I'm honestly struggling to understand the purpose of this -- why would I, as the content author, want to specify policies for the build of my Dockerfile?

To protect yourself from your project becoming a target when one of the upstreams gets attacked or becomes malicious. Same reason why we add digests to critical dependencies in Dockerfile, or why Akihiro wants to improve the security of Moby in PRs like https://github.com/moby/moby/pull/51626/files . Except, 1) this is much more powerful than digest check in the image/git ref 2) you don't need to update to a new digest every time there is an update in upstream or you want to build a different configuration with build-arg 3) digests are not human-readable, making them poor to verify in PR review.

>  feels like a fully out-of-band
> since I'd like to apply polices to the Dockerfiles written by third parties

It's not out-of-band or restricted to upstream only. It is part of the project and loaded automatically. Given the amount of functionality in here, it is impossible to inline something like this into Dockerfile itself, and it would make for very poorly readable Dockerfiles. But they are tightly integrated and a policy is written for a specific (set of) Dockerfiles. As they serve different purposes I think it is very useful that they are not mixed together so you can concentrate on a specific aspect while writing each file.

Additionally you can add your own policies when invoking cli if you want specific additional constraints. This would need a new flag that is not implemented yet, but mentioned in the description.

If you are the author of DOI Dockerfile you can write a policy that protects you, your CI and release bot, and all the other 3rd parties either working with your codebase or building their own versions of the artifacts using your Dockerfile. They don't need to use your policy and can opt-out or set their own more restrictive policy, but you as a repo/Dockerfile author can define their secure-by-default experience.

tonistiigi:
Added `--policy` flag, `buildx policy eval` command from https://github.com/docker/buildx/issues/3567 , and redid the logging through build progressbar.

commands/policy/eval.go
```diff
@@ -0,0 +1,419 @@
...
+	printOutput bool
+	fields      []string
+	builder     *string
+}
+
+func evalCmd(dockerCli command.Cli, rootOpts RootOptions) *cobra.Command {
+	var opts evalOpts
+
+	cmd := &cobra.Command{
+		Use:                   "eval source",
```
crazy-max: Can we make the command usage consistent with other commands? Atm it's a bit opaque:

```
docker buildx policy eval --help
Usage:  docker buildx policy eval source

Evaluate policy for a source

Options:
      --builder string    Override the configured builder instance
  -D, --debug             Enable debug logging
      --fields strings    Fields to evaluate
      --filename string   Policy filename to evaluate (default "Dockerfile")
      --print             Print policy output
```

Something like:

```suggestion
		Use:                   "eval [OPTIONS] SOURCE",
```

commands/policy/eval.go
```diff
@@ -0,0 +1,419 @@
...
+		Short:                 "Evaluate policy for a source",
+		Args:                  cobra.ExactArgs(1),
+		DisableFlagsInUseLine: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			opts.builder = rootOpts.Builder
+			return runEval(cmd.Context(), dockerCli, args[0], opts)
+		},
+	}
+	cmd.Flags().StringVar(&opts.filename, "filename", "Dockerfile", "Policy filename to evaluate")
```
crazy-max, line 51: I guess filename is the policy so default should be:

```suggestion
	cmd.Flags().StringVar(&opts.filename, "filename", "Dockerfile.rego", "Policy filename to evaluate")
```

commands/policy/eval.go
```diff
@@ -0,0 +1,419 @@
...
+	builder     *string
+}
+
+func evalCmd(dockerCli command.Cli, rootOpts RootOptions) *cobra.Command {
+	var opts evalOpts
+
+	cmd := &cobra.Command{
+		Use:                   "eval source",
+		Short:                 "Evaluate policy for a source",
+		Args:                  cobra.ExactArgs(1),
```
crazy-max, line 44: Could we make it optional? So if not defined it checks `Dockerfile` and use `Dockerfile.rego` policy?

commands/policy/eval.go
```diff
@@ -0,0 +1,419 @@
...
+			return runEval(cmd.Context(), dockerCli, args[0], opts)
+		},
+	}
+	cmd.Flags().StringVar(&opts.filename, "filename", "Dockerfile", "Policy filename to evaluate")
+	cmd.Flags().BoolVar(&opts.printOutput, "print", false, "Print policy output")
+	cmd.Flags().StringSliceVar(&opts.fields, "fields", nil, "Fields to evaluate")
+	return cmd
+}
+
+func runEval(ctx context.Context, dockerCli command.Cli, source string, opts evalOpts) error {
```
crazy-max, line 57: If I only set the source I think it should compute the attached rego automatically:

```
$ docker buildx policy eval 1-locals.Dockerfile
ERROR: failed to read policy file Dockerfile.rego: open Dockerfile.rego: no such file or directory
```

Here I think it should look for `1-locals.Dockerfile.rego` instead of default filename.

crazy-max: Also I don't get why setting filename adds the `.rego` extension?

```
$ docker buildx policy eval 1-locals.Dockerfile --filename 1-locals.Dockerfile.rego
ERROR: failed to read policy file 1-locals.Dockerfile.rego.rego: open 1-locals.Dockerfile.rego.rego: no such file or directory
```

Works if I don't specify it but that's confusing imo:

```
$ docker buildx policy eval 1-locals.Dockerfile --filename 1-locals.Dockerfile --print
{
  "local": {
    "name": "context"
  }
}
```

commands/policy/json_schema.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func jsonSchemaCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "json-schema",
+		Short:                 "Print policy JSON schema",
+		Args:                  cobra.NoArgs,
```
crazy-max, line 12: Should we hide this command if not yet implemented?

```suggestion
		Use:                   "json-schema",
		Short:                 "Print policy JSON schema",
		Args:                  cobra.NoArgs,
		Hidden:                true,
```

commands/policy/test.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func testCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "test <path>",
+		Short:                 "Run policy tests",
+		Args:                  cobra.ExactArgs(1),
```
crazy-max, line 12: Same here?

```suggestion
		Use:                   "test <path>",
		Short:                 "Run policy tests",
		Args:                  cobra.ExactArgs(1),
		Hidden:                true,
```

crazy-max (commented):
In the logs:

```
#1 loading policies 10-xx-external.Dockerfile.rego
#1 2.391 checking policy for source local://context
#1 2.401 policy decision for source local://context: ALLOW []
#1 2.472 checking policy for source docker-image://docker.io/library/alpine:latest@sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62
#1 2.472 checking policy for source docker-image://docker.io/tonistiigi/xx:1.8.0@sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 2.481 hack/utils.rego:4: compare sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6 add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 2.482 10-xx-external.Dockerfile.rego:12: tonistiigi/xx true
#1 2.482 hack/utils.rego:4: compare sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6 add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 2.482 policy decision for source docker-image://docker.io/tonistiigi/xx:1.8.0@sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6: ALLOW []
#1 2.482 10-xx-external.Dockerfile.rego:12: alpine false
#1 2.482 policy decision for source docker-image://docker.io/library/alpine:latest@sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62: ALLOW []
#1 ...
```

I guess `alpine false` means it's not evaluated or skipped. Hard to understand from the logs. Could we maybe have something like:

```
#1 2.482 10-xx-external.Dockerfile.rego:12: tonistiigi/xx evaluated
...
#1 2.482 10-xx-external.Dockerfile.rego:12: alpine skipped
```

commands/policy/eval.go
tonistiigi: No, I left it so that `--filename` points to Dockerfile, to be consistent with `build` command.

commands/policy/eval.go
tonistiigi: Arg is not dockerfile, it is the source to be evaluated.

commands/policy/json_schema.go
tonistiigi: I'll either implement or hide it for rc2.

commands/policy/test.go
tonistiigi: Test is already implemented in https://github.com/docker/buildx/pull/3592

commands/policy/eval.go
tonistiigi: You are mixing args here. Usage:

```
 # docker buildx policy eval --print .
{
  "local": {
    "name": "context"
  }
}
 # docker buildx policy eval --print docker-image://alpine:latest
INFO: unresolved fields: image.checksum, image.labels, image.user, image.volumes, image.workingDir, image.env, image.hasProvenance, image.signatures
{
  "image": {
    "ref": "docker.io/library/alpine:latest",
    "host": "docker.io",
    "repo": "alpine",
    "fullRepo": "docker.io/library/alpine",
    "tag": "latest",
    "platform": "linux/arm64",
    "os": "linux",
    "arch": "arm64"
  }
}
 # docker buildx policy eval docker-image://alpine:latest
 # docker buildx policy eval docker-image://alpine:3.16
ERROR: policy denied
```

tonistiigi:
@crazy-max 

> I guess alpine false means it's not evaluated or skipped. Hard to understand from the logs. Could we maybe have something like:

This is a custom debug line from your policy https://github.com/tonistiigi/buildx-rego-examples/blob/main/10-xx-external.Dockerfile.rego#L12 (look at the line number and the filename reported).

commands/policy/eval.go
crazy-max: Yes got confused so `args[0]` is a source being applied to a policy linked to a Dockerfile.

By default it will use `Dockerfile` and attached `Dockerfile.rego` if no filename specified.

If I don't have anything in my workdir:

```
$ ls
$ docker buildx policy eval docker-image://alpine:latest --print
docker buildx policy eval docker-image://alpine:latest --print
INFO: unresolved fields: image.checksum, image.labels, image.user, image.volumes, image.workingDir, image.env, image.hasProvenance, image.signatures
{
  "image": {
    "ref": "docker.io/library/alpine:latest",
    "host": "docker.io",
    "repo": "alpine",
    "fullRepo": "docker.io/library/alpine",
    "tag": "latest",
    "platform": "linux/amd64",
    "os": "linux",
    "arch": "amd64"
  }
}
```

Does it make sense to evaluate? Imo it should fail because it evaluates against nothing?

commands/policy/eval.go
```diff
@@ -0,0 +1,421 @@
...
+	if len(workers) == 0 {
+		return errors.New("no workers available in the builder")
+	}
+
+	defaultPlatform := workers[0].Platforms[0]
+	p := ocispecs.Platform{
+		Architecture: defaultPlatform.Architecture,
+		OS:           defaultPlatform.OS,
+		Variant:      defaultPlatform.Variant,
+	}
```
crazy-max, line 97: It could be handy to specify an arbitrary platform (or more than one?) through a flag like `--platform` otherwise default to worker one. WDYT?

commands/policy/eval.go
tonistiigi: You are using `--print` that doesn't evaluate but prints the source. Like `buildx bake --print` works even if you don't have a builder set up.

commands/policy/eval.go
tonistiigi: Yes, but I don't think critical as policy should almost always work with any platform. This is to make sure the local client platform doesn't become the default, as it could be non-Linux and unsupported by the image.

tonistiigi: Same for things like `--build-arg` etc. that you can match in policy, but make the command bit more confusing with to many extra flags.

build/opt.go
```diff
@@ -55,6 +62,79 @@ var sendGitQueryAsInput = sync.OnceValue(func() bool {
 	return false
 })
 
+type policyProgressLogger struct {
```
crazy-max, line 65: For multiplatform builds should we prefix with `[os/arch]` if there are more than one platform?

```
$ docker buildx build --platform linux/amd64,linux/arm64 -f 10-xx-external.Dockerfile .
...
#1 loading policies 10-xx-external.Dockerfile.rego
#1 0.481 checking policy for source local://context
#1 0.490 policy decision for source local://context: ALLOW []
#1 0.564 checking policy for source docker-image://docker.io/tonistiigi/xx:1.8.0@sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 0.564 checking policy for source docker-image://docker.io/library/alpine:latest@sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62
#1 0.564 checking policy for source docker-image://docker.io/tonistiigi/xx:1.8.0@sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 0.565 checking policy for source docker-image://docker.io/library/alpine:latest@sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62
#1 0.573 hack/utils.rego:4: compare sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6 add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 0.573 hack/utils.rego:4: compare sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6 add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 0.574 10-xx-external.Dockerfile.rego:12: alpine false
#1 0.574 policy decision for source docker-image://docker.io/library/alpine:latest@sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62: ALLOW []
#1 0.574 10-xx-external.Dockerfile.rego:12: alpine false
#1 0.574 policy decision for source docker-image://docker.io/library/alpine:latest@sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62: ALLOW []
#1 0.574 10-xx-external.Dockerfile.rego:12: tonistiigi/xx true
#1 0.574 hack/utils.rego:4: compare sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6 add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 0.574 policy decision for source docker-image://docker.io/tonistiigi/xx:1.8.0@sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6: ALLOW []
#1 0.574 10-xx-external.Dockerfile.rego:12: tonistiigi/xx true
#1 0.574 hack/utils.rego:4: compare sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6 add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6
#1 0.574 policy decision for source docker-image://docker.io/tonistiigi/xx:1.8.0@sha256:add602d55daca18914838a78221f6bbe4284114b452c86a48f96d59aeb00f5c6: ALLOW []
#1 ...
```

crazy-max (dismissed):

build/opt.go
crazy-max: As discussed I will put comments in the issue about policy progress follow-ups

crazy-max (approved):

PR #3546, opened 2025-11-27
driver/kubernetes: change DeploymentOpt.Replicas to int32

thaJeztah:
Directly convert to a int32 instead of later conversion so that we can return an error if the value is out of range.

- This should hopefully silence https://github.com/docker/buildx/security/code-scanning/1


thaJeztah:
cc @crazy-max this will probably silence the remaining code alert

PR #3551, opened 2025-12-04
build: split out node resolver

jedevc:
This PR simply splits out the node resolver into a separate sub-package.

I wanted to use the `resolveDrivers` function in some code that interfaces with `buildx` by importing it. It's useful to access the same logic for booting and platform matching that buildx itself uses (since some platform info is only available once the node is booted).

While the `builder` and `drivers` package are easily importable, the `build` package depends on the entire AWS SDK. Splitting the resolver out into it's own package makes the code more reusable, and also hopefully makes it easier to refactor long-term.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/containerd/containerd/v2/core/images"
 	"github.com/distribution/reference"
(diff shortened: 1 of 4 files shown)


crazy-max (approved):
LGTM thanks!

Issue #608, opened 2021-05-08
Add compatibility with `go install`


sodul:
I tried to get buildx through `go install` and even the `gobin` (now deprecated) utility but could not get it to work.

This is with go 1.16.3 on macOS Intel.

```
> go install github.com/docker/buildx/cmd/buildx@v0.5.1
go install github.com/docker/buildx/cmd/buildx@v0.5.1: github.com/docker/buildx@v0.5.1
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.

> gobin github.com/docker/buildx/cmd/buildx@v0.5.1
failed to run go list -find -json github.com/docker/buildx/cmd/buildx: exit status 1
go: github.com/moby/buildkit@v0.8.1-0.20201205083753-0af7b1b9c693 requires
	github.com/jaguilar/vt100@v0.0.0-20150826170717-2703a27b14ea (replaced by github.com/tonistiigi/vt100@v0.0.0-20190402012908-ad4c4a574305): missing go.sum entry; to add it:
	go mod download github.com/jaguilar/vt100
```

jsternberg:
This should be possible at the moment. We don't have any excludes or replaces in the `go.mod` file at the moment.

PR #3552, opened 2025-12-04
chore: remove unneeded excludes

jsternberg:
The upstream PR that used the pseudo-versions was merged so this no
longer causes an upgrade of those dependencies to pseudo-versions so the
exclude is no longer needed.

Relates to #608.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -178,14 +178,5 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
-exclude (
-	// FIXME(thaJeztah): remoove this once kubernetes updated their dependencies to no longer need this.
-	//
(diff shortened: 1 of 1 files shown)


thaJeztah:
Was hoping we could remove it from Buildkit as well, but looks like `github.com/containerd/accelerated-container-image` still forces it; I recall I started working on that;


```bash
go mod graph | grep ' github.com/davecgh/go-spew@v1.1.2'
github.com/moby/buildkit github.com/davecgh/go-spew@v1.1.2-0.20180830191138-d8f796af33cc
github.com/containerd/accelerated-container-image@v1.3.0 github.com/davecgh/go-spew@v1.1.2-0.20180830191138-d8f796af33cc
```

PR #3562, opened 2025-12-11
auth: add option to load docker config for specific repo/scope

tonistiigi:
Add the possibility to load an alternative Docker config (for registry auth) for a specific repository or scope. These configs can be put in `$DOCKER_CONFIG/buildx/config/registry[/repo/path][@scope[,..]]`.

This should be considered workaround and support for this will NOT be extended to `docker login`. Instead issue tracking support is in https://github.com/docker/cli/issues/6695 and buildx would be updated if that issue progresses. We will add an option to use it in https://github.com/docker/login-action for advanced rate-limiting control that is needed there.

### DIFF
--- a/util/dockerutil/dockerconfig/configprovider.go
+++ b/util/dockerutil/dockerconfig/configprovider.go
@@ -1,0 +1,254 @@
+package dockerconfig
+
+import (
+	"cmp"
+	"context"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/docker/buildx/util/confutil"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/config/configfile"
+	"github.com/docker/cli/cli/config/types"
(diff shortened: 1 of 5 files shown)


util/dockerutil/dockerconfig/configprovider.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+	}
+
+	entry, exists := ap.authConfigCache[key]
+	if exists && (cacheExpireCheck == nil || !cacheExpireCheck(entry.Created, key)) {
+		return *entry.Auth, nil
+	}
+
+	hostKey := host
+	if host == authprovider.DockerHubRegistryHost {
```
crazy-max, line 89: I think we should also support canonical domains `docker.io` and `index.docker.io`:
* https://github.com/docker/cli/blob/874b831c0e4906b0dd94f179c7c00acdb6c6114c/internal/registry/config.go#L47-L50
* https://github.com/docker/cli/blob/874b831c0e4906b0dd94f179c7c00acdb6c6114c/internal/registry/service_v2.go#L11

crazy-max (approved):

crazy-max (approved):

PR #3566, opened 2025-12-11
tests: add integration tests for `dap build`

jsternberg:
This adds integration tests for the `dap build` command to test various
behavior associated with the command. We start the build and the
integration test acts as a dap client to send requests and check that
the output is what we expect.

### DIFF
--- a/tests/dap_build.go
+++ b/tests/dap_build.go
@@ -1,0 +1,1002 @@
+package tests
+
+import (
+	"encoding/json"
+	"os"
+	"os/exec"
+	"path"
+	"runtime"
+	"slices"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/docker/buildx/commands"
+	debug "github.com/docker/buildx/dap"
+	"github.com/docker/buildx/dap/common"
+	"github.com/docker/buildx/util/daptest"
+	"github.com/google/go-dap"
+	"github.com/moby/buildkit/util/testutil/integration"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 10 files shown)


tests/dap_build.go
```diff
@@ -0,0 +1,724 @@
...
+		stackFrames := doStackTrace(t, client, stopped.Body.ThreadId)
+		assertStackTrace(t, stackFrames, exp)
+
+		doStepOut(t, client, stopped.Body.ThreadId)
+	}
+
+	require.NoError(t, done(false))
+}
+
+func testDapBuildExec(t *testing.T, sb integration.Sandbox) {
```
tonistiigi: We don't need to merge in the stubs

tests/dap_build.go
```diff
@@ -0,0 +1,724 @@
...
+	select {
+	case <-time.After(10 * time.Second):
+		require.Fail(t, "timeout reached")
+	case em := <-interruptCh:
+		e, _ = em.(E)
+	}
+	return e
+}
+
+type ExpectedStackFrame struct {
```
tonistiigi: nit: I guess this doesn't need to be public

jsternberg: Renamed to `stackFrameMatcher` and renamed some of the methods since `Expected` is kind of an awkward name.

tests/dap_build.go
jsternberg: Removed.

jsternberg:
@tonistiigi the tests only run on the experimental matrix builds because they don't presently work on the non-experimental branch because dap only works with experimental enabled.

jsternberg:
Including additional unit tests for checking the variable functionality.

PR #3568, opened 2025-12-12
dap: skip the set breakpoints test because of possible deadlock

jsternberg:
The unit test can potentially deadlock in some way that the go runtime
doesn't detect it and it runs for 30 minutes. Skipping the test until we
either fix it or replace it with an equivalent integration test.

### DIFF
--- a/dap/adapter_test.go
+++ b/dap/adapter_test.go
@@ -73,6 +73,8 @@ func TestLaunch(t *testing.T) {
 // TestSetBreakpoints will test sending a setBreakpoints request with no breakpoints.
 // The response should be an empty array instead of null in the JSON.
 func TestSetBreakpoints(t *testing.T) {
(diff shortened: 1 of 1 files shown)


rcjsuen:
Do we have an issue created to remind ourselves to review this?

jsternberg:
Not at the moment but I'll probably end up covering it in the integration tests.

jsternberg:
I found the potential race issue while working on the integration tests which is fixed in this PR: https://github.com/docker/buildx/pull/3566

When the test client makes a request, it registers the request sequence id with a map on the client so it can associate the response to the correct channel. It was adding this sequence id to the map after it sent the message rather than before so receiving the message very quickly could cause the response to be dropped and then code that was polling for the response would hang even though the response had been received.

PR #3583, opened 2026-01-09
build: handle at symbols in an oci-layout path

jsternberg:
The buildx command line will now handle `@` symbols in the
`oci-layout://` path when used with named contexts. Instead of
assuming the `@` symbol is part of the reference digest, it will first
check that it is a valid reference digest. Otherwise, it will assume
it's part of the file path.

Fixes moby/buildkit#6447.

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -558,12 +558,8 @@ func loadInputs(ctx context.Context, d *driver.DriverHandle, inp *Inputs, pw pro
 
 		// handle OCI layout
 		if localPath, ok := strings.CutPrefix(v.Path, "oci-layout://"); ok {
-			localPath, dig, hasDigest := strings.Cut(localPath, "@")
-			localPath, tag, hasTag := strings.Cut(localPath, ":")
-			if !hasTag {
-				tag = "latest"
-			}
-			if !hasDigest {
+			localPath, dig, tag := parseOCILayoutPath(localPath)
+			if dig == "" {
 				dig, err = resolveDigest(localPath, tag)
 				if err != nil {
 					return nil, errors.Wrapf(err, "oci-layout reference %q could not be resolved", v.Path)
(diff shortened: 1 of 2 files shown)


build/opt.go
```diff
@@ -1017,3 +1013,33 @@ func isActive(ce *client.CacheOptionsEntry) bool {
 	}
 	return ce.Attrs["token"] != "" && (ce.Attrs["url"] != "" || ce.Attrs["url_v2"] != "")
 }
+
+// parseOCILayoutPath handles the oci-layout url accepted by buildx.
+func parseOCILayoutPath(s string) (localPath, dgst, tag string) {
```
tonistiigi, line 1018: Add a couple of test cases in unit test to check this parsing.

build/opt.go
```diff
@@ -1017,3 +1013,33 @@ func isActive(ce *client.CacheOptionsEntry) bool {
...
+	// Look for the digest reference. There might be multiple @ symbols
+	// in the path and the @ symbol may be part of the path or part of
+	// the digest. If we find the @ symbol, verify that it's a valid
+	// digest reference instead of just assuming it is because it
+	// might be part of the file path.
+	if i := strings.LastIndex(s, "@"); i >= 0 {
+		after := s[i+1:]
+		if reference.DigestRegexp.MatchString(after) {
```
tonistiigi, line 1028: This is fine but could as well use `digest.Parse`

build/opt.go
jsternberg: Added tests for this.

build/opt.go
jsternberg: I think this method might be better since the same technique is used for checking the tag name and it doesn't require the algorithm to be registered with the binary. The `digest.Parse` method uses `algorithm.Available()` to validate it and I don't think that's necessary here since this code runs in `buildx` and not in `buildkit`.

Issue #3567, opened 2025-12-11
Rego source policy additional UX


tonistiigi:
This issue is to discuss UX changes for #3539 for additional control and helpers.

## Build

`build` and `bake` automatically load a default policy file based on the Dockerfile path(if one exists):

    Dockerfile        → Dockerfile.rego
    foo.Dockerfile    → foo.Dockerfile.rego

When building from a remote context, these files are also loaded from a remote context, like the Dockerfile.

Users may override or extend this behavior using:

    --policy [csv] [--policy [csv]]...

#### `filename=<path>`
Add a policy file.  Can be combined with `reset`.

#### `reset=<bool>`
If set, the currently loaded policy (eg. the default based on Dockerfile) is ignored and the policy set is replaced.  
After `reset=true`, only explicitly provided `filename=` entries are used.

#### `disabled=<bool>`
Disable policy evaluation entirely.  Using this together with other policy flags is an error.

#### `strict=<bool>`
Require policy enforcement.  
If `true`, builds fail when running against a BuildKit that does not support policies.  
Policy file load or compile errors always fail regardless of this setting.
Initially, this will default to `false` but may change somewhere in the future where BuildKit instances without policy support become a rarity.

#### `log-level=<level>`
Set policy logging verbosity (e.g., `debug`).

---

#### `ignore=<expr>`
Example:

    --policy ignore=image.repo=alpine

Potential mechanism for skipping specific sources when generating the policy input.  
Not sure if we need smth. like this.

---

## Policy Subcommand

These commands help writing policies and ensuring that they continue to work correctly after code changes.

### `buildx policy print-schema`
Print the JSON schema describing the policy input object.

### `buildx policy test <path>`
Execute a `_test.rego` file with fixed inputs, similar to `opa test`.  
Inputs are not fetched via BuildKit, but part of the test already.

### `buildx policy eval --filename=<name> [--print] [source]`
Evaluate the specified policy against the input for `source`.

- These are the same source values as can be specified in `--build-context`.  
  The source is resolved in BuildKit and then checked against the policy definition.

- If `--print` is set, the policy isn’t evaluated; the input object is printed instead.  
  Useful for exploring and validating values that policies are written against or for capturing input definitions for tests.

- If `source` is not specified, the command expects an input object JSON on stdin.

@crazy-max @dvdksn @AkihiroSuda @colinhemmings

crazy-max:
> `build` and `bake` automatically load a default policy file based on the Dockerfile path(if one exists):
[...]

PR #3593, opened 2026-01-14
policy: add policy to bake and improve test coverage

tonistiigi:
previously in #3590
part of #3567


crazy-max (approved):
LGTM

Was wondering if we could have a shortand when specifying the policy filename like:

```hcl
target "default" {
  policy = ["extra.rego"]
}
```

Similar to local exporter: https://github.com/docker/buildx/blob/5a5d59f50359dd66566e1aea447d3d465a39bd5b/util/buildflags/export.go#L107-L116

So we don't need to set `filename=`.

PR #3595, opened 2026-01-14
bake: allow disabling env lookup for bake

tonistiigi:
Allow bake to be run in a mode where the local environment variable values can't be captured by bake definition.

In the future, we will likely also add `--var k=v` for an alternative way to pass values to variables.

This is based on https://github.com/moby/buildkit/pull/6463/files#r2690936348

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -15,6 +15,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


jsternberg (approved):

crazy-max:
Needs docs follow-up https://docs.docker.com/build/building/variables/#build-tool-configuration-variables

PR #3610, opened 2026-01-16
bake: add --var flag for setting variable values

tonistiigi:
This provides an alternative way to set values for Bake
variables without adding them to the global environment variables.
This can also be used when environment variable access
is disabled with `BUILDX_BAKE_DISABLE_VARS_ENV_LOOKUP`.


docs/reference/buildx_bake.md
```diff
@@ -32,6 +32,7 @@ Build from a file
 | [`--push`](#push)                   | `bool`        |         | Shorthand for `--set=*.output=type=registry`. Conditional.                                                            |
 | [`--sbom`](#sbom)                   | `string`      |         | Shorthand for `--set=*.attest=type=sbom`                                                                              |
 | [`--set`](#set)                     | `stringArray` |         | Override target value (e.g., `targetpattern.key=value`)                                                               |
+| `--var`                             | `stringArray` |         | Set a variable value (e.g., `name=value`)                                                                             |
```
crazy-max, line 35: Could have basic examples similar to [set](https://github.com/docker/buildx/blob/master/docs/reference/buildx_bake.md#set)

PR #3612, opened 2026-01-16
dockerconfig: add fallback for loading DHI and Scout credentials

tonistiigi:
These are also Docker-managed registries with the same users.

Fall back to the main hostname if not logged in to the exact domain.

Note that this does not currently work with `imagetools` as the auth system there has not yet been combined with new config loader from https://github.com/docker/buildx/pull/3562 .

@crazy-max @cdupuis

### DIFF
--- a/util/dockerutil/dockerconfig/configprovider.go
+++ b/util/dockerutil/dockerconfig/configprovider.go
@@ -18,6 +18,12 @@ import (
 	"github.com/moby/buildkit/session/auth/authprovider"
 )
 
+const (
+	dockerHubRegistryHost   = "registry-1.docker.io"
+	dockerDHIRegistryHost   = "dhi.io"
+	dockerScoutRegistryHost = "registry.scout.docker.com"
+)
+
 func LoadAuthConfig(cli command.Cli) authprovider.AuthConfigProvider {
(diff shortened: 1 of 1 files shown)


jsternberg (approved):

util/dockerutil/dockerconfig/configprovider.go
```diff
@@ -18,6 +18,12 @@ import (
 	"github.com/moby/buildkit/session/auth/authprovider"
 )
 
+const (
+	dockerHubRegistryHost   = "registry-1.docker.io"
```
mathieu-benoit, line 22: Hi @tonistiigi @crazy-max @cdupuis, I just tried this new feature by using the associated released version `v0.31.0` in the `docker/setup-buildx-action`. Out of curiosity, why it's `registry-1.docker.io` and not `docker.io` here?
Context: the default value for the `registry` in `docker/login-action` is `docker.io`.

tonistiigi: That's the actual hostname for main hub registry. https://docs.docker.com/reference/api/registry/latest/#tag/pull `docker.io` is redirect to `docker.com` if you send requests directly to that.

PR #3614, opened 2026-01-16
commands: remove policy json-schema stub

tonistiigi:
Remove stub for now so we don't ship it in a release.


commands/policy/root.go
```diff
@@ -18,7 +18,7 @@ func RootCmd(rootcmd *cobra.Command, dockerCli command.Cli, rootOpts RootOptions
 	}
 
 	cmd.AddCommand(
-		jsonSchemaCmd(),
+		// TODO: jsonSchemaCmd(),
```
jsternberg: Should we just keep the template for this function in the other file since we reference the function in a comment here? Might be easier to just comment out the code in `json_schema.go` instead.

jsternberg (dismissed):
Fine as-is but I think we should just comment out the command code rather than remove it entirely since we reference it in a TODO.

commands/policy/root.go
tonistiigi: The linter will fail if there are unreferenced functions so a big block would need to be commented in. If you want I can modify the comment to just say "policy json-schema command" so it is not pointing to a function.

jsternberg (approved):

PR #3623, opened 2026-01-22
imagetools: avoid excessive HTTP requests on create

tonistiigi:
Two separate fixes to avoid bombarding registry with extra requests.

Potential follow-ups for the next feature release:
- Limit the amount of parallelization. Too many requests are sent in parallel that is not optimal.
- The attestation manifest is still accessed twice from the original repository if the same manifest that is in the index is also returned for the `referrers` command.

### DIFF
--- a/commands/imagetools/create.go
+++ b/commands/imagetools/create.go
@@ -200,25 +200,37 @@ func runCreate(ctx context.Context, dockerCli command.Cli, in createOptions, arg
 	eg, _ := errgroup.WithContext(ctx)
 	pw := progress.WithPrefix(printer, "internal", true)
 
+	tagsByRepo := map[string][]reference.Named{}
 	for _, t := range tags {
+		repo := t.Name()
+		tagsByRepo[repo] = append(tagsByRepo[repo], t)
+	}
+
+	for repo, repoTags := range tagsByRepo {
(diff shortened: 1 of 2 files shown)


tonistiigi:
Added one more optimization for the case where multiple destination tags are set.

commands/imagetools/create.go
```diff
@@ -200,24 +200,39 @@ func runCreate(ctx context.Context, dockerCli command.Cli, in createOptions, arg
...
+						return r.Copy(ctx, &imagetools.Source{
+							Ref:  desc.Source.Ref,
+							Desc: desc.Descriptor,
+						}, seed)
 					})
 				}
 				if err := eg2.Wait(); err != nil {
 					return err
 				}
 				ctx = withMediaTypeKeyPrefix(ctx) // because of containerd bug this needs to be called separately for each ctx/goroutine pair to avoid concurrent map write
```
thaJeztah: Off-topic, but is there a ticket related to this so that we can track if it's still needed?

crazy-max: https://github.com/docker/buildx/pull/3624

crazy-max:
Needs rebase otherwise LGTM

Issue #3368, opened 2025-08-14
Include build context structure when debugging with DAP


rcjsuen:
### Description

When using DAP, I can navigate the file system through the variables view of my DAP client. It would be nice if we could add a node to show the "build context". Lots of Dockerfiles use a simple `COPY . .` but people may not realize what's actually in the build context because of the `.dockerignore` file. Showing the build context when debugging could help demystify this.

rcjsuen:
This should work for a file system or when building with a tarball as a context. I know other types of build contexts are supported as well but not sure which other ones make sense.

jsternberg:
I think this currently works but I just tested it on the change included in https://github.com/docker/buildx/pull/3450 and it definitely works there. If you're on a `COPY` instruction and check the file explorer, you'll see two digests which correspond to the place you're copying from and the place you're copying to. You have to look inside right now to figure out which is which, but it is possible.

I'm thinking to add a bit of a user-facing change that makes the most common situation a bit easier which is identifying sources.

PR #3631, opened 2026-01-28
dap: improve file explorer source names

jsternberg:
Improves the naming for file explorer names when the input relates
directly to a source. This is most common when the input is the context
(which is just usually a simple source like `local://context`).

This should help in most circumstances in determining which input is
which.

Fixes https://github.com/docker/buildx/issues/3368.

### DIFF
--- a/dap/thread.go
+++ b/dap/thread.go
@@ -549,14 +549,25 @@ func (t *thread) solveInputs(ctx context.Context, target *step) (string, map[str
 }
 
 func (t *thread) determineInputName(op *pb.Op, index int, input *pb.Input) string {
-	switch op := op.Op.(type) {
-	case *pb.Op_Exec:
-		for _, m := range op.Exec.Mounts {
+	// Attempt to match the input to one of the mount destinations if we have
+	// an exec operation.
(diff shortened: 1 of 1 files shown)


dap/thread.go
```diff
@@ -549,14 +549,25 @@ func (t *thread) solveInputs(ctx context.Context, target *step) (string, map[str
...
 
 func (t *thread) determineInputName(op *pb.Op, index int, input *pb.Input) string {
-	switch op := op.Op.(type) {
-	case *pb.Op_Exec:
-		for _, m := range op.Exec.Mounts {
+	// Attempt to match the input to one of the mount destinations if we have
+	// an exec operation.
+	if exec, ok := op.Op.(*pb.Op_Exec); ok {
+		for _, m := range exec.Exec.Mounts {
 			if m.Input >= 0 && m.Input == int64(index) {
```
tonistiigi, line 556: Does this assume that root mount is always input `0`? I don't think there is a guarantee for that. Better to check `.Dest == "/"`

tonistiigi: Nvm, it is greater-equal

tonistiigi (approved):

PR #3633, opened 2026-01-28
commands: adds metrics associated with the debugger

jsternberg:
Add metrics associated with the debugger that are reported through the
metrics writer. This adds a few attributes that are only added when a
debugger is used with either the `debug` command or `dap` command.

At the moment, these metrics show up the exact same as a build and we
can't identify if something is using `dap` or `debug` since they use the
same code path.

This also adds a new available metric that can be utilized by plugins to
report additional information. The metrics will check if an environment
variable `BUILDX_DAP_USER_AGENT` is sent and that will get included in
the metrics if they are enabled.

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -239,10 +239,12 @@ const (
 	commandOptionsHash   = attribute.Key("command.options.hash")
 	driverNameAttribute  = attribute.Key("driver.name")
 	driverTypeAttribute  = attribute.Key("driver.type")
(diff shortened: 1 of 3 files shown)


jsternberg:
@chelnak sorry I was trying to type in @crazy-max and auto complete put your name first when I was pressing enter.

tonistiigi (approved):

Issue #2407, opened 2024-04-12
`imagetools create` should return the digest of the new image


Pesa:
### Description

I'm [using](https://github.com/named-data/actions/blob/2907fde08ddbfed9ca5284b517f888a029baeeeb/.github/workflows/docker-image.yml#L192) `docker buildx imagetools create` to create a multi-platform image from previously built single-platform images.

I'd like this command to somehow output the digest of the newly created image, so that subsequent steps and downstream jobs can "atomically" operate (run some checks, inspect, sign, etc.) on that same version of the image without relying on tags, which can be moved.

`docker/build-push-action` already has a `digest` output, so I guess I'm asking for a similar feature in `imagetools create`.

tonistiigi:
@crazy-max Should we add `--metadata-file` to `imagetools create` or do you have better ideas for UX. We can also do it for `--dry-run` that atm. prints the manifest bytes. Digest is just checksum of the same bytes.

crazy-max:
> @crazy-max Should we add `--metadata-file` to `imagetools create` or do you have better ideas for UX. We can also do it for `--dry-run` that atm. prints the manifest bytes. Digest is just checksum of the same bytes.

I think it would be enough to just print the digest imo.

thompson-shaun:
Let's move ahead with the suggestion from @tonistiigi

zanieb:
Just want to give a gentle poke here to say we need this feature over at Astral and am curious if anyone is working on it and / or if it's open for contributions?

tonistiigi:
I think it would be weird to just return digest when `--dry-run` returns the whole manifest bytes. Other suggestions? Or is everyone else ok with that inconsistency?

zanieb:
Is the "workaround" then to just hash the bytes yourself?

I think the alternative would be an opt-in JSON output format with the bytes and digest in a structure form?

tonistiigi:
> Is the "workaround" then to just hash the bytes yourself?

Atm. I think `--dry-run` shows the manifest bytes. Without `--dry-run` there is no stdout output.

Pesa:
> I think it would be weird to just return digest when `--dry-run` returns the whole manifest bytes. Other suggestions? Or is everyone else ok with that inconsistency?

[...]

PR #3638, opened 2026-02-03
imagetools: metadata-file flag

crazy-max:
In the [Docker GitHub Builder ](https://github.com/docker/github-builder) workflows we are merging manifests with the `imagetools create` command. This command doesn't output the pushed digest while `build` or `bake` commands do it.

This PR add a `metadata-file` flag similar to `build` and `bake` commands.

This will be necessary in our github builder so we can have an output for this digest that can then be used in subsequent jobs.

### DIFF
--- a/commands/imagetools/create.go
+++ b/commands/imagetools/create.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/containerd/containerd/v2/core/remotes"
@@ -16,7 +17,9 @@ import (
 	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/cli/cli/command"
+	"github.com/moby/buildkit/exporter/containerimage/exptypes"
 	"github.com/moby/buildkit/util/progress/progressui"
+	"github.com/moby/sys/atomicwriter"
 	"github.com/opencontainers/go-digest"
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -34,6 +37,7 @@ type createOptions struct {
 	progress     string
 	preferIndex  bool
 	platforms    []string
+	metadataFile string
 }
 
 func runCreate(ctx context.Context, dockerCli command.Cli, in createOptions, args []string) error {
@@ -76,10 +80,15 @@ func runCreate(ctx context.Context, dockerCli command.Cli, in createOptions, arg
 	}
 
 	repos := map[string]struct{}{}
-
 	for _, t := range tags {
 		repos[t.Name()] = struct{}{}
 	}
+
+	repoNames := make([]string, 0, len(repos))
+	for repo := range repos {
+		repoNames = append(repoNames, repo)
+	}
+	sort.Strings(repoNames)
 
 	sourceRefs := false
 	for _, s := range srcs {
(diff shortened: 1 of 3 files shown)


thaJeztah:
Oh! Tests are failing though, and could be related?

```
=== Failed
=== FAIL: tests TestIntegration/TestImagetoolsCopyIndex/worker=docker-container (1.64s)
    imagetools.go:148: 
        	Error Trace:	/src/tests/imagetools.go:148
        	            				/src/vendor/github.com/moby/buildkit/util/testutil/integration/run.go:103
        	            				/src/vendor/github.com/moby/buildkit/util/testutil/integration/run.go:251
        	            	invalid checksum digest length
        	Test:       	TestIntegration/TestImagetoolsCopyIndex/worker=docker-container
    --- FAIL: TestIntegration/TestImagetoolsCopyIndex/worker=docker-container (1.64s)

=== FAIL: tests TestIntegration/TestImagetoolsCopyManifest/worker=docker-container (1.87s)
    imagetools.go:63: 
        	Error Trace:	/src/tests/imagetools.go:63
        	            				/src/vendor/github.com/moby/buildkit/util/testutil/integration/run.go:103
        	            				/src/vendor/github.com/moby/buildkit/util/testutil/integration/run.go:251
        	            	invalid checksum digest length
        	Test:       	TestIntegration/TestImagetoolsCopyManifest/worker=docker-container
    --- FAIL: TestIntegration/TestImagetoolsCopyManifest/worker=docker-container (1.87s)
```

tests/imagetools.go
```diff
@@ -80,8 +91,16 @@ func testImagetoolsCopyManifest(t *testing.T, sb integration.Sandbox) {
...
+	stdout = bytes.Buffer{}
+	stderr = bytes.Buffer{}
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	require.NoError(t, cmd.Run(), stdout.String(), stderr.String())
+	outputDigest = strings.TrimSpace(stdout.String())
+	require.NotEmpty(t, outputDigest)
+	t.Logf("digest: %s", outputDigest)
+	_, err = digest.Parse(outputDigest)
+	require.NoError(t, err)
```
thaJeztah: Was the `t.Logf` for debugging? (I always try to avoid logs in tests, and instead only log on failure; e.g.;



```suggestion
	_, err = digest.Parse(outputDigest)
	require.NoError(t, err, "digest: %s", outputDigest)
```

tonistiigi:
I think I prefer `--metadata-file` so that we can extend it if more fields are needed in the future.

crazy-max:
> I think I prefer `--metadata-file` so that we can extend it if more fields are needed in the future.

Should be good now

commands/imagetools/create.go
```diff
@@ -241,6 +244,14 @@ func runCreate(ctx context.Context, dockerCli command.Cli, in createOptions, arg
 		err = err1
 	}
 
+	if err == nil && len(in.metadataFile) > 0 {
+		if err := writeMetadataFile(in.metadataFile, map[string]any{
+			exptypes.ExporterImageDigestKey: desc.Digest.String(),
```
tonistiigi: Maybe already better to just add `ExporterImageDescriptorKey` ?

crazy-max: Added both and image names as well.

tonistiigi: Do we need the `ExporterImageDigestKey`? The value is already in the descriptor.

bo0tzz:
Does this address #2407?

crazy-max:
> Does this address #2407?

Yes good catch we can close this one.

PR #3643, opened 2026-02-04
Dockerfile: update to docker v29.2

thaJeztah:


crazy-max:
```
Error: buildx bake failed with: ERROR: failed to solve: docker.io/moby/moby-bin:29.2: not found
```

Seems this tag does not exist :thinking:

thaJeztah:
Hm... indeed; looks like the last tag is 29.2.0-rc.1 ? https://hub.docker.com/r/moby/moby-bin/tags?name=29.2

/cc @vvoland

thaJeztah:
Also weird, but perhaps I don't understand the UX; is it correct that I cannot filter actions for `tag` events? https://github.com/moby/moby/actions

Screenshot 2026-02-04 at 12 58 47

PR #3648, opened 2026-02-10
bake: derive git auth host from remote URL

crazy-max:
This PR refactors Bake Git authentication secret handling by introducing a dedicated gitauth helper that centralizes how secrets are built from environment variables. The same helper is now used in both build option creation and remote bake file reads, so the behavior is consistent across those paths while preserving existing support for `BUILDX_BAKE_GIT_AUTH_TOKEN` and `BUILDX_BAKE_GIT_AUTH_HEADER`.

It also adds automatic host-based Git auth secret derivation for remote Bake invocations. When a remote URL is in play, Bake now emits both base BuildKit secret IDs and host-scoped IDs (for example `GIT_AUTH_TOKEN.<host>` / `GIT_AUTH_HEADER.<host>`) based on the resolved remote URL logic, enabling per-host auth routing automatically without introducing host-suffixed auth env vars.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -1330,16 +1330,6 @@ func updateContext(t *build.Inputs, inp *Input) {
 	t.ContextPath = inp.URL
 }
 
-func isRemoteContext(t build.Inputs, inp *Input) bool {
-	if urlutil.IsRemoteURL(t.ContextPath) {
-		return true
-	}
-	if inp != nil && urlutil.IsRemoteURL(inp.URL) && !strings.HasPrefix(t.ContextPath, "cwd://") {
-		return true
-	}
-	return false
-}
-
 func collectLocalPaths(t build.Inputs) []string {
 	var out []string
 	if t.ContextState == nil {
@@ -1509,19 +1499,8 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
 	bo.Platforms = platforms
 
 	secrets := t.Secrets
-	if isRemoteContext(bi, inp) {
-		if _, ok := os.LookupEnv("BUILDX_BAKE_GIT_AUTH_TOKEN"); ok {
-			secrets = append(secrets, &buildflags.Secret{
-				ID:  llb.GitAuthTokenKey,
-				Env: "BUILDX_BAKE_GIT_AUTH_TOKEN",
-			})
-		}
-		if _, ok := os.LookupEnv("BUILDX_BAKE_GIT_AUTH_HEADER"); ok {
-			secrets = append(secrets, &buildflags.Secret{
-				ID:  llb.GitAuthHeaderKey,
-				Env: "BUILDX_BAKE_GIT_AUTH_HEADER",
-			})
-		}
+	if inp != nil && shouldAttachGitAuthSecrets(inp.URL, bi.ContextPath) {
+		secrets = append(secrets, gitAuthSecretsFromEnv(inp.URL)...)
 	}
 	bo.SecretSpecs = secrets.Normalize()
 	secretAttachment, err := build.CreateSecrets(bo.SecretSpecs)
--- a/bake/gitauth.go
+++ b/bake/gitauth.go
@@ -1,0 +1,78 @@
+package bake
+
+import (
+	"os"
+	"strings"
+
+	"github.com/docker/buildx/util/buildflags"
+	"github.com/docker/buildx/util/urlutil"
+	"github.com/moby/buildkit/client/llb"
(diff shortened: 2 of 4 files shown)


crazy-max:
~needs docs follow-up: https://docs.docker.com/build/building/variables/#buildx_bake_git_auth_header~

https://github.com/docker/docs/pull/24119

bake/bake.go
```diff
@@ -1330,14 +1330,14 @@ func updateContext(t *build.Inputs, inp *Input) {
 	t.ContextPath = inp.URL
 }
 
-func isRemoteContext(t build.Inputs, inp *Input) bool {
+func remoteContextURL(t build.Inputs, inp *Input) string {
```
tonistiigi: nit: bit cleaner to return `string, bool` from such functions.

bake/gitauth.go
```diff
@@ -0,0 +1,80 @@
...
+	return secrets
+}
+
+func gitAuthSecretsForEnv(secretIDPrefix, envPrefix string, environ []string, hosts []string) buildflags.Secrets {
+	envKey, ok := findGitAuthEnvKey(envPrefix, environ)
+	if !ok {
+		return nil
+	}
+	secrets := make(buildflags.Secrets, 0, len(hosts)+1)
+	secrets = append(secrets, &buildflags.Secret{
```
tonistiigi: I still don't get it.

(assuming this isn't breaking existing users badly), this should be only set if the Bake command is using remote input, not based on if the target is using remote URL or not what seems to happen atm. 

Additionally, there is no point in adding the main secret key and host key with the same value. Only host key should be set, and only for the host that was "bake remote URL", not any host that happened to be remote URL for a bake target.

crazy-max: > there is no point in adding the main secret key and host key with the same value.

Ah right, this should only emit host-scoped git auth secrets (no generic main key).

> (assuming this isn't breaking existing users badly), this should be only set if the Bake command is using remote input, not based on if the target is using remote URL or not what seems to happen atm.

I kept target remote-context handling for backward compatibility for now. I'm a bit hesitant to switch to strict remote input only immediately, because the current behavior follows context resolution in https://github.com/docker/buildx/blob/268f1c7e4ac3268de958fe8cfb6b028aca3b30e2/bake/bake.go#L1294-L1331

And changing that could break existing flows. If you prefer, I can still change it to strict remote input only behavior.

bake/bake.go
```diff
@@ -1509,19 +1509,8 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
...
-				Env: "BUILDX_BAKE_GIT_AUTH_TOKEN",
-			})
-		}
-		if _, ok := os.LookupEnv("BUILDX_BAKE_GIT_AUTH_HEADER"); ok {
-			secrets = append(secrets, &buildflags.Secret{
-				ID:  llb.GitAuthHeaderKey,
-				Env: "BUILDX_BAKE_GIT_AUTH_HEADER",
-			})
-		}
+	if remoteURL, ok := remoteContextURL(bi, inp); ok {
```
tonistiigi: The change in `remote.go` seems right, but this one still seems to use `target.ContextPath` as ther remote URL, what may not be correct. A target context URL and the bake remote definition URL may not be the same and these env secrets should be based on the remote definition URL only iiuc.

crazy-max: > but this one still seems to use `target.ContextPath` as ther remote URL, what may not be correct

Ok I was following the context resolution in https://github.com/docker/buildx/blob/268f1c7e4ac3268de958fe8cfb6b028aca3b30e2/bake/bake.go#L1294-L1331 where ContextPath is set to `inp.URL` but not needed. I'm changing this.

bake/bake.go
```diff
@@ -1509,19 +1499,8 @@ func toBuildOpt(t *Target, inp *Input) (*build.Options, error) {
...
-				Env: "BUILDX_BAKE_GIT_AUTH_TOKEN",
-			})
-		}
-		if _, ok := os.LookupEnv("BUILDX_BAKE_GIT_AUTH_HEADER"); ok {
-			secrets = append(secrets, &buildflags.Secret{
-				ID:  llb.GitAuthHeaderKey,
-				Env: "BUILDX_BAKE_GIT_AUTH_HEADER",
-			})
-		}
+	if inp != nil && urlutil.IsRemoteURL(inp.URL) && !strings.HasPrefix(bi.ContextPath, "cwd://") {
```
tonistiigi: This should do some kind of 1) both are URLs 2) both URLs use same host , check instead of just `!cwd://`

Issue #3639, opened 2026-02-03
Policy: add provenance fields


tonistiigi:
Currently policy only supports `hasProvenance` bool and the signature fields. Should also expose meaningful fields from the provenance attestation directly.

The initial use case would be to check that the build was hermetic and had all the required materials.

This can also be used for extra conditions on the individual materials. E.g., conditions on the Git repo that was the source for the build. In the future, we could even do some kind of recursive verification, so verify the artifact and also verify the materials against the policy.

This requires buildkit update to expose provenance via metadata query.

PR #3652, opened 2026-02-12
policy: add image.provenance input type

tonistiigi:
part one of #3639

### DIFF
--- a/policy/types.go
+++ b/policy/types.go
@@ -128,7 +128,39 @@ type Image struct {
 	WorkingDir  string            `json:"workingDir,omitempty"`
 
 	HasProvenance bool                   `json:"hasProvenance,omitempty"`
+	Provenance    *ImageProvenance       `json:"provenance,omitempty"`
 	Signatures    []AttestationSignature `json:"signatures,omitempty"`
+}
+
+type ImageProvenance struct {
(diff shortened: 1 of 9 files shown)


policy/types.go
```diff
@@ -128,9 +128,41 @@ type Image struct {
 	WorkingDir  string            `json:"workingDir,omitempty"`
 
 	HasProvenance bool                   `json:"hasProvenance,omitempty"`
+	Provenance    *ImageProvenance       `json:"provenance,omitempty"`
 	Signatures    []AttestationSignature `json:"signatures,omitempty"`
 }
 
+type ImageProvenance struct {
+	PredicateType string `json:"predicateType,omitempty"`
```
tonistiigi, line 136: This is intentionally missing materials atm.

I wanted to make every material type `Input` so that policy conditions can run in any level of source material in https://github.com/docker/buildx/commit/c2164fb8d86e6434a554f2d0884163502365c6ab .

But I missed that it is not allowed to return different source from a policy callback. So the current patch only works in `policy eval` and needs a complete redesign.

crazy-max: Ah yes policy callback can't return a different source. Looks good for follow-up.

crazy-max (approved):

PR #3656, opened 2026-02-19
policy: mark policy progress vertex on DENY build failures

### LINKED ISSUE #3644: Policy: associate policy error with the build step progress item
Currently, the policy error just shows up at the end of the build, and none of the build steps are marked as failing. This can be confusing and also hides the last logs for policy evaluation (or makes them appear in the middle of plain progress).

Not really how to implement this yet, as policy evaluation failing doesn't necessarily mean build will fail as well, and build error itself comes from a different codepath than the policy evaluation.

tonistiigi:
fix #3644

Track denied source identifiers during policy evaluation and flag the policy progress vertex as failed when BuildKit returns a matching DENY error pattern.

This improves the progress output of policy error and shows last policy logs with the build error.

### DIFF
--- a/policy/policy_error_test.go
+++ b/policy/policy_error_test.go
@@ -1,0 +1,56 @@
+package policy
+
+import (
+	"errors"
+	"testing"
+
+	gwpb "github.com/moby/buildkit/frontend/gateway/pb"
+	solverpb "github.com/moby/buildkit/solver/pb"
+	"github.com/moby/buildkit/sourcepolicy/policysession"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPolicyIsPolicyErrorMatchesRecordedSource(t *testing.T) {
+	p := NewPolicy(Opt{})
+	req := &policysession.CheckPolicyRequest{
+		Source: &gwpb.ResolveSourceMetaResponse{
+			Source: &solverpb.SourceOp{
+				Identifier: "docker-image://busybox:latest",
+			},
+		},
+	}
+	p.recordDenyIdentifier(req)
+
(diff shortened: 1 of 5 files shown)


policy/policy_error_test.go
```diff
@@ -0,0 +1,56 @@
...
+			Source: &solverpb.SourceOp{
+				Identifier: "docker-image://busybox:latest",
+			},
+		},
+	}
+	p.recordDenyIdentifier(req)
+
+	err := errors.New("failed to solve: error evaluating the source policy: source \"docker-image://alpine:latest\" not allowed by policy: action DENY")
+	require.False(t, p.IsPolicyError(err))
+}
```
crazy-max: nit: these tests look like a good fit for a single table-driven test (`[]struct{name, err, want}`) since setup is identical across cases. And could maybe also moved to `validate_test.go` and named `func TestPolicyIsPolicyError`?

crazy-max:
Could we have an integration test around the `release(err)/policyLogger.Close(inErr)` flow? Similar to https://github.com/docker/buildx/blob/a0103f4278354ae00a2325e26e07f87b1674d1a2/tests/build.go#L645 just to test progress.

tonistiigi:
Added progress integration test and refactored unit test.

crazy-max:
Seems there is a deadlock in sourcemeta resolver, opened #3659

tonistiigi:
Fix for that deadlock is in #3657. If you want, I can pull it out, but #3657 has some other changes to `sourcemeta` as well that would cause rebase conflicts.

crazy-max (approved):

PR #3657, opened 2026-02-20
policy: add artifact attestation and github attestation builtin support 

### LINKED ISSUE #3640: Policy: add artifact attestation verification and github_attestation helper
Allow signed artifacts used via `input.http` to also be verified similar to bundles.

Proposed helpers:

`artifact_attestation(input.http, filename)` - Verifies artifact against signed sigstore bundle file. Returns a struct with signature info fields.

`github_attestation(input.http, repo_name)` - Similar but instead of needing to pass file directly it is pulled by digest with Github API. The signed attestation needs to be uploaded to Github repo.

The sigstore bundle needs to have a valid Fulcio signature, but no specific signer identity is needed for these helpers. Identity can be checked based on the helper return value.

I think we can also extend the current `docker_github_builder`, `docker_github_builder_tag` to directly accept `input.http` and perform `github_attestation` with extra conditions internally. Maybe new `docker_github_builder_bundle(input.http, filename)` to allow file to be passed in directly as well.

All the places that accept filenames could also accept URLs.

tonistiigi:
fixes #3640
~depends on #3656~

### DIFF
--- a/policy/funcs.go
+++ b/policy/funcs.go
@@ -1,12 +1,20 @@
 package policy
 
 import (
+	"bytes"
+	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"maps"
+	"net/url"
 	"strings"
 
 	"github.com/distribution/reference"
+	"github.com/golang/snappy"
+	slsa1 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1"
+	"github.com/moby/buildkit/client/llb"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/solver/pb"
 	"github.com/moby/buildkit/util/gitutil/gitsign"
 	"github.com/open-policy-agent/opa/v1/ast"
@@ -14,6 +22,7 @@ import (
 	"github.com/open-policy-agent/opa/v1/types"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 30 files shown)


policy/validate.go
```diff
@@ -288,6 +288,22 @@ func (p *Policy) CheckPolicy(ctx context.Context, req *policysession.CheckPolicy
 	if err != nil {
 		return nil, nil, err
 	}
+
+	rtUnk := runtimeUnknownInputRefs(st)
```
tonistiigi, line 294: Looks like this could be bug previously if image signatures were the only requirement with no other unknown fields. Similar code is bit above, but it currently only ran in case of the `Partial()` code path.

policy/funcs.go
```diff
@@ -99,6 +109,216 @@ func (p *Policy) initBuiltinFuncs() {
...
+		if err != nil {
+			p.log(logrus.InfoLevel, "%s: failed reading bundle_url %s: %v", funcGithubAttestation, bu, err)
+			continue
+		}
+		bundleRaw = bytes.TrimSpace(bundleRaw)
+		if len(bundleRaw) == 0 || bytes.Equal(bundleRaw, []byte("null")) {
+			continue
+		}
+		if shouldDecodeSnappyBundleURL(bu) {
+			decoded, err := snappy.Decode(nil, bundleRaw)
```
tonistiigi, line 262: fwiw I didn't find any documentation for this. When I wrote similar code couple of months ago in https://github.com/moby/policy-helpers/blob/main/githubapi/pull.go there was no `bundle_url` or snappy compression.

tonistiigi:
Rebased after #3656 merge.

policy/funcs.go
crazy-max: Yes seems gh cli is doing the same: https://github.com/cli/cli/blob/6c3e39ffc481d5b9675451a201ae904b6d42c7cd/pkg/cmd/attestation/api/client.go#L222-L265

Looks like they also have some backoff logic, not sure if we should do the same.

crazy-max (approved):

PR #3662, opened 2026-02-23
policy: add recursive provenance material resolution

### LINKED ISSUE #3639: Policy: add provenance fields

tonistiigi:
closes #3639


policy/materials.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func parseSLSAMaterial(m slsa1.ResourceDescriptor) (*pb.SourceOp, *ocispecs.Platform, error) {
+	uri := m.URI
+	dgst := m.Digest
+	if strings.HasPrefix(uri, "pkg:docker/") {
+		return dockerMaterialSource(uri, dgst)
+	}
+
+	if gu, err := gitutil.ParseURL(uri); err == nil {
```
crazy-max, line 44: Was looking at our gitutil and it only accepts `http|https|ssh|git` so common SLSA-style `git+https://...` URIs fail parse and then fail HTTP detection, so they are treated as unsupported and never resolved: https://slsa.dev/spec/v1.2/build-provenance#builddefinition

```json
"externalParameters": {
    "repository": "https://github.com/octocat/hello-world",
    "ref": "refs/heads/main"
},
"resolvedDependencies": [{
    "uri": "git+https://github.com/octocat/hello-world@refs/heads/main",
    "digest": {"gitCommit": "7fd1a60b01f91b314f59955a4e4d4e80d8edf11d"}
}]
```

policy/provenance.go
```diff
@@ -146,6 +155,48 @@ func parseSLSA02Provenance(dt []byte) (*ImageProvenance, error) {
...
+		rd := slsa1.ResourceDescriptor{
+			URI:    m.URI,
+			Digest: maps.Clone(m.Digest),
+		}
+		if _, _, err := parseSLSAMaterial(rd); err != nil {
+			if logf != nil {
+				logf(logrus.WarnLevel, fmt.Sprintf("skipping unsupported provenance material %q: %v", m.URI, err))
+			}
+			continue
+		}
```
crazy-max, line 173: iiuc malformed material entries are silently filtered out and that includes digest mismatch errors, so invalid/tampered materials are omitted from `input.image.provenance.materials` instead of causing evaluation failure. Could that potentially hide the material records policy is supposed to inspect?

commands/policy/eval.go
```diff
@@ -111,60 +110,66 @@ func runEval(ctx context.Context, dockerCli command.Cli, source string, opts eva
...
-			for _, f := range opts.fields {
-				if slices.Contains(trimmedUnknowns, f) {
-					toReload = append(toReload, f)
-				} else if !doneInvalidCheck {
-					invalidFields = append(invalidFields, f)
-				}
-			}
-			doneInvalidCheck = true
+			toReload, invalid := selectReloadFields(opts.fields, trimmedUnknowns)
+			invalidFields = invalid
```
crazy-max, line 140: `invalidFields` is overwritten on each loop iteration so once a field is resolved and disappears from `unknowns`, it gets reclassified as invalid and may be warned incorrectly?

policy/materials.go
tonistiigi: I don't think buildkit can produce `git+https://` atm and there isn't specific git definition for purl. Just https URLs are used, and `gitutil` figures out if the URL is pointing to a repo.

policy/provenance.go
tonistiigi: This should be preferred as we don't want images to fail automatically when new material types are added. Instead atm it will fail based on policy rules. So if the user has a policy rule expecting a specific material, but it fails to parse in here, then that policy rule will fail the build.

crazy-max:
Lint issue with slsa1: https://github.com/docker/buildx/actions/runs/22424348729/job/64929097618?pr=3662#step:3:431

Needs similar golangci-lint rule: https://github.com/moby/buildkit/blob/373d0adc5c452f65d2e42300a47f328b0a55c40b/.golangci.yml#L127-L129

tonistiigi:
I did some more testing and unfortunately this does not behave quite as well as I would have hoped. The properties of materials that are not initially present are only loaded if the material is accessed via a static index. If it is accessed dynamically, eg.

```
every m in input.image.provenance.materials
  check(m)
```

or

```
every m in input.image.provenance.materials
  allow with m as input
```

Then the rego library can not figure out what fields of the materials are actually required to be loaded.

So while this is working (and maybe more useful in the `policy eval` atm) only subset of policy rules can be used, and actual dynamic recusiveness is hard to achieve atm. I'm not sure if it would be possible to extend rego library or alternatively some of this logic could instead be exposed as a custom helper in the future.

I added some new examples to https://github.com/tonistiigi/buildx-rego-examples

crazy-max (approved):
Ok iiuc recursive material loading is currently reliable only for statically indexed material access, while dynamic iteration (`every m in input.image.provenance.materials`) doesn't give enough hints for lazy resolution.

As follow-up I agree that we should either extend unknown-field discovery in the Rego lib or add a dedicated helper that pre-resolves/iterates material inputs in a policy-friendly way.

PR #3674, opened 2026-02-27
build: ref-count policy FS lifecycle

tonistiigi:
Make memoized policy FS ref-counted so repeated get()/close() pairs don't prematurely close shared handles.

Current implementation could fail when reading same file twice.


build/policy_loader.go
```diff
@@ -228,28 +228,48 @@ func normalizeLocalPolicyPath(name, contextDir string) string {
 
 type memoizedPolicyFS struct {
```
crazy-max, line 229: noticed this small typo:

```suggestion
type memorizedPolicyFS struct {
```

crazy-max (approved):

PR #3685, opened 2026-03-03
vendor: github.com/golang/snappy v1.0.0

thaJeztah:
- relates to https://github.com/docker/buildx/pull/3657

Ensure arm64 frame sizes are 8 (mod 16)

fixes: "Weird failure when building on Raspbian / Debian 10.11"

full diff: https://github.com/golang/snappy/compare/v0.0.4...v1.0.0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/docker/docker v28.5.2+incompatible
 	github.com/docker/go-units v0.5.0
 	github.com/gofrs/flock v0.13.0
-	github.com/golang/snappy v0.0.4
+	github.com/golang/snappy v1.0.0
 	github.com/google/go-dap v0.12.1-0.20250904181021-d7a2259b058b
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/google/uuid v1.6.0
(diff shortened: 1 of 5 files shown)


thaJeztah:
FWIW; this version matches what GitHub uses; https://github.com/cli/cli/blob/7a590331b599a0ac60e55225dc26c79c8b0d51a0/go.mod#L26


I noticed that this module has been archived though; wondering if GitHub (if they define this format to use) is considering alternatives 🤔 

Screenshot 2026-03-03 at 16 32 13

thaJeztah:
Ah! Looks like `klauspost/compress` has alternatives; worth considering perhaps?

https://pkg.go.dev/github.com/klauspost/compress/snappy

> This is a drop-in replacement for github.com/golang/snappy.


And... apparently `s2` may be an alternative as well;

https://pkg.go.dev/github.com/klauspost/compress/s2#section-readme

> S2 is an extension of [Snappy](https://github.com/google/snappy).
> 
> S2 is aimed for high throughput, which is why it features concurrent compression for bigger payloads.
> 
> Decoding is compatible with Snappy compressed content, but content compressed with S2 cannot be decompressed by Snappy. This means that S2 can seamlessly replace Snappy without converting compressed content.

thaJeztah:
Oh! And looks like snappy there is a wrapper around s2, so no need to choose really;
https://github.com/klauspost/compress/blob/v1.18.4/snappy/decode.go#L39-L48


```go
// Decode returns the decoded form of src. The returned slice may be a sub-
// slice of dst if dst was large enough to hold the entire decoded block.
// Otherwise, a newly allocated slice will be returned.
//
// The dst and src must not overlap. It is valid to pass a nil dst.
//
// Decode handles the Snappy block format, not the Snappy stream format.
func Decode(dst, src []byte) ([]byte, error) {
	return s2.Decode(dst, src)
}
```

thaJeztah:
- opened https://github.com/docker/buildx/pull/3686 as alternative

PR #3691, opened 2026-03-04
dap: fix skipped breakpoint when the breakpoint and the entrypoint were the same

jsternberg:
We erroneously skipped a breakpoint when that breakpoint was the same as
the entrypoint and we did not use stop on entry. This is because we only
started evaluating breakpoints after the first step on the entrypoint
instead of at the entrypoint.

### DIFF
--- a/tests/dap_build.go
+++ b/tests/dap_build.go
@@ -78,6 +78,7 @@ var dapBuildTests = []func(t *testing.T, sb integration.Sandbox){
 	testDapBuild,
 	testDapBuildStopOnEntry,
 	testDapBuildSetBreakpoints,
+	testDapBuildEntryBreakpoint,
 	testDapBuildVerifiedBreakpoints,
 	testDapBuildLoadedSource,
 	testDapBuildStepIn,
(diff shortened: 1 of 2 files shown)


tests/dap_build.go
```diff
@@ -200,6 +201,37 @@ func testDapBuildSetBreakpoints(t *testing.T, sb integration.Sandbox) {
...
+		ContextPath: dir,
+	},
+		dap.SourceBreakpoint{Line: 7},
+	)
+
+	stopped := waitForInterrupt[*dap.StoppedEvent](t, interruptCh)
+	threads := doThreads(t, client)
+	require.ElementsMatch(t, []int{stopped.Body.ThreadId}, threads)
+
+	stackTraceResp := <-daptest.DoRequest[*dap.StackTraceResponse](t, client, &dap.StackTraceRequest{
```
tonistiigi: Seems like a place for `doStackTrace` helper

tests/dap_build.go
```diff
@@ -200,6 +201,37 @@ func testDapBuildSetBreakpoints(t *testing.T, sb integration.Sandbox) {
...
+	interruptCh := pollInterruptEvents(client)
+	doLaunch(t, client, commands.LaunchConfig{
+		Dockerfile:  path.Join(dir, "Dockerfile"),
+		ContextPath: dir,
+	},
+		dap.SourceBreakpoint{Line: 7},
+	)
+
+	stopped := waitForInterrupt[*dap.StoppedEvent](t, interruptCh)
+	threads := doThreads(t, client)
```
tonistiigi, line 225: Verify that stop happened for breakpoint reason?

tests/dap_build.go
jsternberg: Done. I also updated the area of code I copied this from. I think the original code was made before I added the helper and I copied the older code rather than the new one.

tests/dap_build.go
jsternberg: Added a check for the reason why it stopped to both this and stop on entry which are very similar tests that do mostly the same thing but for different reasons.

Issue #3636, opened 2026-01-30
Non-deterministic precedence between `--build-arg no_proxy=...` and `~/.docker/config.json` proxy defaults


Wenzel:
### Description

I’m seeing **non-deterministic behavior** in `docker build` where an explicit `--build-arg no_proxy=...` sometimes gets overridden by the proxy default from Docker client `config.json` (`proxies.default.noProxy`).

Repeating the same build command produces different results across runs.

### Reproduce

~~~Dockerfile
# allow to override Docker client proxies if necessary
ARG http_proxy
ARG https_proxy
ARG no_proxy

FROM ubuntu:24.04
# redeclare ARGs
ARG http_proxy
ARG https_proxy
ARG no_proxy

RUN <<EOF
env | grep -i proxy
exit 1
EOF
~~~

`$HOME/.docker/config.json`
~~~JSON
{
    "proxies":
    {
        "default": {
            "noProxy": "my_default_noproxy_value_from_config.json"
        }
    }

}
~~~

Repeatedly run  `docker build --build-arg no_proxy="my_cmdline_value" .`

<img width="1294" height="889" alt="Image" src="https://github.com/user-attachments/assets/7a070861-0fc3-4f30-adfe-fb94c28788dd" />

### Expected behavior

When a user explicitly passes a build-arg on the command line, it should **always take precedence** over proxy defaults coming from `~/.docker/config.json`.

For example, running:

~~~bash
docker build --build-arg no_proxy="my_cmdline_value" .
~~~

should deterministically result in the build environment using:

- `no_proxy=my_cmdline_value` and/or
- `NO_PROXY=my_cmdline_value`

and it should **never** be replaced by `proxies.default.noProxy` (e.g. `my_default_noproxy_value_from_config.json`), regardless of how many times the build is repeated.

### docker version

```bash
Client: Docker Engine - Community
 Version:           25.0.2
 API version:       1.44
 Go version:        go1.21.6
 Git commit:        29cf629
 Built:             Thu Feb  1 00:23:03 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          25.0.2
  API version:      1.44 (minimum version 1.24)
  Go version:       go1.21.6
  Git commit:       fce6e0c
  Built:            Thu Feb  1 00:23:03 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.28
[...]

PR #3697, opened 2026-03-05
build: make proxy build-arg override check case-insensitive

crazy-max:
fixes https://github.com/docker/buildx/issues/3636

Buildx currently checks whether a proxy default should be injected using a case-sensitive map lookup, so `--build-arg no_proxy=...` does not block injecting `NO_PROXY` from `~/.docker/config.json`.

That can result in both `build-arg:no_proxy` and `build-arg:NO_PROXY` being sent to BuildKit and because BuildKit resolves proxy args case-insensitively from a Go map, the winning value can vary by map iteration order.

With this patch, Buildx uses a case-insensitive key existence check before injecting proxy defaults, so an explicit CLI proxy arg consistently takes precedence. This aligns behavior with user expectation that explicit `--build-arg` values should deterministically override config defaults.

### DIFF
--- a/build/opt_test.go
+++ b/build/opt_test.go
@@ -156,3 +156,43 @@ func TestCreateExports_RegistryUnpack(t *testing.T) {
 		})
 	}
 }
+
+func TestProxyArgKeyExists(t *testing.T) {
+	tests := []struct {
+		name      string
+		proxyArgs map[string]string
+		key       string
+		want      bool
+	}{
+		{
+			name:      "exact match",
+			proxyArgs: map[string]string{"NO_PROXY": "cli"},
+			key:       "NO_PROXY",
+			want:      true,
+		},
+		{
+			name:      "case insensitive match",
+			proxyArgs: map[string]string{"no_proxy": "cli"},
+			key:       "NO_PROXY",
+			want:      true,
+		},
+		{
+			name:      "no match",
+			proxyArgs: map[string]string{"HTTP_PROXY": "cli"},
+			key:       "NO_PROXY",
+			want:      false,
+		},
+		{
+			name:      "nil map",
+			proxyArgs: nil,
+			key:       "NO_PROXY",
+			want:      false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, proxyArgKeyExists(tt.proxyArgs, tt.key))
+		})
+	}
+}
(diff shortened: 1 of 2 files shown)


build/opt_test.go
```diff
@@ -156,3 +156,43 @@ func TestCreateExports_RegistryUnpack(t *testing.T) {
...
+	}{
+		{
+			name:      "exact match",
+			proxyArgs: map[string]string{"NO_PROXY": "cli"},
+			key:       "NO_PROXY",
+			want:      true,
+		},
+		{
+			name:      "case insensitive match",
+			proxyArgs: map[string]string{"no_proxy": "cli"},
```
thaJeztah, line 175: I know in the CLI and daemon we only consider `NO_PROXY` and `no_proxy`, but don't consider `nO_pRoXy`; perhaps worth considering.


(I can dig up a link to the code if needed)

thaJeztah: e.g.;

https://github.com/docker/cli/blob/5927d80c76b3ce5cf782be818922966e8a0d87a3/cli/config/configfile/file.go#L218-L253

Also possibly relevant; if either `HTTP_PROXY` or `http_proxy` is set, we set both (because upper/lowercase depends on implementation, and there's no official standard)
https://github.com/moby/moby/blob/83bca512aa7ffc1bb4f37ce1107e0d3e3489ad43/daemon/command/daemon.go#L1119-L1132

```go
func configureProxyEnv(ctx context.Context, cfg config.Proxies) {
	if p := cfg.HTTPProxy; p != "" {
		overrideProxyEnv(ctx, "HTTP_PROXY", p)
		overrideProxyEnv(ctx, "http_proxy", p)
	}
	if p := cfg.HTTPSProxy; p != "" {
		overrideProxyEnv(ctx, "HTTPS_PROXY", p)
		overrideProxyEnv(ctx, "https_proxy", p)
	}
	if p := cfg.NoProxy; p != "" {
		overrideProxyEnv(ctx, "NO_PROXY", p)
		overrideProxyEnv(ctx, "no_proxy", p)
	}
}
```

tonistiigi:
I guess we should also fix it in `proxyEnvFromBuildArgs` where the undeterministic map iteration is. I'm not sure what is the preferred order but main thing would be to just not be random.

crazy-max:
> I guess we should also fix it in `proxyEnvFromBuildArgs` where the undeterministic map iteration is. I'm not sure what is the preferred order but main thing would be to just not be random.

Opened https://github.com/moby/buildkit/pull/6560

PR #3701, opened 2026-03-06
dap: fix the check to determine whether exec will succeed

jsternberg:
This refines the check for determining whether exec will succeed to work
when an error occurs. This check previously relied on the `Ref` being
populated in the result context but this would only happen if we were
paused from a breakpoint or by stepping. An error would not fill in this
field.

The check is now refined to use the new gateway filesystem exec API so
we can create the container and then check even if we don't have a
returned gateway reference. The logic to determine which mount to check
has also been moved.

### DIFF
--- a/build/invoke.go
+++ b/build/invoke.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	_ "crypto/sha256" // ensure digests can be computed
 	"io"
+	"io/fs"
+	"os"
+	"path"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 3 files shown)


build/invoke.go
```diff
@@ -146,6 +149,51 @@ func (c *Container) Exec(ctx context.Context, cfg *InvokeConfig, stdin io.ReadCl
...
+		cmd = cfg.Entrypoint[0]
+	} else if len(cfg.Cmd) > 0 {
+		cmd = cfg.Cmd[0]
+	}
+
+	if cmd == "" {
+		return "no command specified", false
+	}
+
+	for {
```
tonistiigi: Symlink resolution should have a max limit.

build/invoke.go
```diff
@@ -146,6 +149,51 @@ func (c *Container) Exec(ctx context.Context, cfg *InvokeConfig, stdin io.ReadCl
...
+			MountIndex: index,
+		})
+		if err != nil {
+			return fmt.Sprintf("stat error %s: %s", cmd, err), false
+		}
+
+		mode := fs.FileMode(st.Mode)
+		if mode&os.ModeSymlink != 0 {
+			// Follow the link.
+			cmd = st.Linkname
```
tonistiigi: Can't this be relative symlink?

build/invoke.go
```diff
@@ -146,6 +149,51 @@ func (c *Container) Exec(ctx context.Context, cfg *InvokeConfig, stdin io.ReadCl
 	return err
 }
 
+func (c *Container) CanInvoke(ctx context.Context, cfg *InvokeConfig) (string, bool) {
```
tonistiigi: Why isn't this returning error, or `(bool, error)`

build/invoke.go
jsternberg: Added this.

build/invoke.go
jsternberg: Yes. Fixed.

build/invoke.go
jsternberg: The original function did this because it's not really an error it returns but more of a "can this run and what's the reason it can't?" but it's not really compelling enough to avoid an error so I just changed it to return an error and the error gives the description about why it can't.

build/result.go
```diff
@@ -81,38 +80,40 @@ func (r *ResultHandle) NewContainer(ctx context.Context, cfg *InvokeConfig) (gat
...
+		return "", 0, iofs.ErrNotExist
 	}
 
-	slices.SortFunc(candidateMounts, func(a, b gateway.Mount) int {
+	slices.SortFunc(candidateMounts, func(a, b mountCandidate) int {
 		return cmp.Compare(len(a.Dest), len(b.Dest))
 	})
 
 	m := candidateMounts[len(candidateMounts)-1]
 	relpath, err := filepath.Rel(m.Dest, fpath)
```
tonistiigi: Is the usage of `filepath` correct in here (and 116)? `CanInvoke` has `path` (that I think is correct).

jsternberg: I think `path` is correct here since it prepares a URL for the grpc call.

PR #3702, opened 2026-03-06
cobrautil: decouple from bake/build dependency chain

crazy-max:
https://github.com/docker/buildx/pull/3646 was not enough because `cobrautil` imports `bake` which imports `build` package so let's decouple this as well.


crazy-max:
Would like to call `history.BuildName` in Desktop Build backend to avoid drifting. But atm history package imports `completion`: https://github.com/docker/buildx/blob/c423bc738b1178cda935006940eb0216ba745629/commands/history/open.go#L7

Which imports `bake`: https://github.com/docker/buildx/blob/c423bc738b1178cda935006940eb0216ba745629/util/cobrautil/completion/completion.go#L6

Which imports `build`: https://github.com/docker/buildx/blob/c423bc738b1178cda935006940eb0216ba745629/bake/bake.go#L24

But can't import `build` package because of BuildKit session export changes for cloud driver:

```
go: github.com/docker/desktop-build/pkg/buildmanager imports
	github.com/docker/buildx/commands/history imports
	github.com/docker/buildx/util/cobrautil/completion imports
	github.com/docker/buildx/bake imports
	github.com/docker/buildx/build imports
	github.com/moby/buildkit/session/export: module github.com/moby/buildkit@latest found (v0.28.0), but does not contain package github.com/moby/buildkit/session/export
```

tonistiigi:
We should move `BuildName` out of `commands` then as the cli commands packages are not for importing.

tonistiigi:
Is this still needed or should be closed?

crazy-max:
> Is this still needed or should be closed?

I think it still makes sense to decouple but I'm fine closing it if you prefer.

PR #3704, opened 2026-03-06
dap: detect breakpoints for files when the case differs

jsternberg:
Case insensitive filesystems can cause breakpoints to not be seen or
verified. This is particularly true on Windows where the drive letter
can also participate in the filepath.

Change the detection logic for a breakpoint to be case insensitive. At
the same time, report the name of the source as part of the breakpoint
so that the editor can be told which casing we're expecting to be used.

Fixes https://github.com/docker/vscode-extension/issues/235.

### DIFF
--- a/dap/adapter.go
+++ b/dap/adapter.go
@@ -9,6 +9,7 @@ import (
 	"path"
 	"path/filepath"
 	"slices"
+	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


rcjsuen (approved):
Verified by debugging on Windows with `contextPath` having `c` and `C` as the drive letter.

PR #3707, opened 2026-03-07
kubernetes: trim client-go dependency surface

tonistiigi:
Replace the full generated clientset and global Kubernetes scheme with a small local REST client layer and minimal scheme registration.

This keeps the existing kubeconfig/auth and remote exec behavior while significantly reducing the linked and vendored Kubernetes dependency set.


driver/kubernetes/execconn/execconn.go
```diff
@@ -29,7 +29,7 @@ func ExecConn(ctx context.Context, restClient rest.Interface, restConfig *rest.C
 			Stdout:    true,
 			Stderr:    true,
 			TTY:       false,
-		}, scheme.ParameterCodec)
+		}, kubeclient.ParameterCodec())
 	exec, err := remotecommand.NewSPDYExecutor(restConfig, "POST", req.URL())
```
tonistiigi, line 33: @AkihiroSuda Iiuc we could also replace this spdy dependency with https://github.com/kubernetes/enhancements/issues/4006 . Seems that would require k8s 1.31+ . When do you think we could do that?

AkihiroSuda: Some EKS users are still using 1.29 and 1.30, until 23 Jul 2026
https://endoflife.date/amazon-eks

AkihiroSuda: AKS 1.30 until 31 Jul 2026
https://endoflife.date/azure-kubernetes-service

go.mod
```diff
@@ -71,9 +71,9 @@ require (
 	google.golang.org/grpc v1.78.0
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.5.1
 	google.golang.org/protobuf v1.36.11
-	k8s.io/api v0.34.1
-	k8s.io/apimachinery v0.34.1
-	k8s.io/client-go v0.34.1
+	k8s.io/api v0.35.2
+	k8s.io/apimachinery v0.35.2
+	k8s.io/client-go v0.35.2
```
AkihiroSuda, line 76: Alternatively you can just exec kubectl to significantly reduce the binary footprint and Go dependencies.

14M → 6.1M, in the case of Lima:
- https://github.com/lima-vm/lima/pull/4120
- https://github.com/lima-vm/lima/pull/4536

tonistiigi: I think I'd prefer to avoid the extra dependency for buildx users.

AkihiroSuda: I guess it is safe to assume that the users of the buildx kubernetes driver should have already kubectl installed

tonistiigi: I'm not so sure about it. I think it is messier to manage flag parsing and output reads, and I'm not sure how stable the CLI is, even if it is considered.

PR #3344, opened 2025-07-30
e2e: test build records export

crazy-max:
relates to https://github.com/docker/build-push-action/issues/1388

PR #3716, opened 2026-03-11
history: finalize export records on their owning node

crazy-max:
~needs https://github.com/docker/buildx/pull/3724~
fixes https://github.com/docker/build-push-action/issues/1388
closes https://github.com/docker/buildx/pull/3344
closes https://github.com/docker/build-push-action/pull/1392
closes https://github.com/docker/bake-action/pull/336

This fixes `history export --finalize` for multi-node builders such as the Kubernetes driver.

The failure happens when the same logical build ref appears more than once in history, once per node, and export finalization is not applied to the node that owns the selected record. In that case Buildx can later try to export a record whose trace descriptor is still not readable and fail with `NotFound` while loading the trace descriptor:

```
+ docker buildx --builder=buildx-test-c882461ff920a3349084436f290222c8 history ls
}BUILD ID                    NAME                       STATUS      CREATED AT               DURATION   
kh4h2q8qavh3iqe7bvpz869gy   buildx-output.PdsmBG11XK   Completed   Less than a second ago   0.2s       
z8xx18nyoptmh8l0ylyo5bqat   buildx-output.PdsmBG11XK   Completed   1 second ago             1.1s       
z8xx18nyoptmh8l0ylyo5bqat   buildx-output.PdsmBG11XK   Completed   1 second ago             0.9s       
+ docker buildx --builder=buildx-test-c882461ff920a3349084436f290222c8 history export --finalize --all --output /tmp/buildx-output.PdsmBG11XK/records.dockerbuild
ERROR: failed to export: failed to export record: failed to load trace descriptor: failed to get info: NotFound: rpc error: code = NotFound desc = content digest sha256:c257101ad2a226de341bdab5149018946bbea30e8290135b15c9ccf5755f5f60: not found
```

This keeps the existing finalize behavior for records that do not yet expose a trace descriptor, but changes finalization to follow record ownership. Buildx now deduplicates selected records by (ref, node), sends finalize to the owning node for each selected record, re-queries records before export, and collapses duplicate per-node entries down to one logical record per ref.

It also fixes bundle export to use all available node content stores instead of stopping at the first one. That allows the exported records to resolve their referenced content correctly across multi-node builders during archive creation.

### DIFF
--- a/tests/history.go
+++ b/tests/history.go
@@ -19,6 +19,8 @@ import (
 var historyTests = []func(t *testing.T, sb integration.Sandbox){
 	testHistoryExport,
 	testHistoryExportFinalize,
+	testHistoryExportFinalizeMultiNodeRef,
+	testHistoryExportFinalizeMultiNodeAll,
 	testHistoryInspect,
 	testHistoryLs,
 	testHistoryRm,
@@ -29,10 +31,11 @@ var historyTests = []func(t *testing.T, sb integration.Sandbox){
 func testHistoryExport(t *testing.T, sb integration.Sandbox) {
 	ref := buildTestProject(t, sb)
 	require.NotEmpty(t, ref.Ref)
+	requireHistoryRef(t, sb, ref.Ref)
 
 	outFile := path.Join(t.TempDir(), "export.dockerbuild")
 	cmd := buildxCmd(sb, withArgs("history", "export", ref.Ref, "--output", outFile))
-	out, err := cmd.Output()
+	out, err := cmd.CombinedOutput()
 	require.NoError(t, err, string(out))
 	require.FileExists(t, outFile)
 }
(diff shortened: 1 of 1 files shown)


crazy-max:
Hum still fails, looking at the logs: https://github.com/docker/buildx/actions/runs/22947651040/job/66604243143?pr=3716#step:9:321

```
+ docker buildx --builder=buildx-test-30a7ab764361dd1f4482285765e20405 history ls
}BUILD ID                    NAME                       STATUS      CREATED AT               DURATION   
0v82njsviycrtwp3hicwi7ma1   buildx-output.9R23X274Na   Completed   Less than a second ago   0.1s       
i2hwfladqm5qikdpkv1tqrd7x   buildx-output.9R23X274Na   Completed   1 second ago             1.2s       
i2hwfladqm5qikdpkv1tqrd7x   buildx-output.9R23X274Na   Completed   1 second ago             1.0s       
+ docker buildx --builder=buildx-test-30a7ab764361dd1f4482285765e20405 history export --finalize --all --output /tmp/buildx-output.9R23X274Na/records.dockerbuild
ERROR: failed to export: failed to export record: failed to load trace descriptor: failed to get info: NotFound: rpc error: code = NotFound desc = content digest sha256:da1a7755c4ee696f92ea39b1f92465e758d344f0ac6262061c21dea33432d81e: not found
```

`history ls` shows the same build ref twice, once per node. That means export is trying to bundle per-node records for a multi-node build, and one node is advertising a trace digest that its content store still cannot serve or smth around that?

commands/history/export.go
```diff
@@ -103,6 +95,7 @@ func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) e
 			return err
 		}
 		clients = append(clients, c)
+		visited[rec.node] = struct{}{}
```
crazy-max: The collection loop was also updated to deduplicate nodes correctly during export. Seems to be an oversight.

commands/history/export.go
```diff
@@ -53,37 +53,29 @@ func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) e
 			return errors.Errorf("no record found for ref %q", ref)
 		}
 
+		selected := selectExportRecordCandidates(ref, recs, opts.all)
 		if opts.finalize {
-			var finalized bool
-			for _, rec := range recs {
-				if rec.Trace == nil {
```
crazy-max, line 59: @tonistiigi I removed the `if rec.Trace == nil` condition because it doesn't reliably mean the record is finalized. On Kubernetes the trace descriptor can already be present while the trace blob is still unavailable, so export must finalize selected records unconditionally and then re-query them.

util/desktop/bundle/export.go
```diff
@@ -38,7 +38,7 @@ func Export(ctx context.Context, c []*client.Client, w io.Writer, records []*Rec
 		s := proxy.NewContentStore(c.ContentClient())
 		if store == nil {
 			store = s
-			break
+			continue
```
crazy-max, line 41: use all available node content stores instead of stopping at the first one

crazy-max:
Tried to add integration tests but can't make it works with docker runtime atm: https://github.com/docker/buildx/actions/runs/22994422408/job/66762935425#step:7:2418

* [logs.txt](https://github.com/user-attachments/files/25930439/logs.txt)

Will work on multi-nodes support in our integration tests as follow-up: https://github.com/crazy-max/buildx/commit/35166b9e57143e038e7d06373710e0c6ae751475

Edit: opened https://github.com/docker/buildx/pull/3724 and rebased on top of it

commands/history/export.go
```diff
@@ -146,6 +171,26 @@ func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) e
 	return bundle.Export(ctx, clients, w, toExport)
 }
 
+func selectExportRecordCandidates(ref string, recs []historyRecord, all bool) []historyRecord {
```
tonistiigi: It isn't obvious why this new function is here. `queryRecors` already does the empty ref and sorting.

commands/history/export.go
```diff
@@ -53,37 +52,67 @@ func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) e
 			return errors.Errorf("no record found for ref %q", ref)
 		}
 
+		selected := recs
```
tonistiigi: `selected` -> `toExport`
`prefixLen` -> `recCount`

commands/history/export.go
```diff
@@ -53,37 +52,67 @@ func runExport(ctx context.Context, dockerCli command.Cli, opts exportOptions) e
...
+				selected = recs
 			}
 		}
-
-		if ref == "" {
-			slices.SortFunc(recs, func(a, b historyRecord) int {
-				return b.CreatedAt.AsTime().Compare(a.CreatedAt.AsTime())
-			})
+		if len(selected) > 1 {
+			// A logical build ref can appear once per node. Keep one candidate
```
tonistiigi: Not sure I understand this. Multi-node build still has two separate history records with different data. If you are dropping one node here, the data from another node still gets dropped from the export.

crazy-max: Ah yes you're right, I was treating duplicate refs as alternative candidates for the same logical record, but `bundle.Export` does not merge them. It exports one manifest per input record.

PR #3719, opened 2026-03-11
tests: add negative checks for imagetools signature coverage

tonistiigi:
Verify imagetools copies signatures only for attestation manifests and does not copy image-manifest signatures or unsupported referrers.


crazy-max:
Some lints issues otherwise LGTM

tonistiigi:
This was included in #3721

PR #3721, opened 2026-03-12
imagetools: add oci-layout support

tonistiigi:
Add support for oci-layout format in `imagetools create` and `inspect`. Same format as named context in build:
```
oci-layout://<path>[:tag]
oci-layout://<path>[@sha256:...]
```

Can be used both as a source or target parameter. You can mix oci-layout and registry references as you like.

### DIFF
--- a/util/imagetools/inspect.go
+++ b/util/imagetools/inspect.go
@@ -172,7 +172,7 @@ func (r *Resolver) localStore(path string) (content.Store, error) {
 
 func (r *Resolver) FetchReferrers(ctx context.Context, loc *Location, dgst digest.Digest, opts ...remotes.FetchReferrersOpt) ([]ocispecs.Descriptor, error) {
 	if loc.IsOCILayout() {
-		return fetchOCILayoutReferrers(ctx, r.GetDescriptor, loc, dgst)
+		return fetchOCILayoutReferrers(ctx, r.GetDescriptor, loc, dgst, opts...)
 	}
 	f, err := r.registryResolver().Fetcher(ctx, loc.String())
 	if err != nil {
--- a/util/imagetools/inspect_test.go
+++ b/util/imagetools/inspect_test.go
@@ -1,0 +1,53 @@
+package imagetools
+
+import (
+	"context"
+	"testing"
+
+	"github.com/containerd/containerd/v2/core/images"
+	"github.com/containerd/containerd/v2/core/remotes"
+	"github.com/moby/buildkit/client/ociindex"
+	"github.com/opencontainers/go-digest"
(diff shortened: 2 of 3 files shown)


util/imagetools/create.go
```diff
@@ -505,6 +492,100 @@ func (r *Resolver) filterPlatforms(ctx context.Context, dt []byte, desc ocispecs
...
+	if err := os.MkdirAll(ref.OCILayout().Path, 0o755); err != nil {
+		return err
+	}
+	store, err := r.localStore(ref.OCILayout().Path)
+	if err != nil {
+		return err
+	}
+	w, err := store.Writer(ctx, content.WithRef(desc.Digest.String()), content.WithDescriptor(desc))
+	if err != nil {
+		if errdefs.IsAlreadyExists(err) {
```
crazy-max: I think that would break silently and skips the index.json update if we retag, for example:

```
docker buildx imagetools create -t oci-layout:///tmp/demo-layout:v1 docker.io/library/alpine:latest

docker buildx imagetools create -t oci-layout:///tmp/demo-layout:v2 docker.io/library/alpine:latest
```

The second command will succeed but without adding the new tag

util/imagetools/inspect.go
```diff
@@ -113,6 +135,119 @@ func (r *Resolver) GetDescriptor(ctx context.Context, in string, desc ocispecs.D
...
+		return nil, err
+	}
+	r.localStores[path] = store
+	return store, nil
+}
+
+func (r *Resolver) FetchReferrers(ctx context.Context, loc *Location, dgst digest.Digest, opts ...remotes.FetchReferrersOpt) ([]ocispecs.Descriptor, error) {
+	if loc.IsOCILayout() {
+		return fetchOCILayoutReferrers(ctx, r.GetDescriptor, loc, dgst)
+	}
```
crazy-max, line 176: Seems `opts`  are ignored for OCI layout so the filter at: https://github.com/docker/buildx/pull/3721/changes#diff-273961e2b2f5a64c8ad90d15663a519f07200712496ae01f0b8b8d989a0d95aaR448

```
refs, err := r.FetchReferrers(ctx, src.Ref, d, remotes.WithReferrerArtifactTypes(artifactTypeAttestationManifest))
```

Could pull the wrong referrer type if a non-attestation referrer is returned first?

Not quite sure if we can repro this, maybe with a source oci layout that contains an attestation referrer for a platform manifest and other direct referrer on the same platform manifest like a signature?

crazy-max:
Needs docs follow-up for v0.33.0 (cc @dvdksn)

PR #3726, opened 2026-03-12
gitutil: migrate to BuildKit GitCLI API

crazy-max:
~needs https://github.com/moby/buildkit/pull/6624~
closes #2005 

replaces the local compatibility wrapper with the BuildKit-style gitutil API.


tonistiigi (commented):
I'm not sure if this is equivalent as buildkit git implementation sets things like

```
			"GIT_CONFIG_NOSYSTEM=1", // Disable reading from system gitconfig.
			"HOME=/dev/null",        // Disable reading from user gitconfig.
```

because buildkit doesn't want to leak system config.

crazy-max:
> I'm not sure if this is equivalent as buildkit git implementation sets things like
> 
> ```
> 			"GIT_CONFIG_NOSYSTEM=1", // Disable reading from system gitconfig.
> 			"HOME=/dev/null",        // Disable reading from user gitconfig.
> ```
> 
> because buildkit doesn't want to leak system config.

Ah yes you right, the previous wrapper inherited `os.Environ()` and therefore read user/system git config

PR #3728, opened 2026-03-15
docs: clarify Bake environment variable override behavior

allexistence:
## Summary

Clarifies the Bake reference docs for environment variable behavior.

The previous wording and example suggested that values like `$HOME` are
interpolated in the `default` field. In practice, Bake overrides a declared
variable when an environment variable with the same name is set.

## Changes

- Reworded the section to describe name-based environment variable override
- Replaced the ambiguous `default = "$HOME"` example
- Added a pointer to the overrides documentation

Related:
- docker/docs#23762

### DIFF
--- a/docs/bake-reference.md
+++ b/docs/bake-reference.md
@@ -1350,11 +1350,17 @@ to define them.
 
 ### Use environment variable as default
 
-You can set a Bake variable to use the value of an environment variable as a default value:
+If an environment variable exists with the same name as a declared Bake
+variable, Bake uses that environment variable value instead of the declared
+default.
+
+To disable this environment-based variable lookup, set
+`BUILDX_BAKE_DISABLE_VARS_ENV_LOOKUP=1`.
+
 
 ```hcl
 variable "HOME" {
-  default = "$HOME"
+  default = "/root"
 }
 ```
 


docs/bake-reference.md
```diff
@@ -1350,13 +1350,14 @@ to define them.
...
+If an environment variable exists with the same name as a declared Bake
+variable, Bake uses that environment variable value instead of the declared
+default.
 
 ```hcl
 variable "HOME" {
-  default = "$HOME"
+  default = "/root"
 }
-```
```
crazy-max, line 1359: Seems it got removed by mistake

allexistence:
Addressed feedback:
- restored missing code fence
- ensured correct example
- added note for BUILDX_BAKE_DISABLE_VARS_ENV_LOOKUP

PTAL 🙏

docs/bake-reference.md
allexistence: Addressed feedback . PTAL 🙏

allexistence:
> Can you squash your commits please? Thx

PTAL

crazy-max:
> Is this a place where it's worth to mention `BUILDX_BAKE_DISABLE_VARS_ENV_LOOKUP` as well.

Yes this is marked as docs follow-up to document this new var: https://github.com/docker/buildx/pull/3595

@dvdksn Seems we have other things left to cover for docs: https://github.com/docker/buildx/pulls?q=is%3Apr+label%3Astatus%2Fneeds-docs-follow-up+is%3Aclosed

PR #3735, opened 2026-03-19
dap: pass exit code through exited event

jsternberg:
Pass the exit code through the exited event back to the client and
ensure that the printed text is printed completely.

Previously, the exited event just had a big todo and the printer would
sometimes fail to send messages to the connected client. This moves the
printer wait to before the debug adapter is closed to ensure that all
messages get sent through the connection to the editor. While there, I
also plumbed in the exit code to exited. It's not necessarily the real
exit code but it will produce a zero on build success and a non-zero
code on build failure so that should be good enough.

### DIFF
--- a/dap/adapter.go
+++ b/dap/adapter.go
@@ -83,7 +83,7 @@ func (d *Adapter[C]) Start(conn Conn) (C, error) {
 	return resp.Config, resp.Error
 }
 
-func (d *Adapter[C]) Stop() error {
+func (d *Adapter[C]) Stop(retErr error) error {
 	if d.eg == nil {
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


dap/adapter.go
```diff
@@ -94,15 +94,19 @@ func (d *Adapter[C]) Stop() error {
...
-		// 	Body: dap.ExitedEventBody{
-		// 		ExitCode: exitCode,
-		// 	},
-		// }
+
+		exitCode := 0
+		if retErr != nil {
+			exitCode = 1
+		}
+		c.C() <- &dap.ExitedEvent{
```
tonistiigi, line 110: iiuc then "terminated" should come after "exited"?

jsternberg: It's a bit strange but this is the correct order.

https://microsoft.github.io/debug-adapter-protocol/overview

Look at the bottom of the page where there's a diagram. The terminated event is sent when the debugee terminates but hasn't necessarily exited while exited is sent after it has exited. I can add a comment to the docs to clarify this in the code.

dap/adapter.go
```diff
@@ -94,15 +94,27 @@ func (d *Adapter[C]) Stop() error {
...
-		// 	Event: dap.Event{
-		// 		Event: "exited",
-		// 	},
-		// 	Body: dap.ExitedEventBody{
-		// 		ExitCode: exitCode,
-		// 	},
-		// }
+
+		// Send an exit code based on the returned error.
+		// Any error resuls in sending an exit code of 1 while
```
tonistiigi: typo: results

jsternberg: Fixed.

PR #3747, opened 2026-03-27
util/gitutil: Full/ShortCommit(): replace git show with git rev-parse

jegorbunov:
The FullCommit and ShortCommit methods are expected to simply get the hash of HEAD. Using `git show` for this purpose is overkill because `git show` can have side effects that are annoying when `docker build` runs in a CI environment:

1) CI environments (e.g. CircleCI) may use blobless clones, where the target repository is checked out using `git clone --filter=blob:none {url}`.

2) `git show` does more than just discover the hash of a specified revision: in general, it needs parent commits and blobs to compute the diff. When a blobless clone is used, `git show` may try to download these additional blobs from the remote. From my experiments, this happens even when `--quiet` or `--no-patch` is specified.

3) If the `docker build` step runs in an environment where Git is not configured properly to download from the remote, the step may hang indefinitely, as it did in my case with the following prompt:

   The authenticity of host 'github.com (140.82.112.3)' can't be established. ED25519 key fingerprint is SHA256:+.... This key is not known by any other names. Are you sure you want to continue connecting (yes/no/[fingerprint])?

### DIFF
--- a/util/gitutil/gitutil.go
+++ b/util/gitutil/gitutil.go
@@ -101,11 +101,11 @@ func (c *Git) RemoteURL() (string, error) {
 }
 
 func (c *Git) FullCommit() (string, error) {
-	return c.clean(c.run("show", "--format=%H", "HEAD", "--quiet", "--"))
+	return c.clean(c.run("rev-parse", "HEAD"))
 }
 
 func (c *Git) ShortCommit() (string, error) {
-	return c.clean(c.run("show", "--format=%h", "HEAD", "--quiet", "--"))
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-upstream.yml
thaJeztah, line 62: Looks like your PR was opened from an outdated branch and pulled in a commit from https://github.com/docker/buildx/pull/3750 - can you try rebasing?

jegorbunov: Thanks, fixed.

PR #2938, opened 2025-01-21
Kubernetes Driver: switch to 'StatefulSet' from 'Deployment'

ArielLahiany:
I've switched the Kubernetes driver object to 'StatefulSet' from 'Deployment' In order to support long-living persistent data directories between restart, and to meet the needs of the following issue: https://github.com/docker/buildx/issues/2056
Please let me know if there are any other changes that are required, or if a different PR format is required.
Thank you.

PR #3766, opened 2026-03-31
feat: add persistent storage options to k8s driver

### LINKED ISSUE #2056: Support statefulSets with the kubernetes driver
### Description

The kubernetes driver currently only supports creating/using deployments for buildx nodes. This makes it difficult if we want to use local caching - as we can't attach a separate volume to each replica in a deployment (buildkit doesn't support sharing a cache between nodes). A statefulset allows us to attach a volume to each node to store cache, but the kubernetes driver only looks for deployments when looking for existing nodes to use.

So we either have to have a single buildkit node, or store cache in `emptydir`s - neither of which is ideal.

jsternberg:
The k8s driver now supports the `persistent-volume-claim.requests.storage` option.
Setting this option changes the deployment into a statefulset and
creates a persistent volume claim where the buildkit data is stored.

Fixes #2056.
Replaces #2938.

### DIFF
--- a/driver/kubernetes/driver.go
+++ b/driver/kubernetes/driver.go
@@ -44,15 +44,17 @@ type Driver struct {
 
 	// if you add fields, remember to update docs:
 	// https://github.com/docker/docs/blob/main/content/build/drivers/kubernetes.md
-	minReplicas      int
-	deployment       *appsv1.Deployment
-	configMaps       []*corev1.ConfigMap
-	deploymentClient kubeclient.DeploymentClient
-	podClient        kubeclient.PodClient
-	configMapClient  kubeclient.ConfigMapClient
-	podChooser       podchooser.PodChooser
-	defaultLoad      bool
-	timeout          time.Duration
+	minReplicas       int
+	deployment        *appsv1.Deployment
+	statefulSet       *appsv1.StatefulSet
+	configMaps        []*corev1.ConfigMap
+	deploymentClient  kubeclient.DeploymentClient
+	statefulSetClient kubeclient.StatefulSetClient
+	podClient         kubeclient.PodClient
+	configMapClient   kubeclient.ConfigMapClient
+	podChooser        podchooser.PodChooser
+	defaultLoad       bool
+	timeout           time.Duration
 }
 
 func (d *Driver) IsMobyDriver() bool {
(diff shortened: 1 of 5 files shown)


jsternberg:
@AkihiroSuda I was able to test this and it seems to work on my local k3s. Can you take a look at this and make sure it works for you and doesn't change any existing behavior? I don't use the kubernetes driver in my normal workload.

driver/kubernetes/driver.go
```diff
@@ -127,20 +163,69 @@ func (d *Driver) wait(ctx context.Context) error {
...
+			return context.Cause(ctx)
+		case <-timeoutChan:
+			return err
+		case <-ticker.C:
+			stat, err = d.statefulSetClient.Get(ctx, d.statefulSet.Name, metav1.GetOptions{})
+			if err == nil {
+				if stat.Status.ReadyReplicas >= int32(d.minReplicas) {
+					return nil
+				}
+				err = errors.Errorf("expected %d replicas to be ready, got %d", d.minReplicas, stat.Status.ReadyReplicas)
```
AkihiroSuda: This transient state seems fine on `<- ticker.C`?
It still has to be checked on timeout though.

AkihiroSuda: Maybe add a test (after https://github.com/docker/buildx/pull/3345 is merged) to verify this on CI

driver/kubernetes/factory.go
```diff
@@ -199,6 +202,8 @@ func (f *factory) processDriverOpts(deploymentName string, namespace string, cfg
 			deploymentOpt.RequestsMemory = v
 		case k == "requests.ephemeral-storage":
 			deploymentOpt.RequestsEphemeralStorage = v
+		case k == "requests.persistent-storage":
```
AkihiroSuda: The "requests" key prefix is preserved for `ResourceRequirements.Requests` entries
https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements
https://pkg.go.dev/k8s.io/api/core/v1#ResourceName

AkihiroSuda: Maybe:
```suggestion
		case k == "persistent-volume-claim.requests.storage":
```

No strong opinion though

driver/kubernetes/driver.go
jsternberg: I copied this from the deployment code and I think this error only gets sent when the timeout is hit. If the minimum replicas condition is fulfilled, it returns nil and doesn't send this error.

driver/kubernetes/factory.go
jsternberg: Updated the name.

driver/kubernetes/driver.go
jsternberg: Updated this to use a generic function between the two instead of copying the code. The main issue I had was accessing the spec type to get the `ReadyReplicas` attribute for comparison but I was able to fix that by adding a `check` function that gets passed in.

tonistiigi (approved):

crazy-max:
needs docs follow-up https://docs.docker.com/build/builders/drivers/kubernetes/ (cc @dvdksn)

PR #3770, opened 2026-04-02
commands: micro optimization for source date epoch

jsternberg:
Avoids the call to `os.Getenv` when it is unnecessary because it would
be overwritten anyway.

Removes the comments about moving environment variable parsing to a
method for use by library consumers. That method exists in containerd
and this set of code doesn't actually perform any parsing since the
parsing of this time is done within buildkit and not on the client.

* closes #1907
* closes #1908

### DIFF
--- a/commands/bake.go
+++ b/commands/bake.go
@@ -17,6 +17,7 @@ import (
 	"text/tabwriter"
 
 	"github.com/containerd/console"
+	"github.com/containerd/containerd/v2/pkg/epoch"
 	"github.com/containerd/platforms"
 	"github.com/docker/buildx/bake"
 	"github.com/docker/buildx/bake/hclparser"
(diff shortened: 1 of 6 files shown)


thaJeztah:
Ah, yes, I recalled I had a branch, but had to do some work in containerd first! (haven't looked yet at the code, but probably should be good 😅)

commands/bake.go
```diff
@@ -17,6 +17,7 @@ import (
 	"text/tabwriter"
 
 	"github.com/containerd/console"
+	"github.com/containerd/containerd/pkg/epoch"
```
crazy-max:
```suggestion
	"github.com/containerd/containerd/v2/pkg/epoch"
```

commands/build.go
```diff
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/containerd/console"
+	"github.com/containerd/containerd/pkg/epoch"
```
crazy-max:
```suggestion
	"github.com/containerd/containerd/v2/pkg/epoch"
```

commands/bake.go
jsternberg: I have no idea how a rebase changed these to the original version... Fixed.

PR #3771, opened 2026-04-02
fix some linting issues and disable QF1008 (staticcheck)

thaJeztah:
- relates to https://github.com/docker/buildx/pull/3683#issuecomment-3992396545

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -74,6 +74,12 @@ linters:
           alias: "ocispecs"
         - pkg: "github.com/opencontainers/go-digest"
           alias: "digest"
+    staticcheck:
+      # Enable all options, with some exceptions.
+      # For defaults, see https://golangci-lint.run/usage/linters/#staticcheck
+      checks:
+        - all
+        - -QF1008     # Omit embedded fields from selector expression; https://staticcheck.dev/docs/checks/#QF1008
     testifylint:
       disable:
         - empty
--- a/build/build.go
+++ b/build/build.go
@@ -278,9 +278,9 @@ func warnOnNoOutput(ctx context.Context, nodes []builder.Node, opts map[string]O
 	var warnNoOutputBuf bytes.Buffer
 	warnNoOutputBuf.WriteString("No output specified ")
 	if len(noOutputTargets) == 1 && noOutputTargets[0] == "default" {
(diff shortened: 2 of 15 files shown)


.golangci.yml
```diff
@@ -74,6 +74,12 @@ linters:
           alias: "ocispecs"
         - pkg: "github.com/opencontainers/go-digest"
           alias: "digest"
+    staticcheck:
+      # Enable all options, with some exceptions.
+      # For defaults, see https://golangci-lint.run/usage/linters/#staticcheck
+      checks:
+        - all
+        - -QF1008     # Omit embedded fields from selector expression; https://staticcheck.dev/docs/checks/#QF1008
```
crazy-max, line 82: My bigger concern here is not even `-QF1008`, it's switching `staticcheck` from its defaults to `checks: [all, -QF1008]`.

That is a broader policy change than the PR suggests. It doesn't just disable one noisy rule, it also opts this repo into the full staticcheck rule set instead of inheriting upstream defaults. So, we stop following staticcheck's default policy and start owning our own.

It also changes the shape of future `golangci-lint` bumps. By switching to `all`, we opt into the full staticcheck rule set, including rules that may be too opinionated or simply not a good fit for this repo, plus any new checks added upstream later. The default set is usually a more deliberate baseline across projects, whereas `all` pushes us toward a broader policy that may create churn, encourage questionable cleanups, or even move code away from the clearest implementation for this codebase.

I don't see enough benefit here to justify that change imo. If the actual problem is a small number of false positives or overly opinionated reports, I'd rather handle those narrowly than broaden the check set and replace the defaults entirely.

thaJeztah: Yeah, honestly, these settings kinda suck; the default already is `all` with some checks removed, but there's no option to specify `default`, then remove some (or add some); default is;

```console
# Default: ["all", "-ST1000", "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022"]
```

So the `all` will already change when updating, but yeah, it's possible they add more exclusions as a default as well 🫠

crazy-max: Ah I thought default was a minimal set of rules not all of them!

thaJeztah: I can have a look if there's perhaps better ways to configure it. I agree that it'd be much nicer to do something like;

```yaml
- default
- -SOMETHING_I_DONT_WANT
```

Do the other changes look good to you? If so, I could probably move the config changes separate (and can look at options).

PR #3775, opened 2026-04-03
bump compose-go to version v2.10.2

glours:


go.mod
```diff
@@ -224,6 +224,7 @@ require (
 	go.opentelemetry.io/otel/sdk/metric v1.40.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.9.0 // indirect
 	go.yaml.in/yaml/v2 v2.4.3 // indirect
+	go.yaml.in/yaml/v4 v4.0.0-rc.4 // indirect
```
crazy-max, line 227: Outch +15k LOC

Also any reason to consume an RC instead of stable releases?

thaJeztah: Was curious as well when I saw compose was on v4 already; wasn't sure if there were specific features needed though. Initially was hoping they have a non-rc soon, but looks like that repo is moving .. slow.

Currently, most dependencies are either on go.pkg.in/yaml/v3 (deprecated) or go.yaml.in/v3 (not deprecated, drop-in for the old one).

FWIW; looks like testify will move to this module as well, once they have an GA release; currently they're still on the deprecated pkg.go.in/yaml/v3 package, and they stated licensing issues for the go.yaml.in/yaml/v3 package (https://github.com/stretchr/testify/pull/1772#issuecomment-3143937825) that would be resolved with v4, and they're waiting for a stable v4 https://github.com/stretchr/testify/pull/1772#issuecomment-3317995718

glours: Strange, the first RC was introduced months ago https://github.com/compose-spec/compose-go/pull/838
Anyway I don't think there is any changes link to build between the current version used by buildx and this new one.
Feel free to close the PR 😉

crazy-max: Ah yes for `go.pkg.in/yaml/v3` it depends on testify indeed. For `go.yaml.in/yaml/v2` that's from k8s modules:

```
$ go mod why go.yaml.in/yaml/v2  
# go.yaml.in/yaml/v2
github.com/docker/buildx/driver/kubernetes/kubeclient
k8s.io/apimachinery/pkg/runtime
sigs.k8s.io/structured-merge-diff/v6/value
go.yaml.in/yaml/v2
```

No worries @glours I'm fine merging this one, was just questioning if maybe the rc was an oversight.

thaJeztah: Yes, I think this is all temporary; hopefully projects will converge on the new modules "soon", then these will go away.

PR #3777, opened 2026-04-03
driver/docker-container: remove uses of jsonmessage

thaJeztah:
This function was always using `io.Discard` for printing the progress, so we can use the `Wait()` method, which reads the stream, returning any error (similar to jsonmessage.DisplayJSONMessagesStream), and closes the stream either if the context is cancelled, or if the stream ends.


driver/docker-container/driver.go
```diff
@@ -103,8 +102,7 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 		if err != nil {
 			return err
 		}
-		defer resp.Close()
-		return jsonmessage.DisplayJSONMessagesStream(resp, io.Discard, 0, false, nil)
```
crazy-max: #2988 changed this code path to explicitly read the pull JSON stream and return any `errorDetail` entries from the daemon.

That behavior matters because images/create can return 200 OK and then report the real failure later in the streamed JSON response, which is the bug fixed in moby/moby#49453

The current implementation still preserves that because DisplayJSONMessagesStream() returns jm.Error from the stream:
- https://github.com/docker/buildx/blob/182f1f41074daad0f8bf4da8edb553f8b7e9acca/vendor/github.com/moby/moby/client/pkg/jsonmessage/jsonmessage.go#L116-L144
- https://github.com/docker/buildx/blob/182f1f41074daad0f8bf4da8edb553f8b7e9acca/vendor/github.com/moby/moby/client/pkg/jsonmessage/jsonmessage.go#L186-L200

By contrast, resp.Wait(ctx) only returns decoder/context errors and does not inspect jsonstream.Message.Error at all:
- https://github.com/docker/buildx/blob/182f1f41074daad0f8bf4da8edb553f8b7e9acca/vendor/github.com/moby/moby/client/internal/jsonmessages.go#L47-L83
- https://github.com/docker/buildx/blob/182f1f41074daad0f8bf4da8edb553f8b7e9acca/vendor/github.com/moby/moby/api/types/jsonstream/message.go#L8-L14

So this change looks like a regression of #2988: pull failures reported via streamed `errorDetail` would be silently ignored again, and bootstrap would fall through to a later, less accurate error.

thaJeztah:
> By contrast, resp.Wait(ctx) only returns decoder/context errors and does not inspect jsonstream.Message.Error at all:

Hm.. good catch; I think that's a bug!


Let me move this one to draft, and look at that implementation.

crazy-max:
With https://github.com/docker/buildx/pull/3779#pullrequestreview-4055868851 merged I think it can be closed?

thaJeztah:
I think with this fix merged in moby/client;

- https://github.com/moby/moby/pull/52305

(and vendored in https://github.com/docker/buildx/pull/3806), this should be an option again :tada:


Moving it out of draft, but double check if it's now working as intended 😅

crazy-max (approved):

PR #3779, opened 2026-04-03
vendor: moby/client v0.4.0, moby/api v1.54.1

thaJeztah:


driver/docker-container/driver.go
```diff
@@ -104,7 +104,7 @@ func (d *Driver) create(ctx context.Context, l progress.SubLogger) error {
 			return err
 		}
 		defer resp.Close()
-		return jsonmessage.DisplayJSONMessagesStream(resp, io.Discard, 0, false, nil)
+		return jsonmessage.DisplayStream(resp, io.Discard)
```
thaJeztah, line 107: This one just delegates TTY detection (in this case "no TTY"), so that it's not needed to pass it in; see 

- https://github.com/moby/moby/pull/52273
- follow-up: https://github.com/moby/moby/pull/52285

crazy-max: Ah yes sounds better than #3777

PR #3790, opened 2026-04-09
commands: stop dial-stdio when the builder connection closes

### LINKED ISSUE #3668: dial-stdio hangs until write attempt before returning an error when Docker daemon is restarted
### Contributing guidelines

### Description

When connection to Docker daemon is interrupted (e.g. via daemon restarting), dial-stdio hangs indefinitely until something is written to stdin before exiting with an error.

### Expected behaviour

dial-stdio to exit immediately when Docker daemon connection is closed or to re-establish the connection.

### Actual behaviour

dial-stdio hangs indefinitely until something is written to stdin before exiting with an error.

### Buildx version

github.com/docker/buildx 0.31.1 a2675950d46b2cb171b23c2015ca44fb88607531

### Docker info

```text
Client:
 Version:    29.2.1
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  0.31.1
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
    Version:  5.0.2
    Path:     /usr/lib/docker/cli-plugins/docker-compose
  scan: Docker Scan (Docker Inc.)
    Version:  0.9.0
    Path:     /home/invidian/.docker/cli-plugins/docker-scan

Server:
 Containers: 23
  Running: 4
  Paused: 0
  Stopped: 19
 Images: 10
 Server Version: 29.2.1
 Storage Driver: overlayfs
  driver-type: io.containerd.snapshotter.v1
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: dea7da592f5d1d2b7755e3a161be07f43fad8f75.m
 runc version:
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.18.3-zen1-1-zen
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 31.08GiB
 Name: dellxps15mateusz
 ID: RA6A:HWDG:AOTK:PDDB:N2SE:VGNB:GTVU:GR2A:MTIO:7NCA:JMPF:IT5I
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
 Firewall Backend: iptables
```

### Builders list

```text
NAME/NODE     DRIVER/ENDPOINT   STATUS    BUILDKIT   PLATFORMS
default*      docker
 \_ default    \_ default       running   v0.27.1    linux/amd64 (+3), linux/arm64, linux/arm (+2), linux/ppc64le, (6 more)
```

### Configuration

1. Run `docker buildx dial-stdio --progress=plain` in terminal A.
2. Wait until connection to builder is established.
3. Run `sudo systemctl restart docker` in terminal B.
4. Observe the hang.
5. Hit enter in terminal A.
6. Builder should print `#1 ERROR: write unix @->/run/docker.sock: write: broken pipe`.

### Build logs

```text

```

### Additional info

Discovered when [Dalec](https://github.com/project-dalec/dalec) integration tests has been hanging (project-dalec/dalec#970) because of a bug in source policies which caused Docker daemon to crash.

Refs project-dalec/dalec#972

crazy-max:
fixes https://github.com/docker/buildx/issues/3668

Fixes a `dial-stdio` hang that could persist after the builder connection closed. The command now exits when the read side finishes instead of waiting for a new write on stdin to unblock the process.

The old code used two `io.Copy` calls and waited for both to finish, which meant a blocked read from stdin could keep the command alive even after the builder connection had already died. This showed up when the Docker daemon restarted and `dial-stdio` stayed hung until the user pressed Enter.

Tried to add an integration test for this but that's tricky. We would need some wiring with explicit restart/kill hook into the worker to test this regression. So needs changes in BuildKit test framework imo.

cc @invidian @cpuguy83

### DIFF
--- a/commands/dial_stdio.go
+++ b/commands/dial_stdio.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"context"
 	"io"
 	"net"
 	"os"
@@ -15,7 +16,6 @@ import (
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	"golang.org/x/sync/errgroup"
 )
 
 type stdioOptions struct {
@@ -79,27 +79,43 @@ func runDialStdio(dockerCli command.Cli, opts stdioOptions) error {
 			return err
 		}
 
-		defer conn.Close()
-
-		go func() {
-			<-ctx.Done()
-			closeWrite(conn)
-		}()
-
-		var eg errgroup.Group
-
-		eg.Go(func() error {
-			_, err := io.Copy(conn, os.Stdin)
-			closeWrite(conn)
-			return err
-		})
-		eg.Go(func() error {
-			_, err := io.Copy(os.Stdout, conn)
-			closeRead(conn)
-			return err
-		})
-		return eg.Wait()
+		return proxyConn(ctx, conn, os.Stdin, os.Stdout)
 	})
+}
+
+func proxyConn(ctx context.Context, conn net.Conn, stdin io.Reader, stdout io.Writer) error {
+	defer conn.Close()
+
+	stdinDone := make(chan error, 1)
+	stdoutDone := make(chan error, 1)
+
+	go func() {
+		_, err := io.Copy(conn, stdin)
+		closeWrite(conn)
+		stdinDone <- err
+	}()
+	go func() {
+		_, err := io.Copy(stdout, conn)
+		closeRead(conn)
+		stdoutDone <- err
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return context.Cause(ctx)
+		case err := <-stdinDone:
+			if err != nil && !errors.Is(err, net.ErrClosed) && !errors.Is(err, io.ErrClosedPipe) {
+				return err
+			}
+			stdinDone = nil
+		case err := <-stdoutDone:
+			if err != nil && !errors.Is(err, io.EOF) && !errors.Is(err, net.ErrClosed) && !errors.Is(err, io.ErrClosedPipe) {
+				return err
+			}
+			return nil
+		}
+	}
 }
 
 func closeRead(conn net.Conn) error {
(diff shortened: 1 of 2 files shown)


tonistiigi:
This kind of leaves the reader in unknown state. Normally you would read to `io.Discard` in here but not sure if it would hang.

crazy-max:
> This kind of leaves the reader in unknown state. Normally you would read to `io.Discard` in here but not sure if it would hang.

I simplified this by removing the cancelable reader wrapper entirely. The proxy now starts the two `io.Copy` directions directly and waits on their completion channels.

When stdin finishes first, it closes the connection write side and keeps waiting for remote output. When the remote side finishes first, it closes the read side and returns immediately, so `dial-stdio` no longer hangs on a blocked stdin read.

That avoids leaving the original reader in an odd state, since we no longer put a pipe reader in front of it or try to cancel it from the outside. I also removed the extra `errgroup`, so the implementation is much closer to the original copy-loop shape.

I tested it on a real remote builder by closing the connection.

commands/dial_stdio.go
```diff
@@ -79,29 +79,45 @@ func runDialStdio(dockerCli command.Cli, opts stdioOptions) error {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return context.Cause(ctx)
+		case err := <-stdinDone:
+			if err != nil && !errors.Is(err, net.ErrClosed) && !errors.Is(err, io.ErrClosedPipe) {
+				return err
+			}
+			stdinDone = nil
```
cpuguy83, line 111: Doesn't seem necessary.

tonistiigi (approved):

cpuguy83 (commented):
Sorry, I don't recall this one coming in.
LGTM.

cpuguy83 (approved):

PR #3791, opened 2026-04-09
driver/kubernetes: add manifest-patch driver option

### LINKED ISSUE #2626: Feature Request - Add Support for Owner References in Kubernetes Driver
### Description

Hello,

I would like to have the ability to specify Kubernetes Owner References via the Docker Buildx CLI when using the Kubernetes driver. This feature will enhance the integration between Docker Buildx and Kubernetes, allowing resources created during the build process (e.g., Pods, Deployments) to automatically adopt ownership lifecycle policies.

Currently, Docker Buildx allows specifying a variety of options and configurations when using the Kubernetes driver. However, a critical lifecycle management feature is missing: the ability to set ownerReferences on Kubernetes resources like Pods directly through the Buildx CLI.

**CLI and Driver Enhancements:**
- Provide new --driver-opt options to specify ownerReferences.
- Allow users to pass owner reference details (e.g., apiVersion, kind, name, uid, controller, blockOwnerDeletion).

**Use Case:** 
This feature is particularly useful for integrating Docker Buildx with ephemeral GitHub Actions runners or other resources that have dynamic lifecycles. For instance, in CI/CD environments, deployments or Pods created as a part of the build process should be cleaned up automatically when the controlling resource is deleted.

**Current Manual Process:** 
Currently, we are manually adding the owner reference to a Buildx builder deployment to a runner. See an example:

```
> kubectl patch deployment builder-2f073425-0441-48bc-8ed2-2d8a286158001 --type='json' -p='[{"op": "replace", "path": "/metadata/ownerReferences", "value":[{"apiVersion": "actions.github.com/v1alpha1", "kind": "EphemeralRunner", "name": "large-dind-spot-rjz82-runner-4w7sp", "uid": "b636330d-26b7-417a-8464-c2641438feed", "controller": true, "blockOwnerDeletion": false}]}]'

> kubectl get ephemeralrunner large-dind-spot-rjz82-runner-4w7sp
Error from server (NotFound): ephemeralrunners.actions.github.com "large-dind-spot-rjz82-runner-4w7sp" not found
> kubectl get deployment builder-2f073425-0441-48bc-8ed2-2d8a286158001
Error from server (NotFound): deployments.apps "builder-2f073425-0441-48bc-8ed2-2d8a286158001" not found
```

This ensures that when the runner is deleted, the builder is deleted as well. This works much better than running the cleanup at the end of a GitHub Actions job since there are cases where a runner will exit prematurely and never actually run the builder cleanup at the end of a job.

Adding an owner reference ensures the builder deployment will be a part of the GitHub Actions runner lifecycle in Kubernetes.

Thank You

abhay1999:
Add a generic `manifest-patch` driver option that accepts a jq
expression applied to the generated Deployment or StatefulSet
manifest after creation.

This allows users to set arbitrary manifest fields not exposed as
dedicated driver options, such as `ownerReferences`:

    docker buildx create \
      --driver kubernetes \
      --driver-opt 'manifest-patch=.metadata.ownerReferences=[{"apiVersion":"actions.github.com/v1alpha1","kind":"EphemeralRunner","name":"runner-xyz","uid":"..."}]'

The jq expression is evaluated using github.com/itchyny/gojq, a
pure-Go jq implementation. The expression is validated at driver
initialization time. An error is returned if the expression is
invalid or produces no output.

Fixes #2626


driver/kubernetes/factory.go
```diff
@@ -272,6 +274,46 @@ func (f *factory) processDriverOpts(deploymentName string, namespace string, cfg
 
 				deploymentOpt.Tolerations = append(deploymentOpt.Tolerations, t)
 			}
+		case k == "ownerrefs":
```
AkihiroSuda: To avoid inflation of the `switch{}`, can we rather support an arbitrary yq/jq expression to patch the manfiest?

scutuatua-crypto (approved):

scutuatua-crypto (approved):

driver/kubernetes/factory.go
abhay1999: Good point. I dropped the bespoke ownerrefs parsing and switched this to a generic manifest-patch driver option instead, so the owner reference case is now just a patch expression applied after the Kubernetes manifests are generated.

That keeps us from growing the switch further and makes the mechanism reusable for other manifest tweaks too.

I also added coverage for the patch path and reran go test ./driver/kubernetes/....

driver/kubernetes/manifest/manifest.go
AkihiroSuda, line 1: Maybe use https://github.com/itchyny/gojq to avoid maintaining a new equivalent of jq by ourselves

AkihiroSuda: Also, please squash the commits and update the PR description

abhay1999:
@AkihiroSuda I've addressed your feedback and pushed an updated version:

- Replaced the bespoke `ownerrefs` option with a generic `manifest-patch` driver option that accepts any jq expression
- Used `github.com/itchyny/gojq` as the jq engine as you suggested
- The expression is applied to the generated Deployment/StatefulSet after creation by marshaling to JSON, running the jq query, and unmarshaling back
- DCO is passing, all tests pass
- Squashed into a single commit and updated the PR description

Example usage for the original ownerReferences use case:
```
--driver-opt 'manifest-patch=.metadata.ownerReferences=[{"apiVersion":"actions.github.com/v1alpha1","kind":"EphemeralRunner","name":"runner-xyz","uid":"..."}]'
```

Could you take another look when you get a chance?

go.mod
```diff
@@ -155,6 +156,8 @@ require (
 	github.com/hiddeco/sshsig v0.2.0 // indirect
 	github.com/in-toto/attestation v1.1.2 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/itchyny/gojq v0.12.19 // indirect
```
AkihiroSuda: Doesn't seem indirect?

abhay1999: Good catch! Fixed — ran `go mod tidy` which correctly marks `github.com/itchyny/gojq` as a direct dependency. Pushed.

AkihiroSuda (dismissed):

abhay1999:
Friendly ping — @AkihiroSuda has approved the PR and all review threads are resolved. DCO, lint, and all integration tests are passing. The `lint-gopls` failures are pre-existing in files unrelated to this PR (`builder/`, `commands/`, `driver/remote/`) and are not caused by our changes (the `driver/kubernetes/` package passes `go vet` cleanly).

Ready to merge whenever convenient. 🙏

tonistiigi:
@AkihiroSuda What are the alternatives here? This JSON struct inside a key-value string, inside another key-value string is really ugly.

@abhay1999 Linter is failing because of your changes.

PR #3811, opened 2026-04-22
build: fix oci-layout named context serialization

### LINKED ISSUE #3793: Regression on digest-only `oci-layout://` build contexts
`docker buildx build --build-context` no longer accepts a digest-only OCI layout reference in recent versions.

This appears to be a regression in `v0.33.0`. The same kind of reference worked in earlier versions, and the [current documentation](https://docs.docker.com/reference/cli/docker/buildx/build/#source-oci-layout
) explicitly says an OCI layout source may be selected either by tag, or by digest:

## Repro

Given a valid OCI layout directory whose `index.json` contains a descriptor with digest:

```text
sha256:b58316e52c6138f698a0d273bef003fe7e190eef309284fe70e9443340a12c25
```

use it as a build context by digest only:

```bash
docker buildx build \
  --build-context proxy=oci-layout:///home/alpeb/tmp/eraseme@sha256:b58316e52c6138f698a0d273bef003fe7e190eef309284fe70e9443340a12c25 \
  .
```

The build fails with an invalid OCI layout reference:

```text
ERROR: failed to build: failed to solve: could not parse oci-layout reference "bd9j5c2yxngk19855mg3droh0:@sha256:b58316e52c6138f698a0d273bef003fe7e190eef309284fe70e9443340a12c25": invalid reference format
```

Note that using an explicit tag works around the problem.

## Likely cause

The regression seems consistent with the OCI-layout parsing/refactoring introduced around buildx `v0.33.0`, where digest-only refs can now retain an empty tag and later get re-serialized into an invalid `:<empty>@<digest>` form:
https://github.com/docker/buildx/commit/9894189361bc#diff-9e21a311718d1ef95e28d50d589895bb04cd41abf4149b0dbcca6b6b94e86bf9

crazy-max:
fixes #3793 
needs or closes #3812 

The build input path now parses the original OCI layout reference once, resolves the digest when needed, rewrites only the path to the temporary store ID, and serializes the result back through the shared `ocilayout.Ref` API. This keeps the named context flow from rebuilding the reference by hand and accidentally producing an empty tag segment.

The regression in #3793 came from the named context code path turning a digest-only OCI layout reference into an invalid `oci-layout://<store>:@sha256:...` form.

### DIFF
--- a/build/opt_test.go
+++ b/build/opt_test.go
@@ -1,10 +1,16 @@
 package build
 
 import (
+	"context"
 	"testing"
 
 	"github.com/docker/buildx/util/buildflags"
+	"github.com/docker/buildx/util/ocilayout"
+	"github.com/docker/buildx/util/progress"
 	"github.com/moby/buildkit/client"
+	"github.com/moby/buildkit/client/ociindex"
+	"github.com/opencontainers/go-digest"
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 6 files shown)


build/opt_test.go
```diff
@@ -154,3 +161,99 @@ func TestProxyArgKeyExists(t *testing.T) {
 		})
 	}
 }
+
+func TestLoadInputsOCILayoutNamedContext(t *testing.T) {
+	// FIXME: Use a relative temp dir here instead of t.TempDir(). On Windows,
+	//  t.TempDir() returns an absolute path with a drive-letter colon,
+	//  and this test feeds the path through an oci-layout:// ref where
+	//  that colon can be misread as the tag separator.
```
crazy-max: This one looks like a bug in `util/ocilayout` on Windows. I will open a follow-up PR to fix this.

crazy-max: Opened #3812

tonistiigi (approved):

PR #3812, opened 2026-04-22
ocilayout: handle Windows paths in local OCI layout refs

crazy-max:
follow-up https://github.com/docker/buildx/pull/3811#discussion_r3122955087

### DIFF
--- a/util/ocilayout/parse.go
+++ b/util/ocilayout/parse.go
@@ -34,7 +34,7 @@ func Parse(s string) (Ref, bool, error) {
 		}
 	}
 
-	if i := strings.LastIndex(localPath, ":"); i >= 0 {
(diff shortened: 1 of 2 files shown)


util/ocilayout/parse.go
```diff
@@ -51,10 +51,18 @@ func Parse(s string) (Ref, bool, error) {
...
 	}
 	if r.Digest != "" {
-		return s + "@" + r.Digest.String()
+		s += "@" + r.Digest.String()
 	}
 	return s
 }
+
+func isWindowsDrivePath(path string, colon int) bool {
+	if colon != 1 || len(path) < 2 {
```
tonistiigi, line 63: is this really a safe way to detect this? I can have one character subdirs in unix.

PR #3813, opened 2026-04-22
store: replace namesgenerator with neutral builder IDs

crazy-max:
This removes the remaining `github.com/docker/docker` dependency from Buildx by replacing the old `namesgenerator` usage in the store with neutral autogenerated builder names. The new format uses `builder_` plus a short `identity.NewID()` suffix, which keeps generated names predictable enough for current Buildx behavior without carrying the daemon-style human-readable naming scheme.

The existing driver-side runtime name derivation is unchanged, so autogenerated builders still flow through the current `buildx_buildkit_` prefixing path for derived node and container names. In practice that means a builder such as `builder_abc123def456` still becomes `builder_abc123def4560` at the node level and `buildx_buildkit_builder_abc123def4560` at the runtime object level, with Kubernetes continuing to normalize underscores to hyphens when deriving deployment names.

This is the smaller follow-up to the earlier work in #3354. The in-tree human-readable replacement there was not the right long-term direction. The concern raised in the earlier discussion was not just the dependency itself, but also that Buildx should not keep copying the daemon's naming style for builder identities. Using a neutral generated identifier keeps this change scoped to dependency removal, without coupling it to a broader redesign of builder presentation.

This PR intentionally doesn't try to solve the separate TOCTOU and ownership issues discussed in #3354 and the linked follow-ups. Buildx still reserves a builder name in the local store first and later derives node and container names from it, so there is still a window where conflicting objects can appear between reservation and container creation even though this is more unlikely with this change. So that needs explicit follow-up work around ownership labeling and verification before reuse or removal, and it should be handled separately.


store/util.go
```diff
@@ -5,12 +5,16 @@ import (
...
-	"github.com/docker/docker/pkg/namesgenerator"
+	"github.com/moby/buildkit/identity"
 	"github.com/pkg/errors"
 )
 
 var namePattern = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9\.\-_]*$`)
 
+const generatedNamePrefix = "builder_"
+
+var newID = identity.NewID
```
thaJeztah, line 16: Wondering if we need the complexity of `identity` here; would something like this give enough randomness?

```go
import (
	"math/rand/v2"
	"encoding/hex"
)

func NewID() string {
	var b [8]byte // 64 bits
	rand.Read(b[:])
	return hex.EncodeToString(b[:]) // 16 hex chars
}
```

thaJeztah: Oh; and I just recalled I did something similar for k8s some time ago; https://github.com/docker/buildx/blob/d5d3d3d5020d9d55719a552071c189d6318f4454/driver/kubernetes/util/generate.go#L56-L75

store/util.go
```diff
@@ -5,12 +5,16 @@ import (
...
 	"strings"
 
-	"github.com/docker/docker/pkg/namesgenerator"
+	"github.com/moby/buildkit/identity"
 	"github.com/pkg/errors"
 )
 
 var namePattern = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9\.\-_]*$`)
 
+const generatedNamePrefix = "builder_"
```
thaJeztah, line 14: As we're changing, might as well use `-` instead of `_` (which should produce a more correct hostname for the container if there's ever a need for that).

store/util.go
```diff
@@ -5,12 +5,16 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/docker/docker/pkg/namesgenerator"
+	"github.com/moby/buildkit/identity"
 	"github.com/pkg/errors"
 )
 
 var namePattern = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9\.\-_]*$`)
```
thaJeztah, line 12: Not for this PR probably , but looks like we have some different patterns in different packages;

https://github.com/docker/buildx/blob/09f288c71e8b3f0fb748e1a5bc055fae7221d62b/bake/bake.go#L40-L41


I know these regexes are rather memory-hungry; we could even rewrite these to a basic function to loop; I recall we did similar changes elsewhere, e.g.; https://github.com/moby/moby/blob/dff719e3674958407416fd1d8a35db998f128da2/daemon/internal/stringid/stringid.go#L62-L70

store/util.go
crazy-max: I used `builder_` intentionally because Buildx already derives runtime names through the `buildx_buildkit_` prefix, so underscores are already part of the existing naming convention here. And it's a bit easier to read and copy/paste consistently alongside the derived node and runtime names.

jsternberg (approved):

crazy-max:
> This PR intentionally doesn't try to solve the separate TOCTOU and ownership issues discussed in #3354 and the linked follow-ups. Buildx still reserves a builder name in the local store first and later derives node and container names from it, so there is still a window where conflicting objects can appear between reservation and container creation even though this is more unlikely with this change. So that needs explicit follow-up work around ownership labeling and verification before reuse or removal, and it should be handled separately.

Discussed offline with @tonistiigi about this one and this is on purpose as a container builder might be created manually with `docker create` and reused later with `docker buildx create` if it already exists.

thaJeztah (approved):
LGTM

egorman-sray:
hey @crazy-max and @thaJeztah, would you consider undrafting and merging this PR?

kostyay:
Hey
Is this pr still being considered? Would be great to have it merged

PR #3833, opened 2026-05-06
warn on duplicate cache-to targets

### LINKED ISSUE #3819: Bake should not allow multiple cache outputs of the same name
### Description

Currently, bake allows for multiple targets to output to the same cache target. When you do so, the result is non-deterministic and mostly results in your build never caching properly, and it being a total mystery why. I've done this two different ways:
 1. By accident, having a somewhat complex bake file/Dockerfile, and not noticing I did a bad copy paste on a cache name
 2. In a Multiplatform build (see https://github.com/docker/buildx/discussions/1382) This one might already be fixed, or may only be an issue with GHA output, I'm still debugging some cache busting issues. If it's supposed to be fixed, adding validation that we never output cache targets of the same name would be neat.

I'd also say that probably one shouldn't be allowed to name two regular output targets of the same name in the same build, but I haven't actually tried to see if bake lets you.

Is there an actual use case for this? Otherwise it seems like a footgun, and should error. If there's a desire not to break existing "working" bakefiles, could go with a warning for some number of releases, first.

I've been working a lot with various docker internals for my work on clipper.dev, happy to put up a PR to fix this if it would be accepted.

areebahmeddd:
## summary

closes #3819

* adds a warning when multiple build targets try to use the same cache destination

## ai model disclosure

used vscode copilot (claude sonnet 4.6) to understand the issue context and help generate a fix. changes were self-reviewed and verified via:

```bash
make validate-all
go build ./...
```

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -32,6 +32,7 @@ import (
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/frontend/dockerfile/dfgitutil"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/convert"
 )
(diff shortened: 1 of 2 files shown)


tonistiigi (commented):
If added, this should be based on some primary values, eg. `type=local,dest=`, `type=registry,ref=`. Eg. if same ref is used with different mode then it is still a collision. And we shouldn't use the potentially secret values for comparison or log them anyway.

Does this work with target inheritance?

areebahmeddd:
> we shouldn't use the potentially secret values for comparison or log them anyway.

yea sure, i think using a helper func here is enough for comparing the destination key. creds/mode are ignored

> Does this work with target inheritance?

yes, this runs after inheritance gets resolved so shared inherited cache-to values still trigger the warning

areebahmeddd:
gentle ping! @tonistiigi @crazy-max 🙂

PR #3834, opened 2026-05-06
vendor: update buildkit to v0.30.0-rc1

tonistiigi:


cmd/buildx/tracing.go
```diff
@@ -4,7 +4,8 @@ import (
 	"github.com/moby/buildkit/util/tracing/detect"
 	"go.opentelemetry.io/otel"
 
-	_ "github.com/moby/buildkit/util/tracing/env"
+	_ "github.com/moby/buildkit/util/tracing/childprocess"
```
tonistiigi, line 7: @jsternberg please check this

tonistiigi: I removed Jaeger import as it was including some new big packages but not 100% this is compatible with old version atm.

jsternberg (approved):

cmd/buildx/tracing.go
jsternberg: I don't see the jaeger import you're mentioning here. This change from env to childprocess is correct though.

tonistiigi: Initially I had a "github.com/moby/buildkit/util/tracing/detect/jaeger" import here as well but it caused ~25k lines worth of new packages to be imported.

Issue #3849, opened 2026-05-13
[v0.34] buildx bake fails with "cache_to must be a array" when cache_to is an empty []


rushikeshmagar492:
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [ ] ... the documentation does not mention anything about my problem
- [ ] ... there are no open or closed issues that are related to my problem

### Description

When using docker buildx bake with a Compose file where cache_to is explicitly defined as an empty array ([]), the build fails with a schema validation error.

Buildx version: 0.34.0

### Expected behaviour

The bake process should treat cache_to: [] as a valid instruction to not export any cache, or ignore the empty list, and proceed with the build.

### Actual behaviour

ERROR: services.test-buildx.build.cache_to must be a array

### Buildx version

github.com/docker/buildx v0.34.0 3e73561e39785683b31b05eeab1ef645be44ca42

### Docker info

```text
Client: Docker Engine - Community
 Version:    29.4.0
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.34.0
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v5.1.3
    Path:     /usr/libexec/docker/cli-plugins/docker-compose
```

### Builders list

```text
NAME/NODE       DRIVER/ENDPOINT   STATUS    BUILDKIT   PLATFORMS
default*        docker                                 
 \_ default      \_ default       running   v0.29.0    linux/amd64 (+3)
```

### Configuration

Steps to Reproduce
1) Install buildx version 0.34.0
2) Create a docker-compose.yml with cache_to []:
e.g.
services:
  test-buildx:
    build:
      context: .
      cache_from:
        - type=registry,ref=<REDACTED>.dkr.ecr.us-east-1.amazonaws.com/app-cache:arm64
      cache_to: []
      
3) Run the bake command:
docker buildx bake --load test-buildx

ERROR: services.test-buildx.build.cache_to must be a array

### Build logs

```text

```

tonistiigi:
@jsternberg Is this coming from compose or from us?

rushikeshmagar492:
[...]

PR #3852, opened 2026-05-15
bake: preserve empty compose cache lists

tonistiigi:
ref https://github.com/docker/buildx/issues/3849

Workaround for regression in compose-go.

Normalize filtered Compose models before schema validation so empty YAML lists remain arrays through compose-go's JSON validation round trip.


crazy-max (approved):

crazy-max:
looks good for patch release while waiting for compose-go release with the fix https://github.com/compose-spec/compose-go/pull/873

PR #3858, opened 2026-05-18
fix rootless buildkit memory volume path

### LINKED ISSUE #3422: Kubernetes Rootless mode doesn't honor buildkit-root-volume-memory
### Contributing guidelines

### Description

BuildKit in Kubernetes rootless mode does not use RAM disk for builds when requested.
This code should honor the buildkit-root-volume-memory option and create a volume with "Medium" set to "Memory":

https://github.com/docker/buildx/blob/e540b7af14a68a2c0052774988156bc59384bb35/driver/kubernetes/manifest/manifest.go#L302-L305

### Expected behaviour

When buildkit-root-volume-memory is configured, builds should occur on RAM disk, even in rootless mode.

### Actual behaviour

When running BuildKit in rootless mode, /var/lib/buildkit is not used. Instead, data is stored in the non-root user's home directory. This causes image builds to hit the file system instead of memory, even when `buildkit-root-volume-memory` is set.

### Buildx version

github.com/docker/buildx v0.27.0 bac71de

### Docker info

```text

```

### Builders list

```text
kubernetes
```

### Configuration

```
```

### Build logs

```text

```

### Additional info

_No response_

areebahmeddd:
## summary

closes #3422

* in rootless mode, buildkit stores data in `/home/user/.local/share/buildkit` instead of `/var/lib/buildkit`. the `buildkit-root-volume-memory` option was using the wrong path, so builds still used the filesystem even when ram storage was enabled.

## ai model disclosure

used vscode copilot (claude sonnet 4.6) to understand the issue context and help generate a fix. changes were self-reviewed and verified via:

```bash
make validate-all
go build ./...
```

### DIFF
--- a/driver/kubernetes/manifest/manifest.go
+++ b/driver/kubernetes/manifest/manifest.go
@@ -206,7 +206,7 @@ func NewDeployment(opt *DeploymentOpt) (d *appsv1.Deployment, s *appsv1.Stateful
 	}
 
 	if opt.Rootless {
-		if err := toRootless(&podTemplate); err != nil {
+		if err := toRootless(&podTemplate, opt.BuildKitRootVolumeMemory); err != nil {
 			return nil, nil, nil, err
 		}
 	}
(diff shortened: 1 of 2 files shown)


thompson-shaun:
DCO is missing @areebahmeddd

areebahmeddd:
weird. i do use gpg signing automatically while committing but i force pushed my changes with the -s flag @thompson-shaun

AkihiroSuda (approved):

PR #3861, opened 2026-05-19
route bake targets across pods with random loadbalance

### LINKED ISSUE #3382: Bake targets are not distributed across buildkit pods
### Contributing guidelines

### Description

Hi, I'm running docker buildx bake with around 40 targets. 
I'm creating a multi node builder, and I've noticed that all targets are being built on the same node. 
Is there a way to spread the load across the nodes?

### Expected behaviour

Expected the targets to spread across the pods

### Actual behaviour

All of the targets are being built on the same pod

### Buildx version

github.com/docker/buildx v0.25.0 Homebrew

### Docker info

```text
 Version:    28.1.1
 Context:    desktop-linux
 Debug Mode: false
 Plugins:
    Version:  v0.25.0
    Path:     /opt/homebrew/lib/docker/cli-plugins/docker-buildx
    Version:  v2.35.1-desktop.1
    Path:     /Users/morerel/.docker/cli-plugins/docker-compose
Server:
 Containers: 27
  Running: 0
  Paused: 0
  Stopped: 27
 Images: 125
 Server Version: 28.1.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 05044ec0a9a75232cad458027ca83437aae3f4da
 init version: de40ad0
 Security Options:
  seccomp
   Profile: unconfined
  cgroupns
 Kernel Version: 6.10.14-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: aarch64
 CPUs: 11
 Total Memory: 7.654GiB
 Name: docker-desktop
 ID: b7c74a23-00e4-4a33-b9c8-7db46ef250cf
 Debug Mode: false
 HTTP Proxy: http.docker.internal:3128
 HTTPS Proxy: http.docker.internal:3128
 No Proxy: hubproxy.docker.internal
 Labels:
  com.docker.desktop.address=unix:///Users/morerel/Library/Containers/com.docker.docker/Data/docker-cli.sock
 Experimental: false
 Insecure Registries:
  hubproxy.docker.internal:5555
  ::1/128
  127.0.0.0/8
 Live Restore Enabled: false
```

### Builders list

```text
docker buildx ls
NAME/NODE                           DRIVER/ENDPOINT                                                                                     STATUS    BUILDKIT   PLATFORMS
kube*                               kubernetes
 \_ kube0                            \_ kubernetes:///kube?deployment=buildkit-9ece06e2-06e4-4ca5-9075-c44c7cc27c6a-202c1&kubeconfig=   running   v0.23.2    linux/amd64 (+4), linux/386
 \_ kube0                            \_ kubernetes:///kube?deployment=buildkit-9ece06e2-06e4-4ca5-9075-c44c7cc27c6a-202c1&kubeconfig=   running   v0.23.2    linux/amd64 (+4), linux/386
default                             docker
 \_ default                          \_ default                                                                                         running   v0.21.0    linux/amd64 (+2), linux/arm64, linux/arm (+2), linux/ppc64le, (3 more)
desktop-linux                       docker
 \_ desktop-linux                    \_ desktop-linux                                                                                   running   v0.21.0    linux/amd64 (+2), linux/arm64, linux/arm (+2), linux/ppc64le, (3 more)
```

### Configuration

```
// docker-bake.hcl
group "default" {
  targets = ["imageA", "imageB", "imageC", "imageD"]
}

target "imageA" { 
    context = "imageA-dir"
    dockerfile = "Dockerfile" 
    tags = ["image:latest"]
}

target "imageB" { 
    context = "imageB-dir"
    dockerfile = "Dockerfile" 
    tags = ["image:latest"]
}

target "imageC" { 
    context = "imageC-dir"
    dockerfile = "Dockerfile" 
    tags = ["image:latest"]
}

target "imageD" { 
    context = "imageD-dir"
    dockerfile = "Dockerfile" 
    tags = ["image:latest"]
}
```

### Build logs

```text

```

### Additional info

_No response_

areebahmeddd:
`loadbalance=random` on the kubernetes driver was sending all bake targets to the same pod.
the client connection was cached after the first dial, so pod selection only happened once.

fix this by creating a new connection per target when `loadbalance=random` is enabled.
also disable the shared local mount session optimization in this mode.

Closes #3382

### DIFF
--- a/driver/kubernetes/factory.go
+++ b/driver/kubernetes/factory.go
@@ -162,6 +162,7 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
 			StatefulSet: d.statefulSet,
 		}
 	}
+	d.loadbalance = loadbalance
 	return d, nil
 }
 
--- a/driver/manager.go
+++ b/driver/manager.go
@@ -128,6 +128,17 @@ func (d *DriverHandle) Client(ctx context.Context, opt ...client.ClientOpt) (*cl
 	return d.client, d.err
 }
 
+func (d *DriverHandle) UncachedClient(ctx context.Context) (*client.Client, error) {
+	return d.Driver.Client(ctx, d.getClientOptions()...)
+}
+
+func (d *DriverHandle) RequiresUncachedClient() bool {
+	if p, ok := d.Driver.(UncachedClientDriver); ok {
+		return p.RequiresUncachedClient()
+	}
+	return false
+}
+
 func (d *DriverHandle) getClientOptions() []client.ClientOpt {
 	return []client.ClientOpt{
 		client.WithTracerDelegate(delegated.DefaultExporter),
--- a/build/build.go
+++ b/build/build.go
@@ -1030,7 +1030,11 @@ func detectSharedMounts(ctx context.Context, reqs map[string][]*reqForNode) (_ m
 	m := map[string]map[fsKey]*fsTracker{}
 	for _, reqs := range reqs {
 		for _, req := range reqs {
-			nodeName := req.ResolvedNode.Node().Name
+			nodeName := req.Node().Name
+			// skip shared-session optimisation: targets may connect to different replicas.
+			if req.Node().Driver != nil && req.Node().Driver.RequiresUncachedClient() {
+				continue
+			}
 			if _, ok := m[nodeName]; !ok {
 				m[nodeName] = map[fsKey]*fsTracker{}
 			}
--- a/build/resolver/driver.go
+++ b/build/resolver/driver.go
@@ -57,6 +57,14 @@ func (dp ResolvedNode) Platforms() []ocispecs.Platform {
 }
 
 func (dp ResolvedNode) Client(ctx context.Context) (*client.Client, error) {
+	node := dp.resolver.nodes[dp.driverIndex]
+	// loadbalance=random requires a fresh connection per call so each target lands on a different pod.
+	if node.Driver != nil && node.Driver.RequiresUncachedClient() {
+		if _, err := dp.resolver.boot(ctx, []int{dp.driverIndex}, nil); err != nil {
+			return nil, err
+		}
+		return node.Driver.UncachedClient(ctx)
+	}
 	clients, err := dp.resolver.boot(ctx, []int{dp.driverIndex}, nil)
 	if err != nil {
 		return nil, err
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -57,6 +57,10 @@ type Info struct {
 	DynamicNodes []store.Node
 }
 
+type UncachedClientDriver interface {
+	RequiresUncachedClient() bool
+}
+
 type Driver interface {
 	Factory() Factory
 	Bootstrap(context.Context, progress.Logger) error
--- a/driver/kubernetes/driver.go
+++ b/driver/kubernetes/driver.go
@@ -45,6 +45,7 @@ type Driver struct {
 	// if you add fields, remember to update docs:
 	// https://github.com/docker/docs/blob/main/content/build/drivers/kubernetes.md
 	minReplicas       int
+	loadbalance       string
 	deployment        *appsv1.Deployment
 	statefulSet       *appsv1.StatefulSet
 	configMaps        []*corev1.ConfigMap
@@ -59,6 +60,10 @@ type Driver struct {
 
 func (d *Driver) IsMobyDriver() bool {
 	return false
+}
+
+func (d *Driver) RequiresUncachedClient() bool {
+	return d.loadbalance == LoadbalanceRandom
 }
 
 func (d *Driver) Config() driver.InitConfig {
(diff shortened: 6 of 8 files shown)


AkihiroSuda:
> `loadbalance=random` on the kubernetes driver was sending all bake targets to the same pod. the client connection was cached after the first dial, so pod selection only happened once.
> 
> fix this by creating a new connection per target when `loadbalance=random` is enabled.

What about loadbalance=sticky ?

AkihiroSuda:
Also, please squash the commits

areebahmeddd:
> What about loadbalance=sticky ?

no change. it reuses the same shared connection path

> Also, please squash the commits

sure 🙂

AkihiroSuda (dismissed):

crazy-max (dismissed):

driver/manager.go
```diff
@@ -128,6 +128,20 @@ func (d *DriverHandle) Client(ctx context.Context, opt ...client.ClientOpt) (*cl
...
+
+func (d *DriverHandle) NeedsNewClientPerBuild() bool {
+	type perBuildClientDriver interface {
+		NeedsNewClientPerBuild() bool
+	}
+	if p, ok := d.Driver.(perBuildClientDriver); ok {
+		return p.NeedsNewClientPerBuild()
+	}
+	return false
+}
```
crazy-max: Actually after new review this doesn't feel like the right shape for the driver API. `NeedsNewClientPerBuild()` is added as an optional hidden interface inside `DriverHandle`, but the behavior is broader than the name says. The build code also uses it to disable shared local mount sessions.

Can we make that contract explicit near the `Driver` interface instead of hiding it in `manager.go`? Something like an optional `RequiresUncachedClient()` interface would make the build-side behavior easier to reason about. I would also rename `NewClient()` to something like `UncachedClient()` so it is clear this is deliberately bypassing `DriverHandle.Client()` caching, not just another constructor.

crazy-max (commented):
I was looking at https://github.com/docker/buildx/blob/0b358c76374b66d82daaf179a21104c574711e82/driver/kubernetes/podchooser/podchooser.go#L29-L45

Seems to still looks capable of picking the same pod for all targets in one bake no? `ChoosePod()` creates a new RNG on every call and seeds it with `time.Now().Unix()`. With this PR, multiple targets can create fresh clients and dial within the same second, so each call can get the same seed and produce the same first random value iiuc.

Maybe we should seed the RNG once at the package level, with concurrency handled, or use another safe random selection path? Otherwise this PR may still reproduce the original issue.

driver/manager.go
areebahmeddd: > Maybe we should seed the RNG once at the package level, with concurrency handled, or use another safe random selection path? Otherwise this PR may still reproduce the original issue.

yea, re-seeding in ChoosePod() could lead to identical selections for concurrent dials. i'll switch it to use a shared RandSource that's seeded once and reused

> Can we make that contract explicit near the Driver interface instead of hiding it in manager.go? Something like an optional RequiresUncachedClient() interface would make the build-side behavior easier to reason about. I would also rename NewClient() to something like UncachedClient() so it is clear this is deliberately bypassing DriverHandle.Client() caching, not just another constructor.

okay, i've moved it next to the Driver interface as an optional RequiresUncachedClient() interface

areebahmeddd:
gentle ping! @crazy-max @AkihiroSuda 🙂

AkihiroSuda:
> gentle ping! @crazy-max @AkihiroSuda 🙂

https://github.com/docker/buildx/pull/3861#issuecomment-4561738057

areebahmeddd:
ah yes, done : )
cc @AkihiroSuda

AkihiroSuda (dismissed):

driver/kubernetes/factory.go
```diff
@@ -157,11 +158,13 @@ func (f *factory) New(ctx context.Context, cfg driver.InitConfig) (driver.Driver
 		}
 	case LoadbalanceRandom:
 		d.podChooser = &podchooser.RandomPodChooser{
+			RandSource:  rand.NewSource(time.Now().UnixNano()), // #nosec G404 -- no strong seeding required
```
crazy-max: I think this still needs one small fix before merge. `RandomPodChooser` now stores a shared `rand.Source`, but `math/rand.Source` is not safe for concurrent use. In bake, the uncached clients can dial from concurrent build/session/status goroutines, so multiple `ChoosePod()` calls can race on the same source.

Can we either guard the RNG with a mutex, store a `*rand.Rand` behind a lock, or use another concurrency-safe random selection path here? Otherwise the fix trades the old same-seed issue for a data race in the random load-balancer path.

areebahmeddd: fixed. removed `RandSource` from `RandomPodChooser` entirely and switched to `rand.Intn()` directly. the global rand functions in Go 1.20+ use a locked source internally, so concurrent `ChoosePod()` calls are safe without any additional mutex

edit: ci failure is not related to this PR

crazy-max (approved):
LGTM thanks

PR #3872, opened 2026-05-28
build: add --validate=repro reproducibility check

### LINKED ISSUE #2053: Add `--twice` (or `--ntimes=N`) flag to test reproducibility
### Description

The `--twice` (or `--ntimes=N`) flag will run a build twice, and bail if the digests do not match.
The second run will automatically run with `--no-cache` to disable layer caches (but keep cached mounts).

The `SOURCE_DATE_EPOCH` will be set to the time of the command execution if not explicitly set.

lohitkolluri:
## Summary
- Add an extensible `--validate` flag and implement `--validate=repro`.
- Repro validation runs the build twice and fails if the resulting image digest differs.
- The second run forces `--no-cache`.
- If `SOURCE_DATE_EPOCH` is not explicitly set, it is pinned for both runs to make the digest comparison meaningful.

## Test plan
- `go test ./commands/...`
- `go build ./cmd/buildx`

Closes #2053

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"maps"
 	"os"
 	"path/filepath"
 	"slices"
@@ -19,6 +20,7 @@ import (
 
 	"github.com/containerd/console"
 	"github.com/containerd/containerd/v2/pkg/epoch"
+	"github.com/containerd/platforms"
 	"github.com/docker/buildx/build"
 	"github.com/docker/buildx/builder"
 	"github.com/docker/buildx/store"
@@ -29,6 +31,7 @@ import (
 	"github.com/docker/buildx/util/desktop"
 	"github.com/docker/buildx/util/dockerutil"
 	"github.com/docker/buildx/util/dockerutil/dockerconfig"
+	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/buildx/util/ioset"
 	"github.com/docker/buildx/util/metricutil"
 	"github.com/docker/buildx/util/osutil"
@@ -61,6 +64,8 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/protobuf/proto"
+
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 type buildOptions struct {
@@ -95,6 +100,7 @@ type buildOptions struct {
 
 	progress string
 	quiet    bool
+	validate []string
 
 	builder      string
 	metadataFile string
@@ -318,6 +324,11 @@ func runBuild(ctx context.Context, dockerCli command.Cli, debugOpts debuggerOpti
 		return err
 	}
 
+	validate, err := parseValidateFlags(options.validate)
+	if err != nil {
+		return err
+	}
+
 	// Avoid leaving a stale file if we eventually fail
 	if options.imageIDFile != "" {
 		if err := os.Remove(options.imageIDFile); err != nil && !os.IsNotExist(err) {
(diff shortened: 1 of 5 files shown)


commands/build.go
```diff
@@ -488,6 +548,46 @@ func runBuildWithOptions(ctx context.Context, dockerCli command.Cli, opts *Build
...
+			return validateFlags{}, errors.Errorf("unknown --validate value %q (supported: repro)", v)
+		}
+	}
+	return out, nil
+}
+
+func getReproDigest(resp map[string]string) (string, error) {
+	// Prefer the image manifest digest key. This is the digest we want to compare
+	// for reproducibility (not the config digest).
+	if dgst := resp[exptypes.ExporterImageDigestKey]; dgst != "" {
```
AkihiroSuda, line 666: Isn't this an index digest?
Indices are not expected to be reproducible as they may contain SLSA provenances

commands/build.go
```diff
@@ -396,9 +402,63 @@ func runBuild(ctx context.Context, dockerCli command.Cli, debugOpts debuggerOpti
...
+		dgst1, err := getReproDigest(resp1.ExporterResponse)
+		if err != nil {
+			done(err)
+			return err
+		}
+
+		opts2 := cloneBuildOptions(opts)
+		opts2.NoCache = true
+
+		// Second run is the "real" run whose output and artifacts we keep.
```
AkihiroSuda, line 451: Should sleep for one second at least?

commands/build.go
lohitkolluri: You’re absolutely right, thanks for catching that. I’ve patched --validate=repro to detect an index/list via containerimage.descriptor and reject it with guidance to use a single platform / disable attestations, instead of comparing the index digest.

commands/build.go
lohitkolluri: Added a 1s sleep between the two repro runs to avoid same-second time-derived fields affecting the check.

commands/build.go
```diff
@@ -488,6 +554,64 @@ func runBuildWithOptions(ctx context.Context, dockerCli command.Cli, opts *Build
...
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to decode %q", exptypes.ExporterImageDescriptorKey)
+		}
+		var desc ocispecs.Descriptor
+		if err := json.Unmarshal(descJSON, &desc); err != nil {
+			return "", errors.Wrapf(err, "failed to parse %q", exptypes.ExporterImageDescriptorKey)
+		}
+		switch desc.MediaType {
+		case ocispecs.MediaTypeImageIndex, "application/vnd.docker.distribution.manifest.list.v2+json":
+			return "", errors.Errorf("reproducibility validation does not support multi-platform index digest; use a single platform (e.g. --platform=linux/amd64) or disable attestations")
```
AkihiroSuda: Why not compare the manifests for the each of platforms in a multi-platform index?

commands/build.go
AkihiroSuda: > You’re absolutely right, thanks for catching that.

Didn't you catch it in testing?
How did you test this PR?

lohitkolluri: Initial testing was go test ./commands/... + go build ./cmd/buildx and a quick single-platform build; I didn’t test a multi-platform index output, so I missed that containerimage.digest can be an index digest. I’ve updated --validate=repro to compare per-platform manifest digests from the resolved index content instead.

commands/build.go
lohitkolluri: I’ve updated --validate=repro so that for multi-platform index/manifest-list outputs we compare the per-platform manifest digests (sorted platform=digest pairs) instead of the index digest, which can legitimately vary (e.g. provenance/referrers).

lohitkolluri:
I dug into the CI failures and they were mostly a single root cause fanning out across the matrix, plus one integration panic and docs drift.

 - Fixed `lint-gopls` by switching the `BuildArgs` clone to `maps.Copy` (the modernize analyzer was failing the job).
 - Fixed the integration failure by preventing a `send on closed channel` panic in the progress `Printer` during shutdown (writes are now best-effort when closing).
 - Updated the generated docs to include the new `--validate` flag.
 - Made `validate-docs` more deterministic by comparing the generated output to the checked-in docs directly (so failures show a real diff instead of a brittle git-status signal).
 - Normalized one path expectation in `bake/entitlements_test.go` for macOS `/private` path canonicalization.

 Locally verified: `make lint`, `make lint-gopls`, `make validate-docs`, and `go test ./... -short`.

AkihiroSuda:
> I dug into the CI failures and they were mostly a single root cause fanning out across the matrix, plus one integration panic and docs drift.

Thanks, but please submit separate PRs

lohitkolluri:
Per review feedback, I split the follow-up CI fixes into separate PRs and kept this PR focused on `--validate=repro` (+ docs + the small gopls modernize fix).

Split PRs:
- progress shutdown panic fix: https://github.com/docker/buildx/pull/3873
- validate-docs determinism + macOS /private path normalization: https://github.com/docker/buildx/pull/3874

lohitkolluri:
@AkihiroSuda sorry for the trouble. Could you please approve the CI when you get a chance?

lohitkolluri:
Looks like the lint checks are failing across all platforms. I’ll patch the issues and push a follow-up commit shortly.

lohitkolluri:
@AkihiroSuda I tracked down the CI failures, it turned out one file just needed a quick formatting cleanup. I fixed that and pushed an updated (squashed) commit to this PR. CI should be Green now.

lohitkolluri:
@AkihiroSuda looks like CI is passing on all 3 PR's

crazy-max:
Thanks for working on this, but we need to pause this PR for now.

The proposal in #2053 is still being discussed afaik. We have not yet aligned on whether this belongs in the Buildx CLI, whether it should live elsewhere such as the GitHub Action, or what the final flag/API shape should be.

Per our [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md#successful-changes), this kind of large or high-impact changes should be coordinated with maintainers before submitting an implementation PR. Adding a new `buildx build` flag and reproducibility validation behavior falls into that category.

Please continue the design discussion in #2053 and wait for maintainer agreement on the scope and UX before opening or continuing implementation PRs. Otherwise we spend maintainer review and CI time on code that may not be accepted.

I'm going to close this PR for now. Once the proposal is approved and the expected behavior is clearly defined, a focused implementation PR would be welcome.

lohitkolluri:
Understood that this needs maintainer alignment first. I’ll stop work on this PR and continue the design discussion in issue #2053 . Once there’s agreement on the expected behavior, I’m happy to come back with a smaller, focused implementation PR that matches it.

Issue #1176, opened 2022-06-21
Bake: panic while baking


ciaranmcnulty:
Unfortunately I didn't have progress=plain so it's a little garbled, but this was during a fairly parallelised build:


```
 => [php-with-codebase 7/7] COPY --link --from=composer-output vendor vendor                                                                                                                                                                                                                                       3.0s
panic: send on closed channel                                                                                                                                                                                                                                                                                      3.0s
 => => merging                                                                                                                                                                                                                                                                                                     3.0s
 => CACHED [pax-lookup-api composer-installer 3/3] RUN --mount=type=cache,target=/root/.composer     --mount=source=snowfall,target=./snowfall     --mount=from=composer:1,source=/usr/bin/composer,target=/usr/bin/composer     if [ "false" = "false" ]; then        composer validate && composer install --no  0.0s
goroutine 13959 [running]:okup-api 1/8] RUN a2enmod rewrite                                                                                                                                                                                                                                                        0.5s
 => [pax-lookup-tester pax-lookup-tester 1/5] RUN --mount=type=cache,target=/var/cache/apt,sharing=private   curl -sL https://deb.nodesource.com/setup_16.x | bash -   && apt install -y npm git zip                                                                                                              44.0s
github.com/docker/buildx/util/progress.(*Printer).Write(0x140006d01e0, 0x140005cb1a0)                                                                                                                                                                                                                              0.0s
 => CACH/src/util/progress/printer.go:41 +0x3c config config                                                                                                                                                                                                                                                       0.0s
github.com/docker/buildx/util/progress.(*prefixed).Write(0x140001524b0, 0x140005cb1a0)                                                                                                                                                                                                                             0.0s
 => CACH/src/util/progress/multiwriter.go:29 +0x60 web                                                                                                                                                                                                                                                             0.0s
github.com/docker/buildx/util/progress.FromReader({0x104be2d90, 0x140001524b0}, {0x1044348dc, 0x13}, {0x104be12a8, 0x1400073a2c0})                                                                                                                                                                                 0.0s
 => CACH/src/util/progress/fromreader.go:35 +0x448mposer.json composer.lock ./                                                                                                                                                                                                                                     0.0s
github.com/docker/buildx/build.newDockerLoader.func1()mposer-installer /pax/vendor vendor                                                                                                                                                                                                                          0.0s
 => CACH/src/build/build.go:1368 +0x294 --link --from=composer-output vendor vendor                                                                                                                                                                                                                                0.0s
[...]

PR #3873, opened 2026-05-28
progress: avoid panic when closing printer

lohitkolluri:
## Summary
- Prevent `send on closed channel` panics in progress `Printer` shutdown paths by making status writes best-effort during close.

## Test plan
- `go test ./util/progress -short`

### DIFF
--- a/util/progress/printer.go
+++ b/util/progress/printer.go
@@ -29,9 +29,11 @@ type Printer struct {
 	mode progressui.DisplayMode
 	opt  *printerOpts
 
-	status    chan *client.SolveStatus
-	interrupt chan interruptRequest
-	state     printerState
+	status       chan *client.SolveStatus
+	statusMu     sync.Mutex
+	statusClosed bool
+	interrupt    chan interruptRequest
+	state        printerState
 
 	done      chan struct{}
 	closeOnce sync.Once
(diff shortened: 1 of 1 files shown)


crazy-max (changes requested):
Thanks for the patch. Before changing the `Printer` write semantics, can you open an issue with a current reproduction and link it here?

Right now this PR doesn't show the lifecycle that can call `Write` after `Wait` has closed `p.status`, and there is no regression test. The only matching report I found is #1176, which is from 2022 and was closed as stale/possibly SIGTERM-related.

This patch also changes more than shutdown behavior: `Write` becomes best-effort and can drop progress updates whenever the buffer is full, not only while closing. That can lose logs/status/warnings during normal operation.

If this panic is still reproducible, please include the command, buildx/BuildKit versions, progress mode, a minimized Dockerfile or bake file if possible, and the full stack trace. Then we can add a targeted failing test and fix the actual lifecycle race rather than making all progress writes lossy.

lohitkolluri:
Opened an issue with a current reproduction + full stack trace here: https://github.com/docker/buildx/issues/3875

It reproduces in CI under `test-integration (v0.28.1, remote, ./tests, experimental)` and points at a lifecycle race where `Write()` can run after `Wait()` has closed `p.status`.

Next step on my side: add a targeted regression test for that specific lifecycle and adjust the shutdown/lifecycle handling, rather than making all progress writes lossy in normal operation.

lohitkolluri:
Linking for tracking: https://github.com/docker/buildx/issues/3875

lohitkolluri:
Hey,

Just checking in on this PR. Given the discussion in #3875, would you prefer that we keep this PR open, or would it be better to close it for now?

Thanks!

crazy-max:
I think we can close it for now and revisit if it happens in the future

PR #3887, opened 2026-06-05
ci: upload test results to codecov

crazy-max:
This updates the build workflow so Codecov receives test results in addition to coverage reports for both integration and unit test jobs.

Uploading test results alongside coverage gives Codecov the data it needs to show test reporting without changing how coverage is collected or annotated in the existing CI flow.

> Test Analytics offers data on test run times, failure rates, and identifies flaky tests to help decrease the risk of deployment failures and make it easier to ship new features quickly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,14 +148,24 @@ jobs:
           TEST_REPORT_SUFFIX: "-${{ env.TESTREPORTS_NAME }}"
           TESTPKGS: "${{ matrix.pkg }}"
       -
-        name: Send to Codecov
+        name: Upload coverage to Codecov
         if: always()
         uses: codecov/codecov-action@e79a6962e0d4c0c17b229090214935d2e33f8354 # v6.0.1
         with:
           directory: ./bin/testreports
           flags: integration
           token: ${{ secrets.CODECOV_TOKEN }}
-          disable_file_fixes: true
+          disable_file_fixes: true # needed for Go language https://docs.codecov.com/docs/fixing-reports
+      -
+        name: Upload test results to Codecov
+        if: always()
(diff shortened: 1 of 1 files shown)


crazy-max:
ci issue unrelated, see #3888

vvoland (approved):

.github/workflows/build.yml
```diff
@@ -148,14 +148,24 @@ jobs:
...
+          disable_file_fixes: true # needed for Go language https://docs.codecov.com/docs/fixing-reports
+      -
+        name: Upload test results to Codecov
+        if: always()
+        uses: codecov/codecov-action@e79a6962e0d4c0c17b229090214935d2e33f8354 # v6.0.1
+        with:
+          report_type: test_results
+          directory: ./bin/testreports
+          flags: integration
+          token: ${{ secrets.CODECOV_TOKEN }}
```
thaJeztah, line 167: Does this still need a token? I recall they had token-less options for public repositories, but not sure if that applies here; https://docs.codecov.com/docs/codecov-tokens#when-do-i-need-a-token

thaJeztah: And (same page); https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token

Image

crazy-max: Yes, I think we should keep it. Tokenless uploads only apply when Docker has disabled public token auth in Codecov, or when Codecov treats the upload as an unprotected/fork branch. This workflow also runs on pushes to `master`/`v*`, where Codecov requires a token by default. For fork PRs the secret resolves empty, and the v6 action still handles the tokenless fork path, so keeping the input doesn't block tokenless PR uploads.

tonistiigi (approved):

PR #3888, opened 2026-06-05
build: handle buildkit defaulting to oci mediatypes

crazy-max:
follow-up https://github.com/moby/buildkit/pull/6824

Updates Buildx integration behavior for BuildKit versions that default image exports to OCI media types since https://github.com/moby/buildkit/pull/6824

BuildKit compatibility version 30 changes image exports so the default manifest and index media types are OCI media types instead of Docker schema media types. The imagetools integration tests now derive the expected default media type from the BuildKit version, so older BuildKit versions keep expecting Docker media types while newer versions expect OCI media types.

The `TestBuildIndexAnnotationsLoadDocker` case now sets `oci-mediatypes=false` explicitly. This keeps the test focused on the Docker media type conflict without treating `type=docker` itself as a request for Docker schema media types. The default `type=docker` behavior is handled by the BuildKit-side fix in https://github.com/moby/buildkit/pull/6844.

### DIFF
--- a/tests/imagetools.go
+++ b/tests/imagetools.go
@@ -71,7 +71,7 @@ func testImagetoolsCopyManifest(t *testing.T, sb integration.Sandbox) {
 	var mfst ocispecs.Manifest
 	err = json.Unmarshal(dt, &mfst)
 	require.NoError(t, err)
-	require.Equal(t, images.MediaTypeDockerSchema2Manifest, mfst.MediaType)
+	require.Equal(t, defaultManifestMediaType(t, sb), mfst.MediaType)
 
 	registry2, err := sb.NewRegistry()
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


tests/build.go
```diff
@@ -1682,9 +1682,14 @@ func testBuildIndexAnnotationsLoadDocker(t *testing.T, sb integration.Sandbox) {
 	}
 	skipNoCompatBuildKit(t, sb, ">= 0.11.0-0", "annotations")
 	dir := createTestProject(t)
-	out, err := buildCmd(sb, withArgs("--annotation", "index:foo=bar", "--provenance", "false", "--output", "type=docker", dir))
+	out, err := buildCmd(sb, withArgs("--annotation", "index:foo=bar", "--provenance", "false", "--output", "type=docker,oci-mediatypes=false", dir))
```
tonistiigi: > oci-mediatypes=false

Is this needed after we merged https://github.com/moby/buildkit/pull/6844 ? I think it should not be.

crazy-max: No should not be needed. I made this change before it got merged and master image was up to date. So should be good without this now

tonistiigi (approved):

PR #3895, opened 2026-06-10
policy: enable proxy network from source policy caps

tonistiigi:
buildx support for moby/buildkit#6740

Evaluate source policy caps before solve requests so policies can enable
BuildKit proxy networking. Policy can return caps {"exec.proxy": true}
during the caps request to enable proxy network
support for the solve.

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -667,6 +667,7 @@ func configureSourcePolicy(ctx context.Context, np *noderesolver.ResolvedNode, o
 				Data:     policy.DefaultPolicyData(),
 			}},
 		}
+		builtin.SkipCaps = true
 		popts = append([]policyOpt{builtin}, popts...)
 	}
 
@@ -742,6 +743,11 @@ func configureSourcePolicy(ctx context.Context, np *noderesolver.ResolvedNode, o
 			DefaultPlatform:  defaultPlatform(bopts),
 			SourceResolver:   sourceResolver,
 		})
+		if !popt.SkipCaps {
+			if err := applyPolicyCaps(ctx, p, bopts, so); err != nil {
+				return nil, err
+			}
+		}
 		policies = append(policies, p)
 		cbs = append(cbs, p.CheckPolicy)
 		if popt.Strict {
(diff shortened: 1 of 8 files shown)


build/opt.go
```diff
@@ -742,6 +743,11 @@ func configureSourcePolicy(ctx context.Context, np *noderesolver.ResolvedNode, o
 			DefaultPlatform:  defaultPlatform(bopts),
 			SourceResolver:   sourceResolver,
 		})
+		if !popt.SkipCaps {
+			if err := applyPolicyCaps(ctx, p, bopts, so); err != nil {
+				return nil, err
+			}
+		}
```
crazy-max, line 750: I think it could poitentially reject existing policies that never request caps?

Maybe we can treat an undefined or empty decision during caps probing as empty caps?

tests/policy_build.go
```diff
@@ -40,6 +40,82 @@ var policyBuildTests = []func(t *testing.T, sb integration.Sandbox){
...
+	resp := &httpserver.Response{Content: []byte("policy-caps-proxy")}
+	server := httpserver.NewTestServer(map[string]*httpserver.Response{
+		"/file": resp,
+	})
+	defer server.Close()
+
+	dockerfile := []byte(`
+FROM busybox:latest
+RUN wget -O- ` + server.URL + `/file
+`)
```
crazy-max, line 73: seems to run through the normal build path because `server.URL` is a loopback/local endpoint. With the new proxy behavior local services are only reachable in the host+proxy case iiuc

not sure if we can have a proxy-reachable non-loopback test endpoint

tonistiigi: This always fails anyway, but test checks that it fails with "policy deny".

build/opt.go
tonistiigi: If the policies do not define caps then it is as good as having caps `false`. If one policy requires caps and others don't (explicitly or not) then caps still needed.

crazy-max (approved):

PR #3900, opened 2026-06-11
build: improve resource limits output and docs

jirimoravcik:
This PR is a follow-up of #3876, I fixed the things mentioned in the last review that weren't addressed before merging:
- Removed redundant if
- Added to docs
- Added a logic for human-readable print of memory


crazy-max (approved):
Thanks!

PR #3906, opened 2026-06-13
bake: reject targets inside target blocks

### LINKED ISSUE #2437: Invalid bake target syntax still builds
I made a typo when writing a bake definition that contains an invalid `targets` field for `target`:

```hcl
target "app" {
  dockerfile = "app.Dockerfile"
}

target "db" {
  dockerfile = "db.Dockerfile"
}

target "foo" {
  targets = ["app", "db"]
}
```

What's wrong is the bake definition still looks valid:

```
$ docker buildx bake foo --print
{
  "group": {
    "default": {
      "targets": [
        "foo"
      ]
    }
  },
  "target": {
    "foo": {
      "context": ".",
      "dockerfile": "Dockerfile"
    }
  }
}
```

`foo` should be a `group` here:

```hcl
target "app" {
  dockerfile = "app.Dockerfile"
}

target "db" {
  dockerfile = "db.Dockerfile"
}

group "build" {
  targets = ["app", "db"]
}
```

```
$ docker buildx bake foo --print
{
  "group": {
    "default": {
      "targets": [
        "foo"
      ]
    },
    "foo": {
      "targets": [
        "app",
        "db"
      ]
    }
  },
  "target": {
    "app": {
      "context": ".",
      "dockerfile": "app.Dockerfile"
    },
    "db": {
      "context": ".",
      "dockerfile": "db.Dockerfile"
    }
  }
}
```

We should validate this.

amarkdotdev:
## Summary
- reject the  attribute when it is used inside a  block
- point users toward  blocks for grouped bake targets
- add a regression test for issue #2437

## Testing
- go test ./bake -count=1

Closes #2437

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -33,6 +33,7 @@ import (
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/frontend/dockerfile/dfgitutil"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/zclconf/go-cty/cty"
 	"github.com/zclconf/go-cty/cty/convert"
 )
@@ -352,6 +353,7 @@ func ParseFiles(files []File, defaults, vars map[string]string, opts ...ParseOpt
 	var c Config
 	var composeFiles []File
 	var hclFiles []*hcl.File
+	var hclFileNames []string
 	for _, f := range files {
 		isCompose, composeErr := validateComposeFile(f.Data, f.Name, vars)
 		if isCompose {
@@ -367,6 +369,7 @@ func ParseFiles(files []File, defaults, vars map[string]string, opts ...ParseOpt
 					return nil, nil, err
 				}
 				hclFiles = append(hclFiles, hf)
+				hclFileNames = append(hclFileNames, f.Name)
 			} else if composeErr != nil {
 				return nil, nil, errors.Wrapf(err, "failed to parse %s: parsing yaml: %v, parsing hcl", f.Name, composeErr)
 			} else {
@@ -421,6 +424,10 @@ func ParseFiles(files []File, defaults, vars map[string]string, opts ...ParseOpt
 		}
 		c = dedupeConfig(c)
 		pm = *res
+	}
+
+	if err := validateTargetGroups(hclFiles, hclFileNames); err != nil {
+		return nil, nil, err
 	}
 
 	if frel {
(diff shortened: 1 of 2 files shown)


amarkdotdev:
Gentle ping on this one. It's a small, self-contained fix for #2437: `bake` now rejects a `targets` attribute used inside a `target` block (it belongs in a `group` block) with a clear "Unsupported argument" diagnostic, instead of silently building. DCO and labeler are green, and `go test ./bake -count=1` passes locally (full package suite, including the new `TestTargetDisallowsGroupTargetsAttribute`).

The `build` / `validate` / `e2e` / `codeql` workflows are sitting in the "action required" state pending a maintainer's approval to run on this PR. Could a maintainer approve the run and take a look? Happy to address anything that comes up. Closes #2437.

amarkdotdev:
@tonistiigi @crazy-max Gentle ping — rebased on latest `master` and `go test ./bake/...` still passes locally.

Small fix for #2437: `bake` now rejects `targets` inside a `target` block (belongs in a `group` block) with a clear diagnostic instead of silently building. CI was 159/160 green before the rebase.

Would appreciate a look when you have a moment.

crazy-max (changes requested):
Thanks for tackling this. I'm not convinced this fully closes #2437 yet.

This patch special-cases one invalid attribute, `targets` inside a `target` block, and returns a hard parse error after the normal HCL parse succeeds. That fixes the reproducer from #2437, but the issue thread was also about the broader validation contract, including the earlier concern that strict schema validation could be breaking and the later suggestion to warn instead.

So can we make the intended contract explicit with tests for the relevant bake inputs? At minimum I'd expect coverage showing this behaves correctly for merged files and JSON bake syntax too, since otherwise this is easy to mistake for a complete validation fix when it is really a targeted guard.

amarkdotdev:
@crazy-max Fair point — this is intentionally a narrow guard for the #2437 reproducer, not the broader schema validation we discussed in the issue thread. I held off on strict validation given the breaking-change concern you raised there.

Pushed `66aadfa` with tests that spell out what this covers today:

- HCL and JSON bake syntax
- merged files (including HCL + JSON together)
- `group` blocks with `targets` still parse fine

Contract for this PR: `targets` inside a `target` block is a hard parse error with a clear diagnostic, instead of being silently ignored. If you'd rather this be a warning for now (per your earlier note on the issue), happy to switch — just say the word.

amarkdotdev:
@crazy-max The requested coverage is in `bake/hcl_test.go` (`TestTargetDisallowsGroupTargetsAttribute`): HCL, JSON, merged HCL files, and merged HCL+JSON all assert the hard error with file attribution. Happy to switch to a warning instead if you prefer that contract — just let me know.

amarkdotdev:
Pushed warning-based contract per #2437 discussion: `targets` inside a `target` block now logs a file-attributed warning instead of failing parse. Tests updated for HCL, JSON, and merged inputs.

amarkdotdev:
@crazy-max Following up on the warning-based approach from #2437 (pushed Jul 12). Worth a re-review when you get a chance.

amarkdotdev:
@crazy-max Rebased onto latest master and resolved the bake.go conflict while keeping the warning-based contract from the earlier review. `targets` inside a `target` block still emits a file-attributed warning rather than a hard parse error, with coverage for HCL, JSON, and merged inputs. Worth another look when you have a moment.

amarkdotdev:
@crazy-max Gentle ping — this has been on the warning-based contract (file-attributed warning instead of hard parse error) since the Jul 12 update, and was rebased onto latest master since. Mind taking another look / clearing CHANGES_REQUESTED if that direction still looks right?

crazy-max:
Sorry for the delay, will take a look asap

crazy-max:
Thanks for the update, but I'm going to close this PR.

After re-reading #2437 and the follow-up comments, I don't think this is the right direction to land. The problem needs a design for how Bake validates `group` and `target` schemas while still working with composable attributes. This PR is still an ad-hoc post-parse check for one attribute, and the new version emits the warning through `logrus` from `ParseFiles`, which is not the right surface for user-facing Bake diagnostics and gives parsing a global logging side effect.

So this still doesn't close #2437. We are going to write a proper proposal for this first, then use that to decide the validation and warning behavior instead of landing a narrow special case and growing more exceptions around it.

Closing this for now.

PR #3918, opened 2026-06-18
bake: fix duplicated type in bakeEnvFiles lookup param

s3onghyun:
bakeEnvFiles declared its lookup parameter as func(string string) (string, bool) — a stray duplicate that names the param "string". The intended signature is func(string) (string, bool) (the only caller passes os.LookupEnv, a one-arg lookup). Same type, behavior-preserving.


crazy-max (approved):
Thanks

PR #3928, opened 2026-06-28
remote: use endpoint address for buildkit client authority

### LINKED ISSUE #3880: buildx remote driver sends :authority: localhost instead of the endpoint hostname
### Contributing guidelines

## Summary
The Docker buildx remote driver sends gRPC requests with `:authority:` localhost regardless of the configured TCP endpoint. This breaks any HTTP/2 reverse proxy (e.g., Envoy) that routes based on the `:authority` pseudo-header.

## Root cause

The remote driver's Client() method in driver/remote/driver.go passes an empty string as the address to client.New():

```Go
// driver/remote/driver.go, line 96
c, err := client.New(ctx, "", opts...)
```

In client/client.go, when address == "":
1. Line 120-121: address defaults to appdefaults.Address → unix:///run/buildkit/buildkitd.sock
2. Line 135-138: creds is nil (remote driver handles TLS itself via d.Dial(), not via grpc-go transport credentials), so authority is not set from creds.serverName
3. Line 139-141: authority = uri.Host → empty for a unix socket URI
4. Line 149: grpc.WithAuthority("") → grpc-go defaults to localhost
The actual TCP connection works correctly — the remote driver's WithContextDialer + d.Dial() connects to the right host with proper TLS (the servername driver-opt correctly sets TLS SNI). But the gRPC :authority pseudo-header is derived from the empty address, not from the endpoint.

### Expected behaviour

`:authority` should be my-buildkit.example.com:443 (or my-buildkit.example.com), matching the configured endpoint. This would allow HTTP/2 proxies to route based on `:authority`, which is the standard mechanism for virtual hosting in HTTP/2.

### Actual behaviour

With GRPC_GO_LOG_SEVERITY_LEVEL=info, grpc-go logs:
```
original dial target is: "unix:///run/buildkit/buildkitd.sock"
[Channel #1] Channel authority set to "localhost"
```

The outgoing request bears "localhost" as `:authority` instead of the endpoint name.

### Buildx version

buildx: v0.30.1 (9e66234aa13328a5e75b75aa5574e1ca6d6d9c01)

### Docker info

```text

```

### Builders list

```text
N/A
```

### Configuration

```Shell Script
d=$(mktemp -d)
BUILDX_CONFIG=$d buildx create \
    --name repro \
    --driver remote \
    --platform linux/amd64 \
    --driver-opt cacert=/path/to/ca.pem,cert=/path/to/client.pem,key=/path/to/client.pem,servername=my-buildkit.example.com \
    tcp://my-buildkit.example.com:443
BUILDX_CONFIG=$d buildx inspect --bootstrap repro
# Result: "context deadline exceeded" after 20s
BUILDX_CONFIG=$d buildx rm repro
rm -rf "$d"
```

### Build logs

```text

```

### Additional info

_No response_

jarqvi:
The remote driver created the buildkit client with an empty address:

    client.New(ctx, "", opts...)

With an empty address the buildkit client falls back to the system default address (the local unix socket) and derives the gRPC ":authority" pseudo-header from it, which ends up being "localhost". The actual connection was still correct because the remote driver provides its own dialer, but the wrong authority broke HTTP/2 reverse proxies (such as Envoy) that route based on ":authority".

Pass the configured endpoint address to client.New so the authority is derived from the remote endpoint hostname (e.g.
my-buildkit.example.com:443). The custom dialer is preserved, so the dial target and TLS/SNI behavior are unchanged.

Fixes #3880

### DIFF
--- a/driver/remote/driver.go
+++ b/driver/remote/driver.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"net"
+	"net/url"
 	"os"
 	"strings"
 	"sync"
@@ -17,6 +18,7 @@ import (
 	"github.com/moby/buildkit/client/connhelper"
 	"github.com/moby/buildkit/util/tracing/delegated"
 	"github.com/pkg/errors"
+	"google.golang.org/grpc"
 )
 
 type Driver struct {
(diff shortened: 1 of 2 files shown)


tonistiigi (commented):
There is already `servername` driver-opt. Iiuc if it is set then `grpc.WithAuthority(d.tlsOpts.serverName)` should be set instead of getting it from endpoint.

jarqvi:
Good catch.
Updated to use servername for :authority when set, falling back to the endpoint host. Added a test.

driver/remote/driver.go
```diff
@@ -93,13 +95,40 @@ func (d *Driver) Client(ctx context.Context, opts ...client.ClientOpt) (*client.
...
-		c, err := client.New(ctx, "", opts...)
+		// The remote driver establishes the connection itself through a custom
+		// dialer (including TLS), so the buildkit client cannot derive the gRPC
+		// ":authority" pseudo-header from the connection and would fall back to
+		// "localhost". Set it explicitly from the configured endpoint so HTTP/2
+		// reverse proxies (e.g. Envoy) can route on it. Passing the endpoint
+		// address also keeps the gRPC dial target meaningful; the actual dial
+		// target is unaffected as it still goes through the dialer above.
+		if authority := d.clientAuthority(); authority != "" {
+			opts = append(opts, client.WithGRPCDialOption(grpc.WithAuthority(authority)))
```
tonistiigi: nit: ideally this option would go to the beginning of `opts` array, eg `defaultOpts`  so that is same opt is already passed with parameter, then the parameter one wins.

jarqvi:
Done.

tonistiigi (approved):

crazy-max (approved):

PR #3935, opened 2026-07-06
bake: add file-relative path opt-in

### LINKED ISSUE #197: Relative paths in compose file not resolved correctly?
Given the current working directory is
`/src/api-platform/api-platform`

when doing
`docker buildx bake -f docker-compose-prod/docker-compose.build.yml`

then the relative path `../admin` in the build context of the `admin` service
(https://github.com/api-platform/docker-compose-prod/blob/3306ffa4723c2636677ff025e5d703a89347ddc1/docker-compose.build.yml#L52)

should be resolved to
`/src/api-platform/api-platform/admin`

(This works correctly with `docker-compose build -f docker-compose-prod/docker-compose.build.yml`)

### Actual result
`unable to prepare context: path "../admin" not found`

### LINKED ISSUE #1028: buildx bake: Specifying --file in another directory leads to failure
### Behaviour

tl;dr I can get `docker buildx bake` to succeed only when I `cd` to the subdirectory where docker-compose.yml lives.

#### Steps to reproduce this issue

Given the following project directory structure:
```
tests/docker-compose.yml
tests/dockerfiles/debian8/Dockerfile
tests/dockerfiles/debian9/Dockerfile
tests/dockerfiles/debian10/Dockerfile
```
And the following tests/docker-compose.yml file
```
version: "3"
services:
  debian8:
    build: ./dockerfiles/debian8
    cap_add: [ALL]
  debian9:
    build: ./dockerfiles/debian9
    cap_add: [ALL]
  debian10:
    build: ./dockerfiles/debian10
    cap_add: [ALL]
```

When a build definition file is specified in a subdirectory of the project such as above

Then resolution of Dockerfiles fails:

```
docker buildx bake -f  tests/docker-compose.yml
```
```
[+] Building 0.0s (0/0)
error: unable to prepare context: path "dockerfiles/debian9" not found
```

Yes, there is an obvious workaround – Simply first change the working directory to be that of where the build definition file exists.

```
cd tests && docker buildx bake
```
```
cd tests && docker buildx bake
[+] Building 1.4s (23/23) FINISHED
# (output omitted)
```

However, this limits the utility of this tool, especially in places like https://github.com/docker/bake-action.

#### Expected behaviour

`docker buildx bake -f  ./tests/docker-compose.yml` should succeed, given the context in the scenario above.

### Configuration

* See this GitHub Actions Run: https://github.com/sfgeorge/rvm1-ansible/runs/5707843598?check_suite_focus=true

crazy-max:
fixes https://github.com/docker/buildx/issues/1028
fixes https://github.com/docker/buildx/issues/197

This adds an opt-in Bake mode that resolves local directory paths in `target.context` and `target.contexts` relative to the Bake or Compose definition that owns those paths when `BUILDX_BAKE_FILE_RELATIVE_PATHS=1` is set.

Bake now accepts a parser option that records the definition file for local build context paths and rebases them after HCL or Compose files are parsed. The `buildx bake` command enables this option from `BUILDX_BAKE_FILE_RELATIVE_PATHS`, while `cwd://` remains available for paths that should stay relative to the invocation directory.

The opt-in also applies to targets that omit `context`, so the implicit default `.` is resolved from the target definition file instead of the invocation directory.

Compose files follow Compose project semantics for this opt-in, so local build contexts from Compose input are resolved from the directory of the first local Compose file rather than from override file locations.

### DIFF
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -155,7 +155,7 @@ Multiple definitions can be specified by separating them with the system's path 
 
 By default, local directory build contexts in Bake files are resolved from the
 current working directory. To opt in to resolving local directory build contexts
-from the directory of the first Bake file, set
+from the Bake or Compose file that defines each path, set
 `BUILDX_BAKE_FILE_RELATIVE_PATHS=1`. Use the `cwd://` prefix for paths that
 should remain relative to the current working directory.
 
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -155,9 +155,11 @@ Multiple definitions can be specified by separating them with the system's path 
 
 By default, local directory build contexts in Bake files are resolved from the
 current working directory. To opt in to resolving local directory build contexts
-from the Bake or Compose file that defines each path, set
-`BUILDX_BAKE_FILE_RELATIVE_PATHS=1`. Use the `cwd://` prefix for paths that
-should remain relative to the current working directory.
+from the Bake file that defines each path, set
+`BUILDX_BAKE_FILE_RELATIVE_PATHS=1`. Compose files use the first Compose file
+directory as the base, which matches Compose project directory semantics. Use
+the `cwd://` prefix for paths that should remain relative to the current working
+directory.
 
 You can pass the names of the targets to build, to build only specific target(s).
 The following example builds the `db` and `webapp-release` targets that are
--- a/docs/reference/buildx_bake.md
+++ b/docs/reference/buildx_bake.md
@@ -153,6 +153,12 @@ This is mutually exclusive with `-f` / `--file`; if both are specified, the envi
 Multiple definitions can be specified by separating them with the system's path separator
 (typically `;` on Windows and `:` elsewhere), but can be changed with `BUILDX_BAKE_PATH_SEPARATOR`.
 
+By default, local directory build contexts in Bake files are resolved from the
+current working directory. To opt in to resolving local directory build contexts
+from the directory of the first Bake file, set
+`BUILDX_BAKE_FILE_RELATIVE_PATHS=1`. Use the `cwd://` prefix for paths that
+should remain relative to the current working directory.
+
 You can pass the names of the targets to build, to build only specific target(s).
 The following example builds the `db` and `webapp-release` targets that are
 defined in the `docker-bake.dev.hcl` file:
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -449,8 +449,7 @@ func (t *Target) rebaseContextPaths() {
 			t.Context = &contextPath
 		}
 	} else if t.hasDefaultContextBase {
-		contextPath := rebaseContextPath(t.defaultContextBase, ".")
-		t.Context = &contextPath
+		t.useDefaultContextBase = true
 	}
 	for k, v := range t.Contexts {
 		if base, ok := t.contextsBase[k]; ok {
@@ -755,6 +754,9 @@ func (c Config) ResolveTarget(name string, overrides map[string]map[string]Overr
 	t.Inherits = nil
 	if t.Context == nil {
 		s := "."
+		if t.useDefaultContextBase {
+			s = rebaseContextPath(t.defaultContextBase, ".")
+		}
 		t.Context = &s
 	}
 	if t.Dockerfile == nil || (t.Dockerfile != nil && *t.Dockerfile == "") {
@@ -849,6 +851,7 @@ type Target struct {
 
 	defaultContextBase    string
 	hasDefaultContextBase bool
+	useDefaultContextBase bool
 	contextBase           string
 	hasContextBase        bool
 	contextsBase          map[string]string
@@ -973,6 +976,7 @@ func (t *Target) Merge(t2 *Target) {
 	if t2.hasDefaultContextBase {
 		t.defaultContextBase = t2.defaultContextBase
 		t.hasDefaultContextBase = true
+		t.useDefaultContextBase = t2.useDefaultContextBase
 	}
 	if t2.Context != nil {
 		t.Context = t2.Context
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -376,9 +376,12 @@ func ParseFiles(files []File, defaults, vars map[string]string, opts ...ParseOpt
 	}
 
 	if len(composeFiles) > 0 {
-		cfg, cmperr := parseComposeFilesWithBase(composeFiles, vars, frel)
+		cfg, cmperr := ParseComposeFiles(composeFiles, vars)
 		if cmperr != nil {
 			return nil, nil, errors.Wrap(cmperr, "failed to parse compose file")
+		}
+		if frel {
+			setComposeContextBase(cfg, composeFiles)
 		}
 		c = mergeConfig(c, *cfg)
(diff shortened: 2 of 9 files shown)


crazy-max:
Needs docs follow-up: https://docs.docker.com/build/building/variables/#build-tool-configuration-variables

docs/reference/buildx_bake.md
```diff
@@ -153,6 +153,12 @@ This is mutually exclusive with `-f` / `--file`; if both are specified, the envi
 Multiple definitions can be specified by separating them with the system's path separator
 (typically `;` on Windows and `:` elsewhere), but can be changed with `BUILDX_BAKE_PATH_SEPARATOR`.
 
+By default, local directory build contexts in Bake files are resolved from the
+current working directory. To opt in to resolving local directory build contexts
```
tonistiigi, line 157: Why the "first" bake file? Rather than "current" bake file. Even if this is more complicated, I think that should be the difference as the benefit of this method is that there isn't one global target directory.

tonistiigi (commented):
Looks like default `.` does not atm take the relative mode into account

```

	t.Run("default-context-in-subdirectory", func(t *testing.T) {
		bakefile := []byte(`
target "default" {
  dockerfile-inline = <<EOT
FROM scratch
COPY marker /marker
EOT
}
`)

		dir := tmpdir(
			t,
			fstest.CreateDir("definitions", 0700),
			fstest.CreateFile("definitions/docker-bake.hcl", bakefile, 0600),
			fstest.CreateFile("definitions/marker", []byte("marker"), 0600),
		)
		dirDest := t.TempDir()

		out, err := bakeCmd(
			sb,
			withDir(dir),
			withArgs("--file", "definitions/docker-bake.hcl", "--set", "*.output=type=local,dest="+dirDest),
			withEnv("BUILDX_BAKE_FILE_RELATIVE_PATHS=1"),
		)
		require.NoError(t, err, out)
		require.FileExists(t, filepath.Join(dirDest, "marker"))
	})
```

crazy-max:
@tonistiigi Updated so file-relative mode is no longer based on the first Bake file.

The parser now tracks the definition file that contributed each path. For HCL targets, it records the source block filename while decoding the target. For Compose files, it parses files separately in this mode so each generated target keeps the Compose file directory before normal Bake merging happens.

The rebase now uses that metadata per field. An explicit `context` is rebased from the file that defined that `context`. Each named context in `contexts` is rebased from the file that defined that entry. If `context` is omitted, it now treats the implicit default `.` as coming from the target definition file, which fixes the default-context case you pointed out.

`cwd://` remains the escape hatch for paths that should stay relative to the invocation directory, and special context forms such as `target:`, `docker-image://`, `oci-layout://`, remote URLs, and absolute paths are still left untouched.

Also added coverage for the implicit default context case, for multi-file definitions using the defining file instead of the first file, and for the Compose subdirectory case.

crazy-max:
I adjusted the Compose side in the last commit so it follows Compose project directory semantics instead of the HCL Bake definition-file model.

For HCL Bake files, file-relative mode still resolves each path from the Bake file that defines that path. For Compose files, the base is now the first Compose file directory, which better matches what users expect when comparing `docker compose --build` with `docker buildx bake -f compose.yml`.

This also keeps Compose override files from accidentally changing the base directory for paths they add. I added a regression test that covers a second Compose file adding `additional_contexts` and another build target, and both resolve from the first Compose file directory.

tonistiigi (commented):
failing tests for the inherits case

```
func TestInheritedContextRebase(t *testing.T) {
	t.Run("same file", func(t *testing.T) {
		fp := File{
			Name: filepath.Join("subdir", "docker-bake.hcl"),
			Data: []byte(`
target "base" {
  context = "basectx"
}

target "app" {
  inherits = ["base"]
  tags = ["app:latest"]
}`),
		}

		m, _, err := ReadTargets(context.TODO(), []File{fp}, []string{"app"}, nil, nil, nil, &EntitlementConf{}, ParseOpt{
			FileRelativePaths: true,
		})
		require.NoError(t, err)

		// the context path is defined in subdir/docker-bake.hcl, so the
		// inherited value must resolve relative to that file
		require.Equal(t, filepath.ToSlash(filepath.Clean("subdir/basectx")), *m["app"].Context)
	})

	t.Run("cross file", func(t *testing.T) {
		fp1 := File{
			Name: filepath.Join("one", "docker-bake.hcl"),
			Data: []byte(`
target "base" {
  context = "basectx"
}`),
		}
		fp2 := File{
			Name: filepath.Join("two", "docker-bake.hcl"),
			Data: []byte(`
target "app" {
  inherits = ["base"]
  tags = ["app:latest"]
}`),
		}

		m, _, err := ReadTargets(context.TODO(), []File{fp1, fp2}, []string{"app"}, nil, nil, nil, &EntitlementConf{}, ParseOpt{
			FileRelativePaths: true,
		})
		require.NoError(t, err)

		// the context path is defined in one/docker-bake.hcl, so the
		// inherited value must resolve relative to that file, not to the
		// file that defines the inheriting target
		require.Equal(t, filepath.ToSlash(filepath.Clean("one/basectx")), *m["app"].Context)
	})
}
```

tonistiigi:
Not sure if this case can be fixed as well
```

	t.Run("hcl target reference", func(t *testing.T) {
		baseBakefile := []byte(`
target "base" {
  context = "basectx"
}
`)
		appBakefile := []byte(`
target "app" {
  context = target.base.context
  dockerfile-inline = <<EOT
FROM scratch
COPY marker /marker
EOT
}
`)

		dir := tmpdir(
			t,
			fstest.CreateDir("one", 0700),
			fstest.CreateDir("one/basectx", 0700),
			fstest.CreateDir("two", 0700),
			fstest.CreateDir("two/basectx", 0700),
			fstest.CreateFile("one/docker-bake.hcl", baseBakefile, 0600),
			fstest.CreateFile("one/basectx/marker", []byte("source-file"), 0600),
			fstest.CreateFile("two/docker-bake.hcl", appBakefile, 0600),
			fstest.CreateFile("two/basectx/marker", []byte("consumer-file"), 0600),
		)
		dirDest := t.TempDir()

		out, err := bakeCmd(
			sb,
			withDir(dir),
			withArgs("--file", "one/docker-bake.hcl", "--file", "two/docker-bake.hcl", "--set", "app.output=type=local,dest="+dirDest, "app"),
			withEnv("BUILDX_BAKE_FILE_RELATIVE_PATHS=1"),
		)
		require.NoError(t, err, out)

		dt, err := os.ReadFile(filepath.Join(dirDest, "marker"))
		require.NoError(t, err)
		require.Equal(t, "source-file", string(dt))
	})
```

crazy-max:
@tonistiigi The last commit fixes the cross-file target reference.

In file-relative mode, `context = target.base.context` should keep the path base from the target that actually defines the context value. Before this change, the consuming target could rebase the resolved string against its own Bake file, so a target in `two/docker-bake.hcl` referencing `target.base.context` from `one/docker-bake.hcl` could incorrectly look under `two/`.

The fix is intentionally narrow. It only detects exact HCL traversals like `target.base.context` and `target["base"].context`, then reuses the referenced target's explicit context base, or its default context base if the referenced target relies on implicit `.`. It doesn't try to infer provenance through arbitrary expressions or `contexts` lookups, since that would add a lot more logic for cases that we don't really establish atm.

tonistiigi (approved):

PR #3944, opened 2026-07-07
policy: normalize local policy paths

crazy-max:
~needs #3943~

Was testing some policies on multiple machines and it fails in a case on Windows because `fs.FS` paths must use forward slashes, but local policy paths were cleaned with `filepath.Clean`, which produces backslashes on Windows. That made valid policy files fail to load with `invalid argument`:

```powershell
mkdir bx-policy-repro\policy
Set-Content bx-policy-repro\Dockerfile "FROM scratch"
Set-Content bx-policy-repro\policy\allow.rego @'
package docker

default allow := true

decision := {"allow": allow}
'@

cd bx-policy-repro
docker buildx build --check --policy "filename=cwd://policy/allow.rego" .
```
```text
failed to stat policy file cwd://policy/allow.rego: stat policy\allow.rego: invalid argument
```

The policy loader now converts local and `cwd://` policy filenames with `filepath.ToSlash` and `path.Clean` before resolving them through the policy filesystem.

### DIFF
--- a/build/policy_loader.go
+++ b/build/policy_loader.go
@@ -200,7 +200,7 @@ func (p *policyPathFSRef) resolve(name string) (fs.StatFS, string, error) {
 		if err != nil {
 			return nil, "", err
 		}
-		return cwd, filepath.Clean(v), nil
+		return cwd, path.Clean(filepath.ToSlash(v)), nil
 	}
 
 	contextFS, err := p.getContextFS()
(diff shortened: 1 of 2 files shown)


build/policy_loader.go
```diff
@@ -278,11 +279,11 @@ func normalizeLocalPolicyPath(name, contextDir string) string {
 		if rel, err := filepath.Rel(contextDir, name); err == nil {
 			rel = filepath.Clean(rel)
```
tonistiigi: nit: I don't think this clean should be needed here?

tonistiigi (commented):
`SanitizePath` seems to be `Clean(ToSlash)`, with extra WSL logic. Is this WSL logic relevant at all here? If not then I think it is more confusing and direct `Clean(ToSlash)` is clearer.

tonistiigi (approved):

PR #3955, opened 2026-07-16
store: replace namesgenerator with neutral builder IDs

lancefrench:
Removes the remaining `github.com/docker/docker` dependency from Buildx.

Buildx already uses the Moby API and client modules; this removes its only legacy import, `github.com/docker/docker/pkg/namesgenerator`, by using a neutral `builder_` prefix and a BuildKit-generated ID suffix.

This refreshes the approved approach from #3813 on current `master`, with regenerated module and vendor metadata.

Validation:
- `go test ./store`
- `go build ./cmd/buildx`
- `go mod graph | rg github.com/docker/docker` returns only `docker-credential-helpers`, not the legacy Docker daemon module.


lancefrench:
Uh, overeager robot. 😬

PR #3961, opened 2026-07-21
policy: verify BuildKit builder images

tonistiigi:
Extend the built-in policy to validate signed moby/buildkit release and floating tags before docker-container builders are created.

Pull the image first, inspect it through Docker, and bind verification to the descriptor digest. Resolve signature attestations through the BuildKit API embedded in the Docker daemon.

If pulling fails, use a local image while applying the same verification when the containerd image store exposes an immutable descriptor. Keep the classic image-store behavior unchanged because no descriptor is available.

Allow unmanaged repositories and digest-only references unchanged. Add the allow-untrusted-image driver option as an explicit verification bypass.

--

I think in the future this could be optimized to make `policy eval` command share more code with this. Currently some similar code is in cli package of that command that I didn't want to change yet.

### DIFF
--- a/driver/bkimage/bkimage.go
+++ b/driver/bkimage/bkimage.go
@@ -1,7 +1,14 @@
 package bkimage
 
 const (
-	DefaultImage         = "moby/buildkit:buildx-stable-1" // TODO: make this verified
-	QemuImage            = "tonistiigi/binfmt:latest"      // TODO: make this verified
+	DefaultImage         = "moby/buildkit:buildx-stable-1"
+	QemuImage            = "tonistiigi/binfmt:latest" // TODO: make this verified
 	DefaultRootlessImage = DefaultImage + "-rootless"
+
+	// TrustedRepo is the fully-qualified repository whose tags are verified
+	// against the builtin default policy before a builder is created from
+	// them. Images from other repositories pass through unverified; the
+	// allow-untrusted-image driver-opt is only needed when a TrustedRepo tag
+	// that the policy covers does not verify correctly.
+	TrustedRepo = "docker.io/moby/buildkit"
 )
--- a/build/opt.go
+++ b/build/opt.go
@@ -65,18 +65,6 @@ var sendGitQueryAsInput = sync.OnceValue(func() bool {
 	return false
 })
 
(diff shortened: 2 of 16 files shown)


driver/bkimage/bkimage.go
```diff
@@ -1,7 +1,14 @@
 package bkimage
 
 const (
-	DefaultImage         = "moby/buildkit:buildx-stable-1" // TODO: make this verified
-	QemuImage            = "tonistiigi/binfmt:latest"      // TODO: make this verified
+	DefaultImage         = "moby/buildkit:buildx-stable-1"
+	QemuImage            = "tonistiigi/binfmt:latest" // TODO: make this verified
```
crazy-max, line 5: This could be a follow-up but looks tricky to verify QEMU image used by the kubernetes driver during deployment at: https://github.com/docker/buildx/blob/5c10fb7184b0901054bb696b630803e76d3f66c3/driver/kubernetes/manifest/manifest.go#L196-L205

We would need to know beforehand the node platform to perform the verification.

cc @AkihiroSuda

crazy-max: The current policy verifier resolves signatures against a selected platform manifest, so Kubernetes cannot safely use the local default platform here. We'd need either a known pod platform or an all-platform verification story before verifying and pinning this init image.

crazy-max (approved):

driver/bkimage/bkimage.go
```diff
@@ -1,7 +1,14 @@
...
 	DefaultRootlessImage = DefaultImage + "-rootless"
+
+	// TrustedRepo is the fully-qualified repository whose tags are verified
+	// against the builtin default policy before a builder is created from
+	// them. Images from other repositories pass through unverified; the
+	// allow-untrusted-image driver-opt is only needed when a TrustedRepo tag
+	// that the policy covers does not verify correctly.
+	TrustedRepo = "docker.io/moby/buildkit"
```
AkihiroSuda, line 13: Why not just promote `buildkit` (and `binfmt`, `xx`, ...) to DOI?

tonistiigi: How would that help?

PR #3962, opened 2026-07-21
bake: allow overriding declared secret sources

crazy-max:
relates to https://github.com/docker/github-builder/pull/248

This adds `--set target.secret.<id>=...` support to Bake so a caller can override the source of an already declared BuildKit secret without redefining or injecting a new secret into the target.

The override parser now accepts `secret.<id>` as a named field and applies it after normal secret list overrides. The value uses existing Bake secret source attributes, such as `env=AWS_CREDENTIALS` and `src=/path/to/file`, and the override fails if the secret ID is not already declared or if an inline `id=` does not match.

This follows the discussion in https://github.com/docker/github-builder/pull/248, where the reusable workflow needed to provide secret values for Bake without forcing callers to reference workflow-internal environment names. The review concern was that workflow-provided secrets should not define new secrets or accidentally inject secrets into unrelated targets, and this Buildx-side support gives that workflow a cleaner way to change only the source for a secret that the Bake definition already declares.

### DIFF
--- a/bake/bake.go
+++ b/bake/bake.go
@@ -606,9 +606,12 @@ func (c Config) newOverrides(v []string) (map[string]map[string]Override, error)
 					override.Append = appendTo
 					override.ArrValue = append(override.ArrValue, parts[1])
 				}
-			case "resources":
+			case "resources", "secret":
 				if len(keys) != 3 {
-					return nil, errors.Errorf("invalid key %s, resources requires name", parts[0])
+					return nil, errors.Errorf("invalid key %s, %s requires name", parts[0], keys[1])
+				}
+				if appendTo {
+					return nil, errors.Errorf("invalid key %s, %s does not support append", parts[0], keys[1])
 				}
 				override.Value = parts[1]
 			case "args":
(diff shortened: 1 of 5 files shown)


bake/bake.go
```diff
@@ -606,9 +606,12 @@ func (c Config) newOverrides(v []string) (map[string]map[string]Override, error)
...
 					override.ArrValue = append(override.ArrValue, parts[1])
 				}
-			case "resources":
+			case "resources", "secret":
 				if len(keys) != 3 {
-					return nil, errors.Errorf("invalid key %s, resources requires name", parts[0])
+					return nil, errors.Errorf("invalid key %s, %s requires name", parts[0], keys[1])
+				}
+				if appendTo {
+					return nil, errors.Errorf("invalid key %s, %s does not support append", parts[0], keys[1])
```
crazy-max, line 614: I also added a small parser guard so `--set target.secret.foo+=env=FOO` fails instead of being silently treated like `--set target.secret.foo=env=FOO`.

This is intentional because `secret.<id>` is a source replacement for one already declared secret, not an appendable list field. The existing `secrets` field remains the appendable form for full secret definitions. I kept the actual source override application deferred until after regular `secrets` overrides, so combinations like `--set target.secrets=id=foo,src=local` followed by `--set target.secret.foo=env=FOO` resolve against the final declared secret set.

Note that this also affects `resources.<name>+=...`. Today that syntax is already not a real append, because `resources` fields are scalar values and the parser ends up treating `+=` the same as `=`. The new guard makes that explicit instead of silently accepting misleading syntax. I think that's fine because there is no meaningful append behavior for `resources.memory`, `resources.cpu-quota`, or similar fields.

tonistiigi (approved):
This looks good but when integrating with github workflow you need to make sure that only values that are allowed to be passed are the ones coming with secrets, so one can't use this to send existing env to secret and then read it out in dockerfile.

crazy-max:
needs docs follow-up https://docs.docker.com/build/bake/overrides/

Issue #556, opened 2021-03-04
`docker buildx build` got stuck if buildkit container dies


ojab:
buildx-0.5.1, `moby/buildkit:v0.8.1`, `Docker version 20.10.5, build 55c4c88` on linux x86_64.

1. `docker buildx build --platform=local -o . git://github.com/docker/buildx`
2. While build is in progress - `docker exec -ti buildx_buildkit_builder-builder0 kill -s QUIT 1` where `buildx_buildkit_builder-builder0` is the name of buildkit container
3. `docker buildx build` hangs indefinitely

holograph:
Happens with 0.7.1 as well:
github.com/docker/buildx v0.7.1-docker 05846896d149da05f3d6fd1e7770da187b52a247
Docker version 20.10.12, build e91ed57

ceelian:
Have a similar issue when doing a buildx bake. It hangs on random commands. Already investigated 3 days but it is very hard to track it down. 

At first I thought it is something with the hcl files, because when i changed something there and some things in the dockerfile it worked sometimes. But then I changed something again and the previous version has stopped working although it worked a couple of minutes ago. 

So it works/works not based on changes you do in the files but it doesn't matter what you change it is more related to which files you touch, the order of changes of the files. We have split the configuration to multiple files, but I could also reproduce it with a single hcl file and multiple dockerfiles.

So it is probably related to some internal scheduling based on what files are changed (config, dockerfile)? Could that be?

I also found out that it is very likely related to some extend to the `contexts` option int the .hcl files. We use to link to other targets to implement kind of a build order. The problematic (but for our case necessary) Dockerfile commands is the `COPY --from` command.

```
FROM bar AS foo
...
COPY --from=foo ...
```

The overall demo project looks like this:
```
# docker-bake.hcl
target "inh" {
    context   = "."
    cache-from = [
        "type=local,src=/Users/ceelian/tmp/buildx_cache"
    ]
    cache-to = ["type=local,dest=/Users/ceelian/tmp/buildx_cache"]
}

target "base" {
    inherits = ["inh"]
    dockerfile = "baseapp.Dockerfile"
    tags       = ["example.com/base:latest"]
}

target "third" {
    inherits = ["inh"]
    dockerfile = "thirdapp.Dockerfile"
    tags       = ["example.com/third:latest"]
}

target "another" {
    inherits = ["inh"]
    contexts = {
        "thirdapp" = "target:third",
    }
[...]

PR #3966, opened 2026-07-22
driver/kubernetes: close exec-stream pipes when the stream ends

pgimalac:
Ensure the far ends of the command exec pipes are properly closed when the exec stream ends (due to an error or not), to prevent infinite hangs in reader/writer.

Currently, if the remote kubernetes executor gets killed, buildx just hangs forever because it tries to read from pipes which are never closed. With this change, the pipes are immediately closed when `StreamWithContext` finishes, and readers/writers are unblocked.

Refs #556.

Added some unit tests to highlight the fix.

## Repro / testing
Setup:
```sh
# create kind cluster
kind create cluster --name buildkit-repro

# create a builder
docker buildx create --driver kubernetes --name kindbuilder --driver-opt namespace=default,replicas=1 --bootstrap
```

```Dockerfile
FROM alpine:latest
RUN echo "step1 starting" && sleep 120 && echo "step1 done"
```

and build it:
```sh
docker buildx build --builder kindbuilder --progress=plain -f Dockerfile .
```

Wait for the log to show that it's running the sleep step, then open a new shell and kill it:
```sh
POD=$(kubectl get pods -l app=kindbuilder0 -o jsonpath='{.items[0].metadata.name}')
kubectl delete pod "$POD" --grace-period=0 --force
```

The `build` command will then hang forever (in particular won't finish after 2 minutes).

Now build `cmd/buildx` from this PR, and re-run the same steps using this `buildx` binary instead of `docker buildx`, you should observe that the build command logs `err="io: read/write on closed pipe"` when the pod is killed, and fails with the following error:

```
ERROR: failed to build: failed to receive status: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = "error reading from server: command terminated with exit code 137", received prior goaway: code: NO_ERROR, debug data: "graceful_stop"`
```

### DIFF
--- a/driver/kubernetes/execconn/execconn_test.go
+++ b/driver/kubernetes/execconn/execconn_test.go
@@ -1,0 +1,118 @@
+package execconn
+
+import (
+	"context"
+	"errors"
+	"io"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+// fakeExecutor is a fake remotecommand.Executor whose StreamWithContext blocks until
+// unblock is closed, then returns err. It stands in for the real SPDY exec stream
+// to a builder pod, so the pipe-closing behavior in newExecConn can be tested
+// without a real Kubernetes API server.
+type fakeExecutor struct {
+	unblock chan struct{}
+	err     error
+}
+
+func (f *fakeExecutor) Stream(_ remotecommand.StreamOptions) error {
+	panic("unimplemented")
+}
+
+func (f *fakeExecutor) StreamWithContext(ctx context.Context, _ remotecommand.StreamOptions) error {
+	select {
+	case <-f.unblock:
+		return f.err
+	case <-ctx.Done():
+		return context.Cause(ctx)
+	}
+}
+
+func TestNewExecConnPropagatesStreamEnd(t *testing.T) {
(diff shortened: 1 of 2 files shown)


driver/kubernetes/execconn/execconn_test.go
crazy-max, line 1: This test covers the `Read` side well, but the implementation also closes `stdinR` to unblock `kubeConn.Write` when `StreamWithContext` returns. Could we add a small assertion that `conn.Write(...)` also returns after the stream ends? That would lock in both halves of the pipe cleanup this PR is relying on.

crazy-max:
> Fixes #556.

Since #556 originally reports the docker-container driver and this PR fixes the Kubernetes exec path, `Refs #556` may be more accurate than `Fixes #556` unless the original docker-container hang is known to be covered elsewhere.

pgimalac:
> https://github.com/docker/buildx/issues/556 originally reports the docker-container driver

@crazy-max yeah fair, I figured it was a catch-all for both (in particular considering the [last message](https://github.com/docker/buildx/issues/556#issuecomment-2980063930) is about kubernetes), but I'll update the description.

driver/kubernetes/execconn/execconn_test.go
pgimalac: Done

crazy-max (approved):
LGTM thanks!

Issue #2793, opened 2024-11-16
characters ERROR: failed to dial qRPC: rpc error:code = Internal desc = rpc error: code = Internal desC = header key "x-docker-expoSe-session-name" contains value with non-printable ASCI


fsckzy:
### Description

I am trying to build an amd64 arm64 image, but the build failed.
The error message is as shown in the title

### Expected behaviour

Build amd64 arm64 image

### Actual behaviour

```shell
# docker buildx build --platform arm64,amd64 -t dockerhub.kubekey.local/zhjk/xx:8 . --push
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)      docker-container:mybuilder
[+] Building 0.0s (0/0)                                                                                                                                           docker-container:mybuilder
ERROR: failed to dial gRPC: rpc error: code = Internal desc = rpc error: code = Internal desc = header key "x-docker-expose-session-name" contains value with non-printable ASCII characters
```

### Buildx version

github.com/docker/buildx v0.15.1 1c1dbb2

### Docker info

```text
Client: Docker Engine - Community
 Version:    27.0.3
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
[...]

PR #3972, opened 2026-07-26
Build fails when shared context directory contains non-ASCII characters

maxproske:
Initially caught by docker/compose#13965 2 Compose services sharing a build context containing Chinese characters, but after some digging, every build dialed over gRPC fails. Same as #2793 which was closed as not reproducible.

Sanitize the shared key before it reaches gRPC.

Steps to reproduce:
```sh
mkdir -p 早

docker buildx create --name repro --driver docker-container
docker buildx build --builder repro 早

docker buildx bake -f - a b <<'EOF'
target "a" { context = "早" }
target "b" { context = "早" }
EOF

docker compose -f - build <<'EOF'
services:
  a: {build: 早}
  b: {build: 早}
EOF
```

All 3 fail with:
```
failed to dial gRPC: rpc error: code = Internal desc = rpc error: code = Internal desc = header key "x-docker-expose-session-sharedkey" contains value with non-printable ASCII characters
```

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -850,7 +850,7 @@ func loadInputs(ctx context.Context, d *driver.DriverHandle, inp *Inputs, pw pro
 		if p, err := filepath.Abs(sharedKey); err == nil {
 			sharedKey = filepath.Base(p)
 		}
-		target.SharedKey = sharedKey
+		target.SharedKey = sanitizeSharedKey(sharedKey)
 		switch inp.DockerfilePath {
 		case "-":
 			dockerfileReader = inp.InStream.NewReadCloser()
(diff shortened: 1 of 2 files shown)


maxproske:
Newbie question: if approved, worth me guarding downstream in buildkit too?

crazy-max (commented):
Thanks but I think this should probably be fixed in BuildKit by encoding the shared key when it's put into session metadata and decoding it in the session manager, rather than hashing it in Buildx.

crazy-max:
opened https://github.com/moby/buildkit/pull/6989

crazy-max:
Closing this one as https://github.com/moby/buildkit/pull/6989 is probably the right fix but thanks again for your contribution!

maxproske:
Agreed!

PR #3978, opened 2026-07-28
ci: skip windows release checksums and sigstore bundles

crazy-max:
This updates the Buildx release workflow so Windows artifacts are handled like Darwin artifacts before the external signing handoff.

Windows binaries are signed by a downstream workflow, which downloads the Buildx release assets, keeps the Darwin and Windows binaries, signs them, verifies the signatures, and uploads the signed binaries back to the release. Skipping Windows checksums here avoids publishing checksums for pre-signed binaries that will be replaced by the signing workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -336,7 +336,7 @@ jobs:
               mv "provenance.json" "${filename}.provenance.json"
               mv "sbom-binaries.spdx.json" "${filename}.sbom.json"
               find . -name 'sbom*.json' -exec rm {} \;
-              if [[ "$binname" == *darwin* ]]; then
+              if [[ "$binname" == *darwin* ]] || [[ "$binname" == *windows* ]]; then
                 rm -f "provenance.sigstore.json"
               elif [ -f "provenance.sigstore.json" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -336,7 +336,7 @@ jobs:
               mv "provenance.json" "${filename}.provenance.json"
               mv "sbom-binaries.spdx.json" "${filename}.sbom.json"
               find . -name 'sbom*.json' -exec rm {} \;
-              if [[ "$binname" == *darwin* ]]; then
+              if [[ "$binname" == *darwin* ]] || [[ "$binname" == *windows* ]]; then
```
thaJeztah, line 339: Do we have tracking issues for the remaining work for this? (wondering if a TODO + link is warranted here)

thaJeztah (approved):
LGTM

left a suggestion

.github/workflows/build.yml
crazy-max: No public tracking issue that I know of yet. The intent here is to match the existing Darwin flow because Windows binaries will now be signed by the downstream workflow after the Buildx release assets are uploaded, so we should not publish checksums for the unsigned Windows binaries that will be replaced.

I will look to add something in our README.

PR #3991, opened 2026-08-04
build: add OCI artifact default opt-out

crazy-max:
relates to
* https://github.com/moby/buildkit/issues/7014
* https://github.com/moby/buildkit/issues/7007
* https://github.com/moby/buildkit/pull/7015

This adds `BUILDX_NO_DEFAULT_OCI_ARTIFACT=true` as a compatibility escape hatch for Buildx users who need to keep provenance attestations enabled while disabling the newer OCI artifact attestation storage format.

When this environment variable is set, Buildx now sets `oci-artifact=false` when attestation support is available and the user did not already set `oci-artifact` explicitly. This keeps explicit exporter options authoritative, mirrors the existing `BUILDX_NO_DEFAULT_ATTESTATIONS` compatibility pattern, and applies to both `buildx build` and `buildx bake`.

BuildKit started defaulting attestation manifests to OCI artifacts in moby/buildkit#6914 after moby/buildkit#6171 tracked registry support for that format. The issue reported in moby/buildkit#7007 showed that GitLab can still reject a push when the attestation manifest has a `subject` that points at a manifest that is new in the same push. The initial BuildKit-side fix in moby/buildkit#7012 changed the manifest push order, but the follow-up discussion in moby/buildkit#7014 pointed out that this is a registry compliance issue because OCI distribution requires registries to initially accept valid manifests whose `subject` does not exist yet.

This PR does not change the default behavior and does not make Buildx prefer the non-OCI artifact format for compliant registries. It only provides a narrow opt-out for environments that need time to deal with registry compatibility problems, while preserving provenance attestations instead of requiring users to disable them entirely with `BUILDX_NO_DEFAULT_ATTESTATIONS`.

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -65,6 +65,11 @@ var sendGitQueryAsInput = sync.OnceValue(func() bool {
 	return false
 })
 
+const (
+	noDefaultAttestationsEnv = "BUILDX_NO_DEFAULT_ATTESTATIONS"
+	noDefaultOCIArtifactEnv  = "BUILDX_NO_DEFAULT_OCI_ARTIFACT"
+)
+
 // policyExplicitlyDisabled reports whether the user passed `--policy
 // disabled=true`, which suppresses both user-defined and builtin default
 // policies.
@@ -355,12 +360,11 @@ func toSolveOpt(ctx context.Context, np *noderesolver.ResolvedNode, multiDriver 
 	}
 
 	if _, ok := opt.Attests["provenance"]; !ok && supportAttestations {
-		const noAttestEnv = "BUILDX_NO_DEFAULT_ATTESTATIONS"
 		var noProv bool
-		if v, ok := os.LookupEnv(noAttestEnv); ok {
+		if v, ok := os.LookupEnv(noDefaultAttestationsEnv); ok {
 			noProv, err = strconv.ParseBool(v)
 			if err != nil {
-				return nil, nil, errors.Wrap(err, "invalid "+noAttestEnv)
+				return nil, nil, errors.Wrap(err, "invalid "+noDefaultAttestationsEnv)
 			}
 		}
 		if !noProv {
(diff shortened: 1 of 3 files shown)


build/opt.go
```diff
@@ -65,6 +65,11 @@ var sendGitQueryAsInput = sync.OnceValue(func() bool {
 	return false
 })
 
+const (
+	noDefaultAttestationsEnv = "BUILDX_NO_DEFAULT_ATTESTATIONS"
+	noDefaultOCIArtifactEnv  = "BUILDX_NO_DEFAULT_OCI_ARTIFACT"
```
thaJeztah, line 70: Thinking out loud; what are the formats we (currently) support for including attestations? Would it make sense to have a "attestations-format" selection instead of a boolean? (e.g. `BUILDX_ATTESTATIONS_FORMAT=(oci|foo|bar|compat|legacy)`) or was the old format not complying to any standards? (ISTR there were options for legacy registries, so those would still be valid and potentially work around the issue with non-OCI-compliant registries?)

crazy-max: I'd rather keep this as a narrow opt-out for now (specially since this change aims for a patch release). BuildKit currently exposes this as a boolean exporter option, `oci-artifact=true|false`, so mirroring that as `BUILDX_NO_DEFAULT_OCI_ARTIFACT` keeps Buildx from inventing a new format-selection layer on top.

The explicit format selection already exists through `--output type=image,oci-artifact=false` when users want to choose it directly. This env var only changes the default when the option was not set, similar to `BUILDX_NO_DEFAULT_ATTESTATIONS`.

A generic `BUILDX_ATTESTATIONS_FORMAT=legacy|compat|oci` feels a bit too broad to me because Buildx would then need to define what those names mean and keep mapping them to BuildKit exporter behavior. If BuildKit grows more attestation storage formats later, I think that should first be exposed as a BuildKit exporter option, and Buildx can follow that model.

thaJeztah: Yeah, that's ok with me; mostly wanted to avoid having to introduce yet-another env-var if there's more options than enable/disable (not even sure what format the "disable" means and if that's a legacy fallback defined by OCI?)

tests/bake.go
```diff
@@ -2232,6 +2233,41 @@ target "default" {
 	// TODO: test metadata file when supported by multi exporters https://github.com/docker/buildx/issues/2181
 }
 
+func testBakeNoDefaultOCIArtifact(t *testing.T, sb integration.Sandbox) {
+	if isMobyWorker(sb) {
+		t.Skip("attestations are not supported by the docker worker")
+	}
```
thaJeztah, line 2239: Interesting; are these two the same? (I went looking if we had a more specific "supports attestations" instead of driver-name :joy:)

https://github.com/docker/buildx/blob/19888263664a9dca76ff1bc827686a08489f5bd9/tests/integration.go#L126-L129

https://github.com/docker/buildx/blob/19888263664a9dca76ff1bc827686a08489f5bd9/tests/integration.go#L136-L139

thaJeztah (approved):
LGTM

tests/bake.go
crazy-max: Hum indeed, should rework this :sweat_smile:

PR #3992, opened 2026-08-04
build: resolve target links for call methods

### LINKED ISSUE #3343: `bake --check` fails to resolve inter-target dependencies when definitions are split across files
### Contributing guidelines

### Description

The docker `buildx bake --check` command fails to validate a build when target dependencies are split across multiple HCL files. While the actual build command (`docker buildx bake <target>`) works correctly, the validation logic incorrectly reports an `unsupported context source target` error.

### Expected behaviour

The `docker buildx bake --check` command should successfully validate the dependency graph across all provided HCL files without errors, behaving consistently with the actual build process which correctly resolves the same dependencies.

### Actual behaviour

The `docker buildx bake --check` command fails with an `unsupported context source target` error. This indicates the validation logic cannot resolve a dependency specified with `contexts = { my_context = "target:my_target" }` when the referenced target's definition is in another HCL file loaded via the `-f` flag. The regular build command without `--check` succeeds.

### Buildx version

github.com/docker/buildx v0.26.1 1a8287f

### Docker info

```text
 Version:    28.3.3
 Context:    default
 Debug Mode: false
 Plugins:
    Version:  v0.26.1
    Version:  v2.39.1

Server:
 Containers: 4
  Running: 4
  Paused: 0
  Stopped: 0
 Images: 6
 Server Version: 28.3.3
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
 CDI spec directories:
  /etc/cdi
  /var/run/cdi
 Swarm: inactive
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 05044ec0a9a75232cad458027ca83437aae3f4da
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.6.87.1-microsoft-standard-WSL2
 Operating System: Ubuntu 24.04.2 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 4
 Total Memory: 15.62GiB
 Name: volt-lpt
 ID: 9dcce5e7-a038-4a29-b3a1-d2f38b7e795d
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  ::1/128
  127.0.0.0/8
 Registry Mirrors:
  https://mirror.gcr.io/
 Live Restore Enabled: false
```

### Builders list

```text
NAME/NODE          DRIVER/ENDPOINT    STATUS    BUILDKIT   PLATFORMS
default*           docker                                  
 \_ default         \_ default        running   v0.23.2    linux/amd64 (+3), linux/386
```

### Configuration

Here is a minimal setup to reproduce the issue.

1.  File structure

```plaintext
├── build.hcl
├── providers.hcl
└── dockerfiles/
    ├── app/
    │   └── Dockerfile
    ├── base/
    │   └── Dockerfile
    └── builder/
        └── Dockerfile
```

2. File content:

  - providers.hcl

```terraform
# Defines the builder and a base application layer that depends on it.

target "builder" {
  dockerfile = "dockerfiles/builder/Dockerfile"
  tags       = ["my-builder:latest"]
}

target "base" {
  dockerfile = "dockerfiles/base/Dockerfile"
  contexts = {
    # This "target:" reference causes the error
    builder = "target:builder"
  }
  tags = ["my-base:latest"]
}
```

  - build.hcl
```terraform
# Defines the final application, which depends on the base layer.

target "app" {
  dockerfile = "dockerfiles/app/Dockerfile"
  contexts = {
    # This "target:" reference also causes the error
    base = "target:base"
  }
  tags = ["myapp:latest"]
}
```

- Dockerfiles

`dockerfiles/builder/Dockerfile`:

```Dockerfile
# syntax=docker/dockerfile:1
# check=error=true
FROM alpine:latest AS builder

RUN echo "This is the builder" > /builder.txt
```

`dockerfiles/base/Dockerfile`:

```Dockerfile
# syntax=docker/dockerfile:1
# check=error=true
FROM builder AS base
RUN echo "This is the base image" > /base.txt
COPY --from=builder /builder.txt /
```

`dockerfiles/app/Dockerfile`:

```Dockerfile
# syntax=docker/dockerfile:1
# check=error=true
FROM base AS app
RUN echo "This is the final application" > /app.txt
COPY --from=base /base.txt /
```

3. Invoked Command that Fails:

`docker buildx bake -f providers.hcl -f build.hcl app --check`

### Build logs

```text
[+] Building 5.7s (8/8) FINISHED                                                                                                                                                                        docker:default
 => [internal] load local bake definitions                                                                                                                                                                        0.0s
 => => reading providers.hcl 367B / 367B                                                                                                                                                                          0.0s
 => => reading build.hcl 250B / 250B                                                                                                                                                                              0.0s
 => [builder internal] load build definition from Dockerfile                                                                                                                                                      0.0s
 => => transferring dockerfile: 161B                                                                                                                                                                              0.0s
 => [app] resolve image config for docker-image://docker.io/docker/dockerfile:1                                                                                                                                   0.2s
 => CACHED [app] docker-image://docker.io/docker/dockerfile:1@sha256:9857836c9ee4268391bb5b09f9f157f3c91bb15821bb77969642813b0d00518d                                                                             0.0s
 => [builder internal] load metadata for docker.io/library/alpine:latest                                                                                                                                          4.5s
 => [builder internal] load .dockerignore                                                                                                                                                                         0.0s
 => => transferring context: 2B                                                                                                                                                                                   0.0s
 => [base internal] load build definition from Dockerfile                                                                                                                                                         0.0s
 => => transferring dockerfile: 188B                                                                                                                                                                              0.0s
 => [app internal] load build definition from Dockerfile                                                                                                                                                          0.0s
 => => transferring dockerfile: 184B                                                                                                                                                                              0.0s
app

error: unsupported context source target for base
dockerfiles/app/Dockerfile:5
--------------------
   3 |     FROM base AS app
   4 |     RUN echo "This is the final application" > /app.txt
   5 | >>> COPY --from=base /base.txt /
--------------------


base

error: unsupported context source target for builder
dockerfiles/base/Dockerfile:5
--------------------
   3 |     FROM builder AS base
   4 |     RUN echo "This is the base image" > /base.txt
   5 | >>> COPY --from=builder /builder.txt /
--------------------


builder

Check complete, no warnings found.
```

### Additional info

The bug is specific to the `--check` flag. Running the same command without it executes the build successfully, proving that the core dependency resolution logic is working as intended. The issue lies only within the pre-flight validation check.

okhowang:
Targets referencing each other through a "target:" named context are resolved on the client: the linked target is solved first and its result is then rewritten into an "input:" context by waitContextDeps.

A call method such as check or outline replies with the metadata of the subrequest and does not return a build result, so nothing can be linked into the targets depending on it. The "target:" context is then left untouched and reaches the frontend, which rejects it:

  $ docker buildx bake -f providers.hcl -f build.hcl app --check
  app
  error: unsupported context source target for base
  base
  error: unsupported context source target for builder

Solve the build request separately when a target using a call method is linked to by another one, and register that result for them. It is never evaluated, so the definition is only resolved and nothing is built. Checks are skipped for that request as they are already reported by the call method, otherwise a Dockerfile asking for check violations to be errors would fail it before the report is produced.

Fixes #3343

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -607,16 +607,45 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
 						// Capture the error from this build function.
 						defer catchFrontendError(&retErr, &frontendErr)
 
-						if opt.CallFunc != nil {
-							if _, ok := req.FrontendOpt["frontend.caps"]; !ok {
-								req.FrontendOpt["frontend.caps"] = "moby.buildkit.frontend.subrequests+forward"
-							} else {
-								req.FrontendOpt["frontend.caps"] += ",moby.buildkit.frontend.subrequests+forward"
+						rKey := resultKey(dp, k)
+						children := childTargets[rKey]
+
+						// A call method replies with the metadata of the subrequest
+						// instead of a build result, so there would be nothing to link
+						// into the targets using this one as a named context. Solve the
+						// build request on its own in that case and register its result
+						// for them. It is never evaluated, so this only resolves the
+						// definition and does not build anything.
+						linked := false
+						if opt.CallFunc != nil && len(children) > 0 {
+							linkReq := req
+							linkReq.FrontendOpt = maps.Clone(req.FrontendOpt)
+							// checks are reported by the call method below, skip them
+							// here so that this request is not rejected by a Dockerfile
+							// that asks for check violations to be errors
+							linkReq.FrontendOpt["build-arg:BUILDKIT_DOCKERFILE_CHECK"] = "skip=all;error=false"
+							res, err := solve(ctx, c, linkReq)
+							if err != nil {
+								return nil, err
 							}
-							req.FrontendOpt["requestid"] = "frontend." + opt.CallFunc.Name
+							results.Set(rKey, res)
+							linked = true
 						}
 
-						res, err := solve(ctx, c, req)
+						solveReq := req
+						if opt.CallFunc != nil {
+							// keep the initial request untouched, it may have already
+							// been sent for the linked result above
+							solveReq.FrontendOpt = maps.Clone(req.FrontendOpt)
+							if _, ok := solveReq.FrontendOpt["frontend.caps"]; !ok {
+								solveReq.FrontendOpt["frontend.caps"] = "moby.buildkit.frontend.subrequests+forward"
+							} else {
+								solveReq.FrontendOpt["frontend.caps"] += ",moby.buildkit.frontend.subrequests+forward"
+							}
+							solveReq.FrontendOpt["requestid"] = "frontend." + opt.CallFunc.Name
+						}
+
+						res, err := solve(ctx, c, solveReq)
 						if err != nil {
 							return nil, err
 						}
@@ -625,11 +654,12 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
 							callRes = res.Metadata
 						}
 
-						rKey := resultKey(dp, k)
-						results.Set(rKey, res)
+						if !linked {
+							results.Set(rKey, res)
+						}
 
 						forceEval := false
-						if children := childTargets[rKey]; len(children) > 0 {
+						if len(children) > 0 {
 							// wait for the child targets to register their LLB before evaluating
 							_, err := results.Get(ctx, children...)
 							if err != nil {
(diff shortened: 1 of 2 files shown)


crazy-max (approved):
LGTM thanks

Issue #2668, opened 2024-08-31
buildx kubernetes driver sometimes returns `ERROR: error dialing backend: remote error: tls: internal error`


dcherniv:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

On EKS 1.29 specifically using ARM64 nodes, it seems that there exists a race condition somewhere where the node CSR is not signed yet but the node is reported as ready in the cluster. 
The issue goes away after a number of seconds once the CSR is approved and issued.
However, during that time all calls to the pods situated on the node return the above mentioned error. 
Buildx kubernetes driver specifically returns this:
```
ERROR: error dialing backend: remote error: tls: internal error
ERROR: context deadline exceeded

NAME/NODE                             DRIVER/ENDPOINT                                                                                                               STATUS    BUILDKIT   PLATFORMS
buildx-php-release-8.2-2b0efca        kubernetes                                                                                                                                         
 \_ buildx-php-release-8.2-2b0efca0    \_ kubernetes:///buildx-php-release-8.2-2b0efca?deployment=buildkit-8e44b51e-ee72-4a03-873b-8fa232cf24d9-33fdb&kubeconfig=   running   v0.15.2    linux/amd64*
 \_ buildx-php-release-8.2-2b0efca1    \_ kubernetes:///buildx-php-release-8.2-2b0efca?deployment=buildkit-b4c6a857-d35e-4171-b4d3-0e38e1749975-05a58&kubeconfig=   error                linux/arm64*
Failed to get status for buildx-php-release-8.2-2b0efca (buildx-php-release-8.2-2b0efca1): listing workers: failed to list workers: DeadlineExceeded: context deadline exceeded
default*                              docker                                                                                                                                             
 \_ default                            \_ default                                                                                                                   running   v0.15.2    linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/386
```
I was browsing the buildx source code and found the call to list the workers: https://github.com/docker/buildx/blob/master/vendor/github.com/moby/buildkit/client/workers.go#L31 
But it's not clear what the retry logic is here. It seems to me when we get the above tls internal error, the code just dies and buildx quits.
[...]

PR #3995, opened 2026-08-04
kubernetes: jitter dial backoff so builders do not retry in lockstep

1991santhu:
## The problem

`calculateBackoff` is a pure function of the attempt number:

```go
return min(time.Duration(1<<uint(attempt))*baseDelay, maxDelay)
```

so every builder retrying the same condition waits for exactly 500ms, 1s, 2s, 4s, 8s.

That matters specifically because of the condition this retry exists to handle. From the comment on `tryWithBackoff`:

> // This handles the race condition where Kubernetes marks nodes as "Ready" before their
> // Certificate Signing Requests (CSRs) are approved, causing transient TLS errors.

CSR approval lag is a cluster-wide, time-correlated event rather than a per-pod accident. When a node group comes up, every build scheduled onto those nodes hits the transient TLS error at roughly the same moment — and then retries at identical instants. In CI, where parallel builds are the normal case, that concentrates retries on the API server precisely while it is still working through the approval backlog.

The backoff limits how often each builder retries. It does not stop builders retrying together.

## The change

Equal jitter — retain half the computed interval as a floor, randomise the remainder:

```go
d := min(time.Duration(1<<uint(attempt))*baseDelay, maxDelay)
half := d / 2

return half + time.Duration(rand.Int63n(int64(d-half)+1))
```

Half is kept rather than using full jitter so a retry is never issued immediately after a TLS failure. The result never exceeds `maxDelay`, so no builder waits longer than it does today — the change only removes the alignment.

## Testing

`go test -count=2 ./driver/kubernetes/...` passes, `go vet` clean.

The package had no test file, so this adds one covering: the result stays within `[d/2, d]` for attempts 0–5 across 500 samples each; a large attempt count stays capped at `maxDelay`; and successive calls vary, which fails if the implementation regresses to deterministic.

## Note

`math/rand` rather than `crypto/rand` — this is load spreading, not a security boundary. Happy to change if the project prefers otherwise. Context for the original retry: #2668.

### DIFF
--- a/driver/kubernetes/driver.go
+++ b/driver/kubernetes/driver.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	stderrors "errors"
 	"fmt"
+	"math/rand"
 	"net"
 	"strings"
 	"syscall"
@@ -389,9 +390,37 @@ func isTransientConnectionError(err error) bool {
 	return false
 }
 
-// calculateBackoff calculates the delay for the given attempt with exponential backoff.
+// calculateBackoff calculates the delay for the given attempt with exponential
+// backoff and additive jitter, drawing from [d, 2d] capped by maxDelay, where d
+// is the exponential value for the attempt. The exponential value is the floor
+// rather than the midpoint, so a retry is never issued sooner than the schedule
+// would have on its own.
+//
+// The exponential component alone is a pure function of the attempt number, so
+// every builder retrying the same condition waits for exactly the same durations.
+// That matters for the case this backoff exists to handle: CSR approval lagging
+// node readiness is a cluster-wide event, so concurrent builds scheduled onto
+// newly-ready nodes hit the transient TLS error at the same moment and would
+// then retry in unison, concentrating load on the API server while it is already
+// working through the approval backlog.
+//
+// The jitter is added to the interval rather than centred on it, so a retry is
+// never issued sooner than the exponential schedule intended. Centring it would
+// let the first retry fire after baseDelay/2, undercutting a configured minimum
+// while the API server is still working through the CSR backlog. The extra is
+// bounded by the remaining headroom so the result never exceeds maxDelay.
 func calculateBackoff(attempt int, baseDelay, maxDelay time.Duration) time.Duration {
-	return min(time.Duration(1<<uint(attempt))*baseDelay, maxDelay)
+	d := min(time.Duration(1<<uint(attempt))*baseDelay, maxDelay)
+
+	extra := d
+	if headroom := maxDelay - d; headroom < extra {
+		extra = headroom
+	}
+	if extra <= 0 {
+		return d
+	}
+
+	return d + time.Duration(rand.Int63n(int64(extra)+1)) // #nosec G404 -- no strong randomness required for retry jitter
 }
 
 func (d *Driver) Client(ctx context.Context, opts ...client.ClientOpt) (*client.Client, error) {
--- a/driver/kubernetes/driver_test.go
+++ b/driver/kubernetes/driver_test.go
@@ -1,0 +1,58 @@
+package kubernetes
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCalculateBackoffNeverShorterThanSchedule(t *testing.T) {
+	const (
+		baseDelay = 500 * time.Millisecond
+		maxDelay  = 10 * time.Second
+	)
+
+	for attempt := range 6 {
+		schedule := min(time.Duration(1<<uint(attempt))*baseDelay, maxDelay)
+		ceiling := min(2*schedule, maxDelay)
+
+		for range 500 {
+			got := calculateBackoff(attempt, baseDelay, maxDelay)
+			require.GreaterOrEqual(t, got, schedule,
+				"attempt %d must never wait less than the exponential schedule", attempt)
+			require.LessOrEqual(t, got, ceiling,
+				"attempt %d must not exceed twice the schedule, nor maxDelay", attempt)
+		}
+	}
+}
+
+func TestCalculateBackoffRespectsMaxDelay(t *testing.T) {
+	const (
+		baseDelay = 500 * time.Millisecond
+		maxDelay  = 10 * time.Second
+	)
+
+	for range 500 {
+		require.LessOrEqual(t, calculateBackoff(20, baseDelay, maxDelay), maxDelay,
+			"a large attempt count must stay capped at maxDelay")
+	}
+}
+
+func TestCalculateBackoffVaries(t *testing.T) {
+	const (
+		baseDelay = 500 * time.Millisecond
+		maxDelay  = 10 * time.Second
+	)
+
+	seen := make(map[time.Duration]struct{})
+	for range 500 {
+		seen[calculateBackoff(3, baseDelay, maxDelay)] = struct{}{}
+	}
+
+	// A deterministic implementation returns a single value. The range at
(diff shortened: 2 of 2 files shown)


1991santhu:
Corrected the direction of the jitter in my own patch.

It was equal jitter, `[d/2, d]`. At attempt 0 that lets a retry fire at 250ms when `baseDelay` is 500ms, so it hits the API server sooner than configured while it's still working through the CSR backlog. It adds now, `[d, 2d]`, bounded by the headroom to `maxDelay`. With `maxRetries=5` the delays used are 500ms through 4s, so the 10s cap never comes into play.

I ran the numbers rather than just asserting the problem. 100 builders hitting the same CSR lag: without jitter all 100 dial inside the same 100ms window on every attempt, with it about 28 on the first retry and 7 by the fourth. Simulated arrival times, not something I've measured on a real cluster.

`go test -count=2 -run TestCalculateBackoff ./driver/kubernetes/` passes.

driver/kubernetes/driver.go
```diff
@@ -389,9 +390,34 @@ func isTransientConnectionError(err error) bool {
...
+
+	extra := d
+	if headroom := maxDelay - d; headroom < extra {
+		extra = headroom
+	}
+	if extra <= 0 {
+		return d
+	}
+
+	return d + time.Duration(rand.Int63n(int64(extra)+1))
```
crazy-max: The new `math/rand` call should use the same gosec suppression pattern as `podchooser.RandomPodChooser`, since retry jitter doesn't need crypto randomness:

```go
return d + time.Duration(rand.Int63n(int64(extra)+1)) // #nosec G404 -- no strong randomness required for retry jitter
```

crazy-max (changes requested):
The direction makes sense to me.

One small wording nit: this is additive jitter, `[d, 2d]` capped by `maxDelay`, not equal jitter anymore. That behavior is fine if preserving the current exponential delay as the floor is intentional, but the description/comment should call it that directly.

Also, please squash the commits before merge.

1991santhu:
Thanks, all three done.

Switched to the same pattern as `podchooser.RandomPodChooser`:

```go
return d + time.Duration(rand.Int63n(int64(extra)+1)) // #nosec G404 -- no strong randomness required for retry jitter
```

You're right about the wording. The doc comment said "backoff and jitter" without saying which kind, which is worse than useless here since the whole point of the change is that the jitter is additive rather than centred. It now says so directly: additive jitter drawing from `[d, 2d]` capped by `maxDelay`, with the exponential value as the floor rather than the midpoint. That is intentional, for the reason the rest of the comment gives, so a retry can never fire sooner than the schedule alone would have.

Squashed to one commit.

Verified with golangci-lint v2.8.0 built from source against Go 1.26, matching how the lint job builds it: 0 issues. `go test -count=2 -run TestCalculateBackoff ./driver/kubernetes/` passes.

crazy-max:
@chagui if you can take a look :pray:

driver/kubernetes/driver_test.go
```diff
@@ -0,0 +1,58 @@
...
+		maxDelay  = 10 * time.Second
+	)
+
+	seen := make(map[time.Duration]struct{})
+	for range 500 {
+		seen[calculateBackoff(3, baseDelay, maxDelay)] = struct{}{}
+	}
+
+	// A deterministic implementation returns a single value. The range at
+	// attempt 3 is two seconds wide, so one distinct value across 500 draws
```
chagui, line 54: Nit: at attempt 3, d is 4s, so the additive-jitter range is [4s, 8s], four seconds wide not two. I think this comment is left over from the earlier equal-jitter implementation. The test itself looks correct.

```suggestion
	// attempt 3 is four seconds wide, so one distinct value across 500 draws
```

driver/kubernetes/driver.go
```diff
@@ -389,9 +390,37 @@ func isTransientConnectionError(err error) bool {
...
+
+	extra := d
+	if headroom := maxDelay - d; headroom < extra {
+		extra = headroom
+	}
+	if extra <= 0 {
+		return d
+	}
+
+	return d + time.Duration(rand.Int63n(int64(extra)+1)) // #nosec G404 -- no strong randomness required for retry jitter
```
chagui: suggestion: took me some time to understand the function, I think we could simplify the code and use more explicit names to improve readability:

```suggestion
  	delay := min(time.Duration(1<<uint(attempt))*baseDelay, maxDelay)
  	jitterRange := min(delay, maxDelay-delay)
  	if jitterRange <= 0 {
  		return delay
  	}

  	return delay + time.Duration(rand.Int63n(int64(jitterRange)+1)) // #nosec G404 -- no strong randomness required for retry jitter
```

chagui, line 423: suggestion: took me some time to understand the function, I think we could simplify the code and use more explicit names to improve readability:


```suggestion
	delay := min(time.Duration(1<<uint(attempt))*baseDelay, maxDelay)
	jitterRange := min(delay, maxDelay-delay) // never exceed maxDelay
	if jitterRange <= 0 {
		return delay
	}
	// Floored at delay so jitter never issues a retry sooner than
	// plain exponential backoff would have.
	return delay + time.Duration(rand.Int63n(int64(jitterRange)+1)) // #nosec G404 -- no strong randomness required for retry jitter
```

driver/kubernetes/driver.go
```diff
@@ -389,9 +390,37 @@ func isTransientConnectionError(err error) bool {
...
+// rather than the midpoint, so a retry is never issued sooner than the schedule
+// would have on its own.
+//
+// The exponential component alone is a pure function of the attempt number, so
+// every builder retrying the same condition waits for exactly the same durations.
+// That matters for the case this backoff exists to handle: CSR approval lagging
+// node readiness is a cluster-wide event, so concurrent builds scheduled onto
+// newly-ready nodes hit the transient TLS error at the same moment and would
+// then retry in unison, concentrating load on the API server while it is already
+// working through the approval backlog.
```
chagui, line 405: duplicates `tryWithBackoff` (few lines above: https://github.com/1991santhu/buildx/blob/fix%2Fkubernetes-dial-backoff-jitter/driver/kubernetes/driver.go#L316-L318) docs, attaching the TLS/CSR narrative to `calculateBackoff` doesn't really make sense here because calculateBackoff only needs to justify its own local choice (floor vs. midpoint).

driver/kubernetes/driver.go
```diff
@@ -389,9 +390,37 @@ func isTransientConnectionError(err error) bool {
 	return false
 }
 
-// calculateBackoff calculates the delay for the given attempt with exponential backoff.
+// calculateBackoff calculates the delay for the given attempt with exponential
+// backoff and additive jitter, drawing from [d, 2d] capped by maxDelay, where d
+// is the exponential value for the attempt. The exponential value is the floor
+// rather than the midpoint, so a retry is never issued sooner than the schedule
+// would have on its own.
```
chagui, line 397: suggestion: doc comment is much longer than this package's convention and duplicates the caller's documentation. I'd leave the rationale behind the retries to `tryWithBackoff` and focus on  `calculateBackoff` contract and non-obvious implementation choice (floor vs. midpoint).

See http://go.dev/doc/comment
> Doc comments should not explain internal details such as the algorithm used in the current implementation. Those are best left to comments inside the function body.

```suggestion
// calculateBackoff returns a randomized exponential backoff delay for attempt,
// never exceeding maxDelay.
```
(and remove everything below this to keep it to 2 lines)

chagui (commented):
Left some suggestions.
I could not find any occurence of the problem in our environment but the PR is defensively reasonable 👍

PR #3982, opened 2026-07-30
vendor: github.com/moby/go-archive v0.3.0

thaJeztah:
full diff: https://github.com/moby/go-archive/compare/v0.2.1...v0.3.0

v0.3.0

This release fixes CVE-2026-17106 / GHSA-hfg8-hc9c-6c3h, where a crafted tar archive could use links to cause extraction operations to create or overwrite files outside the intended destination directory.

The issue affected Unpack, UnpackLayer, Untar, UntarUncompressed, and the ApplyLayer helpers. Users should upgrade and avoid extracting untrusted archives with earlier versions.

What's Changed

* archive: harden tar extraction against path traversal
* archive: do not follow reparse points in chtimes
* archive: fix creation time updates on Windows
* archive: minor cleanups and godoc touch-up
* archive: RebaseArchiveEntries: fix archive path rebasing

```
Update  github.com/moby/go-archive to v0.3.0 to fix CVE-2026-17106 / GHSA-hfg8-hc9c-6c3h
```

PR #3996, opened 2026-08-05
driver: use stdlib tar for docker-container config files

crazy-max:
closes https://github.com/docker/buildx/pull/3982

This replaces the docker-container driver's config archive path with a focused stdlib tar implementation. The driver now builds the `/etc/buildkit` config archive directly from the in-memory file map before copying it into the BuildKit container.

The first commit removes the direct use of `github.com/moby/go-archive` for docker-container config copies and introduces a small tar writer based on `archive/tar`. The second commit removes the temporary directory staging step, validates archive paths before writing tar headers, writes deterministic directory and file entries.

The old path used a generic archive library and a host temporary directory for a very narrow job. We only need to copy generated BuildKit config files into a container, so staging those bytes on disk and then archiving them added complexity without buying us useful behavior. The new implementation makes the contract explicit: config keys are relative slash paths under `/etc/buildkit`, unsafe names are rejected before tar headers are written, and the copy path no longer depends on host filesystem staging.


driver/docker-container/tar.go
```diff
@@ -0,0 +1,115 @@
...
+	return io.NopCloser(bytes.NewReader(buf.Bytes())), nil
+}
+
+func configTarEntries(configDir string, files map[string][]byte) ([]configTarEntry, []string, error) {
+	entries := make([]configTarEntry, 0, len(files))
+	dirs := map[string]struct{}{}
+	filePaths := make(map[string]struct{}, len(files))
+	if len(files) > 0 {
+		dirs[configDir] = struct{}{}
+	}
```
mickael-docker: configDir is the only name reaching writeConfigTar without passing configArchivePath. It's safe today only because len(files) > 0 makes it a prefix of every validated path. Would it be worth asserting it explicitly as well?

driver/docker-container/tar.go
```diff
@@ -0,0 +1,115 @@
...
+	archivePath := configDir + "/" + name
+	if !fs.ValidPath(archivePath) {
+		return "", errors.Errorf("invalid config file path %q", name)
+	}
+	return archivePath, nil
+}
+
+func writeConfigTar(tw *tar.Writer, dirs []string, entries []configTarEntry) error {
+	for _, dir := range dirs {
+		if err := tw.WriteHeader(&tar.Header{
```
mickael-docker, line 99: Should we force a format [format](https://pkg.go.dev/archive/tar#Format) here? I think Go should select the appropriate one, but something to potentially be aware of.

driver/docker-container/tar_test.go
```diff
@@ -0,0 +1,133 @@
...
+	rc, err := tarConfigFiles(map[string][]byte{
+		"certs":        []byte("file"),
+		"certs/ca.pem": []byte("ca"),
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "conflicts with directory")
+	require.Nil(t, rc)
+}
+
+func readTarEntries(t *testing.T, r io.Reader) map[string]tarEntry {
```
mickael-docker: I guess here the entries are unordered, but does order matter? (e.g. ensure parent is written before child?)

driver/docker-container/tar.go
crazy-max: Yeah, fair point. `configDir` is currently derived from `path.Base(confutil.DefaultBuildKitConfigDir)`, so it's `buildkit` today, but it's still the one path component added to the archive before going through the same explicit validation path. I will tighten that so the archive root is validated too, or refactor it so we don't pass an arbitrary root string around.

driver/docker-container/tar_test.go
crazy-max: Order shouldn't matter for Docker extraction in practice, but the implementation does write all directory entries before file entries, and both sets are sorted for deterministic output. The current test reads into a map, so it doesn't prove that ordering. I will add an explicit assertion for the tar entry sequence so this behavior is covered.

driver/docker-container/tar.go
crazy-max: I think leaving `Format` unset is the right behavior here. The stdlib writer will pick the first format that can encode the header, which keeps simple paths as simple tar headers and still allows longer generated registry paths if needed. Forcing USTAR would make some valid config paths fail, and forcing PAX would add extended headers even when they are not needed.

mickael-docker (approved):
LGTM

driver/docker-container/tar.go
```diff
@@ -0,0 +1,127 @@
...
+	configDir, err := configArchiveRoot(configDir)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	entries := make([]configTarEntry, 0, len(files))
+	dirs := map[string]struct{}{}
+	filePaths := make(map[string]struct{}, len(files))
+	if len(files) > 0 {
+		dirs[configDir] = struct{}{}
```
tonistiigi: Isn't this dead code and these dirs are added anyway below?

tonistiigi (dismissed):

tonistiigi (approved):

PR #3998, opened 2026-08-05
avoid policy dependency from builder loading

crazy-max:
This change stops `builder.LoadNodes` from constructing the default policy verifier itself, so consumers of the `builder` package no longer pull in the policy package and its OPA dependency graph.

The builder package now accepts an optional image verifier through `WithImageVerifier`, and command paths that may boot or use builder images pass the default policy verifier explicitly. The default verifier construction stays in the policy package.

The default policy is runtime opt-in, but the previous unconditional import made downstream users vendor OPA and the signature verification stack even when they did not use policy support. This keeps the buildx CLI behavior while restoring a lighter dependency boundary for builder and driver consumers.

### DIFF
--- a/driver/image_test.go
+++ b/driver/image_test.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"testing"
 
-	"github.com/docker/buildx/policy"
+	"github.com/docker/buildx/util/sourcemeta"
 	"github.com/moby/buildkit/client"
 	digest "github.com/opencontainers/go-digest"
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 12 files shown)


driver/image_test.go
```diff
@@ -36,7 +36,7 @@ func TestVerifyImageRefDigestOnly(t *testing.T) {
 	dgst := digest.FromString("buildkit")
 	called := false
 
-	pinned, applied, err := VerifyImageRef(context.Background(), nopSubLogger{}, "moby/buildkit@"+dgst.String(), nil, nil, func(_ context.Context, _ string, _ *ocispecs.Platform, _ policy.SourceMetadataResolver) (digest.Digest, error) {
+	pinned, applied, err := VerifyImageRef(context.Background(), nopSubLogger{}, "moby/buildkit@"+dgst.String(), nil, nil, func(_ context.Context, _ string, _ *ocispecs.Platform, _ *sourcemeta.Resolver) (digest.Digest, error) {
```
thaJeztah, line 39: If we're breaking the api / signature, perhaps as a follow-up also look if we need all these arguments to be positional arguments or if it could be an option-struct or variadic functional args.

thaJeztah (approved):
LGTM


worth checking if this requires changes in compose, or if it doesn't use this code.

tonistiigi (approved):

PR #4000, opened 2026-08-05
ci: add dependency update workflow

crazy-max:
This adds an `update-deps` workflow that opens dependency update pull requests for the Dockerfile versions used by the test image.

The workflow resolves the latest release tags for Docker, registry, BuildKit, Compose, Scout and Undock, updates the matching Dockerfile arguments, and creates a signed pull request for each dependency. The BuildKit update also refreshes the `test-integration` matrix so it keeps testing `master`, `latest`, `buildx-stable-1`, and the latest three stable minor BuildKit release tags.

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -1,0 +1,436 @@
+name: update-deps
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/update-deps.yml'
+      - '.github/workflows/build.yml'
+      - 'Dockerfile'
+  schedule:
+    - cron: '0 9 * * *'
+  push:
+    branches:
+      - 'master'
+
+jobs:
+  update:
+    runs-on: ubuntu-24.04
+    if: ${{ github.repository == 'docker/buildx' }}
+    timeout-minutes: 10
+    permissions:
+      contents: write
+      pull-requests: write
+    strategy:
+      fail-fast: false
+      matrix:
+        dep:
+          - docker
+          - registry
+          - buildkit
+          - compose
+          - scout
+          - undock
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      -
+        name: Install npm deps
+        run: |
+          install_dir="$RUNNER_TEMP/update-deps-node"
+          mkdir -p "$install_dir"
+
+          cat > "$install_dir/package.json" <<'EOF'
+          {
+            "name": "buildx-update-deps",
+            "version": "0.0.0",
+            "private": true,
+            "dependencies": {
+              "semver": "7.8.3"
+            }
+          }
+          EOF
+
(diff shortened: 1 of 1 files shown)


.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,296 @@
...
+
+              for (let page = 1; page <= 10; page++) {
+                const releases = await github.rest.repos.listReleases({
+                  owner,
+                  repo,
+                  page,
+                  per_page: 100
+                });
+                for (const release of releases.data) {
+                  const version = semver.parse(stripPrefix(release.tag_name, tagPrefix));
```
tonistiigi: Shouldn't this verify that the prefix matched? Just ignoring might be dangerous and use a wrong tag.

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,296 @@
...
+                owner: 'crazy-max',
+                repo: 'undock',
+                path: 'Dockerfile',
+                arg: 'UNDOCK_VERSION',
+                pattern: dockerfileArgPattern('UNDOCK_VERSION'),
+                value: stripLeadingV
+              }
+            };
+
+            const config = dependencyConfigs[dep];
```
tonistiigi, line 342: Can't this just update the defined keys, not replace all.

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,296 @@
...
+            const installDir = path.join(process.env.RUNNER_TEMP, 'update-deps-node');
+            await core.group(`Install npm deps`, async () => {
+              await exec.exec('npm', [
+                'install',
+                '--prefix',
+                installDir,
+                '--loglevel=error',
+                '--no-save',
+                '--package-lock=false',
+                '--ignore-scripts',
```
tonistiigi: This still doesn't seem to be secure to enforce it can't be remotely swapped with a script that runs inside write-token job.

Agent suggest local `semver.js` and `persist-credentials: false`.

.github/workflows/update-deps.yml
crazy-max: Updated this so prefixed tag configs now require the prefix to match before parsing. For Docker, tags that don't start with `docker-` are skipped instead of being silently parsed after a no-op strip, so an unrelated `vX.Y.Z` tag cannot win over the Docker release tags.

.github/workflows/update-deps.yml
crazy-max: Changed this to only replace existing SemVer-looking rows inside the `buildkit` matrix block. The static entries such as `master`, `latest`, and `buildx-stable-1` are left untouched, and the workflow now errors if the matrix doesn't contain exactly the expected number of release rows.

.github/workflows/update-deps.yml
crazy-max: I kept using the real `semver` module rather than adding a local parser, but moved the install to a plain shell step, install it under `$RUNNER_TEMP`, and keep checkout credentials from being persisted with `persist-credentials: false`.

I also looked at splitting this into separate read-only and write-token jobs with an artifact handoff, but that made the workflow more complex. The write-scoped job still needs to create the PR, so this keeps the script dependency explicit while avoiding a checked-in ad hoc SemVer implementation.

crazy-max:
Also added pull_request event to verify the changes: https://github.com/docker/buildx/actions/runs/31488665508/job/93769785673?pr=4000#step:5:11

PR #4003, opened 2026-08-07
Serialize log pauses

alexey-igrychev:
Concurrent Buildx progress printers call `logutil.Pause` against one global logrus logger. `Pause` read and replaced the logger output without serializing the pause intervals, which races under Go's race detector.

Serialize pauses until their buffered output is resumed, and cover concurrent calls with a race regression test.

### DIFF
--- a/util/logutil/pause.go
+++ b/util/logutil/pause.go
@@ -8,7 +8,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var pauseMu sync.Mutex
+
 func Pause(l *logrus.Logger) func() {
+	pauseMu.Lock()
 	// initialize formatter with original terminal settings
 	l.Formatter.Format(logrus.NewEntry(l))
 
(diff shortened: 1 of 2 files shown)


alexey-igrychev:
Verification:

- `CGO_ENABLED=1 go test -race ./util/logutil` passes.
- Mutation: removed the pause mutex → `TestPauseConcurrent` reports the logrus output race.
- `TESTPKGS=./util/logutil TESTFLAGS='-race -run TestPauseConcurrent' make test` builds the test image but its test container sets `CGO_ENABLED=0`, so Go rejects `-race`.

alexey-igrychev:
Closing this PR: the global mutex is held from `Pause()` until the returned resume closure runs. `progress.Printer.updateDisplay` holds that interval while it consumes solve status, normally for the entire build.

Concurrent printers therefore serialize whole builds, and blocked printers stop draining their unbuffered status channels, creating solve-progress backpressure. This is worse than the original race for consumers that build several images in parallel.

A replacement needs concurrent output buffering/multiplexing rather than a process-wide lock around the full pause lifetime.

PR #4028, opened 2026-08-20
build: avoid empty source policy sessions

crazy-max:
Remote Git contexts can enqueue an optional `Dockerfile.rego` lookup even when the build doesn't include a policy file. When that optional file was absent, Buildx resolved zero policy files but still attached a no-op `SourcePolicyProvider`, which made the solve request carry a `SourcePolicySession`. That matters for builders created with `proxyNetwork = true`, because BuildKit treats the presence of a source policy session as part of proxy-network enforcement, so a Git-context could fail with a proxy `403 Forbidden` even though the same target succeeded from a local context.

The issue is reproducible with:

```
docker buildx bake --no-cache --builder <builder> "https://github.com/crazy-max/docker-github-builder.git?ref=proxy-network&subdir=test" proxy-network
```

against a docker-container builder using `moby/buildkit:v0.32.2` and `proxyNetwork = true`.

This change keeps `SourcePolicyProvider` unset when policy resolution loads no policy files, so builds without policies don't accidentally opt into source-policy session behavior.

### DIFF
--- a/build/opt.go
+++ b/build/opt.go
@@ -712,6 +712,10 @@ func configureSourcePolicy(ctx context.Context, np *noderesolver.ResolvedNode, o
 	if err != nil {
 		return nil, err
 	}
+	if len(loadedOpts) == 0 {
+		so.SourcePolicyProvider = nil
+		return defers, nil
+	}
 	var policyFiles []string
 	for _, popt := range loadedOpts {
 		for _, f := range popt.Files {
--- a/tests/policy_bake.go
+++ b/tests/policy_bake.go
@@ -1,15 +1,22 @@
 package tests
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/containerd/continuity/fs/fstest"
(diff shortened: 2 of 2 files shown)


tonistiigi (commented):
Change looks good but the new test `testBakeRemoteGitNoPolicyWithProxyNetwork` seems to pass with or without the patch.

crazy-max:
> Change looks good but the new test `testBakeRemoteGitNoPolicyWithProxyNetwork` seems to pass with or without the patch.

Ah yes it passed without the fix because the request was plain HTTP, and the empty source policy session didn't make that path fail.

~The exec step now calls a local TLS server with `wget --no-check-certificate`. Without the fix, BuildKit returns `HTTP/1.1 403 Forbidden` before the request reaches the test server.~

I updated the test to mirror the original repro more closely by using `https://checkip.amazonaws.com/` from the exec step and asserting the proxy-network request summary contains `GET https://checkip.amazonaws.com/ -> 200`.

tonistiigi (approved):

Issue #1668, opened 2023-03-07
`docker buildx bake` does not support saving multiple targets to a single tarball


negasora:
### Contributing guidelines

- [X] I've read the [contributing guidelines](https://github.com/docker/buildx/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

Attempting to output multiple targets to a single tarball results in a single random target being contained in the tarball, instead of all of them.

### Expected behaviour

The generated tarball contains all images

### Actual behaviour

The generated tarball contains a random image

### Buildx version

github.com/docker/buildx 0.10.3 79e156beb11f697f06ac67fa1fb958e4762c0fab

### Docker info

```markdown
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  0.10.3
    Path:     /usr/lib/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  2.16.0
    Path:     /usr/lib/docker/cli-plugins/docker-compose

Server:
 Containers: 18
  Running: 1
  Paused: 0
  Stopped: 17
 Images: 691
 Server Version: 23.0.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Using metacopy: true
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 2456e983eb9e37e47538f59ea18f2043c9a73640.m
 runc version: 
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 6.2.1-arch1-1
 Operating System: Arch Linux
 OSType: linux
 Architecture: x86_64
 CPUs: 16
 Total Memory: 30.13GiB
 Name: pudge
[...]

Issue #1320, opened 2022-09-16
Bake: Add summary of why build failure happened


ciaranmcnulty:
Currently when baking lots of targets it's common for the logs to contain:

* Lots of successful build lines
* Some error
* Lots of other lines from the parallel jobs being cancelled

When looking at something like CI output (probably with `--progress=plain`) it can be hard to locate the source of the error

It would be great to have a summary that explained clearly:

 * Which target failed
 * Which stage failed
 * What line failed
 * What the error output was (in the case of a `RUN` failure)

Importantly, it would be great to have this at the very end of the output just before buildx exits

ciaranmcnulty:
I'm not sure if there's a case where two failures could happen at the same time? If so we should just summarise both

Ocramius:
Would be good to have the `STDERR`/`STDOUT` of the failed step highlighted, ideally with some sort of delimiter before/after.

Looking into build failures is indeed not really fun, when it becomes an exercise in eye squinting :D

PR #4044, opened 2026-08-31
bake: add execution mode

### LINKED ISSUE #1089: Option to only tag images at end of bake

### LINKED ISSUE #3297: Add a flag to defer errors occurring in matrix build targets
### Description

I have a large matrix build emitting a number of products from multiple sequential 5+ hour build stages.

If one of these matrix targets fails for whatever reason, I'd like the others to run to completion or failure.

Currently, the entire build process dies upon the first error encountered in any individual matrix target, throwing away a bunch of valuable work for other running matrix targets.

I'd like a flag, ```docker buildx bake --defer-exit``` (or whatever) that enables this behavior.

### LINKED ISSUE #3428: add "--continue-on-error" option
### Description

We use a bake file that has hundreds of targets. These targets are all small variations of some common base images with different package versions installed. We would like to rebuild those images from time to time to get updates for the base packages.

However, we have repeatedly hit the issue that some random part of this process fails, for example because some Ubuntu packages archive for arm64 was just updated at the wrong time, some other package mirror is unavailable, etc... This then means that the whole build aborts, because one specific variant could not be build.

For us, it would be acceptable if a failure to build one image does not abort the whole build, but lets the other, unaffected builds continue as normal. We could then add some logic to retry the failed items at a later time.

Would it be possible to implement such a "--continue-on-error" option?

### LINKED ISSUE #3989: CLI flag to limit the number of concurrently building targets in buildx bake / docker compose build
## Description

There is currently **no per-invocation, command-line way to limit build concurrency** when building multiple targets with `docker buildx bake` (or `docker compose build`, which since Compose v5.0.0 delegates all builds to Bake). All targets are dispatched to the builder at once, and the only existing control (`max-parallelism` in the BuildKit worker configuration) must be baked into the builder at creation time.

I am requesting a **client-side concurrency limit**, e.g.:

```bash
docker buildx bake --max-parallelism 2 -f docker-compose.yml
docker compose build --max-parallelism 2
```

that caps how many **targets (build requests)** are in flight simultaneously, independent of which driver or daemon configuration is in use.

## Background / motivating use case

CI pipeline (Jenkins on Kubernetes, ephemeral Docker-in-Docker pods) building multi-service applications from a Compose file. Each service is a multi-stage Dockerfile with resource-heavy compile steps (webpack/node, .NET, etc.).

Different applications sharing the same pipeline have very different resource profiles, so the desired parallelism is a **per-project, per-invocation value** supplied by the CI system — not a property of the builder or the daemon.

With the historical Compose builder this was trivial:

```bash
COMPOSE_PARALLEL_LIMIT=2 docker compose build
```

That capability has been progressively removed:

1. Compose ≥ v2.34 defaults to the Bake backend; `COMPOSE_PARALLEL_LIMIT` / `--parallel` no longer affect builds (docker/compose#13043, docker/desktop-feedback#584).
2. `COMPOSE_BAKE=false` was deprecated in Compose v2.39.0 (docker/compose#13065).
3. Compose v5.0.0 removed the internal builder entirely (docker/compose#13056), so the last escape hatch is gone.

There is now **no supported equivalent** of the old behavior anywhere in the CLI surface.

## Current workaround and why it is insufficient

The documented approach is to set worker-level parallelism when creating a builder:

```bash
docker buildx create --name ci --driver docker-container \
  --buildkitd-flags '--oci-max-parallelism=2' --bootstrap --use
```

It has significant drawbacks:

- **Not per-invocation.** The value is frozen at builder-creation time. Changing it means destroying and recreating the builder (losing its local cache) or pre-provisioning one builder per concurrency tier and selecting with `--builder`.
- **Not available on the default `docker` driver.** The BuildKit instance embedded in dockerd exposes no `max-parallelism` in `daemon.json` and no dockerd flag. Users on the default driver have *no* option at all.
- **Forces the `docker-container` driver in environments where its costs are highest.** In ephemeral CI (DinD pods) this means creating a builder per job and paying the `--load` export/import penalty (minutes for large images) or restructuring pipelines around `--push`, purely to obtain a concurrency limit.

### 1. `--max-parallelism <N>` on `docker buildx bake` (and `build` for completeness)

Client-side scheduling: Bake already computes the target dependency graph (`contexts = { x = "target:y" }`); instead of dispatching all ready targets at once, dispatch at most N concurrently. Because this is done in the CLI before solves are submitted:

- it works with **every driver**, including the default `docker` driver;
- it requires **no daemon reconfiguration or builder lifecycle changes**;
- it composes with (does not replace) worker-level `max-parallelism`, which remains the right tool for capping *step* concurrency inside a single large multi-stage build.

### 2. Bake-file attribute for per-group/per-target defaults

```hcl
group "default" {
  targets = ["app-a", "app-b", "app-c"]
  max-parallelism = 2
}
```

with the CLI flag taking precedence, consistent with how other bake attributes are overridden via `--set`.

### 3. Compose passthrough

`docker compose build --max-parallelism N` (and/or reviving `COMPOSE_PARALLEL_LIMIT` mapped onto it), restoring pre-Bake parity now that Compose v5 has no internal builder.

## Prior art / related issues

- docker/buildx#2581 — Add concurrency control to matrix builds (open; proposes `--max-parallel` and a per-target `concurrency` attribute)
- moby/buildkit#6038 — Add a per-build flag to limit concurrency/parallelism (open; notes the docker/for-win#2049 daemon-wide config "does not meet this need")
- moby/buildkit#5981 — Specify max parallelism for multi-stage build (open)
- docker/buildx#3006 — Limit concurrent build requests on a single BuildKit pod (open; worker `max-parallelism` does not cap accepted build requests)
- docker/compose#13043 — `COMPOSE_PARALLEL_LIMIT` no longer limiting concurrent builds
- docker/desktop-feedback#584 — All parallelism-limiting mechanisms ineffective once Bake is the builder (100 services building simultaneously)
- moby/buildkit#2049 — original daemon-wide `max-parallelism` config (shipped; explicitly daemon-scoped)

The recurring theme across these issues: the only shipped control is daemon-scoped and creation-time-static, while the demand is for an invocation-scoped, client-side limit.

## Expected behavior

```bash
docker buildx bake --max-parallelism 1 -f compose.yml svc-a svc-b
```

builds `svc-a` and `svc-b` strictly sequentially (subject to dependency order), on any driver, without any builder or daemon reconfiguration. Default remains unlimited (current behavior), so the flag is fully backward compatible.

## Environment where this was hit

- Docker Engine 28.3.x (DinD), Buildx ≥ v0.23, Compose v5.3.1
- Jenkins on Kubernetes, ephemeral DinD build pods, one builder per job

crazy-max:
- fixes #1089
- fixes #3297
- fixes #3428
- fixes #3989
- closes #1197

This carries forward the synchronized-output work from #1197 and adds explicit execution modes and target concurrency control to Bake. `--execution=fail-fast` remains the default, preserving cancellation of other targets when a build fails.

`--execution=defer-output` waits until all participating targets have successfully evaluated their build results before allowing exports to begin. This addresses the premature publication scenario in #1089, including failures discovered during evaluation rather than Dockerfile parsing. Exports are not transactional: export failures can still leave partial output, completed exports are not rolled back, and Buildx-side multi-node manifest merging and pushing remain outside the synchronization barrier. It does not combine multiple targets into a shared archive, as requested in #1668.

`--execution=defer-error` lets independent targets finish even when another target fails, covering the workflows in #3297 and #3428. Bake still exits with an error, but successful targets can export their outputs and retain their entries in `--metadata-file`. Failed multi-target builds end with an `[internal] target results` progress step distinguishing succeeded, failed, aborted, and unstarted targets. These outcomes are also available through `--progress=rawjson`; quiet mode remains silent. This partially addresses #1320, without adding its broader proposed summary of failing stages, source lines, and command logs.

`--jobs=N`, also available as `-j=N`, controls target concurrency independently of the execution mode. `-j=1` runs independent targets sequentially, while `0` retains unlimited concurrency. This provides the per-invocation Bake control requested in #3989 without changing BuildKit internal step parallelism or requiring builder reconfiguration. Bake-file defaults and Compose passthrough are outside this change. With deferred output or linked targets, a nonzero limit smaller than the total participating target count is rejected to avoid deadlocks; that count includes implicit targets referenced through `target:` contexts.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -465,6 +465,7 @@ type (
 const (
 	ExecutionModeFailFast   ExecutionMode = "fail-fast"
 	ExecutionModeSyncOutput ExecutionMode = "sync-output"
+	ExecutionModeDeferError ExecutionMode = "defer-error"
 )
 
 func newLinkedTargetState(parents, children map[string][]string) *linkedTargetState {
@@ -614,7 +615,12 @@ func Build(ctx context.Context, nodes []builder.Node, opts map[string]Options, d
 		return nil, err
 	}
 
-	eg, ctx := errgroup.WithContext(ctx)
+	var eg *errgroup.Group
+	if bh != nil && bh.Execution.Mode == ExecutionModeDeferError {
+		eg = &errgroup.Group{}
+	} else {
+		eg, ctx = errgroup.WithContext(ctx)
+	}
 	reqForNodes, release, err := newBuildRequests(ctx, docker, cfg, drivers, w, opts)
 	if err != nil {
 		return nil, err
@@ -1381,9 +1387,12 @@ func waitContextDeps(ctx context.Context, node *noderesolver.ResolvedNode, resul
 		if !ok {
 			continue
 		}
+		if err, ok := r.(error); ok {
+			return err
+		}
 		rr, ok := r.(*gateway.Result)
 		if !ok {
-			return errors.Errorf("invalid result type %T", rr)
+			return errors.Errorf("invalid result type %T", r)
 		}
 		if so.FrontendAttrs == nil {
 			so.FrontendAttrs = map[string]string{}
--- a/build/build.go
+++ b/build/build.go
@@ -6,6 +6,7 @@ import (
 	_ "crypto/sha256" // ensure digests can be computed
 	"encoding/base64"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"maps"
@@ -433,9 +434,14 @@ func toRepoOnly(in string) (string, error) {
 }
 
 type (
+	ExecutionMode string
+	Execution     struct {
+		Mode ExecutionMode
+	}
 	EvaluateFunc func(ctx context.Context, name string, c gateway.Client, res *gateway.Result, opt Options) error
 	Handler      struct {
-		Evaluate EvaluateFunc
+		Evaluate  EvaluateFunc
+		Execution Execution
 	}
 	linkedTargetState struct {
 		results   *waitmap.Map
@@ -444,6 +450,21 @@ type (
 		parents   map[string][]string
 		children  map[string][]string
 	}
+	linkedTargetHooks struct {
+		preEvaluate  func() error
+		evaluate     func() error
+		postEvaluate func() error
+	}
+	syncTargetState struct {
+		targets   []string
+		results   *waitmap.Map
+		evaluated *waitmap.Map
+	}
+)
+
+const (
+	ExecutionModeFailFast   ExecutionMode = "fail-fast"
+	ExecutionModeSyncOutput ExecutionMode = "sync-output"
 )
 
 func newLinkedTargetState(parents, children map[string][]string) *linkedTargetState {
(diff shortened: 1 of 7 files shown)


build/build.go
```diff
@@ -538,7 +636,31 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
 	// loop below, before any goroutines are spawned; no mutex needed.
 	linkedClients := make(map[string]*client.Client)
 
-	for k, opt := range opts {
+	var syncState *syncTargetState
+	if bh != nil && bh.Execution.Mode == ExecutionModeSyncOutput {
+		if bh.Execution.Parallel > 0 && bh.Execution.Parallel < len(opts) {
+			return nil, errors.Errorf("sync-output execution requires parallelism to be unlimited or at least the number of targets")
+		}
```
crazy-max, line 670: I kept `sync-output` compatible with `parallel=N` only when the limit is unlimited or at least the number of selected targets. A smaller limit can deadlock because `sync-output` waits for every target to reach the output boundary together.

Does it match what we want? Or should `sync-output` reject any explicit parallel limit for now?

build/build.go
```diff
@@ -538,7 +636,31 @@ func BuildWithResultHandler(ctx context.Context, nodes []builder.Node, opts map[
...
+		// every ref evaluation before exporters can run, so output is only written
+		// after all targets have reached the output boundary successfully.
+		syncState = newSyncTargetState(opts, drivers)
+	}
+
+	var targetLimit chan struct{}
+	if bh != nil && bh.Execution.Parallel > 0 {
+		if bh.Execution.Parallel < len(opts) && linkedTargets.hasLinks() {
+			return nil, errors.Errorf("limited parallelism is not supported with linked targets")
+		}
```
crazy-max, line 681: I rejected limited target parallelism when linked targets are present because linked target evaluation can require parent and child targets to be alive together.

Does it make sense for this first version or should linked targets be allowed when the requested limit is high enough to cover each linked group?

commands/bake.go
```diff
@@ -561,6 +573,7 @@ func bakeCmd(dockerCli command.Cli, rootOpts *rootOptions) *cobra.Command {
 	flags.StringArrayVar(&options.vars, "var", nil, `Set a variable value (e.g., "name=value")`)
 	flags.StringVar(&options.callFunc, "call", "build", `Set method for evaluating build ("check", "outline", "targets")`)
 	flags.StringArrayVar(&options.allow, "allow", nil, "Allow build to access specified resources")
+	flags.StringVar(&options.execution, "execution", "fail-fast", `Set target execution behavior (format: "mode[,parallel=N]")`)
```
crazy-max, line 579: I used `--execution` as the flag for target failure behavior, output synchronization, and target parallelism, but I'm not fully convinced this is the best name. Any ideas?

cc @dvdksn

tonistiigi: `--jobs` (also `-j N`), `--strategy` (Ansible).

PR #4058, opened 2026-09-04
build: fix partial multi-node image pushes

tomasperezmolina:
## Why

A failed multi-node registry build must not leave the requested tag pointing to a single-platform image. Commit [`de05d88f`](https://github.com/docker/buildx/commit/de05d88f6053ae750e22ce5c2a22ebea8f131128), `build: let drivers prepare solve options`, correctly gave each node an independent copy of its exporter attributes, exposing code that still depended on those attributes being shared. Only one node was switched to a repository-only digest push; another node could publish the final tag before every platform completed.

## Reproduction

Assume `repro` is a two-node builder where the first node handles `linux/arm64` and the second handles `linux/amd64`. Start with a tag that does not exist.

```dockerfile
FROM alpine
ARG TARGETARCH
RUN if [ "$TARGETARCH" = arm64 ]; then sleep 30; exit 1; fi
RUN echo "$TARGETARCH" > /platform
```

```console
docker buildx build \
  --builder repro \
  --platform linux/arm64,linux/amd64 \
  --tag registry.example/repro:test \
  --push .

# The build fails after the amd64 node has finished exporting.
docker buildx imagetools inspect registry.example/repro:test
```

On an affected version, the tag exists and contains only `linux/amd64`, even though the requested multi-platform build failed. Buildx never reaches the final manifest-list assembly because the arm64 solve failed.

## Fix

Before 0.37, every node shared the same exporter attribute map, so rewriting the first node implicitly rewrote the others. Once those maps were cloned, the target-wide `pushNames` guard caused later nodes to skip the rewrite.

`BuildWithResultHandler` loops over `dps`, one entry per node, and calls `prepareMultiDriverExports` with that node's `SolveOpt` on every iteration. Inside the helper, `pushPrepared` is local and therefore resets for each node, while `pushNames` remains target-wide and captures the final tag only once. Each node's first pushed image exporter is consequently changed to a repository-only digest push. Buildx publishes the requested tag only after every node succeeds and the manifest list is assembled. Additional image exporters keep their existing behavior.

## Notes

The first commit only extracts the existing exporter preparation. The second commit changes its behavior and adds the unit regression test. The third commit adds a `remote+multinode` integration test for the failed-push behavior.

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -433,29 +433,35 @@ func toRepoOnly(in string) (string, error) {
 }
 
 func prepareMultiDriverExports(so *client.SolveOpt, pushNames *string, insecurePush *bool) error {
-	for i, e := range so.Exports {
+	var pushPrepared bool
+	for i := range so.Exports {
+		e := &so.Exports[i]
 		switch e.Type {
 		case "oci", "tar":
 			return errors.Errorf("%s for multi-node builds currently not supported", e.Type)
 		case "image":
-			if *pushNames == "" && e.Attrs["push"] != "" {
-				if ok, _ := strconv.ParseBool(e.Attrs["push"]); ok {
-					*pushNames = e.Attrs["name"]
-					if *pushNames == "" {
-						return errors.Errorf("tag is needed when pushing to registry")
-					}
-					names, err := toRepoOnly(e.Attrs["name"])
-					if err != nil {
-						return err
-					}
-					if ok, _ := strconv.ParseBool(e.Attrs["registry.insecure"]); ok {
-						*insecurePush = true
-					}
-					e.Attrs["name"] = names
-					e.Attrs["push-by-digest"] = "true"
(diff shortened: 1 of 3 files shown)


crazy-max (commented):
I pushed an additional integration test commit to my fork that should be useful for this PR: https://github.com/crazy-max/buildx/commit/64818ad1acb0354042b99627bde89d660ce438ac

It adds a `remote+multinode` regression test that intentionally fails one platform during a multi-platform `--push`, then verifies the requested tag was not published. Since I can't push to the PR branch directly, please cherry-pick it if that looks good to you.

tomasperezmolina:
Thanks! Cherry-picked as 4b2dbec82bd89be9b56820892bd12297a235828a.

crazy-max (approved):
LGTM thanks!

PR #4068, opened 2026-09-09
policy: normalize HTTP schemes, host casing, and default ports

### LINKED ISSUE #4061: OPA policies should normalize input.http.host for well-known ports
### Description

As requested by @tonistiigi I am opening this as a feature request here.

I am currently playing with the buildkit exec proxy and depending on the tools I am using (for instance curl), the requested URLs include the default HTTP(S) ports.

As an example the following Dockerfile and policy fails:
```
FROM ubuntu:26.04
RUN apt-get update -y && apt-get install -y curl
ARG CACHE_BUSTER
RUN curl -qf "https://example.com"
```
policy:
```
package docker

default allow := false

allowed_hosts := [
    "security.ubuntu.com",
    "archive.ubuntu.com",
    "example.com"
]

allow if input.local
allow if input.image

allow if {
  input.http.host in allowed_hosts
}

decision := {
	"allow": allow,
	"caps": {"exec.proxy": true},
}
```

build via a buildkit builder with network proxy enabled:
```
docker buildx build --builder=buildkitd "--build-arg=CACHE_BUSTER=$(date)" --progress=plain .
```

and observe the following failure:
```
#7 [3/3] RUN curl -qf "https://example.com"
#7 0.086   % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
#7 0.086                                  Dload  Upload  Total   Spent   Left   Speed
  0      0   0      0   0      0      0      0                              0
#7 0.103 curl: (22) The requested URL returned error: 403
#7 ERROR: process "/bin/sh -c curl -qf \"https://example.com\"" did not complete successfully: exit code: 22

#1 loading policies Dockerfile.rego
#1 0.196 checking policy for source local://context
#1 0.200 policy decision for source local://context: ALLOW
#1 0.209 checking policy for source docker-image://docker.io/library/ubuntu:26.04@sha256:2260313b31c8c011cd2eebe728008efac1b3982be73eb71348ea2648d2c0e09b (linux/amd64)
#1 0.216 policy decision for source docker-image://docker.io/library/ubuntu:26.04@sha256:2260313b31c8c011cd2eebe728008efac1b3982be73eb71348ea2648d2c0e09b (linux/amd64): ALLOW
#1 0.323 checking policy for source https://example.com:443/
#1 0.328 policy decision for source https://example.com:443/: DENY
#1 DONE 0.3s
```

If I change the policy to include `example.com:443` as `allowed_host` then it works. I think it would be great if the default ports of 80 and 443 are stripped. If a non-default port is used it could stay included or the port could move into a different variable completely?

crazy-max:
fixes #4061 
closes #4065 

Normalize HTTP policy input so hostname allowlists match requests regardless of scheme casing, host casing, or explicit default ports, including curl requests through the exec proxy. Strip `:80` for HTTP and `:443` for HTTPS, including zero-padded forms, while preserving non-default ports, IPv6 brackets and zone casing, and the original URL. Existing policies matching explicit default ports or mixed-case hostnames must use the normalized host instead.

This supersedes #4065, retaining its original commit and adding simplified normalization, case handling, and regression coverage through source conversion and policy evaluation.

### DIFF
--- a/policy/validate.go
+++ b/policy/validate.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/fs"
 	"maps"
-	"net"
 	"net/url"
 	"path"
 	"slices"
@@ -533,22 +532,13 @@ func (p *Policy) Print(ctx print.Context, msg string) error {
 	return nil
 }
 
-// httpHostForPolicy returns the host used in policy input.http.host.
-// Well-known default ports (http/80, https/443) are stripped so curl-style
-// URLs like https://example.com:443 match allow-lists that contain example.com.
-func httpHostForPolicy(scheme, host string) string {
-	h, port, err := net.SplitHostPort(host)
-	if err != nil {
-		return host
-	}
+func normalizeHTTPHost(u *url.URL) string {
+	port := strings.TrimLeft(u.Port(), "0")
 	switch {
-	case scheme == "https" && port == "443", scheme == "http" && port == "80":
-		if strings.Contains(h, ":") {
-			return "[" + h + "]"
-		}
-		return h
+	case u.Scheme == "http" && port == "80", u.Scheme == "https" && port == "443":
+		return strings.TrimSuffix(u.Host, ":"+u.Port())
 	default:
-		return host
+		return u.Host
 	}
 }
 
@@ -574,7 +564,7 @@ func sourceToInput(ctx context.Context, getVerifier PolicyVerifierProvider, src 
 		inp.HTTP = &HTTP{
 			URL:    src.Source.Identifier,
 			Schema: scheme,
-			Host:   httpHostForPolicy(scheme, u.Host),
+			Host:   normalizeHTTPHost(u),
 			Path:   u.Path,
 			Query:  u.Query(),
 		}
--- a/policy/validate.go
+++ b/policy/validate.go
@@ -533,13 +533,16 @@ func (p *Policy) Print(ctx print.Context, msg string) error {
 }
 
 func normalizeHTTPHost(u *url.URL) string {
+	host := u.Host
 	port := strings.TrimLeft(u.Port(), "0")
-	switch {
-	case u.Scheme == "http" && port == "80", u.Scheme == "https" && port == "443":
-		return strings.TrimSuffix(u.Host, ":"+u.Port())
-	default:
-		return u.Host
+	if u.Scheme == "http" && port == "80" || u.Scheme == "https" && port == "443" {
+		host = strings.TrimSuffix(host, ":"+u.Port())
 	}
+	// Preserve IPv6 zone identifiers, which may be case-sensitive interface names.
+	if i := strings.IndexByte(host, '%'); strings.HasPrefix(host, "[") && i >= 0 {
+		return strings.ToLower(host[:i]) + host[i:]
+	}
+	return strings.ToLower(host)
 }
 
 func sourceToInput(ctx context.Context, getVerifier PolicyVerifierProvider, src *gwpb.ResolveSourceMetaResponse, platform *ocispecs.Platform, logf func(logrus.Level, string)) (Input, []string, error) {
@@ -554,6 +557,7 @@ func sourceToInput(ctx context.Context, getVerifier PolicyVerifierProvider, src 
 	if !ok {
 		return inp, nil, errors.Errorf("invalid source identifier: %s", src.Source.Identifier)
 	}
+	scheme = strings.ToLower(scheme)
 
 	switch scheme {
 	case "http", "https":
--- a/policy/validate.go
+++ b/policy/validate.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/fs"
 	"maps"
+	"net"
 	"net/url"
 	"path"
 	"slices"
@@ -532,6 +533,25 @@ func (p *Policy) Print(ctx print.Context, msg string) error {
 	return nil
 }
 
+// httpHostForPolicy returns the host used in policy input.http.host.
+// Well-known default ports (http/80, https/443) are stripped so curl-style
+// URLs like https://example.com:443 match allow-lists that contain example.com.
+func httpHostForPolicy(scheme, host string) string {
+	h, port, err := net.SplitHostPort(host)
+	if err != nil {
+		return host
+	}
+	switch {
+	case scheme == "https" && port == "443", scheme == "http" && port == "80":
+		if strings.Contains(h, ":") {
+			return "[" + h + "]"
+		}
+		return h
+	default:
+		return host
+	}
+}
+
 func sourceToInput(ctx context.Context, getVerifier PolicyVerifierProvider, src *gwpb.ResolveSourceMetaResponse, platform *ocispecs.Platform, logf func(logrus.Level, string)) (Input, []string, error) {
 	var inp Input
 	var unknowns []string
@@ -554,7 +574,7 @@ func sourceToInput(ctx context.Context, getVerifier PolicyVerifierProvider, src 
 		inp.HTTP = &HTTP{
 			URL:    src.Source.Identifier,
 			Schema: scheme,
-			Host:   u.Host,
+			Host:   httpHostForPolicy(scheme, u.Host),
 			Path:   u.Path,
 			Query:  u.Query(),
 		}
--- a/policy/validate_test.go
+++ b/policy/validate_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"testing"
 	"time"
 
@@ -14,6 +15,8 @@ import (
 	slsa1 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1"
 	gwpb "github.com/moby/buildkit/frontend/gateway/pb"
 	"github.com/moby/buildkit/solver/pb"
+	policyaction "github.com/moby/buildkit/sourcepolicy/pb"
+	"github.com/moby/buildkit/sourcepolicy/policysession"
 	policyverifier "github.com/moby/policy-helpers"
 	policyimage "github.com/moby/policy-helpers/image"
 	policytypes "github.com/moby/policy-helpers/types"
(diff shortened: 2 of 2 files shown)


policy/validate.go
```diff
@@ -532,6 +532,20 @@ func (p *Policy) Print(ctx print.Context, msg string) error {
...
+func httpHostForPolicy(scheme, host string) string {
+	switch scheme {
+	case "https":
+		return strings.TrimSuffix(host, ":443")
+	case "http":
+		return strings.TrimSuffix(host, ":80")
+	default:
+		return host
+	}
+}
```
thaJeztah: While... probably not very common, it's ... valid to have leading zeros;

```bash
docker run -d --name foo -p 8080:80 nginx:alpine

curl -s http://localhost:0008080 | grep 'Welcome'
<title>Welcome to nginx!</title>
<h1>Welcome to nginx!</h1>
```



So perhaps something like this?

```go
func httpHostForPolicy(scheme string, u *url.URL) string {
	port := strings.TrimLeft(u.Port(), "0")

	switch {
	case scheme == "http" && port == "80",
		scheme == "https" && port == "443":
		return u.Hostname()
	default:
		return u.Host
	}
}
```

policy/validate.go
thaJeztah, line 619: We probably need to normalize Git URLs as well? i.e.; `https://github.com:443/docker/buildx.git` or `https://github.com:000443/docker/buildx.git` would also need normalising.

policy/validate_test.go
```diff
@@ -23,6 +25,28 @@ import (
...
+	tests := []struct {
+		scheme, host, want string
+	}{
+		{"https", "example.com", "example.com"},
+		{"https", "example.com:443", "example.com"},
+		{"http", "example.com:80", "example.com"},
+		{"https", "example.com:80", "example.com:80"},
+		{"http", "example.com:443", "example.com:443"},
+		{"https", "example.com:8443", "example.com:8443"},
+		{"http", "example.com:8080", "example.com:8080"},
```
thaJeztah, line 54: Host can be empty; perhaps we should have a test for that; (`http://:80` / `https://:443`)

policy/validate_test.go
```diff
@@ -23,6 +25,28 @@ import (
...
+		{"https", "[2001:db8::1]", "[2001:db8::1]"},
```
thaJeztah, line 55: I think Go1.26 rejects them early; not sure if we have a test-case anywhere for the old "permissive" format without brackets? (`http://::1:443`)

policy/validate_test.go
```diff
@@ -23,6 +25,28 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestHTTPHostForPolicy(t *testing.T) {
+	tests := []struct {
+		scheme, host, want string
+	}{
+		{"https", "example.com", "example.com"},
+		{"https", "example.com:443", "example.com"},
+		{"http", "example.com:80", "example.com"},
```
thaJeztah, line 42: Hostnames and schemes are case-insensitive; we may want to add some normalisation for that as well;

```
http://example.com
Http://Example.com
hTtP://eXaMpLe.CoM
HTTP://EXAMPLE.COM
```

thaJeztah (changes requested):
I'd probably squash the commits; if you amend the original commit, it still keeps the author (if we think it's worth preserving), so that's fine.

I left some suggestions for things we're currently missing.

policy/validate.go
crazy-max: Yes agreed that Git over HTTP(S) should use the same normalization. I'd prefer to handle that in a follow-up, covering both remote URL paths and ensuring SSH/SCP remotes retain their existing behavior.

policy/validate.go
```diff
@@ -532,6 +532,19 @@ func (p *Policy) Print(ctx print.Context, msg string) error {
 	return nil
 }
 
+func normalizeHTTPHost(u *url.URL) string {
+	host := u.Host
+	port := strings.TrimLeft(u.Port(), "0")
+	if u.Scheme == "http" && port == "80" || u.Scheme == "https" && port == "443" {
+		host = strings.TrimSuffix(host, ":"+u.Port())
```
thaJeztah: Why not use `u.Hostname()` here? That already provides the cleaned host; https://go.dev/play/p/rp1kJtUNKDl

```go
package main

import (
	"fmt"
	"net/url"
)

func main() {
	inputs := []string{
		"https://github.com:0800/foo.git",
		"https://[fe80::abcd%25Eth0]:8443",
	}

	for _, s := range inputs {
		u, err := url.Parse(s)
		if err != nil {
			panic(err)
		}

		fmt.Println(u.Host)
		fmt.Println(u.Hostname())
		fmt.Println(u.Port())

	}
}
```

policy/validate.go
```diff
@@ -532,6 +532,19 @@ func (p *Policy) Print(ctx print.Context, msg string) error {
...
+	}
+	// Preserve IPv6 zone identifiers, which may be case-sensitive interface names.
+	if i := strings.IndexByte(host, '%'); strings.HasPrefix(host, "[") && i >= 0 {
+		return strings.ToLower(host[:i]) + host[i:]
+	}
```
thaJeztah: This looks incorrect; `url.Parse` would reject a plain `%` as not url-encoded; https://go.dev/play/p/HQdmhO3HbpP

```go
package main

import (
	"fmt"
	"net/url"
)

func main() {
	inputs := []string{
		"https://[fe80::abcd%25Eth0]:8443",
		"https://[fe80::abcd%Eth0]:8443",
	}

	for _, s := range inputs {
		u, err := url.Parse(s)
		if err != nil {
			panic(err)
		}

		fmt.Println(u.Host)
		fmt.Println(u.Hostname())
		fmt.Println(u.Port())

	}
}
```

Outputs;

```
[fe80::abcd%Eth0]:8443
fe80::abcd%Eth0
8443
panic: parse "https://[fe80::abcd%Eth0]:8443": invalid URL escape "%Et"

goroutine 1 [running]:
main.main()
	/tmp/sandbox2041763268/prog.go:17 +0x197
```


If it's an IP-address, we should parse it, then use its canonical form; https://go.dev/play/p/3y39iXxLpbv

```go
package main

import (
	"fmt"
	"net/netip"
)

func main() {
	addrs := []string{
		"FE80::ABCD%Eth0",
		"FE80:0000:0000:0000:0000:0000:0000:ABCD%Eth0",
		"2001:0DB8:0000:0000:0001:0000:0000:0001",
		"0000:0000:0000:0000:0000:0000:0000:0001",

		"192.0.2.128",
		"192.168.001.001", // invalid: IPv4 octets must not have leading zeros

		"::FFFF:192.0.2.128",
		"::FFFF:C000:0280",
	}

	for _, s := range addrs {
		addr, err := netip.ParseAddr(s)
		if err != nil {
			fmt.Printf("%-50s -> error: %v\n", s, err)
			continue
		}

		fmt.Printf("%-50s -> %s\n", s, addr.Unmap())
	}
}
```

Outputs;

```
FE80::ABCD%Eth0                                    -> fe80::abcd%Eth0
FE80:0000:0000:0000:0000:0000:0000:ABCD%Eth0       -> fe80::abcd%Eth0
2001:0DB8:0000:0000:0001:0000:0000:0001            -> 2001:db8::1:0:0:1
0000:0000:0000:0000:0000:0000:0000:0001            -> ::1
192.0.2.128                                        -> 192.0.2.128
192.168.001.001                                    -> error: ParseAddr("192.168.001.001"): IPv4 field has octet with leading zero
::FFFF:192.0.2.128                                 -> 192.0.2.128
::FFFF:C000:0280                                   -> 192.0.2.128
```

policy/validate_test.go
```diff
@@ -23,6 +26,44 @@ import (
...
+		{"https", "example.com:008443", "example.com:008443"},
+		{"http", "example.com:000", "example.com:000"},
+		{"https", "example.com:80", "example.com:80"},
+		{"http", "example.com:443", "example.com:443"},
+		{"https", "example.com:8443", "example.com:8443"},
+		{"http", "example.com:8080", "example.com:8080"},
+		{"https", "[2001:db8::1]", "[2001:db8::1]"},
+		{"https", "[2001:DB8::ABCD]:443", "[2001:db8::abcd]"},
+		{"https", "[FE80::ABCD%Eth0]:000443", "[fe80::abcd%Eth0]"},
+		{"https", "[FE80::ABCD%Eth0]:8443", "[fe80::abcd%Eth0]:8443"},
```
thaJeztah, line 67: See my other comment; this is an invalid address, and would be rejected.

thaJeztah (changes requested):

policy/validate.go
crazy-max: `Hostname()` also removes IPv6 brackets and non-default ports, which we currently preserve. I'll use it as the starting point for IP canonicalization, then restore the brackets and retain non-default ports.

policy/validate.go
crazy-max: This operates on the parsed `u.Host`, where `url.Parse` has already decoded `%25` into `%`, as your example shows. The `netip` suggestion makes sense though; it would also normalize expanded IPv6 addresses and preserve zone casing without the manual handling. I'd use `Addr.String()` without `Unmap()` to preserve the distinction between IPv4-mapped IPv6 and plain IPv4.

policy/validate_test.go
crazy-max: This test constructs `url.URL{Host: ...}` directly, so the decoded `%Eth0` is intentional. The `https-ipv6-zone-case-preserved` source-conversion test separately parses a URL containing `%25Eth0` and verifies the decoded host.

policy/validate.go
```diff
@@ -532,6 +535,34 @@ func (p *Policy) Print(ctx print.Context, msg string) error {
...
+			return r
+		}, host)
+	}
+	port := u.Port()
+	normalizedPort := strings.TrimLeft(port, "0")
+	if u.Scheme == "http" && normalizedPort == "80" || u.Scheme == "https" && normalizedPort == "443" {
+		port = ""
+	}
+	if port != "" || strings.HasSuffix(u.Host, ":") {
+		return net.JoinHostPort(host, port)
```
thaJeztah, line 565: Should this join the normalised port? i.e., `000123` should likely be normalised to `123` to get a normalised reference.

thaJeztah (commented):
found one possible issue

PR #3369, opened 2025-08-15
Add support for native grpc on the docker socket

dmcgowan:
- relates to / addresses https://github.com/moby/moby/issues/49836
- relates to https://github.com/moby/buildkit/issues/5927
- Related to https://github.com/moby/moby/pull/50744

PR #4069, opened 2026-09-09
docker: use native gRPC for supported local endpoints

crazy-max:
- relates to / addresses https://github.com/moby/moby/issues/49836
- relates to https://github.com/moby/buildkit/issues/5927
- Related to https://github.com/moby/moby/pull/50744
- supersedes, closes https://github.com/docker/buildx/pull/3369
- supersedes, closes https://github.com/docker/buildx/pull/3780

Automatically use native gRPC for supported Docker daemons over Unix sockets and named pipes, using the daemon support introduced in [moby/moby#50744](https://github.com/moby/moby/pull/50744). Retain the legacy `/grpc` and `/session` transport for remote endpoints and daemons without native gRPC support, preserving Docker's existing TLS and SSH handling. Concurrent callers share a bounded capability probe, and failed or canceled probes remain retryable. Explicit authentication and authorization failures are returned without falling back.

### DIFF
--- a/driver/docker/driver.go
+++ b/driver/docker/driver.go
@@ -5,12 +5,18 @@ import (
 	"net"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/docker/buildx/driver"
 	"github.com/docker/buildx/util/progress"
 	"github.com/moby/buildkit/client"
+	"github.com/moby/buildkit/util/flightcontrol"
+	"github.com/moby/buildkit/util/grpcerrors"
 	dockerclient "github.com/moby/moby/client"
+	"github.com/moby/moby/client/pkg/versions"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc/codes"
 )
 
 type Driver struct {
@@ -21,6 +27,7 @@ type Driver struct {
 	// https://github.com/docker/docs/blob/main/content/build/drivers/docker.md
 	features    features
 	hostGateway hostGateway
+	nativeGRPC  flightcontrol.CachedGroup[bool]
 }
 
 func (d *Driver) Bootstrap(ctx context.Context, l progress.Logger) error {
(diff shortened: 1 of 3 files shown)


tests/docker.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+			out, err := dockerCmd(sb, withArgs("context", "create", name, "--docker", endpoint)).CombinedOutput()
+			require.NoError(t, err, string(out))
+			t.Cleanup(func() {
+				out, err := dockerCmd(sb, withArgs("context", "rm", "-f", name)).CombinedOutput()
+				require.NoError(t, err, string(out))
+			})
+			if trusted {
+				// Moby's TCP listener overwrites TLS ALPN with HTTP/1.1, so expect fallback until fixed upstream.
+				testDockerGRPCBuild(t, sb, name, false)
```
crazy-max: Native gRPC over TLS currently falls back to `/grpc` because Moby's TCP listener overwrites the configured ALPN protocols. The daemon configures `h2` and `http/1.1`, but [`sockets.NewTCPSocket` replaces them with only `http/1.1`](https://github.com/moby/moby/blob/3ce5872b7950c63ba2ffbc5123101019ff3e6682/vendor/github.com/docker/go-connections/sockets/tcp_socket.go#L18). This reproduces on Docker 29.8 with `tls: client requested unsupported application protocols (["h2"])`.

The upstream fix should preserve the configured ALPN protocols when [Moby creates its TCP listener](https://github.com/moby/moby/blob/3ce5872b7950c63ba2ffbc5123101019ff3e6682/daemon/listeners/listeners_linux.go#L30), with coverage that verifies successful `h2` negotiation over TLS. Until then, the Buildx TLS integration test expects legacy fallback while still checking successful builds and rejection of untrusted certificates.

driver/docker/driver.go
```diff
@@ -64,16 +76,88 @@ func (d *Driver) Dial(ctx context.Context) (net.Conn, error) {
 }
 
 func (d *Driver) Client(ctx context.Context, opts ...client.ClientOpt) (*client.Client, error) {
+	// Native gRPC bypasses the daemon's HTTP authorization middleware.
+	// Keep it opt-in until Moby enforces authorization on that path.
+	enabled, _ := strconv.ParseBool(os.Getenv("BUILDX_DOCKER_NATIVE_GRPC"))
```
crazy-max: Native gRPC currently bypasses Moby's authorization plugins. [`httpHandler.ServeHTTP` dispatches gRPC requests directly to the gRPC server or extension proxy](https://github.com/moby/moby/blob/3ce5872b7950c63ba2ffbc5123101019ff3e6682/daemon/command/httphandler.go#L38-L47), bypassing the HTTP API handler where [the authorization middleware is registered](https://github.com/moby/moby/blob/3ce5872b7950c63ba2ffbc5123101019ff3e6682/daemon/command/daemon.go#L885-L886). Consequently, these requests do not pass through its [`AuthZRequest` check](https://github.com/moby/moby/blob/3ce5872b7950c63ba2ffbc5123101019ff3e6682/pkg/authorization/middleware.go#L74-L79).

This is why native gRPC remains opt-in here. Moby needs to enforce authorization before dispatching native RPCs, covering both local services and the extension proxy while preserving streaming behavior. Keeping Buildx on the legacy transport by default preserves its existing authorization path, but doesn't prevent other clients from accessing the daemon's native gRPC endpoint.

tests/docker.go
crazy-max: relates to https://github.com/docker/buildx/actions/runs/34351778177/job/102466555883#step:7:2918

```
=== Failed
=== FAIL: tests TestIntegration/TestDockerNativeGRPCTLS/worker=docker/trusted=true/native (0.10s)
    docker.go:73: 
        	Error Trace:	/src/tests/docker.go:73
        	Error:      	"DEBUG: docker driver: native gRPC unavailable, using /grpc: failed to list workers: Unavailable: connection error: desc = \"transport: authentication handshake failed: remote error: tls: no application protocol\"\nDEBUG: docker driver: native gRPC unavailable, using /grpc: failed to list workers: Unavailable: connection error: desc = \"transport: authentication handshake failed: remote error: tls: no application protocol\"\n\n" does not contain "docker driver: using native gRPC"
        	Test:       	TestIntegration/TestDockerNativeGRPCTLS/worker=docker/trusted=true/native
            --- FAIL: TestIntegration/TestDockerNativeGRPCTLS/worker=docker/trusted=true/native (0.10s)
...
...
     sandbox.go:205: 2026/09/09 12:41:58 http: TLS handshake error from 127.0.0.1:37134: tls: client requested unsupported application protocols (["h2"])
```

driver/docker/driver.go
thaJeztah: The /grpc endpoint is deprecated; no need to make it opt-in. Buildx isn't responsible for protecting the API because anything already can do.

crazy-max: Removed the opt-in. I kept the `DialMeta` guard because Desktop uses `X-Docker-Desktop-idle: background` on the `/grpc` upgrade to let Resource Saver distinguish background connections from interactive builds (https://github.com/docker/desktop-build/pull/312). That's separate from Moby's authorization middleware. These clients retain the legacy transport until equivalent native gRPC behavior is supported and verified.

PR #4077, opened 2026-09-14
kubernetes: wait for exec connection readiness

### LINKED ISSUE #4040: driver/kubernetes: same first-connection race as #4021 — Dial() returns before SPDY stream is ready
The `kubernetes` driver has the structural bug #4021 identified in the `docker-container` driver: `Dial()` returns synchronously while the transport underneath isn't ready, so the failure surfaces at the first gRPC RPC. The retry added by #3493 wraps `Dial()`, so it doesn't cover this path.

Symptom is #2668's error, on buildx v0.36.1 (which includes #3493):

```
ERROR: unable to upgrade connection: error dialing backend:
       remote error: tls: internal error
ERROR: failed to build: listing workers for Build: failed to list workers:
       Unavailable: connection error: desc = "transport: failed to write
       client preface: ..."
```

### Why #3493 doesn't catch it

[`ExecConn`](https://github.com/docker/buildx/blob/v0.36.1/driver/kubernetes/execconn/execconn.go) runs `StreamWithContext` in a goroutine and returns success immediately:

```go
go func() {
    serr := exec.StreamWithContext(ctx, ...)
    if serr != nil && serr != context.Canceled {
        logrus.Error(serr)   // only logged
    }
}()
return kc, nil               // always success
```

The apiserver→kubelet TLS handshake happens inside that goroutine. If it fails (kubelet-serving CSR not yet signed on a fresh node, etc.), the goroutine logs and closes the pipe, but `ExecConn` has already returned success. `tryWithBackoff` sees no error. gRPC's later `Write(clientPreface)` on the broken conn is what surfaces `tls: internal error`, outside the retry scope.

Same shape as #4021: dial succeeds, transport isn't ready, first RPC fails, no retry.

### Evidence

Hit on EKS 1.35 during a CA scale-up. Buildkit pods landed on freshly-joined nodes. `buildx create --bootstrap` succeeded; `buildx build` failed **259 ms** later — shorter than `tryWithBackoff`'s first 500 ms backoff, consistent with the retry not firing.

Cluster's kubelet-serving CSR sign latency during the scale-up: 15–33s, which is the window during which the apiserver→kubelet TLS handshake fails.

### Suggested fix

Mirror #4027: probe readiness before returning from `Dial()`, so async setup failures surface synchronously and `tryWithBackoff` can retry them.

Happy to send a PR if that direction looks right.

### Environment

- buildx `v0.36.1`, moby/buildkit `v0.32.2` (vendored)
- EKS 1.35, `serverTLSBootstrap: true`, `eks:certificate-controller` as CSR signer

jarqvi:
fixes #4040

The kubernetes driver dials a builder pod by running `buildctl dial-stdio`
through the exec API. `ExecConn` started `StreamWithContext` in a goroutine and
returned the connection immediately, before the SPDY stream was up. If the
upgrade failed (e.g. `tls: internal error` on a fresh node whose kubelet cert
isn't signed yet), `Dial()` had already reported success, so `tryWithBackoff`
never retried and the error surfaced later as a gRPC client-preface error.

Wait until the stream is established before returning. The executor reads stdin
only once its streams are created, so the first stdin read is used as the
readiness signal — no sleeps or probes. Transient setup failures are now
returned synchronously so `tryWithBackoff` can retry, and context cancellation
still unblocks the wait.

Added unit tests for `newExecConn` and `tryWithBackoff` using a fake executor.


driver/kubernetes/execconn/execconn_test.go
```diff
@@ -4,47 +4,166 @@ import (
...
+		return context.Cause(ctx)
+	}
+
+	go func() {
+		if f.echo {
+			_, _ = io.Copy(opts.Stdout, opts.Stdin)
+		} else {
+			_, _ = io.Copy(io.Discard, opts.Stdin)
+		}
+	}()
```
crazy-max, line 65: This goroutine keeps reading stdin after the fake executor returns, so a write can still succeed when the test expects an error. I reproduced this by running the stream-end tests with `-count=10000`. Could we make the fake stop consuming stdin once it has signaled readiness?

jarqvi: Thank u for the note, done

crazy-max (changes requested):
Looks good thanks. Please squash the two commits before merging.

jarqvi:
> Looks good thanks. Please squash the two commits before merging.

Thanks, done

crazy-max (approved):
LGTM, PTAL @AkihiroSuda

AkihiroSuda (approved):

PR #4079, opened 2026-09-15
build: detect zstd-compressed context from stdin

nrvate:
**Problem**

`docker build -` decides whether stdin carries a build context or a raw Dockerfile by sniffing magic bytes (`isArchive` in `build/utils.go`). The table knows bzip2, gzip, xz and plain tar — but not zstd. A zstd-compressed context on stdin is therefore parsed as a Dockerfile and the build fails immediately:

```
$ tar cf ctx.tar Dockerfile && zstd -q ctx.tar -o ctx.tar.zst
$ docker build --check - < ctx.tar.zst
1 | >>> (▒/▒d '5 ▒F`i...
```

gzip and plain tar on the same path work.

**Fix**

Add the zstd frame magic (`0x28, 0xB5, 0x2F, 0xFD`) to the table. Nothing else changes: for the default docker driver the stream is passed through to the daemon, and the daemon already decompresses zstd build contexts — verified against Docker Engine 29.7.2 by building the same tar.zst context through the `/build` API directly (it succeeds today; only this client-side sniff rejects it).

**Testing**

- new `TestIsArchive` covering plain tar, gzip, bzip2, xz, zstd, a Dockerfile-from-stdin negative, and a truncated-magic negative
- ```go test ./build/ -run TestIsArchive``` passes (Go 1.26, per go.mod)

Related: moby/go-archive#13 (broader zstd support in the archive library — this PR is the narrower client-side sniff gap).


crazy-max (changes requested):
I initially thought moby/buildkit#7069 covered this, but Buildx needs to recognize the archive before uploading it, so this fix is needed.

I’ve extracted a shared `compression.IsArchive` helper from moby/buildkit#7034 in https://github.com/moby/buildkit/pull/7148, including zstd skippable-frame support. Once that lands and we vendor it here, we can replace Buildx local detector with the shared helper and cover zstd contexts with a stdin integration test. That should keep the two implementations from drifting again.

PR #4103, opened 2026-09-18
git: honor advice controls for build contexts and remote Bake

crazy-max:
follow-up https://github.com/moby/buildkit/pull/6998

Propagate BUILDKIT_GIT_ADVICE to Git source states and add BUILDX_BAKE_GIT_ADVICE for fetching remote Bake definitions.


crazy-max:
needs docs follow-up for BUILDX_BAKE_GIT_ADVICE in https://docs.docker.com/build/building/variables/#build-tool-configuration-variables

